PR #35, opened 2023-08-13
fix(or remove): dimensions bench

Ethan-000:
hi :)

`dimensions.rs` in benches is not working rn. hopefully this will fix it


moodlezoup (approved):
thanks!

PR #37, opened 2023-08-15
Made less than subtable generic

puma314:


sragss (commented):
Looking good.

Few comments:
1) rename `lt.rs` -> `cmp.rs` or similar
2) add a test similar to the ex

sragss:
Closing as we've adjusted the `SubtableStrategy` to `JoltInstruction` to support Jolt.

PR #42, opened 2023-08-22
feat: parallelized to_grand_products

shryasss:
Parallelized `to_grand_products` - Issue https://github.com/a16z/Lasso/issues/33

`trait SubtableStrategy` only had constants and functions which did not modify any state, so just making it `Sync` and using a `par_iter` worked. 

Not super confident about this, so any remarks are appreciated!


moodlezoup:
superseded by #51

PR #44, opened 2023-08-22
feat(parallelize): parallelize prove_arbitrary

### LINKED ISSUE #31: Parallelize Sumcheck::prove_arbitrary summation
Feature gate with "multicore"

PatStiles:
Closes #31


PatStiles:
I wrapped elected to wrap `eval_points` in a `Mutex` for simplicity. It may not be ideal due to thread contention.

PR #56, opened 2023-09-30
Jolt Circom/R1CS files

arasuarun:
Added the jolt circom file. The test in `src/r1cs/snark.rs` compiles it using `circom_scotia` and generates a proof using `spartan2`. Circom must be installed (https://docs.circom.io/getting-started/installation/#installing-dependencies) but the files are compiled using scripts in the test. 

When we decide on a transcript structure, it simply needs to be loaded into the `JoltCircuit` struct as a vector of field elements.


sragss (approved):
LGTM

A few items to be handled post-merge:
- More tests (test successful manually written trace, test failing malicious trace)
- `JoltCircuit::synthesize` should not run the proof, rather just build. This should be moved to `JoltCircuit::prove`

PR #58, opened 2023-10-17
Bn254

liyuange:
Use bn254 instead of curve25519.


sragss:
Thanks for the PR – Multiple curves are intended to be supported via the generics as your PR demonstrates. For now we'd like to keep the benchmarking on 25519.

PR #61, opened 2023-10-25
Add instr

arasuarun:
Added support for the ADD instruction with the identity (IDEN) and the LOWER9 subtables. The test seems to be working for both the instruction and the subtables.


src/jolt/subtable/lower9.rs
```diff
@@ -0,0 +1,59 @@
...
+
+use super::LassoSubtable;
+
+#[derive(Default)]
+pub struct LOWER9Subtable<F: PrimeField> {
+  _field: PhantomData<F>,
+}
+
+impl<F: PrimeField> LOWER9Subtable<F> {
+  pub fn new() -> Self {
```
moodlezoup, line 16: I'd prefer this to be a generic "lower N bits" subtable, so it can be agnostic to `C` and `M`
If we add `C` and `M` as parameters to `JoltInstruction.subtables`, we can compute the number of lower bits we need and pass that in to the constructor here

src/jolt/instruction/add.rs
```diff
@@ -0,0 +1,57 @@
...
+  fn add_instruction_e2e() {
+    let mut rng = test_rng();
+    const C: usize = 8;
+    const M: usize = 1 << 16;
+
+    let u64_mask: u64 = 0xffffffffffffffff;
+
+    for _ in 0..256 {
+      let (x, y) = (rng.next_u64(), rng.next_u64());
+      jolt_instruction_test!(ADDInstruction(x, y), ((x+y) & u64_mask).into());
```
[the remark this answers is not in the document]
arasuarun: Okay I changed it to use `overflowing_add`. Lemme know if that's what you meant.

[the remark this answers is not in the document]
moodlezoup: yep, thanks

src/jolt/instruction/add.rs
```diff
@@ -0,0 +1,55 @@
...
+    concatenate_lookups([&vals[C..(C+1)], &vals[1..C]].concat().as_slice(), C, M)
+  }
+
+  fn g_poly_degree(&self, _: usize) -> usize {
+    1
+  }
+
+  fn subtables<F: PrimeField>(&self) -> Vec<Box<dyn LassoSubtable<F>>> {
+    vec![Box::new(IDENSubtable::new()), Box::new(LOWERKSubtable::new(9))]
+  }
```
moodlezoup, line 27: sorry I wasn't clear –– I meant that this "9" should be calculated from the word size, `C`, and `M` (which requires passing `C` and `M` as parameters to this function)

```suggestion
  fn subtables<F: PrimeField>(&self, C: usize, M: usize) -> Vec<Box<dyn LassoSubtable<F>>> {
    let k: usize = 64 % (log2(M) / 2);
    vec![Box::new(IDENSubtable::new()), Box::new(LOWERKSubtable::new(k))]
  }
```

src/jolt/subtable/test.rs
```diff
@@ -18,4 +18,22 @@ macro_rules! subtable_materialize_mle_parity_test {
       }
     }
   };
+  ($test_name:ident, $subtable_type:ty, $F:ty, $M:expr, $A:expr) => {
```
moodlezoup, line 21: Actually I think we can use a variadic here to avoid the duplication
```suggestion
  ($test_name:ident, $subtable_type:ty, $F:ty, $M:expr, $($constructor_args:expr),*) => {
```

moodlezoup (approved):

PR #66, opened 2023-10-30
Feat(zeromorph): Add Zeromorph PCS

PatStiles:
Adds an implementation of the [Zeromorph](https://eprint.iacr.org/2023/917.pdf) PCS based on the implementation done by the wonderful folks at [Aztec](https://github.com/AztecProtocol/barretenberg/blob/master/cpp/src/barretenberg/honk/pcs/zeromorph/zeromorph.hpp).


sragss:
Let's try to stay close to the `arkworks/poly-commit` interface. Currently their `PolynomialCommitment` trait doesn't support multilinear polynomial commitment schemes, but work is being tracked [here](https://github.com/arkworks-rs/poly-commit/issues/116) and it will likely look quite similar to the univariate trait.

`arkworks-rs/poly-commit`:
```rust
let pp = PCS::setup(max_degree, None, rng).unwrap(); // trusted setup - we should hide this from the interface

// 2. PolynomialCommitment::trim
// Since the setup produced pp with a max degree of 16, and our poly is of degree 10, we can trim the SRS to tailor it to this example.
let (ck, vk) = PCS::trim(&pp, degree, 2, Some(&[degree])).unwrap(); 

// 3. PolynomialCommitment::commit
// The prover commits to the polynomial using their committer key `ck`.
let (comms, rands) = PCS::commit(&ck, [&labeled_poly], Some(rng)).unwrap(); 

let challenge_generator: ChallengeGenerator<<Bls12_377 as Pairing>::ScalarField, Sponge_Bls12_377> = ChallengeGenerator::new_univariate(&mut test_sponge);

// 4a. PolynomialCommitment::open
// Opening proof at a single point.
let proof_single = PCS::open(&ck, [&labeled_poly], &comms, &point_1, &mut (challenge_generator.clone()), &rands, None).unwrap(); 

// 5a. PolynomialCommitment::check
// Verifying the proof at a single point, given the commitment, the point, the claimed evaluation, and the proof.
assert!(PCS::check(&vk, &comms, &point_1, [secret_poly.evaluate(&point_1)], &proof_single, &mut (challenge_generator.clone()), Some(rng)).unwrap()); 
```
(From the [README](https://github.com/arkworks-rs/poly-commit#readme))

When we do this refactor, I think we should hide `gens` from the high level interfaces. If we need trusted setup the `MutliPoly.commit()` function should be able to statically access the trusted setup parameters of the required size on their own. This would handle steps 1 and 2 of the `arkworks-rs/poly-commit` interface.

PR #68, opened 2023-11-01
[JOLT-34] Fix Surge 2.0 non-power-of-two operations

sragss:
Fixes bug where Surge 2.0 only worked for even power of two operations. Now we pad read_cts and final_cts with reads into the 0th element of the table.


moodlezoup (approved):
lgtm

PR #69, opened 2023-11-01
[JOLT-45] Make SLLInstruction generic-ish over C

moodlezoup:
Requires threading C through the `subtables` method


sragss (approved):
lgtm – wish we could use const generics here, but trait object issues...

PR #70, opened 2023-11-01
Jolt: Subtable Reuse

sragss:
Subtable reuse via summing correct instruction_flags in verifier `FingerprintStrategy`.

Includes separate tests in `test_vm.rs`.

### DIFF
--- a/src/jolt/vm/test_vm.rs
+++ b/src/jolt/vm/test_vm.rs
@@ -95,14 +95,22 @@ mod tests {
   use rand_chacha::rand_core::RngCore;
   use std::collections::HashSet;
 
-  use crate::jolt::instruction::bge::BGEInstruction;
+  use crate::jolt::instruction::and::ANDInstruction;
+use crate::jolt::instruction::bge::BGEInstruction;
   use crate::jolt::instruction::bgeu::BGEUInstruction;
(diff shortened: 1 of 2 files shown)


src/jolt/vm/test_vm.rs
```diff
@@ -148,4 +166,56 @@ mod tests {
...
+  fn subtable_reuse() {
+    // Setup VM / Instructions / Subtables
+    #[repr(u8)]
+    #[derive(Copy, Clone, EnumIter, EnumCountMacro)]
+    #[enum_dispatch(JoltInstruction)]
+    pub enum ReuseInstructionSet {
+      BNE(BNEInstruction),
+      BGEU(BGEUInstruction),
+    }
+    impl Opcode for ReuseInstructionSet {}
```
moodlezoup, line 180: why not use the macro for this?

moodlezoup (approved):
just one q

src/jolt/vm/test_vm.rs
sragss: It was running into reimport issues. We have to sort out the macro export at some point if we want to keep using the macro. IIRC this is difficult due to the non-idempotence of enum_dispatch.

PR #72, opened 2023-11-02
Delete old Lasso code

moodlezoup:
Deletes all the old subtables, `surge.rs`, `densified.rs`, plus some files from an earlier iteration of Jolt. 
Also deletes some now-unused code from `grand_product.rs` and `sumcheck.rs`


sragss (approved):
lgtm. can you rename surge_2.rs -> surge.rs as well?

PR #73, opened 2023-11-03
[JOLT-43] instruction_util docs and tests

sragss:
Adds `instruction_utils.rs` tests + doc comments.

### DIFF
--- a/jolt-core/src/jolt/instruction/sll.rs
+++ b/jolt-core/src/jolt/instruction/sll.rs
@@ -71,9 +71,10 @@ mod test {
     let operand_size = 22 / 2;
 
     for _ in 0..8 {
-      let (x, y) = (rng.next_u64(), rng.next_u64() % operand_size);
+      let (x, y) = (rng.next_u64(), rng.next_u64());
 
-      let entry: u64 = x.checked_shl((y) as u32).unwrap_or(0);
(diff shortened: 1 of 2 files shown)


src/jolt/instruction/sll.rs
```diff
@@ -62,11 +62,12 @@ mod test {
...
     const C: usize = 6;
     const M: usize = 1 << 22;
+    let operand_size = 22 / 2;
 
     for _ in 0..8 {
-      let (x, y) = (rng.next_u64(), rng.next_u64());
+      let (x, y) = (rng.next_u64(), rng.next_u64() % operand_size);
 
-      let entry: u64 = x.checked_shl((y % 64) as u32).unwrap_or(0);
+      let entry: u64 = x.checked_shl((y) as u32).unwrap_or(0);
```
moodlezoup, line 70: I think this was right before; the SLL instruction is specified to ignore all but the last 5 bits of y: https://jemu.oscc.cc/SLL

moodlezoup (approved):
one more nit

src/jolt/instruction/sll.rs
sragss: Ok, removed this and the assertion that y <= operand_size in `utils;:chunk_and_concatenate_for_shift`.

PR #74, opened 2023-11-04
refactor: migrate to cargo workspace

ncitron:
Basic migration to a workspace.

- I chose to keep the features for ark-lasso inside the ark-lasso crates features section rather than the global namespace, but if we want to change this down the line we can.
- Would it make sense to migrate Jolt to it's own crate in the workspace? One thing to note about crates in a workspace is we cannot create a dependency cycle (Jolt cannot import Lasso if Lasso needs to import Jolt). 
- If we add more crates with shared dependencies we can actually move them to the workspace cargo file to make version changes easier.
- What version number should the top level crate be? 0.1.0? I see that 0.4.0 was chosen for Lasso to match Spartan.


Cargo.toml
```diff
@@ -18,57 +18,8 @@ repository = "https://github.com/a16z/lasso"
...
-[lib]
-name = "liblasso"
-path = "src/lib.rs"
-
-[features]
-default = ["ark-ec/parallel", "ark-ff/parallel", "ark-std/parallel", "ark-ff/asm", "multicore"]
-multicore = ["rayon"]
-ark-msm = [] # run with arkworks MSM without small field element optimization
+[workspace]
+members = ["ark-lasso"]
```
sragss, line 22: Let's rename to "lasso" rather than ark-lasso. I don't think the arkworks reference needs to be primary in the naming scheme.

sragss (changes requested):
One comment on naming.

Cargo.toml
ncitron: One problem with that is that the top level workspace is also named lasso. We are not allowed to have a crate with the same name as the workspace. Any thoughts on what to change the crate/workspace to?

moodlezoup: `jolt` and `jolt-core`? it's no longer just Lasso anymore

ncitron: done

Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
-name = "ark-lasso"
+name = "jolt"
 # sync up with Arkwork's version
```
sragss, line 3: Nit: rm

jolt-core/Cargo.toml
```diff
@@ -0,0 +1,71 @@
...
+# sync up with Arkwork's version
+version = "0.4.0"
+authors = [
+    # author of original Spartan paper and code base 
+    "Srinath Setty <srinath@microsoft.com>",
+    # authors who contributed to the Arkworks Spartan fork
+    "Zhenfei Zhang <zhenfei.zhang@hotmail.com>",
+    # authors who contributed to Lasso
+    "Michael Zhu <mzhu@a16z.com>",
+    "Sam Ragsdale <sragsdale@a16z.com>",
```
sragss, line 12: Nit: +Noah

jolt-core/Cargo.toml
```diff
@@ -0,0 +1,71 @@
...
+]
+edition = "2021"
+description = "LASSO SNARK lookup argument, built on Arkworks"
```
sragss, line 15: The lookup singularity. Based on Spartan; built on Arkworks.

sragss (approved):
LGTM

PR #76, opened 2023-11-06
duplicated jolt circom file

arasuarun:
The jolt circuit with all (a, v, t) values explicitly provided as witnesses without de-duplication. For example, in the old version, the program code memory checks are all done with the same address and need to be provided only once (but requires optimizations when calculating the MLE). This version has it all explicitly provided as separate elements and verifies they're equal in the circuit. This is simpler when calculating MLEs later on.

This duplication is done for all memory reads.

### DIFF
--- a/src/r1cs/circuits/compile_jolt.sh
+++ b/src/r1cs/circuits/compile_jolt.sh
@@ -3,4 +3,5 @@
 cd src/r1cs/circuits
 mkdir -p jolt
 circom jolt.circom --r1cs --wasm --sym --c --output jolt
+circom jolt_duplicated.circom --r1cs --wasm --sym --c --output jolt
 cd -
--- a/src/r1cs/circuits/jolt_duplicated.circom
+++ b/src/r1cs/circuits/jolt_duplicated.circom
@@ -1,0 +1,316 @@
+pragma circom 2.1.6;
+
+/*  Constants: written as functions because circom.
+*/
+function N_CHUNKS() {return 6;}
+function L_CHUNK() {return 11;}
+function L_MS_CHUNK() {return 9;}
+function N_FLAGS() {return 15;}
+
+function WIT_PER_STEP() {return 6 * 3 + 1 + 2 * 3 + N_FLAGS() + 8 * 3 + 1 + 2 + 4 * N_CHUNKS();} // 54 for N=1, NC=6, NF=15
+
+function ALL_ONES() {return 0x1111111111111111;} 
+
+// state: [field; 2] = [step_num, pc]
+function STEP_NUM_IDX() {return 0;}
+function PC_IDX() {return 1;}
+
+// Utility functions
+template subarray(S, L, N) {
+    signal input in[N];
+    signal output out[L];
+
+    for (var i=S; i<S+L; i++) {
+        out[i-S] <== in[i];
+    }
+}
+
+template submatrix(S, R, C, N) {
+    signal input in[N];
+    signal output out[R][C];
+
+    for (var r=0; r<R; r++) {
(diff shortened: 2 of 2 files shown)


src/r1cs/circuits/jolt_duplicated.circom
```diff
@@ -0,0 +1,316 @@
...
+    signal input in[N * WIT_PER_STEP()];
+
+    component jolt_loop; 
+    jolt_loop = JoltLoop(N); 
+    jolt_loop.in <== in;
+
+ }
+
+component main {public [in]} 
+    = JoltMain(1);
```
sragss, line 316: Eventually I think we'll need to make this file a template which gets written by `build.rs` such that we can set the constants at runtime. Fine for now.

src/r1cs/circuits/jolt_duplicated.circom
```diff
@@ -0,0 +1,316 @@
...
+    /* These Per-step witnesses
+    */
+
+    signal prog_read_avt[6][3];
+    signal op_flags_packed;
+    signal op_flag[N_FLAGS()]; // just advice, not read from code
+
+    /* Register reading witnesses 
+    */
+    signal rs1_read_avt[3];
```
sragss, line 91: What are the expected values here? Is this a special (a,v,t) from RAM memory checking where a == rs1?

src/r1cs/circuits/jolt_duplicated.circom
```diff
@@ -0,0 +1,316 @@
...
+    signal rs2_read_avt[3];
+
+
+    signal mem_read_write_avt[8][3];
+    signal mem_read_val; // advice for loads; asserted for stores
+
+    signal lookup_output;
+    signal instr_lookup_query;
```
sragss, line 99: What are these two values? Could you add a comment?

src/r1cs/circuits/jolt_duplicated.circom
```diff
@@ -0,0 +1,316 @@
...
+    /* The rest of this function has three broad components: 
+    1. Parse the witness to get the individual elements 
+    2. Write out the constraints
+    */
+
+    /*********************** 1. PARSE WITNESSES *******************************/
+    /* These Per-step witnesses
+    */
+
+    signal prog_read_avt[6][3];
```
sragss, line 85: Could you add a comment describing this value?

src/r1cs/circuits/jolt_duplicated.circom
```diff
@@ -0,0 +1,316 @@
...
+    signal op_flags_packed;
+    signal op_flag[N_FLAGS()]; // just advice, not read from code
+
+    /* Register reading witnesses 
+    */
+    signal rs1_read_avt[3];
+    signal rs2_read_avt[3];
+
+
+    signal mem_read_write_avt[8][3];
```
sragss, line 95: Comment explaining the dimensionality "8".

src/r1cs/circuits/compile_jolt.sh
```diff
@@ -3,4 +3,5 @@
 cd src/r1cs/circuits
 mkdir -p jolt
 circom jolt.circom --r1cs --wasm --sym --c --output jolt
+circom jolt_duplicated.circom --r1cs --wasm --sym --c --output jolt
```
sragss, line 6: Is there a reason we're keeping the old one around? Is it better in some cases?

sragss (commented):
Nits mostly around naming and comments.

I figure this is the source of truth for what the R1CS portion handles, so we may as well document all the input signals thoroughly inline.

PR #77, opened 2023-11-07
Jolt trace interface

sragss:


jolt-core/src/jolt/vm/pc.rs
```diff
@@ -138,23 +154,23 @@ pub struct PCPolys<F: PrimeField> {
 
 impl<F: PrimeField> PCPolys<F> {
   // TODO(sragss): precommit PC strategy
-  pub fn new_program(program: Vec<ELFRow>, trace: Vec<ELFRow>) -> Self {
-	assert!(is_power_of_two(program.len()));
-	assert!(is_power_of_two(trace.len()));
+  pub fn new_program(mut program: Vec<ELFRow>, mut trace: Vec<ELFRow>) -> Self {
+    Self::validate_program(&program, &trace);
```
moodlezoup, line 158: maybe we should put this behind `#[cfg(test)]`?

jolt-core/src/jolt/trace/rv.rs
```diff
@@ -0,0 +1,721 @@
...
+
+#[derive(Debug, PartialEq)]
+pub enum RV32InstructionFormat {
+  R,
+  I,
+  S,
+  SB,
+  U,
+  UJ,
+}
```
moodlezoup, line 90: Instead of having all the Options on `RVTraceRow`, why not change these enum variants to structs?
```suggestion
struct RInstruction {
    pc: u64;
    opcode: RV32IM;
    rd: u64;
    rs1: u64;
    rs2: u64;
    rd_pre_val: u64;
    rd_post_val: u64;
    rs1_val: u64;
    rs2_val: u64;
}
// etc...
pub enum RV32InstructionFormat {
  R(RInstruction),
  I(IInstruction),
  S(SInstruction),
  SB(SBInstruction),
  U(UInstruction),
  UJ(UJInstruction),
}
```

jolt-core/src/jolt/trace/rv.rs
```diff
@@ -0,0 +1,721 @@
...
+      F::from(flag_5),
+      F::from(flag_6),
+      F::from(flag_7),
+      F::from(flag_8),
+      F::from(flag_9),
+      F::from(flag_10),
+      F::from(flag_11),
+      F::from(flag_12),
+      F::from(flag_13),
+      F::from(flag_13),
```
arasuarun, line 676: flag 13 is stored twice

jolt-core/src/jolt/vm/pc.rs
sragss: We could – I was thinking the cost was fairly minimal compared to our other activities

jolt-core/src/jolt/vm/pc.rs
```diff
@@ -138,23 +154,23 @@ pub struct PCPolys<F: PrimeField> {
...
+    Self::preprocess(&mut program, &mut trace);
+
+    // Preprocessing should deal with padding.
+    assert!(is_power_of_two(program.len()));
+    assert!(is_power_of_two(trace.len()));
```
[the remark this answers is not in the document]
sragss: Done

jolt-core/src/jolt/trace/rv.rs
sragss: I went back and forth on this. If you look at the match statements below, we mostly don't want to match on the InstructionFormat which would add unnecessary unwrapping in most cases.

It's a bit more memory efficient, and slightly cleaner in specific cases (R instructions for example), but at the cost of readability.

jolt-core/src/jolt/trace/rv.rs
sragss: Good catch!

sragss:
Closing as this is handled in V2.

PR #78, opened 2023-11-07
Build jolt.circom via Cargo

sragss:
- Stores the compiled artifacts to `./target/<build type>/jolt-core-<post-fix>/out/circom/...`
- Stores this path to `CIRCUIT_DIR`
- `snark.rs` reads `CIRCUIT_DIR`

### DIFF
--- a/jolt-core/src/r1cs/snark.rs
+++ b/jolt-core/src/r1cs/snark.rs
@@ -7,9 +7,7 @@ use spartan2::{
   SNARK,
 };
 
-use bellpepper_core::{
-  Circuit, ConstraintSystem, SynthesisError,
-};
+use bellpepper_core::{Circuit, ConstraintSystem, SynthesisError};
 use core::marker::PhantomData;
 use ff::PrimeField;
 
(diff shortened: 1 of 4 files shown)


jolt-core/src/r1cs/snark.rs
```diff
@@ -26,9 +24,10 @@ struct JoltCircuit<F: PrimeField> {
...
   fn synthesize<CS: ConstraintSystem<F>>(self, cs: &mut CS) -> Result<(), SynthesisError> {
-    let root = current_dir().unwrap().join("src/r1cs/circuits/jolt/");
-    let r1cs_path = root.join("jolt.r1cs");
-    let wtns_path = root.join("jolt_js/jolt.wasm");
+    let circuit_dir = std::env::var("CIRCUIT_DIR").expect("failed to read CICUIT_DIR, should've been set by build.rs");
+    let circuit_dir = std::path::PathBuf::from(&circuit_dir);
+    let r1cs_path = circuit_dir.join("jolt.r1cs");
+    let wtns_path = circuit_dir.join("jolt_js/jolt.wasm");
 
     let cfg = CircomConfig::new(wtns_path, r1cs_path.clone()).unwrap();
```
arasuarun, line 33: This is probably not relevant to this build PR but we'd need some way to feed in the trace to the `JoltCircuit` struct. I just made it send in the all 0s for now below. How do you imagine we connect to the other parts like `RVTraceRow` later? We probably need to store the trace as a field of `JoltCircuit` cause it needs to accessed within the `synthesize()` function.

sragss:
self-LGTM.

PR #82, opened 2023-11-09
[WIP] Init RV32IM instruction set

sragss:


sragss:
Moved to #77 bc requires access to `trace.imm`.

PR #83, opened 2023-11-10
updated to new circom format

arasuarun:
Changed the circom file to take inputs with the new format consistent with the vectors produced by the rest of Jolt. Compiler variables to be set are in a function at the top. Currently hardcoded to test that circom compiles, but I can change it in build.rs once we're okay with the design. 

It's currently 35 r1cs constraints btw. Haven't tried to optimize yet but it's still really good for now.

I've tried to add more documentation to the variables and constraints but it'll definitely need more. When you guys have a look, lemme know what's missing or confusing.

There are some minor TODOs: 
1. If regs and RAM are in the same address space, we need to reflect how they're encoded here.
2. Loads require sign-extension which I haven't added yet.

### DIFF
--- a/jolt-core/src/r1cs/circuits/jolt.circom
+++ b/jolt-core/src/r1cs/circuits/jolt.circom
@@ -1,12 +1,21 @@
 pragma circom 2.1.6;
 
-/*  Constants: written as functions because circom.
-*/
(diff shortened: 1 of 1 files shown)


sragss (approved):
LGTM!

PR #84, opened 2023-11-10
feat: add tracing crate

ncitron:
Implements basic tracing in the tracing crate. The crate only exports the `trace` function and the a few types that make up the traces.


sragss:
Safe to assume everything in `tracer/src/emulator` was copied directly and otherwise new for the `tracing` crate?

tracer/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "tracer"
+version = "0.2.0"
+authors = [
+    # author of the original riscv-rust codebase
+    "Takahiro <hogehoge@gachapin.jp>",
+    # authors of the mofications for Jolt
```
sragss, line 7: *modifications

tracer/src/emulator/mmu.rs
```diff
@@ -0,0 +1,1065 @@
...
+  }
+}
+
+impl Mmu {
+  /// Creates a new `Mmu`.
+  ///
+  /// # Arguments
+  /// * `xlen`
+  /// * `terminal`
+  pub fn new(xlen: Xlen, terminal: Box<dyn Terminal>, tracer: Rc<Tracer>) -> Self {
```
sragss, line 91: nit: tracer comment

tracer/src/lib.rs
```diff
@@ -0,0 +1,42 @@
...
+    elf_file.read_to_end(&mut elf_contents).unwrap();
+
+    emulator.setup_program(elf_contents);
+
+    let mut prev_pc = 0;
+    loop {
+        let pc = emulator.get_cpu().read_pc();
+        emulator.tick();
+
+        if prev_pc == pc {
```
sragss, line 29: nit: comment about infinite loop for program completion

tracer/src/trace.rs
```diff
@@ -0,0 +1,107 @@
...
+
+        let mut rows = self.rows.try_borrow_mut().unwrap();
+        let row = rows.last_mut().unwrap();
+        if let Some(rd) = row.instruction.rd {
+            row.register_state.rd_pre_val = Some(reg[rd]);
+        }
+    }
+
+    pub fn capture_post_state(&self, reg: [i64; 32]) {
+        if !*self.open.try_borrow().unwrap() {
```
sragss, line 73: What is the case where this is called and `open == false`?

sragss (approved):

PR #86, opened 2023-11-17
feat: add example programs

ncitron:
Adds example programs for fibonacci and keccak hashing. To build either, run `cargo build --release -p hash` or `cargo build --release -p fibonacci`. The binary elf files should then be in `target/riscv32i-unknown-none-elf/release`.


sragss:
LGTM

PR #87, opened 2023-11-18
feat: common crate

sragss:


ncitron (approved):
LGTM!

One thing that might be a good addition here though is to modify L15 of tracer/src/lib.rs to set the emulator's xlen value based on the XLEN constant?

For the register count, the emulator has it hard coded to 32, with a separate set of registers for the floating point extension. Is the plan here just to have this for convenience but never update it?

Finally what is the register start address reffering to?

sragss:
> One thing that might be a good addition here though is to modify L15 of tracer/src/lib.rs to set the emulator's xlen value based on the XLEN constant?
Done.

> For the register count, the emulator has it hard coded to 32, with a separate set of registers for the floating point extension. Is the plan here just to have this for convenience but never update it?
Yes, I figure we'll use these constants where we can to avoid hardcoding and track these references for validation. I'm sure we won't be perfect and swapping register count / register sizes won't be as easy as updating the contants.

> Finally what is the register start address reffering to?
My thinking is that we'll have all constants associated to our magic register addresses here. Those being [0, 1, ... 31]. Again we won't be perfect but it will centralize the references rather than have magic values.

PR #88, opened 2023-11-18
feat: decode elf files

ncitron:
Adds a `decode` function which decodes an elf file and returns `Vec<common::Instruction>`. One important thing to note is that even though I think the instructions should be in order with no gaps, there is a chance the compiler doesn't do this (there is no requirement that the different text sections be laid out in order without gaps, but in my testing they seem to). I'd just go by the `address` of each instruction rather than the location in the vector. 

I'm happy to make some changes here to make understanding the different sections (rather than concatenating everything) easier. Alternatively we can check for consistency and panic if anything is not laid out nicely, and cross that bridge if it ever comes up.

### DIFF
--- a/tracer/Cargo.toml
+++ b/tracer/Cargo.toml
@@ -17,5 +17,6 @@ edition="2021"
 
 [dependencies]
 fnv = "1.0.7"
-common = { path = "../common" }
 object = "0.32.1"
+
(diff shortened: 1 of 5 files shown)


sragss:
LGTM. The tracing PR will run validation and we can tweak as need be.

PR #91, opened 2023-11-21
Benches + Sumcheck speedup

sragss:
Init benches

~60% sumcheck speedup with parallelism.

### DIFF
--- a/jolt-core/src/lib.rs
+++ b/jolt-core/src/lib.rs
@@ -9,11 +9,11 @@
 pub mod benches;
 pub mod lasso;
 mod msm;
-mod poly;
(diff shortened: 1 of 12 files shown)


jolt-core/src/lib.rs
```diff
@@ -9,11 +9,11 @@
 pub mod benches;
 pub mod lasso;
 mod msm;
-mod poly;
+pub mod poly;
```
sragss, line 12: Not thrilled about all these new exports in order to support benchmarking from `cargo bench`.

moodlezoup (approved):
nice work

PR #96, opened 2023-11-27
Lookup output function

arasuarun:
Given a JoltInstruction object, this function produces the lookup entry. It does so by obtaining each subtable entry using the MLE evaluations (not by materializing the subtables) and then combining them, as before.

I've added a test for each instruction on top of the existing test (which materializes subtables). 

This function will be needed to generate the list of lookup outputs to feed into the R1CS.

### DIFF
--- a/jolt-core/src/jolt/instruction/mod.rs
+++ b/jolt-core/src/jolt/instruction/mod.rs
@@ -9,6 +9,22 @@ pub trait JoltInstruction {
   fn g_poly_degree(&self, C: usize) -> usize;
   fn subtables<F: PrimeField>(&self, C: usize) -> Vec<Box<dyn LassoSubtable<F>>>;
   fn to_indices(&self, C: usize, log_M: usize) -> Vec<usize>;
+  fn lookup_entry<F: PrimeField>(&self, C: usize, M: usize) -> F {
+    let log_M = ark_std::log2(M) as usize;
+
+    let subtable_lookup_indices = self.to_indices(C, ark_std::log2(M) as usize);
+
+    let subtable_lookup_values: Vec<F> = self.subtables::<F>(C)
+      .into_iter()
+      .flat_map(|subtable| {
+          subtable_lookup_indices.iter().map(move |&lookup_index| {
+              subtable.evaluate_mle(&crate::utils::index_to_field_bitvector(lookup_index, log_M))
+          })
+      })
+      .collect();
+
+    self.combine_lookups(&subtable_lookup_values, C, M)
+  }
 }
 
 pub trait Opcode {
--- a/jolt-core/src/jolt/subtable/srl.rs
+++ b/jolt-core/src/jolt/subtable/srl.rs
@@ -74,10 +74,17 @@ impl<F: PrimeField, const CHUNK_INDEX: usize> LassoSubtable<F> for SrlSubtable<F
         0
       };
 
-      let shift_x_by_k = (m..b)
+      // the most significant chunk might be shorter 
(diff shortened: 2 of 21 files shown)


jolt-core/src/jolt/instruction/mod.rs
```diff
@@ -9,6 +9,22 @@ pub trait JoltInstruction {
...
   fn to_indices(&self, C: usize, log_M: usize) -> Vec<usize>;
+  fn lookup_entry<F: PrimeField>(&self, C: usize, M: usize) -> F {
+    let log_M = ark_std::log2(M) as usize;
+
+    let subtable_lookup_indices = self.to_indices(C, ark_std::log2(M) as usize);
+
+    let subtable_lookup_values: Vec<F> = self.subtables::<F>(C)
+      .into_iter()
+      .flat_map(|subtable| {
+          subtable_lookup_indices.iter().map(move |&lookup_index| {
```
sragss, line 20: I believe you can do `.into_iter()` here to avoid the move.

jolt-core/src/jolt/instruction/mod.rs
```diff
@@ -9,6 +9,22 @@ pub trait JoltInstruction {
...
+              subtable.evaluate_mle(&crate::utils::index_to_field_bitvector(lookup_index, log_M))
```
sragss, line 21: `import crate::utils;` at the top then `utils::index_to_field_bitvector`

jolt-core/src/jolt/subtable/srl.rs
```diff
@@ -74,10 +74,17 @@ impl<F: PrimeField, const CHUNK_INDEX: usize> LassoSubtable<F> for SrlSubtable<F
...
+        b - ((b * (CHUNK_INDEX + 1)) - MAX_SHIFT)
+      } else {
+        b
+      };
+
+      let shift_x_by_k = (m..this_chunk_length)
         .enumerate()
-        .map(|(_, j)| F::from(1_u64 << (b * CHUNK_INDEX + j - k)) * x[b - 1 - j])
-        .fold(F::zero(), |acc, val| acc + val);
+        .map(|(_, j)| F::from((2_u64.pow((b * CHUNK_INDEX + j - k) as u32))) * x[b - 1 - j])
```
sragss, line 86: Aren't these equivalent?

jolt-core/src/jolt/subtable/srl.rs
```diff
@@ -74,10 +74,17 @@ impl<F: PrimeField, const CHUNK_INDEX: usize> LassoSubtable<F> for SrlSubtable<F
         0
       };
 
-      let shift_x_by_k = (m..b)
+      // the most significant chunk might be shorter 
+      let this_chunk_length = if (b * (CHUNK_INDEX + 1)) > MAX_SHIFT {
```
sragss, line 78: Let's call this `chunk_length`

sragss (commented):
Some nits, otherwise LGTM.

jolt-core/src/jolt/subtable/srl.rs
arasuarun: Yeah the original is cleaner. I'll revert to that.

jolt-core/src/jolt/instruction/mod.rs
arasuarun: (Blaming copilot on that one). 

But yeah, I tried something else that seems to work without having to use move or clone. The tests pass so lemme know if it looks okay.

PR #99, opened 2023-11-28
SLTU lookups to prove validity of read timestamps

moodlezoup:


jolt-core/src/jolt/vm/mod.rs
```diff
@@ -119,7 +126,15 @@ pub trait Jolt<F: PrimeField, G: CurveGroup<ScalarField = F>, const C: usize, co
...
 
-    todo!("Lasso lookups to enforce timestamp validity")
+    let timestamp_validity_lookups: Vec<SLTUInstruction> = read_timestamps
+      .iter()
+      .enumerate()
+      .map(|(i, &ts)| SLTUInstruction(ts, i as u64 + 1))
+      .collect();
+
+    let timestamp_validity_proof =
+      <Surge<F, G, SLTUInstruction, 1, MAX_TRACE_SIZE>>::new(timestamp_validity_lookups)
```
arasuarun, line 136: I think we need `c=2` here. If each `read_timestamp` is 22 bits long, we also need to treat each global time `i` as being 22 bits. Having c=1 would make the subtables too large (44 bits) but c=2 should work. Does this sound right?

jolt-core/src/jolt/vm/mod.rs
```diff
@@ -119,7 +126,15 @@ pub trait Jolt<F: PrimeField, G: CurveGroup<ScalarField = F>, const C: usize, co
       &mut random_tape,
     );
 
-    todo!("Lasso lookups to enforce timestamp validity")
+    let timestamp_validity_lookups: Vec<SLTUInstruction> = read_timestamps
+      .iter()
+      .enumerate()
+      .map(|(i, &ts)| SLTUInstruction(ts, i as u64 + 1))
```
arasuarun, line 132: It should also work with `ceil(i / 7)` here, which is when we increment the counter only once per CPU step.

PR #104, opened 2023-11-30
Finalize trace validation

sragss:


jolt-core/src/jolt/trace/rv.rs
```diff
@@ -345,13 +369,33 @@ impl RVTraceRow {
       ensure!(self.memory_bytes_before.is_some(), "Line {}: memory_bytes_before is None", line!());
       ensure!(self.memory_bytes_after.is_none(), "Line {}: memory_bytes_after is not None", line!());
       ensure!(self.memory_bytes_before.as_ref().unwrap().len() == size, "Line {}: memory_bytes_before length does not match size", line!());
+
+      let mut rd = self.rd_post_val.unwrap();
+      let mask = 0b11111111;
+      for i in 0..size {
+        let expected_byte = (rd & mask) as u8;
```
moodlezoup, line 376: I don't think you need the mask here, `as u8` should suffice

jolt-core/src/jolt/trace/rv.rs
```diff
@@ -345,13 +369,33 @@ impl RVTraceRow {
...
       ensure!(self.memory_bytes_before.is_some(), "Line {}: memory_bytes_before is None", line!());
       ensure!(self.memory_bytes_after.is_some(), "Line {}: memory_bytes_after is None", line!());
       ensure!(self.memory_bytes_before.as_ref().unwrap().len() == size, "Line {}: memory_bytes_before length does not match size", line!());
       ensure!(self.memory_bytes_after.as_ref().unwrap().len() == size, "Line {}: memory_bytes_after length does not match size", line!());
+
+      // Check that memory_bytes_after == rs2
+      let mut store_val = self.rs2_val.unwrap();
+      let mask = 0b11111111;
+      for i in 0..size {
+        let expected_byte = (store_val & mask) as u8;
```
moodlezoup, line 394: same here

moodlezoup (approved):
looks good, just two nits
should we consider pulling the trace validation stuff into separate trait/file? `rv.rs` is getting pretty big

jolt-core/src/jolt/trace/rv.rs
sragss: done.

PR #109, opened 2023-12-06
[JOLT-73] Load bytecode to memory on init

moodlezoup:
Also closes JOLT-69

### DIFF
--- a/jolt-core/src/jolt/vm/read_write_memory.rs
+++ b/jolt-core/src/jolt/vm/read_write_memory.rs
@@ -103,9 +103,10 @@ impl<F: PrimeField, G: CurveGroup<ScalarField = F>> ReadWriteMemory<F, G> {
             .unwrap_or(0);
         let max_bytecode_address = bytecode
             .iter()
-            .map(|instr| remap_address(instr.address + 3))
+            .map(|instr| remap_address(instr.address))
             .max()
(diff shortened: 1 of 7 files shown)


jolt-core/src/jolt/vm/read_write_memory.rs
```diff
@@ -71,17 +75,54 @@ where
...
+            .iter()
+            .map(|op| match op {
+                MemoryOp::Read(a, _) => remap_address(*a),
+                MemoryOp::Write(a, _) => remap_address(*a),
+            })
+            .max()
+            .unwrap_or(0);
+        let max_bytecode_address = bytecode
+            .iter()
+            .map(|instr| remap_address(instr.address + 3))
```
sragss, line 106: Could we move the + 3 to after the max and add a comment along the lines: "For RV32, 4 bytes of memory addressable raw instructions, max address of bytecode would be instruction + 3."

jolt-core/src/jolt/vm/read_write_memory.rs
```diff
@@ -71,17 +75,54 @@ where
...
+            .max()
+            .unwrap_or(0);
```
sragss, line 108: Can just unwrap here, should fail if we can't find a max.

sragss (approved):
LGTM!

PR #112, opened 2023-12-07
Reduce DensePoly commit time

sragss:


sragss:
Moved into `sragss/accelerate-3`.

PR #117, opened 2023-12-08
Move fib_e2e to rv32i_vm and fix bytecode preprocessing

moodlezoup:


jolt-core/src/jolt/vm/bytecode.rs
```diff
@@ -174,15 +173,14 @@ impl<F: PrimeField, G: CurveGroup<ScalarField = F>> BytecodePolynomials<F, G> {
...
 
         // Preprocessing should deal with padding.
         assert!(is_power_of_two(bytecode.len()));
         assert!(is_power_of_two(trace.len()));
 
         let num_ops = trace.len().next_power_of_two();
-        let code_size = bytecode.len().next_power_of_two();
-
-        println!("BytecodePolynomials::new num_ops {num_ops} code_size {code_size}");
+        let code_size = (max_bytecode_address + 1).next_power_of_two();
```
sragss, line 183: Let's stick with pre-processing (`Self::preprocess` above) doing all padding and not adjust trace or num_ops sizes here implicitly. (I made this issue in an earlier PR)

jolt-core/src/jolt/trace/rv.rs
sragss, line 1: Can we handle JOLT(71) using your new constants? Line 337 suggests we should just assert the memory addresses are greater than RAM_START_ADDRESS

jolt-core/src/jolt/vm/bytecode.rs
```diff
@@ -238,8 +233,21 @@ impl<F: PrimeField, G: CurveGroup<ScalarField = F>> BytecodePolynomials<F, G> {
...
+        for instruction in trace.iter_mut() {
+            assert!(instruction.address >= RAM_START_ADDRESS as usize);
+            assert!(instruction.address % BYTES_PER_INSTRUCTION == 0);
+            instruction.address -= RAM_START_ADDRESS as usize;
+            instruction.address /= BYTES_PER_INSTRUCTION;
+        }
+
         // Bytecode: Add single no_op instruction at adddress | ELF + 1 |
-        let no_op_address = bytecode.last().unwrap().address + ADDRESS_INCREMENT;
+        let no_op_address = bytecode.last().unwrap().address + 1;
```
sragss, line 250: We make the no_op_address necessarily expensive to commit to by putting it at bytecode + 1, rather than one of the early register slots. We can handle this later with a smarter leaf hashing / committing algo or make it smaller here.

No need to handle in this PR but could we add a linear issue? See JOLT(CachedHasher) for some context.

jolt-core/src/jolt/vm/bytecode.rs
```diff
@@ -238,8 +233,21 @@ impl<F: PrimeField, G: CurveGroup<ScalarField = F>> BytecodePolynomials<F, G> {
     }
 
     fn preprocess(bytecode: &mut Vec<ELFRow>, trace: &mut Vec<ELFRow>) {
+        for instruction in bytecode.iter_mut() {
+            assert!(instruction.address >= RAM_START_ADDRESS as usize);
+            assert!(instruction.address % BYTES_PER_INSTRUCTION == 0);
```
sragss, line 238: Do this need to be spaced by`BYTES_PER_INSTRUCTION`? Could we make them sequential to get some cost savings?

(Don't need to handle here, but good to discuss / track)

jolt-core/src/jolt/vm/bytecode.rs
moodlezoup: this isn't pre-processing; the `+ 1` here is because the bytecode addresses are 0-indexed

moodlezoup: I'll add a comment

jolt-core/src/jolt/vm/bytecode.rs
moodlezoup: it's handled two lines below:
```
instruction.address /= BYTES_PER_INSTRUCTION;
```

jolt-core/src/jolt/vm/bytecode.rs
moodlezoup: JOLT-85

jolt-core/src/jolt/trace/rv.rs
moodlezoup: let's address in a follow-up PR

PR #121, opened 2023-12-15
Generating Spartan proofs in VM module using R1CS module 

arasuarun:
Adds the `prove_r1cs` function in the Jolt trait. Able to generate proofs using the

### DIFF
--- a/common/src/field_conversion.rs
+++ b/common/src/field_conversion.rs
@@ -2,11 +2,16 @@ use ark_bn254::Fr as ArkFr;
 use ark_ff::{fields::PrimeField as ArkPrimeField, BigInteger};
 
 use ff::PrimeField as GenericPrimeField;
-use spartan2::provider::bn256_grumpkin::bn256::Base as Spartan2Fr;
+use spartan2::provider::bn256_grumpkin::bn256::Scalar as Spartan2Fr;
 
-pub fn ark_to_spartan(ark: ArkFr) -> Spartan2Fr {
-    let bigint = ark.into_bigint();
-    Spartan2Fr::from_raw(bigint.0)
+pub fn ark_to_spartan<ArkF: ArkPrimeField>(ark: ArkF) -> Spartan2Fr {
+    let bigint: <ArkF as ArkPrimeField>::BigInt = ark.into_bigint();
+    let bytes = bigint.to_bytes_le();
+    let mut array = [0u64; 4];
+    for (i, chunk) in bytes.chunks(8).enumerate() {
+        array[i] = u64::from_le_bytes(chunk.try_into().unwrap());
+    }
+    Spartan2Fr::from_raw(array)
 }
 
 pub fn spartan_to_ark(bell: Spartan2Fr) -> ArkFr {
--- a/jolt-core/src/jolt/trace/rv.rs
+++ b/jolt-core/src/jolt/trace/rv.rs
@@ -181,15 +181,6 @@ impl RVTraceRow {
         }
     }
 
-    // Arasu: does the same as above but pre-processes the imm 
-    pub fn from_common_r1cs(common: common::RVTraceRow) -> Self {
-        let mut res = Self::from_common(common); 
-        res.imm = match res.imm_u64() {
-            (imm) => Some(imm as u32),
-        };
-        res 
-    }
-
     fn RType(
         pc: u64,
         opcode: RV32IM,
--- a/jolt-core/src/jolt/trace/rv.rs
+++ b/jolt-core/src/jolt/trace/rv.rs
@@ -869,7 +869,7 @@ impl JoltProvableTrace for RVTraceRow {
         // 11: Instruction asserts lookup output as false
         // 12: Instruction asserts lookup output as true
         // 13: Sign-bit of imm
-        // 14: Instruction is lui
+        // 14: Is concat (Note: used to be is_lui)
 
         let mut flags = vec![false; 15];
 
@@ -971,8 +971,18 @@ impl JoltProvableTrace for RVTraceRow {
             _ => false,
         };
 
+        // flags[14] = match self.opcode {
+        //     RV32IM::LUI => true,
+        //     _ => false,
+        // };
+
         flags[14] = match self.opcode {
-            RV32IM::LUI => true,
+            RV32IM::XOR | RV32IM::XORI | RV32IM::OR | RV32IM::ORI | RV32IM::AND | RV32IM::ANDI | 
+            RV32IM::SLL | RV32IM::SRL | RV32IM::SRA | RV32IM::SLLI | RV32IM::SRLI | RV32IM::SRAI |
+            RV32IM::SLT | RV32IM::SLTU | RV32IM::SLTI | RV32IM::SLTIU | 
+            RV32IM::BEQ | RV32IM::BNE | RV32IM::BLT | RV32IM::BGE | 
+            RV32IM::BLTU | RV32IM::BGEU 
+            => true, 
             _ => false,
         };
 
--- a/jolt-core/src/jolt/trace/rv.rs
+++ b/jolt-core/src/jolt/trace/rv.rs
@@ -181,6 +181,15 @@ impl RVTraceRow {
         }
     }
 
+    // Arasu: does the same as above but pre-processes the imm 
+    pub fn from_common_r1cs(common: common::RVTraceRow) -> Self {
+        let mut res = Self::from_common(common); 
+        res.imm = match res.imm_u64() {
+            (imm) => Some(imm as u32),
+        };
+        res 
+    }
+
     fn RType(
         pc: u64,
         opcode: RV32IM,
@@ -686,9 +695,11 @@ impl JoltProvableTrace for RVTraceRow {
       RV32IM::BGE  => vec![BGEInstruction(self.rs1_val.unwrap(), self.rs2_val.unwrap()).into()],
       RV32IM::BGEU => vec![BGEUInstruction(self.rs1_val.unwrap(), self.rs2_val.unwrap()).into()],
 
-      RV32IM::JAL  => vec![JALInstruction(self.pc, self.imm_u64()).into()],
-      RV32IM::JALR => vec![JALRInstruction(self.rs1_val.unwrap(), self.imm_u64()).into()],
+    //   RV32IM::JAL  => vec![JALInstruction(self.pc, self.imm_u64()).into()],
+    //   RV32IM::JALR => vec![JALRInstruction(self.rs1_val.unwrap(), self.imm_u64()).into()],
 
+      RV32IM::JAL  => vec![ADDInstruction::<32>(self.pc, self.imm_u64()).into()],
+      RV32IM::JALR => vec![ADDInstruction::<32>(self.rs1_val.unwrap(), self.imm_u64()).into()],
       RV32IM::AUIPC => vec![ADDInstruction::<32>(self.pc, self.imm_u64()).into()],
 
       _ => vec![]
@@ -870,11 +881,14 @@ impl JoltProvableTrace for RVTraceRow {
         // 12: Instruction asserts lookup output as true
         // 13: Sign-bit of imm
         // 14: Is concat (Note: used to be is_lui)
+        // Arasu: Extra to get things working
+        // 15: is lui or auipc
+        // 16: is jal
 
-        let mut flags = vec![false; 15];
+        let mut flags = vec![false; 17];
 
         flags[0] = match self.opcode {
-            RV32IM::JAL | RV32IM::JALR | RV32IM::LUI | RV32IM::AUIPC => true,
+            RV32IM::JAL | RV32IM::LUI | RV32IM::AUIPC => true,
             _ => false,
         };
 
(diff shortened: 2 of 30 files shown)


jolt-core/src/jolt/trace/rv.rs
```diff
@@ -686,9 +686,11 @@ impl JoltProvableTrace for RVTraceRow {
       RV32IM::BGE  => vec![BGEInstruction(self.rs1_val.unwrap(), self.rs2_val.unwrap()).into()],
       RV32IM::BGEU => vec![BGEUInstruction(self.rs1_val.unwrap(), self.rs2_val.unwrap()).into()],
 
-      RV32IM::JAL  => vec![JALInstruction(self.pc, self.imm_u64()).into()],
-      RV32IM::JALR => vec![JALRInstruction(self.rs1_val.unwrap(), self.imm_u64()).into()],
+    //   RV32IM::JAL  => vec![JALInstruction(self.pc, self.imm_u64()).into()],
+    //   RV32IM::JALR => vec![JALRInstruction(self.rs1_val.unwrap(), self.imm_u64()).into()],
 
+      RV32IM::JAL  => vec![ADDInstruction::<32>(self.pc, self.imm_u64()).into()],
```
sragss, line 692: Why did we swap these out for ADD?

jolt-core/src/jolt/vm/bytecode.rs
```diff
@@ -82,6 +84,45 @@ impl ELFRow {
...
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+        ).to_circuit_flags();
+
+        // convert each element of circuit_flags to a u8 cause they must be 0 or 1
```
sragss, line 104: Nit: rm comment

jolt-core/src/jolt/vm/bytecode.rs
```diff
@@ -175,6 +216,44 @@ impl<F: PrimeField> FiveTuplePoly<F> {
             self.imm.evaluate(r),
         ]
     }
+
+    fn from_elf_r1cs(elf: &Vec<ELFRow>, N_SKIP: usize) -> Vec<F> {
+        // Do not pad.
```
sragss, line 221: Comment -> "Padding should be handled elsewhere."

Let's also assert the length is a power of 2 here.

jolt-core/src/jolt/vm/bytecode.rs
```diff
@@ -175,6 +216,44 @@ impl<F: PrimeField> FiveTuplePoly<F> {
             self.imm.evaluate(r),
         ]
     }
+
+    fn from_elf_r1cs(elf: &Vec<ELFRow>, N_SKIP: usize) -> Vec<F> {
```
sragss, line 220: Nit: Comment about why N_SKIP exists.

common/src/field_conversion.rs
```diff
@@ -2,11 +2,16 @@ use ark_bn254::Fr as ArkFr;
...
 
 use ff::PrimeField as GenericPrimeField;
-use spartan2::provider::bn256_grumpkin::bn256::Base as Spartan2Fr;
-
-pub fn ark_to_spartan(ark: ArkFr) -> Spartan2Fr {
-    let bigint = ark.into_bigint();
-    Spartan2Fr::from_raw(bigint.0)
+use spartan2::provider::bn256_grumpkin::bn256::Scalar as Spartan2Fr;
+
+pub fn ark_to_spartan<ArkF: ArkPrimeField>(ark: ArkF) -> Spartan2Fr {
```
sragss, line 7: Does this work for the generic version? `ark_to_ff`

If so let's convert that one as well.

jolt-core/src/jolt/vm/bytecode.rs
```diff
@@ -175,6 +216,44 @@ impl<F: PrimeField> FiveTuplePoly<F> {
...
+            opcodes.push(F::from(row.opcode));
+            rds.push(F::from(row.rd));
+            rs1s.push(F::from(row.rs1));
+            rs2s.push(F::from(row.rs2));
+            imms.push(F::from(row.imm));
+            // circuit_flags.push(row.circuit_flags_packed::<F>());
+        }
+
+        // skip the first N_SKIP elements of each vector 
+        let opcodes = opcodes.iter().skip(N_SKIP).cloned().collect::<Vec<_>>();
```
sragss, line 242: Clones are very expensive (especially when the vectors are this large. Let's instead deal with this in the for loop above and never add the first `N_SKIP` elements to the vectors.

jolt-core/src/jolt/trace/rv.rs
```diff
@@ -930,14 +937,19 @@ impl JoltProvableTrace for RVTraceRow {
...
             | RV32IM::JALR
             | RV32IM::LUI => false,
             _ => true,
         };
 
         flags[7] = match self.opcode {
-            RV32IM::ADD | RV32IM::ADDI | RV32IM::JAL | RV32IM::JALR | RV32IM::AUIPC => true,
+            RV32IM::ADD 
+            | RV32IM::ADDI 
+            // Arasu: don't count these here.
```
sragss, line 949: Nit: Make this comment `TODO(github-handle)` that way we can `cmd + F` these later and ensure they get converted to tracking bugs (linear) or removed. We don't want comments directed at an individual at time of publish.

jolt-core/src/jolt/trace/rv.rs
```diff
@@ -870,12 +871,15 @@ impl JoltProvableTrace for RVTraceRow {
         // 11: Instruction asserts lookup output as false
         // 12: Instruction asserts lookup output as true
         // 13: Sign-bit of imm
-        // 14: Instruction is lui
+        // 14: Is concat (Note: used to be is_lui)
+        // Arasu: Extra to get things working
```
sragss, line 875: Nit: Same as next comment – `TODO(github-handle)`

jolt-core/src/jolt/trace/rv.rs
```diff
@@ -972,8 +984,28 @@ impl JoltProvableTrace for RVTraceRow {
             _ => false,
         };
 
+        // flags[14] = match self.opcode {
```
sragss, line 987: Rm

jolt-core/src/jolt/vm/mod.rs
```diff
@@ -186,8 +188,114 @@ pub trait Jolt<F: PrimeField, G: CurveGroup<ScalarField = F>, const C: usize, co
...
+        instructions: Vec<Self::InstructionSet>,
+        mut bytecode_rows: Vec<ELFRow>,
+        mut trace: Vec<ELFRow>,
+        bytecode: Vec<ELFInstruction>,
+        memory_trace: Vec<MemoryOp>,
+        circuit_flags: Vec<F>,
+        transcript: &mut Transcript,
+        random_tape: &mut RandomTape<G>,
+    ) {
+        let N_SKIP = 3;  // the instructions at the beginning of the trace to be skipped
```
sragss, line 201: Nit: Comment on why

jolt-core/src/jolt/vm/mod.rs
```diff
@@ -186,8 +188,114 @@ pub trait Jolt<F: PrimeField, G: CurveGroup<ScalarField = F>, const C: usize, co
...
+        mut bytecode_rows: Vec<ELFRow>,
+        mut trace: Vec<ELFRow>,
+        bytecode: Vec<ELFInstruction>,
+        memory_trace: Vec<MemoryOp>,
+        circuit_flags: Vec<F>,
+        transcript: &mut Transcript,
+        random_tape: &mut RandomTape<G>,
+    ) {
+        let N_SKIP = 3;  // the instructions at the beginning of the trace to be skipped 
+        let N_FLAGS = 17;
```
sragss, line 202: Nit: Consts here

jolt-core/src/jolt/vm/mod.rs
```diff
@@ -186,8 +188,114 @@ pub trait Jolt<F: PrimeField, G: CurveGroup<ScalarField = F>, const C: usize, co
...
+        memreg_a_rw.drain(..N_SKIP * 7);
+        memreg_v_reads.drain(..N_SKIP * 7);
+        memreg_v_writes.drain(..N_SKIP * 7);
+
+        let (chunks_x, chunks_y): (Vec<F>, Vec<F>) = 
+            instructions
+            .iter()
+            .flat_map(|op| {
+                let chunks_xy = op.operand_chunks(C, log_M);
+                let chunks_x = chunks_xy[0].clone();
```
sragss, line 235: Clones are likely to be expensive. You can unpack directly: `let [chunks_x, chunks_y] = op.operand_chunks(C, log_M);`

jolt-core/src/jolt/vm/mod.rs
```diff
@@ -186,8 +188,114 @@ pub trait Jolt<F: PrimeField, G: CurveGroup<ScalarField = F>, const C: usize, co
...
+
+        let instructions = &instructions[N_SKIP..];
+        let circuit_flags = &circuit_flags[N_FLAGS * N_SKIP..];
+
+        let [mut prog_a_rw, mut prog_v_rw, prog_t_reads] = 
+            BytecodePolynomials::<F, G>::r1cs_polys_from_bytecode(bytecode_rows, trace, N_SKIP);
+        prog_v_rw.extend(circuit_flags
+            .chunks(N_FLAGS)
+            .map(|x| {
+                let mut packed = F::ZERO;
```
sragss, line 215: Nit: Use `F::zero()` to be consistent with rest of codebase.

jolt-core/src/jolt/vm/mod.rs
```diff
@@ -186,8 +188,114 @@ pub trait Jolt<F: PrimeField, G: CurveGroup<ScalarField = F>, const C: usize, co
...
+        let TRACE_LEN = trace.len()-N_SKIP;
+
+        let log_M = log2(M) as usize;
+
+        let instructions = &instructions[N_SKIP..];
+        let circuit_flags = &circuit_flags[N_FLAGS * N_SKIP..];
+
+        let [mut prog_a_rw, mut prog_v_rw, prog_t_reads] = 
+            BytecodePolynomials::<F, G>::r1cs_polys_from_bytecode(bytecode_rows, trace, N_SKIP);
+        prog_v_rw.extend(circuit_flags
```
sragss, line 212: I'm a bit unclear whats happening here – could you add a comment on why we're raising the flags to 2^(1-flag_index)

jolt-core/src/jolt/vm/mod.rs
```diff
@@ -186,8 +188,114 @@ pub trait Jolt<F: PrimeField, G: CurveGroup<ScalarField = F>, const C: usize, co
...
+            memreg_v_reads, 
+            memreg_v_writes,
+            chunks_x, 
+            chunks_y,
+            chunks_query,
+            lookup_outputs,
+            circuit_flags.to_vec(),
+        ];
+
+        // TODO: move this conversion to the r1cs module
```
sragss, line 278: Let's do this before merging.

jolt-core/src/r1cs/circuits/jolt.circom
```diff
@@ -1,24 +1,22 @@
 pragma circom 2.1.6;
 
 /* Compiler Variables */
```
sragss, line 3: Nit: Let's have a comment on what each of these are and why they exist.

tracer/src/emulator/cpu.rs
```diff
@@ -1818,7 +1819,9 @@ fn normalize_is_imm(value: i64) -> u32 {
 }
 
 fn normalize_b_imm(value: u64) -> u32 {
-    ((value as i32) >> 1) as u32
+    // TODO: Hack – value should be unnormalized in the tracer
```
sragss, line 1822: Nit: `TODO(github-handle): Hack`

jolt-core/src/jolt/trace/rv.rs
arasuarun: Turns out I misunderstood the semantics on the Jumps. It's sufficient to just use ADDs there. I think it's safe to remove the JAL and JALR instructions, but we could wait till later just to make sure.

jolt-core/src/jolt/vm/bytecode.rs
arasuarun: We don't need it to be a power of 2 for the R1CS polys. In fact, things get complicated when padding. 

We might need it when we feed polys directly into Spartan. but I'll just delete the comment for now.

jolt-core/src/jolt/vm/mod.rs
arasuarun: Changed it here but wasn't possible in parts of `r1cs/snark.rs` though. Might be cause that's ff::PrimeField.

jolt-core/src/jolt/vm/bytecode.rs
sragss: Can you provide more detail on the complication?

All of the polys that are derived from Lasso things are already powers of 2. So I think we should just assert that all of them are a power of 2 length as to avoid future bugs.

common/src/field_conversion.rs
arasuarun: Yup. Works with that conversion function.

jolt-core/src/jolt/vm/bytecode.rs
arasuarun: Padding will make all the witnesses 0, which isn't a satisfying assignment for the R1CS. We can work around it and modify the R1CS constraints to "turn off" when we want it to but that'll take some effort. I think padding is also internally in Spartan so we'll have to revisit this when plugging polynomial commitments directly from Jolt to Spartan.

PR #122, opened 2023-12-17
Remove batched polynomials

sragss:


sragss:
Rather than plumb through a bunch of lifetimes, let's just leave parent as the holder of `Vec<DensePolynomial<F>>` then have a batch commit, batch open which takes a `&Vec<DensePolynomial<F>>` rather than `&DensePolynomial`. This way no cloning for merges, no lifetime plubming.

sragss:
Still needs a rework of `CombinedTableEvalProof`. I'm thinking the attempt to use `Vec<&DensePolynomial>` rather than `CombinedPolynomial<&a'> { Vec<&a', DensePolynomial<F>> }` was a mistake. 

Would be much cleaner to implement these functions on a `CombinedPolynomial`, likely could create a joint iterator over `DensePolynomial.evals_ref()`.

sragss:
Another alternative here (with Prover / Verifier implications) is that we commit separately and do batched opening proofs.

moodlezoup:
Now that we're committing to the majority of polynomials individually, the efficiency gains from this are marginal

PR #127, opened 2023-12-19
feat: Sum Trace fmt

sragss:


jolt-core/src/benches/sum_timer.rs
```diff
@@ -0,0 +1,107 @@
...
+        if let Some(span) = ctx.span(id) {
+            span.extensions_mut().insert(Instant::now());
+        }
+    }
+
+    fn on_exit(&self, id: &tracing::span::Id, ctx: Context<'_, S>) {
+        if let Some(span) = ctx.span(id) {
+            let span_name = span.name().to_string();
+            if let Some(start_time) = span.extensions_mut().remove::<Instant>() {
+                let duration = start_time.elapsed().as_millis();
```
moodlezoup, line 97: nit: store duration in micros and nanos, but still print out the cumulative numbers in millis (with a couple decimal places of precision)
otherwise I think we'll be underestimating stuff like `eval_loop`, `binding`

moodlezoup (approved):
nit: "sum" => "cumulative"

jolt-core/src/benches/sum_timer.rs
sragss: Yup, good call.

PR #128, opened 2023-12-20
Easy prove_memory optimizations

moodlezoup:


jolt-core/src/jolt/vm/read_write_memory.rs
```diff
@@ -499,66 +513,52 @@ where
 
     #[tracing::instrument(skip_all, name = "ReadWriteMemory::read_leaves")]
     fn read_leaves(&self, polynomials: &Self, gamma: &F, tau: &F) -> Vec<DensePolynomial<F>> {
```
sragss, line 515: We can probably go further and make these "no_op optimized". Not necessary for this PR though.

sragss (approved):
LGTM

PR #131, opened 2023-12-23
Optimize/sumcheck

moodlezoup:
Sprinkling rayon in a few more places

### DIFF
--- a/jolt-core/src/subprotocols/sumcheck.rs
+++ b/jolt-core/src/subprotocols/sumcheck.rs
@@ -12,6 +12,7 @@ use ark_ff::PrimeField;
 use ark_serialize::*;
 use ark_std::One;
 use merlin::Transcript;
+use rayon::prelude::*;
 
 #[cfg(feature = "ark-msm")]
 use ark_ec::VariableBaseMSM;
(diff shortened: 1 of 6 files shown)


jolt-core/src/subprotocols/sumcheck.rs
```diff
@@ -416,13 +395,11 @@ impl<F: PrimeField> SumcheckInstanceProof<F> {
                     let len = poly_A.len() / 2;
 
                     // In the case of a flagged tree, the majority of the leaves will be 1s, optimize for this case.
-                    let (eval_point_0, eval_point_2, eval_point_3) = (0..len).into_iter()
+                    let (eval_point_0, eval_point_2, eval_point_3) = (0..len).into_par_iter()
```
sragss, line 398: I think we want to pick either parallelism across the batch or parallelism across the MLE, but not both. When I benched this I believe that parallelism across the batch performed better but it was when we had the randomness bug meaning the trees were *vast* majority 1s.

sragss (approved):
One comment on batched sumcheck parallelism. Otherwise LGTM.

jolt-core/src/subprotocols/sumcheck.rs
moodlezoup: in my testing, I found that using parallelism in both places had the best performance (in both isolated sumcheck benchmarks and e2e-except-r1cs)

PR #133, opened 2023-12-28
feat: Sparse GKR

sragss:
Create a version of the Read / Write multi-set hash evaluation for Jolt which exploits the fact that these GKR trees have a high probability of evaluating to one at each multiplication gate.

The bottom layer of the Read / Write multi-set hash of Jolt looks as follows:
```
leaf[i] = hash(address[i], value[i], timestamp[i])
flagged_leaf[i] = flag[i] * leaf[i] + (1-flag[i])
```
`flag[i]` is non-zero for a 1-2 subtables per instruction. Assuming roughly 30 subtables this means that `P(flagged_leaf[i] == 1) ~= 1 - 1/30 = 0.97`. 

The GKR tree then does a uniform tree-ed multiplication circuit over `flagged_leaf[i]`.

Representing each of these leaves with 256 bits is hugely wasteful. After the first layer `P(flagged_leaf_layer_1[i] == 1) = 0.97^2 = 0.94`. You can see how this progresses.

This PR represents the polynomials more intelligently by not storing the `1` values and does computations over these sparse polynomials more intelligently.

There are three relevant algorithms this speeds up:
1. GKR tree construction (multiply out GKR trees per memory)
2. Sumcheck evaluation loop (eval[0, 2, 3])
3. Polynomial binding (`SparsePoly::bound_poly_var_top`)


sragss:
There are two problems with c3dfc8dfc962e6bc547605a3bd03f9ff9ad3f84b which stores the indices separately:
a) indices is `poly.len()` sized, regardless of sparsity, taking up space and costing iterations
b) we must point to `self.indices[i]` / `self.indices[i + self.mid]` / `self.entries[sparse_i]` / `self.entries[sparse_i + self.mid]` simultaneously, all in different locations in memory

I believe that 66cc9e4636095e64814ade7f2a30adbaec725167 was 2x faster due to the fact that 2 `self.indices` pointers were not required, thus more cache hits on page on the processing we actually need (non-sparse `F`)

If this theory is true, we could theoretically bring all of the required items (poly[low], poly[high], low) for the 3 algos
- GKR tree construction
- Poly evaluation (0,2,3)
- Poly binding (r)
into the same area of memory.

Here's a shot at this datastructure:
```
impl<F> SparsePair<F> {
    pub fn new(low_index: usize, low_val: Option<F>, high_val: Option<F>) -> Self {
        Self {
            low_index,
            low_val,
            high_val
        }
    }
}

pub struct PairedSparsePoly<F> {
    entries: Vec<SparsePair<F>>,
    num_vars: usize,
    len: usize
}

impl<F: PrimeField> PairedSparsePoly<F> {
    pub fn new(entries: Vec<SparsePair<F>>, num_vars: usize) -> Self {
        let len = 1 << num_vars;
        assert!(entries.len() < len);
        Self { entries, num_vars, len }
    }

    /// Does not need to parallelize due to always used in a batch.
    pub fn bound_poly_var_top(&mut self, r: &F) {
        let mut new_entries = Vec::with_capacity(self.entries.len()); // TODO(sragss): Overshoots: Prune by the count of left only

        self.entries.into_iter().map(|entry| {
            let new_entry = match (entry.low_val, entry.high_val) {
                (Some(low_val), None) => {
                    let m = F::one() - low_val;
                    m * r
                },
                (None, Some(high_val)) => {
                    let m = high_val - F::one();
                    F::one() + m * r
                },
                (Some(low_val), Some(high_val)) => {
                    let m = high_val - low_val;
                    low_val + m * r
                },
                (None, None) => {
                    panic!("Shouldn't be any dual sparse indices here.")
                }
            };
            // TODO(sragss): Construct the `SparsePair`
           let is_high = entry.low_index > self.len / 4;
           let new_low_index = is_high ? entry.low_index - half : entry.low_index;
            new_entries.push(new_entry);
        });

        self.entries = new_entries;
        self.num_vars -= 2;
        self.len /= 2;
    }
}
```

The primary issue is that after binding we aren't guaranteed an ordering any more. This would mean that the access pattern to the `EqPolynomial::evals: Vec<F>` during the `cubic_sumcheck` would be random.

PR #136, opened 2024-01-03
Synthesize constraints using a single step (circom) r1cs circuit 

arasuarun:
The circom r1cs file is only for a single step of r1cs (and not a loop for all NUM_STEPS). The looping is instead handed in the synthesize function in `r1cs/snark.rs`. 

The FIbonacci example runs and verifies. Need to do some slight tweaks to ensure correctness, but the approach is sane. Will make it a proper PR when ready. 

We also need to modify the compilation pipeline now as the circom circuits doesn't need to know the PC_START_ADDR or NUM_STEPS. These only need be set in the jolt_circuit object created in prove_r1cs, which should be much easier.


jolt-core/src/jolt/vm/rv32i_vm.rs
```diff
@@ -321,11 +321,151 @@ mod tests {
...
+            .flat_map(|row| row.to_jolt_instructions())
+            .collect();
+
+        let instructions_r1cs: Vec<RV32I> = converted_trace
+            .clone()
+            .into_iter()
+            .flat_map(|row| {
+                let instructions = row.to_jolt_instructions();
+                if instructions.is_empty() {
+                    vec![ADDInstruction::<32>(0_u64, 0_u64).into()]
```
sragss, line 433: Let's move this into RV32I::to_jolt_instructions so that all of this logic is in one place. Let's also make an explicit `RV32I::no_op() -> RV32I` which creates this instruction.

jolt-core/src/jolt/vm/rv32i_vm.rs
```diff
@@ -321,11 +321,151 @@ mod tests {
...
+            .map(|common| RVTraceRow::from_common(common))
+            .collect();
+
+        let bytecode_trace: Vec<ELFRow> = converted_trace
+            .iter()
+            .map(|row| row.to_bytecode_trace())
+            .collect();
+
+        let instructions: Vec<RV32I> = converted_trace
+            .clone()
```
sragss, line 422: I believe we can remove these clones.

jolt-core/src/jolt/vm/rv32i_vm.rs
```diff
@@ -321,11 +321,151 @@ mod tests {
...
+            .into_iter()
+            .flat_map(|row| row.to_jolt_instructions())
+            .collect();
+
+        let instructions_r1cs: Vec<RV32I> = converted_trace
+            .clone()
```
sragss, line 428: Same here.

jolt-core/src/jolt/vm/rv32i_vm.rs
```diff
@@ -321,11 +321,151 @@ mod tests {
...
+            .map(|common| RVTraceRow::from_common(common))
+            .collect();
+
+        let bytecode_trace: Vec<ELFRow> = converted_trace
+            .iter()
+            .map(|row| row.to_bytecode_trace())
+            .collect();
+
+        let instructions: Vec<RV32I> = converted_trace
+            .clone()
```
sragss, line 355: We shouldn't need this clone.

jolt-core/src/jolt/vm/rv32i_vm.rs
```diff
@@ -321,11 +321,151 @@ mod tests {
...
+            .into_iter()
+            .flat_map(|row| row.to_jolt_instructions())
+            .collect();
+
+        // Emulator sets register 0xb to 0x1020 upon initialization for some reason,
+        // something about Linux boot requiring it...
+        let mut memory_trace: Vec<MemoryOp> = vec![MemoryOp::Write(11, 4128)];
```
sragss, line 362: If we do this every time we should likely move into a default function called "gen_memory_trace" or something of that nature. (Can deal with in later PR, add a TODO here)

jolt-core/src/jolt/vm/rv32i_vm.rs
```diff
@@ -321,11 +321,151 @@ mod tests {
...
+            .into_iter()
+            .map(|common| RVTraceRow::from_common(common))
+            .collect();
+
+        let bytecode_trace: Vec<ELFRow> = converted_trace
+            .iter()
+            .map(|row| row.to_bytecode_trace())
+            .collect();
+
+        let instructions: Vec<RV32I> = converted_trace
```
sragss, line 421: `instructions` is unused

jolt-core/src/jolt/vm/rv32i_vm.rs
```diff
@@ -321,11 +321,151 @@ mod tests {
...
+                }
+            })
+            .collect();
+    
+        let memory_trace_r1cs = converted_trace.clone().into_iter().flat_map(|row| row.to_ram_ops()).collect_vec();
+
+        let circuit_flags = converted_trace.clone()
+            .iter()
+            .flat_map(|row| {
+                let mut flags: Vec<Fr> = row.to_circuit_flags();
```
sragss, line 445: Don't need this to be mutable, nor the `.into_iter()`

jolt-core/src/r1cs/snark.rs
```diff
@@ -74,10 +80,11 @@ impl<F: PrimeField> Circuit<F> for JoltCircuit<F> {
   fn synthesize<CS: ConstraintSystem<F>>(self, cs: &mut CS) -> Result<(), SynthesisError> {
     // TODO(sragss): These paths should not be hardcoded.
     let circuit_dir = JoltPaths::circuit_artifacts_path();
-    let r1cs_path = circuit_dir.join("jolt-fibonacci.r1cs");
-    let wtns_path = circuit_dir.join("jolt-fibonacci_js/jolt-fibonacci.wasm");
+    let r1cs_path = circuit_dir.join("jolt_single_step.r1cs");
```
sragss, line 83: Given these only need to be compiled once, we should add this to the build flow. We can do this in a few places:
- simple build script
- build.rs
- compiler crate

jolt-core/src/r1cs/snark.rs
```diff
@@ -74,10 +80,11 @@ impl<F: PrimeField> Circuit<F> for JoltCircuit<F> {
...
+    let wtns_path = circuit_dir.join("jolt_single_step_js/jolt_single_step.wasm");
```
sragss, line 84: Move full paths into the `common::JoltPaths` module.

jolt-core/src/r1cs/snark.rs
sragss: Currently: `./jolt-core/src/r1cs/scripts/compile_jolt.sh ./jolt-core/src/r1cs/circuits/jolt_single_step.circom ./target/circuits/`

jolt-core/src/jolt/vm/rv32i_vm.rs
```diff
@@ -321,11 +321,151 @@ mod tests {
...
+        RV32IJoltVM::prove_r1cs(
+            instructions_r1cs, 
+            bytecode_rows,
+            bytecode_trace,
+            bytecode, 
+            memory_trace_r1cs, 
+            circuit_flags,
+            &mut transcript,
+            &mut random_tape,
+            954,
```
sragss, line 463: We need to be able to compute this from compilation artifacts.

sragss:
LGTM

PR #140, opened 2024-01-15
Accelerate Witness Generation

sragss:
@philsippl is working on the following change to `circom-witness-rs`: https://github.com/philsippl/circom-witness-rs/pull/5

Implementing this will speed up our witness computation time by 10x.

The linked PR is not yet complete.


sragss:
Speeds up witness generation another 3x. Remaining slowness due to conversion. I believe there's another 25% to be squeezed but we're below 0.1% of E2E time.

jolt-core/src/r1cs/snark.rs
```diff
@@ -73,73 +73,54 @@ impl<F: PrimeField<Repr = [u8; 32]>> Circuit<F> for JoltCircuit<F> {
...
+      "prog_v_rw".to_string(), 
+      "memreg_a_rw".to_string(), 
+      "memreg_v_reads".to_string(), 
+      "memreg_v_writes".to_string(), 
+      "chunks_x".to_string(), 
+      "chunks_y".to_string(), 
+      "chunks_query".to_string(), 
+      "lookup_output".to_string(), 
+      "op_flags".to_string(),
+      "input_state".to_string()
```
arasuarun, line 87: Nit: Can we move input state to the top? Just feels right.

arasuarun:
LGTM. The benches pass on my system, too.

jolt-core/src/r1cs/snark.rs
sragss: Unfortunately non-trivial to do this due to the zip.

Could add some chaining logic but I can't make it pretty.

arasuarun: Yeah I figured. I think that's why I kept it last too.

PR #141, opened 2024-01-16
Include read-only data sections in joltbytecode

moodlezoup:


tracer/src/lib.rs
```diff
@@ -98,34 +98,36 @@ pub fn decode(elf: &PathBuf) -> Vec<ELFInstruction> {
...
-                });
+
+            if let Ok(inst) = decode_raw(word) {
+                if let Some(trace) = inst.trace {
+                    let inst = trace(&inst, &get_xlen(), word, address);
+                    instructions.push(inst);
+                    continue;
+                }
             }
+            // Unrecognized instruction, or from a ReadOnlyData section
```
moodlezoup, line 121: feels sus to encode read-only data as instructions... may need to revisit

PR #142, opened 2024-01-16
feat: decode data sections

ncitron:


common/src/lib.rs
```diff
@@ -40,6 +40,25 @@ pub enum MemoryState {
...
 
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub enum Section {
+    Text {
+        instructions: Vec<ELFInstruction>,
+        address: u64,
+        size: u64,
+    },
+    Data {
+        data: Vec<u8>,
```
sragss, line 51: We should consider using `serde_bytes` here in the future if we have performance issues.

ncitron: Good idea. Although I had to use the hex crate instead of serde_bytes to do this. It seems serde_bytes doesn't use a more efficient format when serializing to json, but the hex crate lets us serialize to hexadecimal.

moodlezoup:
we no longer need this, as we just load the `raw` instruction data to RAM as `v_init`

PR #144, opened 2024-01-16
Jump instructons cleanup

arasuarun:
Removed the superfluous `is_jal` flag because we double the immediate in the rust code and provide it as an input to the circuit directly. 

And made the jump instructions cleaner by putting them under the `is_add_instr` flag to generate the instruction query. Removed `instruction/jal.rs`, `instruction/jalr.rs` because they weren't necessary


sragss (commented):
`src/benches/bench.rs` needs the removal of `JALInstruction` / `JALRInstruction` imports. 

I'll make the change.

PR #146, opened 2024-01-16
synthesize only variables during proof (and not shape)

arasuarun:
This synthesizes only variables when calling `JoltCircuit.synthesize`, and not the constraints. Synthesizing constraints in only needed when generating the shape and that's handled in `JoltSkeleton.synthesize`. 

For the 32k hash trace, the total time spent in synthesize halves (670 ms -> 310 ms) on my system. It's overall not a major cost now anyway but I think it'll be significant as the trace lengths improve. 

(We didn't need to modify any of circom scotia or Spartan for this.)

### DIFF
--- a/jolt-core/src/r1cs/snark.rs
+++ b/jolt-core/src/r1cs/snark.rs
@@ -4,7 +4,7 @@ use spartan2::{
   traits::{snark::RelaxedR1CSSNARKTrait, Group},
   SNARK, errors::SpartanError,
 };
-use bellpepper_core::{Circuit, ConstraintSystem, LinearCombination, SynthesisError, Variable, Index};
+use bellpepper_core::{Circuit, ConstraintSystem, LinearCombination, SynthesisError, Variable, Index, num::AllocatedNum};
 use ff::PrimeField;
 use ruint::aliases::U256;
 use circom_scotia::r1cs::CircomConfig;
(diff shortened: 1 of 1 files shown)


jolt-core/src/r1cs/snark.rs
```diff
@@ -151,12 +151,13 @@ impl<F: PrimeField<Repr = [u8; 32]>> Circuit<F> for JoltCircuit<F> {
...
       let _guard = span.enter();
-      let _ = circom_scotia::synthesize(
-          &mut cs.namespace(|| format!("jolt_step_{}", i)),
-          cfg.r1cs.clone(),
-          Some(jolt_witnesses[i].clone()),
-      )
-      .unwrap();
+      {
+        let witness = &jolt_witnesses[i];
+        (1..cfg.r1cs.num_inputs).chain(cfg.r1cs.num_inputs..cfg.r1cs.num_inputs + cfg.r1cs.num_aux).for_each(|i| {
```
sragss, line 156: Is the chain necessary here?

Can't we just do:
``
let total vars = cfg.r1cs.num_inputs + cfg.r1cs.num_aux;
(1..total_vars).for_each(...)
```

File: jolt-core/src/r1cs/snark.rs [hunk 2]
```diff
@@ -151,12 +151,13 @@ impl<F: PrimeField<Repr = [u8; 32]>> Circuit<F> for JoltCircuit<F> {
     for i in 0..NUM_STEPS {
       let span = tracing::span!(tracing::Level::INFO, "circom_scotia::synthesize");
       let _guard = span.enter();
-      let _ = circom_scotia::synthesize(
-          &mut cs.namespace(|| format!("jolt_step_{}", i)),
-          cfg.r1cs.clone(),
-          Some(jolt_witnesses[i].clone()),
-      )
-      .unwrap();
+      {
```
sragss, line 154: Context block unneeded. Often added by copilot when adding spans.

sragss (approved):
LGTM – Fix changes before merging

sragss:
Fixed my comments.

PR #151, opened 2024-01-23
Cleanup dense poly

mmagician:
Slowly preparing for [generalizing](https://github.com/a16z/Lasso/issues/8) the [PCS](https://github.com/a16z/Lasso/pull/66), which would require switching to `MultilinearExtension` from `ark-poly` and deprecating the custom `DensePolynomial` used now.

This PR should be useful regardless of whether the above happens or not:
1. Removes `merge_dual` in favor of calling `merge` on the concatenated slices
2. Removes unused `extend`
3. msm-based functionality (`flag_msm` & `sm_msm`) are not methods on `DensePolynomial` but rather stand-alone functions. Moves them to the `msm` module.

### DIFF
--- a/jolt-core/src/jolt/vm/instruction_lookups.rs
+++ b/jolt-core/src/jolt/vm/instruction_lookups.rs
@@ -122,7 +122,7 @@ where
     fn batch(&self) -> Self::BatchedPolynomials {
         use rayon::prelude::*;
         let (batched_dim_read, (batched_final, batched_E, batched_flag)) = rayon::join(
-            || DensePolynomial::merge(&[self.dim.as_slice(), self.read_cts.as_slice()].concat()),
+            || DensePolynomial::merge(self.dim.iter().chain(&self.read_cts)),
             || {
                 let batched_final = DensePolynomial::merge(&self.final_cts);
                 let (batched_E, batched_flag) = rayon::join(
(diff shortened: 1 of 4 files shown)


mmagician:
Dev on `ark-poly` in that direction: https://github.com/arkworks-rs/algebra/pull/763

jolt-core/src/jolt/vm/instruction_lookups.rs
```diff
@@ -122,7 +122,7 @@ where
     fn batch(&self) -> Self::BatchedPolynomials {
         use rayon::prelude::*;
         let (batched_dim_read, (batched_final, batched_E, batched_flag)) = rayon::join(
-            || DensePolynomial::merge_dual(self.dim.as_ref(), self.read_cts.as_ref()),
+            || DensePolynomial::merge(&[self.dim.as_slice(), self.read_cts.as_slice()].concat()),
```
sragss, line 125: Can we drop the concat here?

I'm concerned that we're going to allocate a vector of length `self.dim.len() + self.read_cts.len()` twice. Once after the `.concat()` call and once in `DensePolynomial::merge`. Recall that these are likely to be massive (2^28 -> 2^32).

sragss (commented):
One comment on allocation. Otherwise LGTM.

sragss:
LGTM

PR #155, opened 2024-01-26
feat: jolt sdk

ncitron:
Initial implementation of jolt-sdk.

- Inputs and outputs are written to `examplename.joltdevice`
- Inputs are expected at the memory location `common::constants::INPUT_START_ADDRESS`
- Outputs are expected at memory location `common::constants::OUTPUT_START_ADDRESS`
- A 1 is written to `common::constants::PANIC_ADDRESS` if there was a panic at any point during the program execution
- Inputs are assumed to be initialized to 0 if not serialized into `examplename.joltdevice`
- Currently no way to feed inputs in using the compiler
    - Will address this in a follow up PR
    - This shouldn't be an issue since unset inputs are assumed to zero


common/src/constants.rs
```diff
@@ -3,4 +3,8 @@ pub const REGISTER_COUNT: u64 = 32;
 pub const REGISTER_START_ADDRESS: usize = 0;
 pub const RAM_START_ADDRESS: u64 = 0x80000000;
 pub const BYTES_PER_INSTRUCTION: usize = 4;
-pub const MEMORY_OPS_PER_INSTRUCTION: usize = 7;
\ No newline at end of file
+pub const MEMORY_OPS_PER_INSTRUCTION: usize = 8;
```
ncitron, line 6: Did I change this or did this get modified recently?

ncitron:
Also another follow up PR I need to look at here is to determine stable serialization of the inputs/outputs. Right now we use `mem::size_of` which has two problems. First, it will not work for pointer types (like `Vec<T>`) since those will just show up as the size of a pointer, which is wrong. Second, there is no guarantee that `size_of` is stable across compilations according to the docs. To address we should use some sort of serialization/deserialization scheme instead. We probably want one that is as efficient as possible. One choice that I think makes sense is bincode but open to suggestions.

common/src/constants.rs
sragss: This should be 7. 

4x byte addressable, rd, rs1, rs2.

moodlezoup:
squashed and incorporated into https://github.com/a16z/Lasso/pull/201

PR #158, opened 2024-01-30
Feat/commitment salad

moodlezoup:
Refactors `vm/mod.rs` so that all relevant commitments and polynomials can be passed to `prove_r1cs`
Also includes changes to `read_write_memory` and `timestamp_range_check` to group read-write memory polynomials by `rs1`, `rs2`, `rd`, and the four bytes of RAM accessed per CPU instruction


sragss (approved):
LGTM

PR #161, opened 2024-02-07
optimize: efficiently compute lookup outputs

sragss:
Saves 5-7% e2e by computing the lookup outputs in `u64`.

### DIFF
--- a/jolt-core/src/jolt/instruction/add.rs
+++ b/jolt-core/src/jolt/instruction/add.rs
@@ -60,6 +60,16 @@ impl<const WORD_SIZE: usize> JoltInstruction for ADDInstruction<WORD_SIZE> {
         add_and_chunk_operands(self.0 as u128, self.1 as u128, C, log_M)
     }
 
+    fn lookup_entry_u64(&self) -> u64 {
+        if WORD_SIZE == 32 {
+            (self.0 as u32).overflowing_add(self.1 as u32).0.into()
+        } else if WORD_SIZE == 64 {
+            self.0.overflowing_add(self.1).0
+        } else {
+            panic!("only implemented for u32 / u64")
+        }
+    }
+
     fn random(&self, rng: &mut StdRng) -> Self {
         use rand_core::RngCore;
         Self(rng.next_u32() as u64, rng.next_u32() as u64)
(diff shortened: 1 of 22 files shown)


sragss:
Needs a parity test between `Subtable::materialize` / `Instruction::operand_chunks` / `Instruction::combine` and `Instruction::lookup_entry_u64`.

Additionally the function name `lookup_entry_u64` is poor.

jolt-core/src/jolt/instruction/add.rs
```diff
@@ -72,26 +82,71 @@ mod test {
...
+            let instruction = ADDInstruction::<32>(x, y);
+            let expected = instruction.lookup_entry_u64();
+            jolt_instruction_test!(
+                instruction,
+                expected.into()
+            );
+            assert_eq!(
+                instruction.lookup_entry::<Fr>(C, M),
+                expected.into()
+            );
```
moodlezoup, line 108: Let's remove the existing `lookup_entry` method, rename `lookup_entry_u64` to `lookup_entry`, and remove these assertions from the instruction e2e tests

jolt-core/src/jolt/instruction/test.rs
```diff
@@ -28,3 +28,40 @@ macro_rules! jolt_instruction_test {
...
+where
+    F: PrimeField,
+    I: JoltInstruction 
+{
+    for (index, instruction) in instructions.into_iter().enumerate() {
+        let fr = instruction.lookup_entry::<F>(TEST_C, TEST_M);
+        let u = instruction.lookup_entry_u64();
+        assert_eq!(fr, F::from(u), "[{index}] Entries did not match for {instruction:?}. Left: lookup_entry(), Right: lookup_entry_u64()");
+    }
+}
```
moodlezoup, line 67: let's delete these; we really only need `lookup_entry_u64` and the `jolt_instruction_test` macro

moodlezoup (changes requested):
`lookup_entry` is only being used in two places in `rv.rs` right now, which can easily be changed to use `lookup_entry_u64` so let's just move completely over

jolt-core/src/jolt/instruction/test.rs
sragss: done

moodlezoup (approved):

PR #162, opened 2024-02-07
Rewrite Hyrax opening proofs without ZK

moodlezoup:
Deletes a bunch of vestigial ZK stuff that we weren't using, including parts of Hyrax that were only necessary for ZK
As a result, we're finally able to delete RandomTape, simplifying some function signatures throughout
Also renames/refactors some structs/files to Hyrax-related code is more clearly delineated


sragss (approved):
Beautiful.

PR #164, opened 2024-02-07
optimize: speedy drops

sragss:


sragss:
Threading the drops (`params.poly_As.into_iter().for_each(drop)`) does save 50% of drop time on it's own and returns the memory to the allocator.

sragss:
Closing due to better version in #175

PR #168, opened 2024-02-08
build-fast profile

moodlezoup:
4x faster hot builds, with no perceptible change in runtime speed


sragss (approved):
LGTM

I'm unclear on why this speeds up build – is there anything we're doing to make the debug symbols massive or is that natural?

PR #172, opened 2024-02-10
feat: integrated proofs

sragss:
Use `vm::prove()` for every end-to-end flow rather than piecemeal calls to `prove_memory()` / `prove_bytecode()` / `prove_r1cs()` / ...


moodlezoup (approved):
lgtm

PR #174, opened 2024-02-13
feat: Factor out generators

sragss:


sragss:
Rather than computing all of this stuff manually, we could just include the generator step in the call to `StructuredPolynomials::batch() -> (Self::BatchedPolynomials, Self::Generators)`

PR #175, opened 2024-02-14
Optimize: drop sumcheck params in separate thread

moodlezoup:
https://abrams.cc/rust-dropping-things-in-another-thread


sragss (approved):
LGTM

PR #178, opened 2024-02-16
feat: gens factor init reference

sragss:


jolt-core/src/poly/pedersen.rs
```diff
@@ -4,20 +4,26 @@ use digest::{ExtendableOutput, Input};
...
+    pub generators: Vec<G>
 }
 
-impl<G: CurveGroup> PedersenGenerators<G> {
-    pub fn new(n: usize, label: &[u8]) -> Self {
+impl<G: CurveGroup> PedersenInit<G> {
+    #[tracing::instrument]
+    pub fn new(max_num_vars: usize, label: &[u8]) -> Self {
+        // TODO(sragss): Realistically this needs to be moved upstream
+        // as it's a hyrax implementation detail.
```
moodlezoup, line 25: let's address this TODO before merging

jolt-core/src/poly/structured_poly.rs
```diff
@@ -20,7 +22,9 @@ pub trait BatchablePolynomials {
     /// uses `DensePolynomial::merge` to combine polynomials of the same size.
     fn batch(&self) -> Self::BatchedPolynomials;
     /// Commits to batched polynomials, typically using `DensePolynomial::combined_commit`.
-    fn commit(batched_polys: &Self::BatchedPolynomials) -> Self::Commitment;
+    fn commit(batched_polys: &Self::BatchedPolynomials, initalizer: &PedersenInit<G>) -> Self::Commitment;
+
+    fn max_generator_size(&self) -> usize;
```
moodlezoup, line 27: I think the implementations will be cleaner if we change the signature to:
```suggestion
    fn max_generator_size(&Self::BatchedPolynomials) -> usize;
```

jolt-core/src/benches/bench.rs
```diff
@@ -9,7 +9,8 @@ use crate::jolt::vm::read_write_memory::{
 use crate::jolt::vm::rv32i_vm::{RV32IJoltVM, C, M, RV32I};
 use crate::jolt::vm::Jolt;
 use crate::poly::dense_mlpoly::bench::{init_commit_bench, run_commit_bench};
-use ark_curve25519::{EdwardsProjective, Fr};
+// use ark_curve25519::{EdwardsProjective, Fr};
```
sragss, line 12: nit: fix jank

jolt-core/src/poly/pedersen.rs
sragss: Done

jolt-core/src/benches/bench.rs
sragss: Done

jolt-core/src/poly/structured_poly.rs
sragss: I did it this way so we could theoretically split the call to `batch()` and the call to `max_generator_size()`.

sragss: It's not too bad, lmk what you think #184

PR #179, opened 2024-02-17
if_else using only 1 constraint, wire

arasuarun:
Optimized if-else that just uses one extra constraint and wire. (It was two before). The total number of wires per step goes from 93 to 82 now. Updated the graph.bin accordingly.


jolt-core/src/r1cs/circuits/jolt_single_step.circom
```diff
@@ -50,8 +50,11 @@ template if_else() {
     signal a <== in[1];
     signal b <== in[2];
 
-    signal _please_help_me_god <== (1-zero_for_a) * a;
-    out <== _please_help_me_god + (zero_for_a) * b;
+    // signal _please_help_me_god <== (1-zero_for_a) * a;
```
sragss, line 53: nit: rm

sragss (approved):
rm unused (and alarming) `_please_help_me_god` variable

Otherwise LGTM

PR #181, opened 2024-02-20
Use zips instead of indexing into vectors

moodlezoup:
Some micro-optimizations to avoid bounds checks

### DIFF
--- a/jolt-core/src/poly/dense_mlpoly.rs
+++ b/jolt-core/src/poly/dense_mlpoly.rs
@@ -75,12 +75,32 @@ impl<F: PrimeField> DensePolynomial<F> {
 
     pub fn bound_poly_var_top(&mut self, r: &F) {
         let n = self.len() / 2;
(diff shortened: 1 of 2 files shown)


jolt-core/src/poly/dense_mlpoly.rs
```diff
@@ -75,12 +75,32 @@ impl<F: PrimeField> DensePolynomial<F> {
 
     pub fn bound_poly_var_top(&mut self, r: &F) {
         let n = self.len() / 2;
+        let (left, right) = self.Z.split_at_mut(n);
+
+        left.iter_mut().zip(right.iter()).for_each(|(a, b)| {
```
moodlezoup, line 80: experimented with turning these into parallel iterators, but threads are saturated one layer up wherever `bound_poly_var_top` and `bound_poly_var_top_many_ones` are used

sragss (approved):

PR #182, opened 2024-02-20
optimize: GKR eq additions

sragss:


sragss:
Current commits simply remove the additions (but do not replace with the new logic). Savings are 5-10%.

PR #185, opened 2024-02-26
Fast u64 montgomery conversion

moodlezoup:
Uses our fork of `arkworks/algebra` to leverage fast conversion from u64s to field elements (in Montgomery form)
Companion PR: https://github.com/a16z/arkworks-algebra/pull/2


sragss (approved):
LGTM

PR #187, opened 2024-02-29
implement the bn256 curve conversion between ark and halo2curves

nanne007:
@sragss reached me about my [implementation](https://github.com/zkmove/halo2-verifier.move/blob/main/crates/vk-gen-examples/src/to_ark.rs) of the conversion between ark and halo2curves.

He suggests I provide the impl to the repo, so here we are.


sragss (approved):
LGTM, thanks for the PR!

PR #188, opened 2024-02-29
feat: Ark -> Spartan G1 Conversion

sragss:
Adds onto @nanne007's PR #187 to allow conversion from Arkworks -> Spartan in the generator preprocess function.

### DIFF
--- a/common/src/field_conversion.rs
+++ b/common/src/field_conversion.rs
@@ -1,6 +1,6 @@
 use ark_bn254::Fr as ArkFr;
 use ark_ec::short_weierstrass::SWCurveConfig;
-use ark_ec::{AffineRepr, CurveConfig};
+use ark_ec::CurveConfig;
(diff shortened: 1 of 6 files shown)


common/src/field_conversion.rs
```diff
@@ -82,37 +89,64 @@ pub trait IntoArk: CurveAffine {
...
+            }
+        }
+    }
+}
+
+/// Transforms an `ark_ec` elliptic curve point into `halo2curves` format for Spartan compatibility.
+///
+/// Serializes `ark_ec` point coordinates and deserializes them into `SpartanAffine` points.
+/// Returns the identity point for the curve's zero element, or a new `SpartanAffine` point otherwise.
+pub trait IntoSpartan: ark_ec::AffineRepr {
```
moodlezoup, line 132: we don't need both `FromArk` and `IntoSpartan` right?

moodlezoup (approved):
just one nit, otherwise lgtm

PR #191, opened 2024-03-08
Speed up RAM poly construction

sragss:


moodlezoup (approved):
thanks, while you're at it could you do the same for bytecode?

PR #196, opened 2024-03-12
feat: Surge example

sragss:


moodlezoup (commented):
I'm not a huge fan of having this in the same directory as the example guest programs. Can we put these under `book/` or a separate `book_examples/` directory?

PR #202, opened 2024-03-20
use compressed PC value

arasuarun:
Uses the mapped PC value as input that is much smaller without the extra 0x80....00. This should lead to a noticeable decrease in commitment costs.


moodlezoup:
Done in https://github.com/a16z/Lasso/pull/203

PR #214, opened 2024-03-26
feat: linker automation

ncitron:
Linker is now automatically generated and build arguments are passed in by the sdk rather than the build script.

This additionally adds an optional `memory_size` attribute to the macro which allows us to specify exactly how much memory will be reserved by the linker.


jolt-core/src/host/mod.rs
```diff
@@ -45,9 +44,16 @@ impl Program {
         self
     }
 
+    pub fn memory_size(mut self, len: usize) -> Self {
+        self.memory_size = len;
+        self
+    }
```
moodlezoup, line 50: nit:
```suggestion
    pub fn set_memory_size(&mut self, len: usize) {
        self.memory_size = len;
    }
```

jolt-sdk/macros/src/lib.rs
```diff
@@ -1,298 +1,363 @@
...
+            let input_ptr = #INPUT_START_ADDRESS as *const u8;
+            let input_slice = unsafe {
+                core::slice::from_raw_parts(input_ptr, #max_input_len)
             };
+        };
+
+        let args = &self.func_args;
+        let args_fetch = args.iter().map(|(name, ty)| {
+            quote! {
+                let x = input_slice[0];
```
moodlezoup, line 220: does this line do anything?

jolt-sdk/macros/src/lib.rs
```diff
@@ -1,298 +1,363 @@
...
+                let (program, preprocessing) = #preprocess_fn_name();
+                let program = std::rc::Rc::new(program);
+                let preprocessing = std::rc::Rc::new(preprocessing);
+
+                let program_cp = program.clone();
+                let preprocessing_cp = preprocessing.clone();
+
+                let prove_closure = move |#inputs| {
+                    let program = (*program).clone();
+                    let preprocessing = (*preprocessing).clone();
```
moodlezoup, line 81: is there any way to avoid these clones? otherwise these will be included in our reported benchmark times

ncitron: We do need them as the prove and verify functions take ownership of their inputs.

Fortunately if you look a couple lines above I moved the cloned data into a `Rc` which makes the cloning costs insignificant.

moodlezoup: ah gotcha

moodlezoup (approved):
good stuff, ty

PR #218, opened 2024-03-27
Instruction flags constraints, cleanup

arasuarun:
working, but needs a TON of cleanup

### DIFF
--- a/common/src/rv_trace.rs
+++ b/common/src/rv_trace.rs
@@ -235,18 +235,6 @@ pub struct ELFInstruction {
 pub const NUM_CIRCUIT_FLAGS: usize = 17;
 
 impl ELFInstruction {
-    pub fn default_from_opcode(opcode: RV32IM) -> Self {
-        Self {
-            address: 0,
-            opcode,
-            raw: 0,
-            rs1: None,
-            rs2: None,
-            rd: None,
-            imm: None,
-        }
-    }
-
     #[rustfmt::skip]
     pub fn to_circuit_flags(&self) -> [bool; NUM_CIRCUIT_FLAGS] {
         // Jolt Appendix A.1
--- a/common/src/rv_trace.rs
+++ b/common/src/rv_trace.rs
@@ -235,6 +235,18 @@ pub struct ELFInstruction {
 pub const NUM_CIRCUIT_FLAGS: usize = 17;
 
 impl ELFInstruction {
+    pub fn default_from_opcode(opcode: RV32IM) -> Self {
+        Self {
+            address: 0,
+            opcode,
+            raw: 0,
+            rs1: None,
+            rs2: None,
+            rd: None,
+            imm: None,
+        }
+    }
+
     #[rustfmt::skip]
     pub fn to_circuit_flags(&self) -> [bool; NUM_CIRCUIT_FLAGS] {
         // Jolt Appendix A.1
--- a/jolt-core/src/host/mod.rs
+++ b/jolt-core/src/host/mod.rs
@@ -10,7 +10,7 @@ use serde::Serialize;
 
 use common::{
     constants::MEMORY_OPS_PER_INSTRUCTION,
-    rv_trace::{JoltDevice, MemoryOp, RV32IM},
+    rv_trace::{JoltDevice, MemoryOp, NUM_CIRCUIT_FLAGS, RV32IM},
 };
 use tracer::ELFInstruction;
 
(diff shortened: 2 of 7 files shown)


jolt-core/src/jolt/vm/mod.rs
```diff
@@ -505,8 +507,8 @@ pub trait Jolt<F: PrimeField, G: CurveGroup<ScalarField = F>, const C: usize, co
         let inputs: R1CSInputs<F> = R1CSInputs::new(
             padded_trace_len,
             bytecode_a,
-            bytecode_v,
-            packed_flags,
+            bytecode_v.clone(),
```
sragss, line 510: Looks like these are cloned just to be able to print afterwards

arasuarun:
Ready for review. Now, both the instruction flags and circuit flags are sent to the R1CS and are checked against the packed flags value `op_flags_packed` which replaces the opcode in the program code. (We don't actually need the opcode in the circuit.)

jolt-core/src/host/mod.rs
```diff
@@ -121,11 +122,24 @@ impl Program {
...
             .flat_map(|row| {
-                row.instruction
+                let mut bitvector = row.instruction
                     .to_circuit_flags()
                     .iter()
                     .map(|&flag| flag.into())
-                    .collect::<Vec<F>>()
+                    .collect::<Vec<F>>();
+
+                let N_FLAGS = bitvector.len();
```
moodlezoup, line 131: Let's use `NUM_CIRCUIT_FLAGS` from `rv_trace.rs`

jolt-core/src/r1cs/constraints.rs
```diff
@@ -7,7 +7,7 @@ use common::constants::{RAM_START_ADDRESS, RAM_WITNESS_OFFSET};
 
 /* Compiler Variables */
 const C: usize = 4; 
-const N_FLAGS: usize = 17; 
+const N_FLAGS: usize = 17 + 19;
```
moodlezoup, line 10: `NUM_CIRCUIT_FLAGS + RV32I::COUNT`

jolt-core/src/host/mod.rs
```diff
@@ -121,11 +122,24 @@ impl Program {
...
+
+                // instruction flag
+                if let Ok(jolt_instruction) = InstructionSet::try_from(&row.instruction) {
+                    let instruction_index = InstructionSet::enum_index(&jolt_instruction);
+                    // bitvector <<= instruction_index;
+                    // bitvector |= 1;
+                    // bitvector <<= InstructionSet::COUNT - instruction_index - 1;
+                    bitvector[N_FLAGS + instruction_index] = F::from(1 as u64);
+                } 
+                bitvector
```
moodlezoup, line 142: We already have the instruction flag polynomials from `instruction_lookups`, shouldn't we be feeding those to the r1cs instead of recreating them here?

moodlezoup (changes requested):

common/src/rv_trace.rs
```diff
@@ -235,6 +235,18 @@ pub struct ELFInstruction {
...
+            address: 0,
+            opcode,
+            raw: 0,
+            rs1: None,
+            rs2: None,
+            rd: None,
+            imm: None,
+        }
+    }
+
```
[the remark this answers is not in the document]
arasuarun: You're right. Removed it.

jolt-core/src/host/mod.rs
arasuarun: Yeah this isn't ideal. I don't think the flags (either circuit or instr) should be in this `trace` function. It would be ideal to just use the opcode and get all the circuit and instruction flags in `mod.rs`. I tried doing something like this but I wasn't able to call `bitflags()` in bytecode.rs because it is expecting an ELFInstruction while we only have access to an RV32I enum in the `prove_r1cs` function. 

I might just be missing something, though. We could try to do a full refactor (no flags in trace) or a partial one for now and do the full later. What do you think?

moodlezoup: no what I mean is, we still need to compute `row.instruction.to_circuit_flags()` here in `trace`. But we don't need these instruction flags in the bitvector, because we already have them in the form of `InstructionPolynomials.instruction_flag_polys`
We should be passing those `instruction_flag_polys` into `prove_r1cs`, and using those in the circuit. As it is right now, we're recommitting to these instruction flag bits

arasuarun: Ah right. Yeah that should be the way to do it. Lemme try adding that now.

jolt-core/src/host/mod.rs
moodlezoup, line 1: nit: can we revert the changes to this file?

moodlezoup (approved):
one nit but otherwise lgtm

jolt-core/src/host/mod.rs
arasuarun: Done

PR #219, opened 2024-03-27
feat: Serialize proofs

sragss:
The strategy is to use `ark_serialize::{CanonicalSerialize, CanonicalDeserialize}` recursively as low as possible via macros. If we'd like to expose `serde::Serialize` we should do this at the highest level. If we try to do `serde` at a lower level we'll have to use serde's `serialize_with` on nearly every field.

The lack of compatibility between these two appears contentious:
- https://github.com/arkworks-rs/algebra/issues/178
- https://github.com/arkworks-rs/algebra/pull/506
- https://github.com/EspressoSystems/jellyfish/issues/288

Issues remaining: 
- Does not yet remove generators
- R1CS proof contains the `UniformSpartanKey` which should not be in the proof, but rather "known" by the verifier


sragss:
The PR manually implements `ark_serialize::{CanonicalSerialize, CanonicalDeserialize}` for types including `PhantomData<T>`. I believe the macros will work for these as long as `T: Send + Sync` based on the arkworks default impls.

https://github.com/arkworks-rs/algebra/blob/bb663bc27e776af81e66c94d9016bde2dbba9968/serialize/src/impls.rs#L275-L282

sragss:
`T: Send + Sync` strategy described above works.

moodlezoup (approved):

PR #226, opened 2024-03-28
feat: Poly_ABC acceleration

sragss:


jolt-core/src/r1cs/spartan.rs
```diff
@@ -305,16 +311,29 @@ impl<F: PrimeField, G: CurveGroup<ScalarField = F>> UniformSpartanProof<F, G> {
...
+            let span = tracing::span!(tracing::Level::TRACE, "poly_ABC_big_RLC_evals");
+            let _enter = span.enter();
+            // small_RLC_evals is 30% ones.
+            RLC_evals.par_chunks_mut(n_steps).take(key.num_vars_total / n_steps).enumerate().for_each(|(chunk_index, rlc_chunk)| {
+                if !small_RLC_evals[chunk_index].is_zero() {
+                    for (eq_index, item) in rlc_chunk.iter_mut().enumerate() {
+                        *item = eq_rx_ts[eq_index] * small_RLC_evals[chunk_index];
+                    }
+                }
+            });
```
moodlezoup, line 331: Did this help? I'm pretty I tried something similar but it was slower

moodlezoup (approved):

jolt-core/src/r1cs/spartan.rs
sragss: high variance but 25% to 50% quicker. small_RLC_evals is also 25% zeros

PR #229, opened 2024-03-28
Hyrax batch commit

moodlezoup:
Smooths out the parallelism when committing to a batch of polynomials. Significant improvement for commit_segments


sragss (approved):
LGTM

PR #235, opened 2024-03-29
Constraints/state pc loads

arasuarun:
1. Removes the step counter from the state. 
2. Enforces the correct program counter value (doesn't map to a lower address space, though).
3. Enforces the correct lookup query for load sign-extension lookups

### DIFF
--- a/jolt-core/src/r1cs/constraints.rs
+++ b/jolt-core/src/r1cs/constraints.rs
@@ -14,6 +14,7 @@ const N_FLAGS: usize = NUM_CIRCUIT_FLAGS + RV32I::COUNT;
 const W: usize = 32;
 const LOG_M: usize = 16; 
 const MEMORY_ADDRESS_OFFSET: usize = (RAM_START_ADDRESS - RAM_WITNESS_OFFSET) as usize; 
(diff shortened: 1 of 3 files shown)


jolt-core/src/r1cs/constraints.rs
```diff
@@ -412,6 +445,11 @@ impl R1CSBuilder {
             smallvec![(combined_z_chunks, 1), (x, -1), (y, 1), (0, -1 * (ALL_ONES + 1))],
             smallvec![]
         ); 
+        R1CSBuilder::constr_abc(instance,
+            smallvec![(is_load_instr, 1)], 
+            smallvec![(combined_z_chunks, 1), (load_or_store_value, -1)],
+            smallvec![]
+        );
```
moodlezoup, line 452: I think we actually don't need this constraint. I changed the `is_add_instr` flag to include store/load instructions (https://github.com/a16z/Lasso/blob/jolt/common/src/rv_trace.rs#L330-L339), and for store/load instructions we set x to 0. So the constraint above for add instructions should be sufficient

moodlezoup (approved):

arasuarun:
LGTM.

PR #236, opened 2024-03-29
remove unnecessary M extension flags, constraints

arasuarun:
Removes the unnecessary flags and constraints for the M extension.

### DIFF
--- a/common/src/rv_trace.rs
+++ b/common/src/rv_trace.rs
@@ -232,7 +232,7 @@ pub struct ELFInstruction {
     pub imm: Option<u32>,
 }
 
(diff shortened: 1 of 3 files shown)


common/src/rv_trace.rs
```diff
@@ -232,7 +232,7 @@ pub struct ELFInstruction {
     pub imm: Option<u32>,
 }
 
-pub const NUM_CIRCUIT_FLAGS: usize = 17;
+pub const NUM_CIRCUIT_FLAGS: usize = 14;
```
moodlezoup, line 235: 13

moodlezoup (approved):
lgtm, just change `NUM_CIRCUIT_FLAGS` to 13

PR #238, opened 2024-03-29
feat: add jolt tool

ncitron:
To install run: `cargo install --git https://github.com/a16z/Lasso --branch jolt --force` (will only work after merge)

To install toolchain run: `jolt install-toolchain`

To create new project run `jolt new <NAME>` and a new directory will be created with the name <NAME> and a simple fibonacci example.

Note: we will need to update the branch in the toml file once we change the branch from jolt to master.

For some reason I have to import jolt_sdk directly rather than jolt or the no_std in the guest compilation is not respected. I'd like to fix this but am unsure how. If anyone has any ideas let me know.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,6 +40,16 @@ members = [
     "examples/multi-function/guest",
 ]
 
+[lib]
+path = "./src/lib.rs"
+
+[[bin]]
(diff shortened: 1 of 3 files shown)


moodlezoup (approved):
nice work

PR #239, opened 2024-03-29
misc: Faster Allocs

sragss:


sragss:
These changes have been cleaned and moved to #243

PR #244, opened 2024-03-31
optimize: 4x DensePolynomial::merge

sragss:
Accelerate polynomial merging.

This is close to as fast as we'll get. Makes the interface a bit gross. Should likely remove the merging / batching steps entirely.


moodlezoup:
https://github.com/a16z/Lasso/pull/233 gets rid of `DensePolynomial::merge` entirely

sragss:
That was efficient

PR #245, opened 2024-04-01
IO consistency constraints

arasuarun:
Adds a constraint per step that checks that the output PC of step i is the input PC to step i+1. This is only enforced for the true (non padded) steps of the trace, which is passed as a new variable. 

This requires modifying the r1cs matrix-vector multiplication and the MLE evaluations for both the prover and verifier. 

(There are optimizations that improve overall performance and also avoid this, but require changing the trace.)


jolt-core/src/r1cs/r1cs_shape.rs
```diff
@@ -192,7 +192,8 @@ impl<F: PrimeField> R1CSShape<F> {
     pub fn multiply_vec_uniform<P: IndexablePoly<F>>(
         &self,
         full_witness_vector: &P,
-        num_steps: usize,
+        true_num_steps: usize,
```
sragss, line 195: rm

PR #253, opened 2024-04-02
Optimize read-write memory commitments

moodlezoup:
Avoids a couple of commitments and constraints by being smarter about how we compute (a, v, t) tuples for read-write memory, namely:
- rs1 and rs2 are read-only, so we know `v_write = v_read` and `t_write = global_timestamp`
- rd is write-only, so we know `t_write = global_timestamp + 1`
- The four bytes of RAM are contiguous, so we only need to commit to the address of the first byte


common/src/rv_trace.rs
```diff
@@ -173,7 +173,7 @@ impl Into<[MemoryOp; MEMORY_OPS_PER_INSTRUCTION]> for &RVTraceRow {
                 RV32IM::SB => [
                     rs1_read(),
                     rs2_read(),
-                    MemoryOp::no_op(),
+                    MemoryOp::Write(0, 0),
```
sragss, line 176: nit: comment "rd is always a write"

PR #259, opened 2024-04-05
refactor timestamp.commitments from proof -> commitments

sragss:


moodlezoup:
I think ultimately we'll want to do the opposite, move the commitments inside their respective proof structs

PR #261, opened 2024-04-05
Fiat Shamir

sragss:


jolt-core/src/jolt/vm/bytecode.rs
```diff
@@ -330,6 +331,18 @@ pub struct BytecodeCommitment<G: CurveGroup> {
     pub t_final_commitment: HyraxCommitment<1, G>,
 }
 
+impl<G: CurveGroup> AppendToTranscript<G> for BytecodeCommitment<G> {
+    fn append_to_transcript<T: ProofTranscript<G>>(&self, label: &'static [u8], transcript: &mut T) {
+        <T as ProofTranscript<G>>::append_protocol_name(transcript, label);
+
+        for commitment in &self.read_write_commitments {
```
fmhall, line 338:
```suggestion
        for commitment in &self.read_write_own_commitments {
```

fmhall: bah dum tssss

ncitron: These commitments are being borrowed in to the scope which implies that ownership is not being transferred. At best these are `read_write_borrow_commitments`. Maybe we can consider tokenizing these commitments to allow shared ownership without the need for borrowing (renting).

PR #265, opened 2024-04-07
Refactor/fiat shamir

moodlezoup:
Refactor so that all polynomials are committed to at the beginning of `Jolt::prove`, and all commitments are appended to the transcript. 
Also includes an optimization to reuse the bytecode v polynomials corresponding to rs1, rs2, rd in read_write_memory, instead of separately committing to `a_rs1`, `a_rs2`, and `a_rd`

### DIFF
--- a/jolt-core/src/jolt/vm/mod.rs
+++ b/jolt-core/src/jolt/vm/mod.rs
@@ -2,7 +2,6 @@ use ark_ec::CurveGroup;
 use ark_ff::PrimeField;
 use ark_serialize::{CanonicalDeserialize, CanonicalSerialize};
 use ark_std::log2;
-use common::constants::NUM_R1CS_POLYS;
 use common::rv_trace::JoltDevice;
 use itertools::max;
 use merlin::Transcript;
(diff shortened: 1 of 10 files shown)


jolt-core/src/jolt/vm/mod.rs
```diff
@@ -157,78 +269,99 @@ pub trait Jolt<F: PrimeField, G: CurveGroup<ScalarField = F>, const C: usize, co
         JoltProof<C, M, F, G, Self::InstructionSet, Self::Subtables>,
         JoltCommitments<G>,
     ) {
-        println!("Jolt::prove({}", instructions.len());
```
sragss, line 160: Idk @ncitron really likes this print

jolt-core/src/jolt/vm/mod.rs
```diff
@@ -157,78 +286,109 @@ pub trait Jolt<F: PrimeField, G: CurveGroup<ScalarField = F>, const C: usize, co
...
+        let trace_length = instructions.len();
+        let padded_trace_length = trace_length.next_power_of_two();
+        println!("Trace length: {}", trace_length);
```
nazar-pc, line 291: Did you mean to keep it in the code?

PR #267, opened 2024-04-08
added r1cs shape to transcript

arasuarun:


sragss:
Let's fix the `vk.digest` in this PR too.

PR #271, opened 2024-04-08
feat: add allocator support

ncitron:


jolt-sdk/src/lib.rs
```diff
@@ -66,3 +68,54 @@ impl Proof {
...
+
+    pub fn free_memory(&self) -> usize {
+        heap_start() + (self.offset.get() as usize)
+    }
+
+    pub fn fake_alloc(&self, layout: Layout) -> *mut u8 {
+        unsafe {
+            self.alloc(layout)
+        }
+    }
```
moodlezoup, line 102: what is this for?

jolt-sdk/src/lib.rs
```diff
@@ -66,3 +68,54 @@ impl Proof {
...
+        alloc_start as *mut u8
+    }
+
+    unsafe fn dealloc(&self, _ptr: *mut u8, _layout: Layout) {
+        
+    }
+}
+
+fn align_up(addr: usize, align: usize) -> usize {
+    (addr + align - 1) & !(align - 1)
```
moodlezoup, line 120: should we be asserting that `align` is a power of two here?

ncitron: Align is guaranteed to be a power of two so no need.

jolt-sdk/src/lib.rs
ncitron: Oh my bad that's from something I was testing. Removed.

moodlezoup (approved):

PR #280, opened 2024-04-09
Add Collatz example

moodlezoup:


examples/collatz/guest/src/lib.rs
```diff
@@ -0,0 +1,29 @@
...
+        let num_steps = collatz_convergence(n);
+        if num_steps > max_num_steps {
+            max_num_steps = num_steps;
+        }
+    }
+    max_num_steps
+}
+
+#[jolt::provable]
+fn collatz_convergence(n: u128) -> u128 {
```
ncitron, line 17: Does this need to be marked as provable? It is not used in the host so I think we can remove it.

PR #289, opened 2024-04-12
feat: Save program traces to file

sragss:
Should make it easier to debug ppls issues without requiring sharing a full repo.

Additionally down the road will give us better integration tests and binary searching for failing instructions within a program.


moodlezoup (approved):
nice

PR #300, opened 2024-04-17
feat: standard library support

ncitron:
Adds support for the standard library in guests. This is done by adding the `guest-std` flag when importing jolt in the guest. Also note that the flag supplied to the sdk when called from the host has changed from `std` to `host` to avoid confusion (although existing programs will need to be updated).

To use this, you need to have a custom rust toolchain by running `jolt install-toolchain` with this version of the jolt tool.

I've also noticed a few bugs that crop up when tracing and proving some programs generated with the standard library which we should get to the bottom of.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,4 +1,7 @@
-use std::{fs::{self, File}, io::Write};
+use std::{
+    fs::{self, File},
+    io::Write,
+};
 
 use clap::{Parser, Subcommand};
 use eyre::Result;
@@ -68,7 +71,11 @@ fn link_toolchain() {
             "toolchain",
             "link",
             "riscv32i-jolt-zkvm-elf",
-            dirs::home_dir().unwrap().join(".jolt/rust/build/host/stage2").to_str().unwrap(),
+            dirs::home_dir()
+                .unwrap()
+                .join(".jolt/rust/build/host/stage2")
+                .to_str()
+                .unwrap(),
         ])
         .output()
         .expect("failed to link toolchain");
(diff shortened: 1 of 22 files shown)


moodlezoup (approved):
awesome stuff

src/main.rs
```diff
@@ -40,13 +42,86 @@ fn create_project(name: String) {
...
+
+fn download_jolt_toolchain(client: &Client, url: &str) {
+    let bytes = client.get(url).send().unwrap().bytes().unwrap();
+    let jolt_dir = dirs::home_dir().unwrap().join(".jolt");
+    if !jolt_dir.exists() {
+        fs::create_dir(&jolt_dir).unwrap();
+    }
+
+    let path = jolt_dir.join("rust-toolchain.tar.gz");
+    fs::write(&path, &bytes).unwrap();
```
moodlezoup, line 108: nit:
```suggestion
    fs::write(path, &bytes).unwrap();
```

Levike34:
@ncitron 
> To use this, you need to have a custom rust toolchain by running `jolt install-toolchain` with this version of the jolt tool.

which custom toolchain? I get an error about the target `riscv32i-jolt-zkvm-elf` not being available for the specified toolchain in the repo.

ncitron:
The toolchain is `riscv32i-jolt-zkvm-elf` which is downloaded from [a16z/rust](https://github.com/a16z/rust). To install make sure you have the latest version of the Jolt tool and run `jolt install-toolchain`. You can reinstall the jolt tool to update it by running `cargo +nightly install --git https://github.com/a16z/jolt --force --bins jolt`.

We should probably just check on the fly to see if it is available and install it if not though to avoid these issues.

Issue #293, opened 2024-04-15
Parallelize InstructionLookups::polynomialize


sragss:
For a 64 core machine at a cycle count of ~16M Jolt spends ~4% of its time on segment called `InstructionLookups::polynomialize` [here](https://github.com/a16z/jolt/blob/main/jolt-core/src/jolt/vm/instruction_lookups.rs#L985). This segment allocates and computes the offline memory checking (a,v,t) polynomials used for the RV Instruction lookups. 

Currently it is computed entirely serially. We should parallelize this to get a speedup as the core count increases.

Similar to #292.

shreyas-londhe:
Hi @sragss I'm willing to work on this issue.

YashBit:
@sragss I am interested. Could you assign it to me?

mahmudsudo:
Hi , can i take on this issue ?

PR #302, opened 2024-04-17
refac(perf): refactor parallelize polynomialize()

PatStiles:
Took a shot at making polynomialize() less serial as described in #293. After inlining and combining compute_lookup_outputs() and subtable_lookup_indices()  I did not observe any improvements in performance.

### DIFF
--- a/jolt-core/src/jolt/vm/instruction_lookups.rs
+++ b/jolt-core/src/jolt/vm/instruction_lookups.rs
@@ -990,10 +990,13 @@ where
 
         let subtable_lookup_indices: Vec<Vec<usize>> = Self::subtable_lookup_indices(ops);
 
-        let polys: Vec<(DensePolynomial<F>, DensePolynomial<F>, DensePolynomial<F>)> = (0
+        //TODO:
+        // - Parellelize inner operation
+        let (read_cts, final_cts, E_polys): (Vec<DensePolynomial<F>>, Vec<DensePolynomial<F>>, Vec<DensePolynomial<F>>) = (0
             ..preprocessing.num_memories)
             .into_par_iter()
-            .map(|memory_index| {
+            .fold(|| (Vec::new(), Vec::new(), Vec::new()),
+            |(mut read_acc, mut final_acc, mut E_acc), memory_index| {
                 let dim_index = preprocessing.memory_to_dimension_index[memory_index];
                 let subtable_index = preprocessing.memory_to_subtable_index[memory_index];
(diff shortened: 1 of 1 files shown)


jolt-core/src/jolt/vm/instruction_lookups.rs
```diff
@@ -988,85 +988,98 @@ where
...
+                |(mut read_acc, mut final_acc, mut E_acc), memory_index| {
+                    let dim_index = preprocessing.memory_to_dimension_index[memory_index];
+                    let subtable_index = preprocessing.memory_to_subtable_index[memory_index];
+                    let access_sequence: &Vec<usize> = &subtable_lookup_indices[dim_index];
+
+                    let mut final_cts_i = vec![0usize; M];
+                    let mut read_cts_i = vec![0usize; m];
+                    let mut subtable_lookups = vec![F::zero(); m];
+
+                    for (j, op) in ops.iter().enumerate() {
```
sragss, line 1051: It may be faster if instead of the current pattern:
```
for memory in memories {
    for op in ops {
         if(op.memories.contains(memory)) { // probably linear in the size of memories (4 -> 12 with current params)
                  // do stuff
         }
    }
}
```

Doing (bad psuedo code to show idea):
```
// Precompute list of instruction indices for each memory
let memory_A_op_indices = [];
let memory_B_op_indices = [];
let memory_ops = [memory_A_op_indices, memory_B_op_indices,...];
...
for op in ops {
     for memory in ops.memories_used {
          memory_ops[memory].push(op);
     }
} 

// Now compute the memory counters directly in parallel
for memory in memories {
                    let mut final_cts_i = vec![0usize; M];
                    let mut read_cts_i = vec![0usize; m];
                    let mut subtable_lookups = vec![F::zero(); m];

                  for op in memory_ops[memory] {
                       // update relevant counters
                  }

                 // TODO: construct polynomials
}
```

I'm thinking this will have better cache performance (less stuff is looked up in the hot loop) and get rid of the `memories_used.contains`.

sragss (commented):
I do see a 20% improvement on the big machine. I've added some ideas in the review. Probably worth creating a little `iai-callgrind` benchmark. 

One other suggestion would be to add more granular tracing. Trace `instruction_flag_bitvectors` construction, `instruction_flag_polys` construction, and even within the big loop. NUM_MEMORIES should only be around 80 so we should be able to sort what's going on in the trace. 

Looks to me like 50% of e2e time is spend on `DensePolynomial::from` which isn't going to get much faster so we can consider that a ceiling.

PR #305, opened 2024-04-17
Flesh out wiki

moodlezoup:


Cargo.toml
```diff
@@ -80,6 +80,6 @@ inherits = "release"
 debug = false
 
 [patch.crates-io]
-ark-ff = { git = "https://github.com/a16z/arkworks-algebra", branch = "optimize/field-from-u64" }
```
sragss, line 83: Don't think we want to commit this.

moodlezoup: oops, not sure how this snuck in

PR #306, opened 2024-04-17
Clean up ProofTranscript trait

sragss:
ProofTranscript was introduced back in the day to provide a layer of indirection so we could mock the transcript challenges.  We had exactly 1 sumcheck test that did this and it added a ton of bloat messy generic signatures.


moodlezoup (approved):
ty, this has been irking me

Issue #295, opened 2024-04-15
ReadWriteMemory::get_polys_r1cs by reference


sragss:
`ReadWriteMemory::get_polys_r1cs()` is extremely slow ([here](https://github.com/a16z/jolt/blob/main/jolt-core/src/jolt/vm/read_write_memory.rs#L761)) because it clones all of the underlying not once but twice. On a ~16M cycle count trace on a 64 core machine it takes 2% of end-to-end time.

First: remove the double clones from `self.v_read` and `self.v_write_rd`. Double clones come from a clone in `DensePolynomial::evals()` then another in `.flat_map(...).collect()`. 

Second: refactor `get_polys_r1cs()` to return lifetimed references to the underlying vectors via `DensePolynomial::evals_ref()` rather than cloning / reallocating at all.

It may be helpful to review the [tracing strategy](https://jolt.a16zcrypto.com/dev/tools.html#tracing) for performance testing.

MatteoMer:
hey! Would be happy to take this!

sragss:
Great! Feel free to ask questions here.

moodlezoup:
Closed by #307

PR #307, opened 2024-04-18
Refactoring get_polys_r1cs to use references instead of cloning

MatteoMer:
Solves #295

I had to refactor `R1CSInputs` a bit, to accept references

Doing benchmarks on my machine using `cargo run --profile build-fast -p jolt-core trace --name sha3 --format chrome` `get_poly_r1cs` gets a ~5x improvement (`138ms` to `38ms`)


moodlezoup (approved):
nice, thank you Matteo!

PR #309, opened 2024-04-19
Refactor generics

sragss:
Refactors generics to support multiple fields and polynomial commitment schemes.

## `JoltField`
New field trait in `jolt-core/src/poly/field.rs`. Will allow us to wrap types of fields and make sure they implement Jolt specific functionality. We can support `ark_ff::PrimeField`, `ff::PrimeField`, `binius_field::TowerField` all in one place.

## `CommitmentScheme`
The other big change is that we've added a `CommitmentScheme` trait in `jolt-core/src/poly/commitment/commitment_scheme.rs`. This is passed around as a static generic everywhere it's needed. Also allows us to remove the `CommitmentScheme` specific `ark_ff::CurveGroup` generic and apply that to only Curve based commitment schemes.


jolt-core/src/poly/field.rs
```diff
@@ -0,0 +1,91 @@
...
+    fn is_one(&self) -> bool;
+    fn zero() -> Self;
+    fn one() -> Self;
+    fn from_u64(n: u64) -> Option<Self>; // TODO(sragss): Consider removing the option here.
+    fn square(&self) -> Self;
+    fn from_bytes(bytes: &[u8]) -> Self;
+}
+
+impl JoltField for ark_bn254::Fr {
+    const BYTES: usize = 64;
```
moodlezoup, line 57: shouldn't this be 32?

moodlezoup: also nit: `NUM_BYTES` or `BYTE_LEN` or something

jolt-core/src/poly/field.rs
```diff
@@ -0,0 +1,91 @@
...
+    fn one() -> Self {
+        <Self as ark_std::One>::one()
+    }
+
+    fn from_u64(n: u64) -> Option<Self> {
+        <Self as ark_ff::PrimeField>::from_u64(n)
+    }
+
+    fn square(&self) -> Self {
+        self * self
```
moodlezoup, line 84: I think `self.square()` might be more efficient

jolt-core/src/poly/commitment/commitment_scheme.rs
```diff
@@ -0,0 +1,97 @@
...
+    type Proof: Sync + Send + CanonicalSerialize + CanonicalDeserialize;
+    type BatchedProof: Sync + Send + CanonicalSerialize + CanonicalDeserialize;
+
+    fn generators(shapes: &[GeneratorShape]) -> Self::Generators;
+    fn commit(poly: &DensePolynomial<Self::Field>, gens: &Self::Generators) -> Self::Commitment;
+    fn batch_commit(
+        evals: &[&[Self::Field]],
+        gens: &Self::Generators,
+        batch_type: BatchType,
+    ) -> Vec<Self::Commitment>;
```
moodlezoup, line 47: in general, a batched commitment isn't necessarily a vector of `Self::Commitment`, so we may want to have a separate associated type for `BatchedCommitment`

jolt-core/src/poly/commitment/commitment_scheme.rs
```diff
@@ -0,0 +1,97 @@
...
+pub enum BatchType {
+    Big,
+    Small,
+    SurgeInitFinal,
+    SurgeReadWrite,
+}
+
+pub trait CommitmentScheme: Clone + Sync + Send + 'static {
+    type Field: JoltField;
+    type Generators: Clone + Sync + Send;
```
moodlezoup, line 36: In general a commitment scheme may not use generators. 
At the risk of using `Preprocessing` too much: 
```suggestion
    type Preprocessing: Clone + Sync + Send;
```

jolt-core/src/jolt/vm/mod.rs
```diff
@@ -226,13 +223,14 @@ pub trait Jolt<F: PrimeField, G: CurveGroup<ScalarField = F>, const C: usize, co
...
-            ReadWriteMemory::<F, G>::num_generators(max_memory_address, max_trace_length);
-        let timestamp_range_check_generators =
-            TimestampValidityProof::<F, G>::num_generators(max_trace_length);
+    ) -> JoltPreprocessing<F, CS> {
+        let bytecode_generator_shapes =
+            BytecodePolynomials::<F, CS>::generator_shapes(max_bytecode_size, max_trace_length);
+        let ram_generator_shapes =
+            ReadWriteMemory::<F, CS>::generator_shapes(max_memory_address, max_trace_length);
+        let timestamp_range_check_generator_shapes =
+            TimestampValidityProof::<F, CS>::generator_shapes(max_trace_length);
```
moodlezoup, line 232: Maybe instead of these `generator_shapes` functions (which will end up being a misnomer for Binius), we should instead have `pcs_preprocess` functions which mutate a common `CS::Preprocessing` variable

moodlezoup (commented):
looking good, my main suggestion is to rework the PCS-related preprocessing to be more generic (for Hyrax preprocessing == setting up the generators, but in general it may be something different or even a no-op)

jolt-core/src/poly/commitment/commitment_scheme.rs
sragss: Switch to `Setup` to avoid conflicting with vm::Preprocessing

jolt-core/src/jolt/vm/mod.rs
sragss: Switching to `GeneratorShape` -> `CommitShape`. Idea is that the commitments have a constant shape and then `PCS::setup(shapes: &[CommitShape]) -> PCS::Setup` can figure out how to do setup based on that. For EC commits this will be doing generator stuff, for hash based, no_op as you mention above.

In practice, we don't use the `StructuredCommitment` anymore so the `commitment_shape() -> Vec<CommitShape>` function on {bytecode, RAM, instruction lookups} should really just be on `VM`. I'm going to punt on that for now.

jolt-core/src/poly/commitment/commitment_scheme.rs
sragss: Good point.

We depend on the ability to flatten the batch in a few places but notably `R1CSProof::format_commitments`. I'm not exactly sure what that means in the general case. We could have a `PCS::flatten_commitments(Vec<PCS::Commitment>) -> PCS::BatchCommitment`.

I suggest we punt on this until we need to handle this case for a specific PCS.

moodlezoup (approved):
lgtm!

PR #310, opened 2024-04-19
Fix bug in ram_byte_read closure

moodlezoup:


ncitron (approved):
LGTM! Looks like the string serialization issues are unrelated so I think we're good to merge. Will keep hunting down the postcard issue.

PR #318, opened 2024-04-23
Refactor Hyrax Batching

sragss:
Includes a `batch_size` param in calls to `CommitmentScheme::{batch_commit, batch_prove}` and includes the `batch_size` param in the `CommitmentScheme::BatchedProof` struct if needed.

Two unfortunate things:
- Leaky abstraction: Hyrax is the only commitment scheme we know about that needs the `batch_size` param to be used as the Hyrax sizing ratio. Theoretically this could be refactored to be an associated type on `CommitmentScheme` called `BatchCommitConfig` which could store other info, we just aren't aware of another commitment scheme that needs it.
- Bad parameter naming: `batch_size` isn't quite the correct parameter name. Currently we call `CommitmentScheme::{batch_commit, batch_prove}` for a range of *real* batch sizes {163, 40}. Across all of these `NUM_R1CS_POLYS=64` is close enough to get desired performance. Also `batch_prove` is often passed a subset of the batch's polynomials, but still needs the matching `batch_size` param to whatever was committed. We could return to `RATIO` but only makes sense in the context of Hyrax.

### DIFF
--- a/common/src/constants.rs
+++ b/common/src/constants.rs
@@ -2,7 +2,6 @@ pub const XLEN: usize = 32;
 pub const REGISTER_COUNT: u64 = 32;
 pub const BYTES_PER_INSTRUCTION: usize = 4;
 pub const MEMORY_OPS_PER_INSTRUCTION: usize = 7;
-pub const NUM_R1CS_POLYS: usize = 64;
 
 pub const RAM_START_ADDRESS: u64 = 0x80000000;
 pub const DEFAULT_MEMORY_SIZE: u64 = 10 * 1024 * 1024;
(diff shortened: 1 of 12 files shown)


sragss:
Currently `batch_commit` is called from `impl StructuredOpeningProof for {Bytecode, RAM, InstructionLookups}` / `VM::Commit` / `R1CSProof::compute_witness_commit`. 

`StructuredOpeningProof` is not used anymore for Read/Write across {Bytecode, RAM, InstructionLookups} in favor of `VM::commit`. It is still used for Init / Final but these have a `batch_size` / `RATIO` of 1.

`R1CSProof::compute_witness_commit` should likely be refactored to `R1CSProof::compute_witness` and `VM::commit` should handle the batch commitment.

A decent solution would be to switch the `batch_size` param to an enum `CS::BigBatch / CS::SmallBatch` which for Hyrax would correspond to `NUM_R1CS_POLYS` / `1`.

PR #327, opened 2024-04-25
Update links and add more instructions to install the book locally

MatteoMer:
- A link of the book is invalid: https://jolt.a16zcrypto.com/how_it_works.html (Jolt link)
- The Background note is missing the batched polynomial opening notes in the summary https://jolt.a16zcrypto.com/background.html
- Added more instructions on how to install the book locally https://jolt.a16zcrypto.com/dev/install.html


moodlezoup (approved):
thanks!

PR #329, opened 2024-04-26
Fix: Chaining Hashes in `sha2-chain` Example

shreyas-londhe:
This PR fixes the chaining of hashes in the `sha2-chain` example.

### DIFF
--- a/examples/sha2-chain/src/main.rs
+++ b/examples/sha2-chain/src/main.rs
@@ -1,23 +1,9 @@
-use sha2::{Digest, Sha256};
-
-fn compute_native(input: [u8; 32], num_iters: u32) -> [u8; 32] {
-    let mut hash = input;
-    for _ in 0..num_iters {
-        let mut hasher = Sha256::new();
-        hasher.update(&hash);
-        let res = &hasher.finalize();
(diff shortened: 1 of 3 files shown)


examples/sha2-chain/src/main.rs
```diff
@@ -1,11 +1,28 @@
...
+
+    hash
+}
+
 pub fn main() {
     let (prove_sha2_chain, verify_sha2_chain) = guest::build_sha2_chain();
 
     let input = [5u8; 32];
     let iters = 100;
+    let native_output = compute_native(input, iters);
```
ncitron, line 20: You should be able to run `guest::sha2_chain` which will support native computation of the function rather than rewriting.

shreyas-londhe:
Made the required changes!

moodlezoup:
@shreyas-londhe clippy and cargo fmt are still failing, could you run `cargo clippy --fix && cargo fmt`?

shreyas-londhe:
@moodlezoup I did run it in the `jolt` directory, it does not show any errors for me.

sragss (approved):
LGTM, thanks for the assist!

PR #333, opened 2024-04-26
Wiki: Writeup of ZK

### LINKED ISSUE #330: Wiki: Explanation of ZK Strategy
ZK is not a high priority for us at the moment, but we should add an explanation of how we would achieve ZK in different settings.

sragss:
Fixes #330 

Written by @GUJustin


sragss:
Wow, nice. LGTM!

Issue #292, opened 2024-04-15
Parallelize RAM Timestamp Count Initialization


sragss:
For a 64 core machine at a cycle count of ~16M, Jolt spends ~3.5% of its time in a segment called `memory_trace_processing` [here](https://github.com/a16z/jolt/blob/main/jolt-core/src/jolt/vm/read_write_memory.rs#L384). This segment allocates and computes the offline memory checking (a,v,t) polynomials used for the combined registers and RAM. Some additional details can be found in the [wiki](https://jolt.a16zcrypto.com/how/memory-checking.html#offline-memory-checking). 

Currently this ~300 line segment takes ~3.5% of end-to-end time because it is computed completely serially. No use of additional CPU cores. We should parallelize this to get up to a `NUM_CPU`x speedup.

The goal is to fill out the following.
Trace Length sized
- `a_ram`
- `v_read`
- `v_write_rd`
- `v_write_ram`
- `t_read`
- `t_write_ram`
Memory sized
- `v_final`
- `t_final`

PatStiles:
Interested!

PR #338, opened 2024-05-01
Parallelize Memory Trace Processing

tahsintunan:
### Issue
https://github.com/a16z/jolt/issues/292

### Approach
Parallelized the processing of registers and RAM, improving the performance of the `memory_trace_processing` segment (**~30% improvement on M2 Pro**). However, the overall performance is still largely bottlenecked by the RAM task, which takes longer to complete compared to the register task.

Ideally, the next step would be to further subdivide the RAM task into smaller subtasks based on address ranges to distribute the workload across multiple cores. However, testing with `fib_e2e` and `sha3_e2e` revealed an uneven distribution of RAM accesses across the address space. That means, if we break down the >=32 address space into **_N_** fixed-sized blocks, for example, and assign one CPU per block, then practically only 2-3 cores will be working while the rest will sit idle.


common/src/constants.rs
```diff
@@ -2,6 +2,9 @@ pub const XLEN: usize = 32;
 pub const REGISTER_COUNT: u64 = 32;
 pub const BYTES_PER_INSTRUCTION: usize = 4;
 pub const MEMORY_OPS_PER_INSTRUCTION: usize = 7;
+pub const REG_OPS_PER_INSTRUCTION: usize = 3;
+pub const RAM_OPS_PER_INSTRUCTION: usize = 4;
+pub const NUM_R1CS_POLYS: usize = 64;
```
moodlezoup, line 7: this looks like a merge conflict artifact
```suggestion
```

common/src/constants.rs
```diff
@@ -2,6 +2,9 @@ pub const XLEN: usize = 32;
 pub const REGISTER_COUNT: u64 = 32;
 pub const BYTES_PER_INSTRUCTION: usize = 4;
 pub const MEMORY_OPS_PER_INSTRUCTION: usize = 7;
+pub const REG_OPS_PER_INSTRUCTION: usize = 3;
+pub const RAM_OPS_PER_INSTRUCTION: usize = 4;
```
moodlezoup, line 6: for clarity: 
```suggestion
pub const REG_OPS_PER_INSTRUCTION: usize = 3;
pub const RAM_OPS_PER_INSTRUCTION: usize = 4;
pub const MEMORY_OPS_PER_INSTRUCTION: usize = REG_OPS_PER_INSTRUCTION + RAM_OPS_PER_INSTRUCTION;
```

jolt-core/src/jolt/vm/read_write_memory.rs
```diff
@@ -286,6 +300,26 @@ where
...
+) -> [Vec<u64>; MEMORY_OPS_PER_INSTRUCTION] {
+    assert!(MEMORY_OPS_PER_INSTRUCTION == REG_OPS_PER_INSTRUCTION + RAM_OPS_PER_INSTRUCTION);
+    let mut merged_arr: [Vec<u64>; MEMORY_OPS_PER_INSTRUCTION] =
+        std::array::from_fn(|_| Vec::with_capacity(memory_trace_len));
+
+    merged_arr.par_iter_mut().enumerate().for_each(|(i, v)| {
+        if i < REG_OPS_PER_INSTRUCTION {
+            *v = reg_arr[i].clone();
+        } else {
+            *v = ram_arr[i - REG_OPS_PER_INSTRUCTION].clone();
```
moodlezoup, line 316: I think we can use `std::mem::take` instead of cloning here

jolt-core/src/jolt/vm/read_write_memory.rs
```diff
@@ -23,7 +28,8 @@ use crate::{
 use ark_serialize::{CanonicalDeserialize, CanonicalSerialize};
 use common::constants::{
     memory_address_to_witness_index, BYTES_PER_INSTRUCTION, MEMORY_OPS_PER_INSTRUCTION,
-    RAM_START_ADDRESS, REGISTER_COUNT,
+    NUM_R1CS_POLYS, RAM_OPS_PER_INSTRUCTION, RAM_START_ADDRESS, REGISTER_COUNT,
```
moodlezoup, line 31:
```suggestion
    RAM_OPS_PER_INSTRUCTION, RAM_START_ADDRESS, REGISTER_COUNT,
```

jolt-core/src/jolt/vm/read_write_memory.rs
```diff
@@ -355,354 +383,549 @@ impl<F: JoltField, C: CommitmentScheme<Field = F>> ReadWriteMemory<F, C> {
...
+                                        t_final_ram[remapped_a_index],
+                                    ));
+                                }
+                                #[cfg(test)]
+                                {
+                                    w_tuples_ram
+                                        .lock()
+                                        .unwrap()
+                                        .insert((remapped_a, v, timestamp));
+                                }
```
moodlezoup, line 483: nit (here and elsewhere)
```suggestion
                                #[cfg(test)]
                                {
                                    r_tuples_ram.lock().unwrap().insert((
                                        remapped_a,
                                        v,
                                        t_final_ram[remapped_a_index],
                                    ));
                                    w_tuples_ram
                                        .lock()
                                        .unwrap()
                                        .insert((remapped_a, v, timestamp));
                                }
```

moodlezoup (commented):
Nice work! Just a couple of nits. Could you also make sure to run `cargo fmt` and clippy? 🙏

jolt-core/src/jolt/vm/read_write_memory.rs
tahsintunan: I separated it into two blocks to release mutex locks immediately after each write, but I think it should be okay to keep them in a single block for readability (since it's just tests).

Issue #343, opened 2024-05-03
Potential Bug: timestamp_range_check commitments length does not match openings length


sragss:
For `fib_e2e` (and presumably elsewhere), timestamp_range_check passes 35 openings and 38 commitments to Hyrax.

Relevant snippet [here](https://github.com/a16z/jolt/blob/main/jolt-core/src/jolt/vm/timestamp_range_check.rs#L708-L733).

sragss:
After fixing lets add an assert to the batched hyrax opening proof to avoid in the future.

PR #344, opened 2024-05-04
Init Binius

sragss:
- Sets up the Binius fields
- Moves Jolt field to `src/field/mod.rs`
- Creates `MockCommitmentScheme`
- Generalize `rv32i_vm.rs – fib_e2e` to support multiple fields
- Implement Binius subtable tests
- Fix `lt` and `ltu` subtables which had used 2 * x * y as a shortcut
- Updates nightly version (Binius requires newer features)


rust-toolchain.toml
```diff
@@ -1,4 +1,4 @@
 [toolchain]
-channel = "nightly-2023-09-22"
+channel = "nightly-2024-04-20"
```
moodlezoup, line 2: nice

jolt-core/benches/iai.rs
```diff
@@ -45,7 +42,8 @@ fn bench_msm<G: CurveGroup>(input: (Vec<G>, Vec<G::ScalarField>)) -> G {
 #[bench::long(bound_poly_setup::<Fr>(4096))]
 fn bench_polynomial_binding<F: JoltField>(input: (DensePolynomial<F>, F)) {
     let (mut poly, val) = input;
-    black_box(poly.bound_poly_var_top(&val));
+    poly.bound_poly_var_top(&val);
+    black_box(());
```
moodlezoup, line 46:
```suggestion

```

jolt-core/src/subprotocols/sumcheck.rs
```diff
@@ -100,11 +100,11 @@ impl<F: JoltField> CubicSumcheckParams<F> {
         if *l == F::one() && *r == F::one() {
             *eq
         } else if *l == F::one() {
-            *r * eq
+            *r * *eq
```
moodlezoup, line 103: we should make it so that you can perform arithmetic operations on `&F` values without dereferencing
https://stackoverflow.com/a/59547146

jolt-core/src/jolt/subtable/eq_msb.rs
```diff
@@ -57,4 +59,10 @@ mod test {
         Fr,
         256
     );
+    subtable_materialize_mle_parity_test!(
+        eq_binius_msb_materialize_mle_parity,
```
moodlezoup, line 63:
```suggestion
        eq_msb_binius_materialize_mle_parity,
```

jolt-core/src/poly/commitment/mod.rs
```diff
@@ -1,3 +1,5 @@
+pub mod binius;
 pub mod commitment_scheme;
 pub mod hyrax;
+pub mod mock;
```
moodlezoup, line 4: can we put this under `#[cfg(test)]`?

jolt-core/src/jolt/vm/rv32i_vm.rs
```diff
@@ -189,8 +191,26 @@ mod tests {
         );
     }
 
+    // TODO(#343): Solve bug, reimplement.
```
moodlezoup, line 194: #343 should be a one-line fix,
Change [this](https://github.com/a16z/jolt/blob/main/jolt-core/src/jolt/vm/timestamp_range_check.rs#L719) to:
```
[1 + MEMORY_OPS_PER_INSTRUCTION + 5..1 + 2 * MEMORY_OPS_PER_INSTRUCTION + 5];
```

(and optionally add that assert to hyrax)

jolt-core/src/poly/commitment/mock.rs
```diff
@@ -0,0 +1,133 @@
...
+#[derive(CanonicalSerialize, CanonicalDeserialize)]
+pub struct MockProof<F: JoltField> {
+    opening_point: Vec<F>,
+}
+
+#[derive(CanonicalSerialize, CanonicalDeserialize)]
+pub struct MockBatchedProof {}
+
+#[derive(Clone)]
+pub struct None {}
```
moodlezoup, line 41: nit:
```suggestion
pub struct None;
```

jolt-core/src/poly/commitment/mock.rs
```diff
@@ -0,0 +1,133 @@
...
+    }
+}
+
+#[derive(CanonicalSerialize, CanonicalDeserialize)]
+pub struct MockProof<F: JoltField> {
+    opening_point: Vec<F>,
+}
+
+#[derive(CanonicalSerialize, CanonicalDeserialize)]
+pub struct MockBatchedProof {}
```
moodlezoup, line 38: nit:
```suggestion
pub struct MockBatchedProof;
```

moodlezoup: I don't think we use this anywhere actually

jolt-core/src/field/binius.rs
```diff
@@ -0,0 +1,185 @@
...
+        let field_element = bytemuck::try_from_bytes::<F>(bytes).unwrap();
+        Self(field_element.to_owned())
+    }
+}
+
+impl<F: BiniusSpecific> Neg for BiniusField<F> {
+    type Output = Self;
+
+    fn neg(self) -> Self {
+        BiniusField(self.0.neg())
```
moodlezoup, line 76: I assume the Binius field implement the `std::ops` traits?
if so we should do this for readability imo:
```suggestion
        BiniusField(-self.0)
```

jolt-core/src/field/binius.rs
```diff
@@ -0,0 +1,185 @@
...
+    }
+}
+
+impl<F: BiniusSpecific> Add for BiniusField<F> {
+    type Output = Self;
+
+    fn add(self, other: Self) -> Self {
+        BiniusField(self.0.add(other.0))
```
moodlezoup, line 84: same thing here and below
```suggestion
        BiniusField(self.0 + other.0)
```

moodlezoup (commented):
looking good

rust-toolchain.toml
sragss: ;)

jolt-core/src/subprotocols/sumcheck.rs
sragss: Binius' fields don't implement these by reference so it's hard for us to do it with the same automatic method. #380

sragss:
Fixes #343

PR #354, opened 2024-05-09
move criterion to dev-dependencies

flyq:
- move criterion to dev-dependencies
- ~~move sumcheck bench code out of src code~~(The recent commit in the main branch has deleted the relevant bench code and significantly modified some interfaces in sumcheck, so sumcheck bench code will be deleted directly.)


moodlezoup (approved):
thanks!

PR #355, opened 2024-05-10
refact: EqPolynomial::evals

### LINKED ISSUE #351: Refactor `EqPolynomial::evals`
https://github.com/a16z/jolt/blob/main/jolt-core/src/poly/eq_poly.rs

Currently the function signature is `pub fn evals(&self) -> Vec<F>`
If we change this to `pub fn evals(r: &[F]) -> Vec<F>`, we should be able to avoid some unnecessary cloning throughout the codebase.

SuccinctPaul:
* Resolved
 #351


moodlezoup:
Thanks for this! could you fix the clippy error? otherwise looks good to me

SuccinctPaul:
> Thanks for this! could you fix the clippy error? otherwise looks good to me

Thx for quick response. It's been done.

moodlezoup (approved):
Nice work!

PR #357, opened 2024-05-10
Efficent opening proofs in verifier

### LINKED ISSUE #350: Efficient opening proofs in `read_write_memory.rs`
There are three `TODO`s in [`read_write_memory.rs`](https://github.com/a16z/jolt/blob/main/jolt-core/src/jolt/vm/read_write_memory.rs) of the form "TODO: Compute opening without instantiating * polynomial itself"

Currently, these openings require the verifier to do work linear in the memory used. These openings should be linear in `|bytecode| + |program IO|`. Assuming the verifier knows the bytecode and program inputs/outputs, it can evaluate the necessary openings without reconstructing the full memory MLE.

aleph-v:
Resolves two of the three todos reducing the memory use of the verifier when calculating the evaluations of the initial memory polynomials. Resolves #350

### DIFF
--- a/jolt-core/src/jolt/vm/read_write_memory.rs
+++ b/jolt-core/src/jolt/vm/read_write_memory.rs
@@ -1621,21 +1621,30 @@ where
 
         let memory_layout = &preprocessing.program_io.as_ref().unwrap().memory_layout;
 
-        // TODO(moodlezoup): Compute openings without instantiating io_witness_range polynomial itself
-        let memory_size = proof.num_rounds.pow2();
-        let io_witness_range: Vec<_> = (0..memory_size as u64)
+        let nonzero_memory_size = memory_layout.ram_witness_offset as usize;
+        let log_nonzero_memory_size = nonzero_memory_size.trailing_zeros() as usize;
+        assert_eq!(
+            1 << log_nonzero_memory_size,
+            nonzero_memory_size,
+            "Ram witness offset must be a power of two"
+        );
+
+        let io_witness_range: Vec<_> = (0..nonzero_memory_size as u64)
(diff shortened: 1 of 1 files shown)


jolt-core/src/jolt/vm/read_write_memory.rs
```diff
@@ -1621,21 +1621,30 @@ where
 
         let memory_layout = &preprocessing.program_io.as_ref().unwrap().memory_layout;
 
-        // TODO(moodlezoup): Compute openings without instantiating io_witness_range polynomial itself
-        let memory_size = proof.num_rounds.pow2();
-        let io_witness_range: Vec<_> = (0..memory_size as u64)
+        let nonzero_memory_size = memory_layout.ram_witness_offset as usize;
+        let log_nonzero_memory_size = nonzero_memory_size.trailing_zeros() as usize;
```
moodlezoup, line 1625:
```suggestion
        let log_nonzero_memory_size = nonzero_memory_size.log_2();
```

jolt-core/src/jolt/vm/read_write_memory.rs
```diff
@@ -1621,21 +1621,30 @@ where
...
+        assert_eq!(
+            1 << log_nonzero_memory_size,
+            nonzero_memory_size,
+            "Ram witness offset must be a power of two"
+        );
```
moodlezoup, line 1630:
```suggestion
        assert!(
            nonzero_memory_size.is_power_of_two(),
            "Ram witness offset must be a power of two"
        );
```

moodlezoup (commented):
Two nits but otherwise nice work!

PR #358, opened 2024-05-12
feat(jolt-core): Adding MOVSIGN instruction

mw2000:
We now support the MOVSIGN instruction as part of the RISCV M extensions

### DIFF
--- a/jolt-core/src/jolt/instruction/movsign.rs
+++ b/jolt-core/src/jolt/instruction/movsign.rs
@@ -1,0 +1,119 @@
+use rand::prelude::StdRng;
+use rand::RngCore;
+use serde::{Deserialize, Serialize};
+
+use super::JoltInstruction;
+use crate::{
+    jolt::{
+        instruction::SubtableIndices,
+        subtable::{sign_extend::SignExtendSubtable, LassoSubtable},
+    },
+    poly::field::JoltField,
+    utils::instruction_utils::chunk_operand_usize,
+};
+
+#[derive(Copy, Clone, Default, Debug, Serialize, Deserialize)]
+pub struct MOVSIGNInstruction<const WORD_SIZE: usize>(pub u64);
+
+// Constants for 32-bit and 64-bit word sizes
+const NEGATIVE_32: u64 = 0xFFFF_FFFF;
+const NEGATIVE_64: u64 = 0xFFFF_FFFF_FFFF_FFFF;
+const POSITIVE: u64 = 0;
+const SIGN_BIT_32: u64 = 0x8000_0000;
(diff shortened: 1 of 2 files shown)


jolt-core/src/jolt/instruction/movsign.rs
```diff
@@ -0,0 +1,119 @@
...
+
+impl<const WORD_SIZE: usize> JoltInstruction for MOVSIGNInstruction<WORD_SIZE> {
+    fn operands(&self) -> (u64, u64) {
+        (self.0, 0)
+    }
+
+    fn combine_lookups<F: JoltField>(&self, vals: &[F], C: usize, _: usize) -> F {
+        let val = vals[0];
+        let mut result = F::zero();
+        for i in 0..C {
```
moodlezoup, line 33: actually could we change this to: 
```suggestion
        for i in 0..WORD_SIZE / 16 {
```

jolt-core/src/jolt/instruction/movsign.rs
```diff
@@ -0,0 +1,119 @@
...
+
+    use crate::{jolt::instruction::JoltInstruction, jolt_instruction_test};
+
+    use super::MOVSIGNInstruction;
+
+    #[test]
+    fn movsign_instruction_32_e2e() {
+        let mut rng = test_rng();
+        const C: usize = 2;
+        const M: usize = 1 << 16;
```
moodlezoup, line 97: Once we make the above change, we should be able to change this to:
```suggestion
        const C: usize = 4;
        const M: usize = 1 << 16;
```

jolt-core/src/jolt/instruction/movsign.rs
```diff
@@ -0,0 +1,119 @@
...
+            let instruction = MOVSIGNInstruction::<32>(x);
+            jolt_instruction_test!(instruction);
+        }
+    }
+
+    #[test]
+    fn movsign_instruction_64_e2e() {
+        let mut rng = test_rng();
+        const C: usize = 4;
+        const M: usize = 1 << 16;
```
moodlezoup, line 111: And similarly: 
```suggestion
        const C: usize = 8;
        const M: usize = 1 << 16;
```

moodlezoup: (for context, Jolt assumes elsewhere that `C * log2(M) = 2 * WORD_SIZE`)

moodlezoup (commented):
Nice work, just a couple of nits

mw2000:
@moodlezoup addressed all nits, ready for re-review

moodlezoup (approved):
Thanks!

PR #360, opened 2024-05-14
forward-compatibility with future ark_ff

slumber:
We'd like to integrate Jolt as a dependency for [Nexus zkVM](https://github.com/nexus-xyz/nexus-zkvm). 
However, arkworks 0.4 has crucial bugs such as https://github.com/arkworks-rs/r1cs-std/issues/94 and several others.

Thus, nexus code is forced to use patched ark-* crates. Luckily, there aren't many changes needed to be forward compatible with potential arkworks 0.5, hence this PR.

(note that
```rust
    // From JoltField 

    fn double(&self) -> Self;
```
is unused and depends on `AdditiveGroup` instead of `Field` in the future. It's my assumption that it's safe to be removed.)

### DIFF
--- a/jolt-core/src/poly/field.rs
+++ b/jolt-core/src/poly/field.rs
@@ -49,7 +49,6 @@ pub trait JoltField:
     fn zero() -> Self;
     fn one() -> Self;
     fn from_u64(n: u64) -> Option<Self>;
-    fn double(&self) -> Self;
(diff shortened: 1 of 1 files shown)


moodlezoup:
I think we want to keep these methods around; we can patch as necessary when arkworks 0.5 is released

PR #368, opened 2024-05-21
feat: R1CS Refactor

sragss:
R1CS refactor for declarative R1CS programming. Supports appending constraints by any implementations of `R1CSConstraintBuilder` which will be important with the RISC-V extensions and Binius custom SNARKs.

Example:
```rust
struct TestConstraints();
impl<F: JoltField> R1CSConstraintBuilder<F> for TestConstraints {
    type Inputs = TestInputs;
    fn build_constraints(&self, builder: &mut R1CSBuilder<F, Self::Inputs>) {
        builder.constrain_prod(TestInputs::OpFlags0, TestInputs::OpFlags1, TestInputs::BytecodeA);
        let _aux = builder.allocate_prod(TestInputs::OpFlags2, TestInputs::OpFlags3);
    }
}
```

Each aux variable that is appended to the builder is appended along with a lambda to compute the aux from inputs.

Auxiliary variables are computed directly from linear combinations (`LC`) and (Az, Bz, Cz) are computed directly from linear combinations of the corresponding constraints. This appears to be slightly less efficient than before costing 1-2% end-to-end. 

Devex: `R1CSBuilder.assert_valid` will describe the exact constraint and step which failed.

PR introduces a massive number of tests for each and every part of R1CS and Spartan. Ignoring these tests the PR is diff neutral.


jolt-core/src/r1cs/builder.rs
```diff
@@ -0,0 +1,1443 @@
...
+    a: LC<I>,
+    b: LC<I>,
+    c: LC<I>,
+}
+
+impl<I: ConstraintInput> Constraint<I> {
+    #[cfg(test)]
+    fn is_sat(&self, inputs: &Vec<i64>) -> bool {
+        // Find the number of variables and the number of aux. Inputs should be equal to this combined length
+        let num_input = I::COUNT;
```
moodlezoup, line 33:
```suggestion
        let num_inputs = I::COUNT;
```

moodlezoup (commented):
still going through `builder.rs`...

jolt-core/src/r1cs/builder.rs
```diff
@@ -0,0 +1,1443 @@
...
+                let values = if let Some(range) = self.input_to_flat[input_index].clone() {
+                    &values[range]
+                } else {
+                    &[]
+                };
+                input_lc.evaluate(values)
+            })
+            .collect();
+        (self.compute)(&computed_inputs)
+    }
```
moodlezoup, line 163: it might be annoying to get this to work with the Rust compiler, but it would be cool if, instead of having this as a method, we generate this as a closure in `new` and store it as the `compute` field in `AuxComputation`. Then we could get rid of the `input_to_flat` field I think

moodlezoup: Alternatively, we could pass in all the inputs/aux values for a given step into this function. I'm not a huge fan of having `flat_vars` and `input_to_flat` as fields in the `AuxComputation` struct.

jolt-core/src/r1cs/builder.rs
```diff
@@ -0,0 +1,1443 @@
...
+        let constraint = Constraint { a, b, c };
+        self.constraints.push(constraint);
+    }
+
+    pub fn constrain_binary(&mut self, value: impl Into<LC<I>>) {
+        let one: LC<I> = Variable::Constant.into();
+        let value: LC<I> = value.into();
+        // value * (1 - value)
+        let constraint = Constraint {
+            a: value.clone(),
```
moodlezoup, line 249: I don't think we need this clone, because of the `into()` above? Here and in the other `constrain` functions

jolt-core/src/r1cs/builder.rs
```diff
@@ -0,0 +1,1443 @@
...
+            a: a_sparse,
+            b: b_sparse,
+            c: c_sparse,
+            num_vars: I::COUNT + self.num_aux(),
+            num_rows: self.constraints.len(),
+        }
+    }
+}
+
+pub type OffsetLC<I> = (bool, LC<I>);
```
moodlezoup, line 504: doc comment

jolt-core/src/r1cs/builder.rs
```diff
@@ -0,0 +1,1443 @@
...
+            .iter()
+            .filter(|term| matches!(term.0, Variable::Input(_) | Variable::Auxiliary(_)))
+            .for_each(|term| {
+                condition.offset_vars.push((
+                    self.uniform_builder.variable_to_column(term.0),
+                    constraint.condition.0,
+                    from_i64::<F>(term.1),
+                ))
+            });
+        if let Some(term) = sorted_condition.last() {
```
moodlezoup, line 644: I think this can be an unwrap? condition should never be empty

jolt-core/src/r1cs/builder.rs
```diff
@@ -0,0 +1,1443 @@
...
+            });
+
+        // Handle constants
+        lhs.terms().iter().for_each(|term| {
+            assert!(
+                !matches!(term.0, Variable::Constant),
+                "Constants only supported in RHS"
+            )
+        });
+        if let Some(term) = rhs.sorted_terms().last() {
```
moodlezoup, line 682: same here

jolt-core/src/r1cs/builder.rs
```diff
@@ -0,0 +1,1443 @@
...
+            }
+        }
+
+        NonUniformR1CS::new(eq, condition)
+    }
+
+    /// inputs should be of the format [[I::0, I::0, ...], [I::1, I::1, ...], ... [I::N, I::N]]
+    /// aux should be of the format [[Aux(0), Aux(0), ...], ... [Aux(self.next_aux - 1), ...]]
+    #[tracing::instrument(skip_all, name = "CombinedUniformBuilder::compute_spartan")]
+    pub fn compute_spartan(&self, inputs: &[Vec<F>], aux: &[Vec<F>]) -> (Vec<F>, Vec<F>, Vec<F>) {
```
moodlezoup, line 694: `compute_matrix_vector_products` or `compute_spartan_Az_Bz_Cz` or something

jolt-core/src/r1cs/ops.rs
```diff
@@ -0,0 +1,627 @@
...
+    }
+
+    pub fn terms_in_field<F: JoltField>(&self) -> Vec<F> {
+        self.terms()
+            .iter()
+            .map(|term| from_i64::<F>(term.1))
+            .collect()
+    }
+
+    pub fn sorted_terms(&self) -> Vec<Term<I>> {
```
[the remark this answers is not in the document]
sragss: Yup doing the latter

jolt-core/src/r1cs/ops.rs
```diff
@@ -0,0 +1,627 @@
...
+            LC::new(vec![self, other])
+        }
+    }
+}
+
+impl<I: ConstraintInput> std::ops::Sub for Term<I> {
+    type Output = LC<I>;
+
+    fn sub(self, other: Self) -> Self::Output {
+        LC::new(vec![self, -other])
```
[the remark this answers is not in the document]
sragss: Technically this is caught by the `#[cfg(test)] assert_no_duplicate` terms in `LC::new()` but we should probably be more explicit.

jolt-core/src/r1cs/jolt_constraints.rs
```diff
@@ -0,0 +1,307 @@
...
+            JoltIn::LookupOutput + 4,
+            4 * JoltIn::PcIn + PC_START_ADDRESS + 4,
+        );
+
+        let next_pc_jump_branch = cs.allocate_if_else(
+            branch_and_lookup_output,
+            4 * JoltIn::PcIn + PC_START_ADDRESS + imm_signed,
+            next_pc_jump,
+        );
+        assert_static_aux_index!(next_pc_jump_branch, PC_BRANCH_AUX_INDEX);
```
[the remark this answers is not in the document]
sragss: It constraints it to the constant `PC_BRANCH_AUX_INDEX` at runtime. This can then be used in `OffsetEqConstraints`.

TODO: Move the singular `OffsetEqConstraint` from `vm/mod.rs` into `jolt_constraints.rs`.

jolt-core/src/r1cs/builder.rs
sragss: Unfortunately we don't have ops by reference. Super cheap though so going to leave it.

jolt-core/src/r1cs/builder.rs
sragss: Pivoted to LC::constant_term()

jolt-core/src/r1cs/inputs.rs
```diff
@@ -0,0 +1,265 @@
...
+            chunks_y,
+            chunks_query,
+            lookup_outputs,
+            circuit_flags_bits,
+            instruction_flags_bits,
+        }
+    }
+
+    #[tracing::instrument(skip_all, name = "R1CSInputs::clone_to_trace_len_chunks")]
+    pub fn clone_to_trace_len_chunks(&self) -> Vec<Vec<F>> {
```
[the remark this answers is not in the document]
sragss: Mitigated with an assert on sizing. Leaving this issue to address when we have better static info about memory checking instances: #378 378

jolt-core/src/r1cs/inputs.rs
```diff
@@ -0,0 +1,265 @@
...
+        let bytecode_trace_commitments = &jolt_commitments.bytecode.trace_commitments;
+        let memory_trace_commitments = &jolt_commitments.read_write_memory.trace_commitments
+            [..1 + MEMORY_OPS_PER_INSTRUCTION + 5]; // a_read_write, v_read, v_write
+        let instruction_lookup_indices_commitments =
+            &jolt_commitments.instruction_lookups.trace_commitment[..C];
+        let instruction_flag_commitments = &jolt_commitments.instruction_lookups.trace_commitment
+            [jolt_commitments.instruction_lookups.trace_commitment.len() - RV32I::COUNT - 1
+                ..jolt_commitments.instruction_lookups.trace_commitment.len() - 1];
+
+        let mut combined_commitments: Vec<&C::Commitment> = Vec::new();
```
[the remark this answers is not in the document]
sragss: #378 have not dealt with this instance as it includes aux.

jolt-core/src/r1cs/jolt_constraints.rs
```diff
@@ -0,0 +1,307 @@
...
+    OpFlags_IsJmp,
+    OpFlags_IsBranch,
+    OpFlags_LookupOutToRd,
+    OpFlags_SignImm,
+    OpFlags_IsConcat,
+    OpFlags_IsVirtualSequence,
+    OpFlags_IsVirtual,
+
+    // Instruction Flags
+    // Should match JoltInstructionSet
```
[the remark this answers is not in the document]
sragss: #378

jolt-core/src/r1cs/key.rs
```diff
@@ -0,0 +1,627 @@
...
+                    acc * if bit == '0' {
+                        F::one() - r_var[j]
+                    } else {
+                        r_var[j]
+                    }
+                });
+
+                product * segment_evals[var_index]
+            })
+            .sum::<F>();
```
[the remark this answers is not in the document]
sragss: Smart, yes

jolt-core/src/jolt/vm/mod.rs
```diff
@@ -515,17 +533,96 @@ pub trait Jolt<F: JoltField, PCS: CommitmentScheme<Field = F>, const C: usize, c
...
+            BatchType::Big,
+        );
+        drop(_guard);
+
+        let io_comms = PCS::batch_commit(&[inputs.pc.as_ref()], generators, BatchType::Big);
+        let aux_comms = PCS::batch_commit(
+            &aux.iter().map(AsRef::as_ref).collect::<Vec<&[F]>>(),
+            generators,
+            BatchType::Big,
+        );
```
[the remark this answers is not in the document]
sragss: Yeah there's some clean up to do here, opened #382

jolt-core/src/r1cs/builder.rs
sragss: These have been removed as of #384

PR #370, opened 2024-05-27
feat: virtual instructions

Ethan-000:
i could be implementing them wrong. any review or comments are much appeciated :)


jolt-core/src/jolt/instruction/lte.rs
```diff
@@ -0,0 +1,134 @@
...
+            LTEInstruction(0, u32_max),
+            LTEInstruction(u32_max, 0),
+            LTEInstruction(u32_max, u32_max),
+            LTEInstruction(u32_max, 1 << 8),
+            LTEInstruction(1 << 8, u32_max),
+        ];
+        for instruction in instructions {
+            jolt_instruction_test!(instruction);
+        }
+    }
```
mw2000, line 133: We should ideally break this into seperate tests for u32 and u64 respectively to keep it consistent

jolt-core/src/jolt/instruction/virtual_assert_eq_signs.rs
```diff
@@ -0,0 +1,88 @@
...
+
+use super::{JoltInstruction, SubtableIndices};
+use crate::{
+    jolt::subtable::{eq_msb::EqMSBSubtable, LassoSubtable},
+    utils::instruction_utils::chunk_and_concatenate_operands,
+};
+
+#[derive(Copy, Clone, Default, Debug, Serialize, Deserialize)]
+pub struct EQSIGNSInstruction(pub u64, pub u64);
+
```
mw2000, line 14: Can we call this `ASSERTEQSIGNSInstruction` to stay consistent with the pattern?

jolt-core/src/jolt/instruction/virtual_assert_lt_abs.rs
```diff
@@ -0,0 +1,114 @@
...
+use crate::{
+    jolt::subtable::{
+        eq::EqSubtable, eq_abs::EqAbsSubtable, lt_abs::LtAbsSubtable, ltu::LtuSubtable,
+        LassoSubtable,
+    },
+    utils::instruction_utils::chunk_and_concatenate_operands,
+};
+
+#[derive(Copy, Clone, Default, Debug, Serialize, Deserialize)]
+pub struct LTABSInstruction<const WORD_SIZE: usize>(pub u64, pub u64);
```
mw2000, line 16: Can we call this `ASSERTLTABSInstruction` to stay consistent with the pattern?

jolt-core/src/jolt/instruction/lte.rs
```diff
@@ -0,0 +1,134 @@
+use rand::prelude::StdRng;
```
mw2000, line 1: Interestingly, I don't see the Paper describing this instruction explicitly, however we do use assert_lte in DIVU, would love to know if we want to switch this to `virtual_assert_lte.rs`

jolt-core/src/jolt/instruction/mulhsu.rs
```diff
@@ -0,0 +1,192 @@
...
+
+        let r_x = rng.next_u64() % 32;
+        let r_y = rng.next_u64() % 32;
+        let rd = rng.next_u64() % 32;
+
+        let x = rng.next_u32() as u64;
+        let y = if r_x == r_y { x } else { rng.next_u32() as u64 };
+        let result = ((i128::from(x as i32) * i128::from(y)) >> 32) as u32;
+
+        let mulh_trace_row = RVTraceRow {
```
mw2000, line 131: should be`mulhsu_trace_row`

mw2000 (changes requested):
Mostly nits and naming issues from my end for now, I think you've gotten this under wraps :)

Ethan-000:
done. thanks for the review :)

jolt-core/src/jolt/instruction/mulhsu.rs
```diff
@@ -0,0 +1,192 @@
...
+/// Perform signed*unsigned multiplication and return the upper WORD_SIZE bits
+pub struct MULHSUInstruction<const WORD_SIZE: usize>;
+
+impl<const WORD_SIZE: usize> VirtualInstructionSequence for MULHSUInstruction<WORD_SIZE> {
+    fn virtual_sequence(trace_row: RVTraceRow) -> Vec<RVTraceRow> {
+        assert_eq!(trace_row.instruction.opcode, RV32IM::MULHSU);
+        // MULHSU operands
+        let x = trace_row.register_state.rs1_val.unwrap();
+        let y = trace_row.register_state.rs2_val.unwrap();
+        // MULH source registers
```
moodlezoup, line 19:
```suggestion
        // MULHSU source registers
```

jolt-core/src/jolt/instruction/virtual_advice.rs
```diff
@@ -0,0 +1,72 @@
...
+    use ark_std::test_rng;
+    use rand_chacha::rand_core::RngCore;
+
+    use super::ADVICEInstruction;
+    use crate::{jolt::instruction::JoltInstruction, jolt_instruction_test};
+
+    #[test]
+    fn advice_instruction_32_e2e() {
+        let mut rng = test_rng();
+        const C: usize = 8;
```
moodlezoup, line 63:
```suggestion
        const C: usize = 4;
```

jolt-core/src/jolt/instruction/virtual_assert_eq_signs.rs
```diff
@@ -0,0 +1,88 @@
...
+        vec![(Box::new(EqMSBSubtable::new()), SubtableIndices::from(0))]
+    }
+
+    fn to_indices(&self, C: usize, log_M: usize) -> Vec<usize> {
+        chunk_and_concatenate_operands(self.0, self.1, C, log_M)
+    }
+
+    fn lookup_entry(&self) -> u64 {
+        ((self.0 as i32).signum() == (self.1 as i32).signum()
+            || (self.0 as i32).signum() >= 0 && (self.1 as i32).signum() >= 0)
```
moodlezoup, line 42: I think the first clause is unnecessary?
```suggestion
        (self.0 as i32).signum() >= 0 && (self.1 as i32).signum() >= 0)
```

jolt-core/src/jolt/instruction/virtual_advice.rs
```diff
@@ -0,0 +1,72 @@
...
+
+    fn subtables<F: JoltField>(
+        &self,
+        C: usize,
+        _: usize,
+    ) -> Vec<(Box<dyn LassoSubtable<F>>, SubtableIndices)> {
+        vec![(
+            Box::new(IdentitySubtable::new()),
+            SubtableIndices::from(0..C),
+        )]
```
moodlezoup, line 35: this is pretty subtle, but for the ADVICE instruction we need to range-check that prover-supplied value is in `[0, 2^WORD_SIZE)`
By using `C` IdentitySubtables here, we're checking that the value is in the range `[0, M^C)` instead. For RV32 we use M=2^16 and C=4, so this range would be `[0, 2^64)` instead of `[0, 2^32)`. 

Instead, we can do what [`ADDInstruction`](https://github.com/a16z/jolt/blob/main/jolt-core/src/jolt/instruction/add.rs#L33-L49) does, using a combination of TruncateOverflowSubtable and IdentitySubtable to ensure that the output is at most `WORD_SIZE` bits

moodlezoup (commented):
Nice work! This is looking really good. One thing to change for ADVICE + a couple of nits

jolt-core/src/jolt/instruction/virtual_assert_eq_signs.rs
Ethan-000: was trying to check the case where both signs are negative. made it more explicit https://github.com/a16z/jolt/pull/370/commits/90f5cb5b9f88423ea096e12ebbef45d129c56f79

moodlezoup (approved):
Nice work! 🚀

PR #372, opened 2024-05-29
Memory usage utils

moodlezoup:
Adds util functions to print current RAM usage and report memory usage deltas for "memory tracing spans"


sragss (approved):
Is there any overhead associated with memory_stats calls?

moodlezoup:
> Is there any overhead associated with memory_stats calls?

I think there's a little overhead, so if we want to add memory tracing spans we should put them behind a gate

PR #339, opened 2024-05-03
Feat(Zeromorph PCS): Add Zeromorph Multilinear PCS Scheme

PatStiles:
Refactors #66 for new Jolt CommitmentScheme Trait and addresses #249.

Issue #208, opened 2024-03-25
Add support for Zeromorph and/or HyperKZG commitments in addition to Hyrax


GUJustin:
Zeromorph is described in this paper: https://eprint.iacr.org/2023/917. But it's very focused on hiding/zk, which is not an initial priority for Jolt. 

HyperKZG is described and implemented here: https://github.com/microsoft/Nova/blob/main/src/provider/hyperkzg.rs

Switching to these commitment schemes will reduce verifier costs to logarithmic and slightly speed up the prover too.

Ethan-000:
ill give a try at porting hyperkzg based on https://github.com/a16z/jolt/pull/339 if nobody else is working on it

moodlezoup:
That would be amazing!

PR #373, opened 2024-06-03
feat: hyperkzg pcs

Ethan-000:
based on this branch https://github.com/a16z/jolt/pull/339

and modifications of https://github.com/microsoft/Nova/blob/main/src/provider/hyperkzg.rs

related to https://github.com/a16z/jolt/issues/208


moodlezoup (commented):
Thanks for this! I know this is probably an artifact of the Nova codebase it's forked from, but could you refactor the closures out of `open` and `verify`? I find the closures (and especially the nested closures) hard to read/maintain.

Ethan-000:
> Thanks for this! I know this is probably an artifact of the Nova codebase it's forked from, but could you refactor the closures out of `open` and `verify`? I find the closures (and especially the nested closures) hard to read/maintain.

will do :)

Issue #314, opened 2024-04-22
Add wasm target


Levike34:
Would be very useful if jolt could be compiled to a WASM executable

sebasti810:
I’ve started to look into this, but want to get some clarification before I spend too much time on the wrong path - should we also have first class support for the prover, or is the verifier sufficient ([as done here](https://github.com/a16z/jolt/issues/209#issuecomment-2055581251))? I’m assuming the big selling point here would be embedding verifiers into webapps, but does it make sense to make the prover wasm compatible as well?

One way we could do this is by introducing a new jolt command, e.g. `jolt create-wasm …`, which could take over the process of creating the required components and then generating the WASM-compatible binaries… should I continue down this road?

larron:
@sebasti810 Support for in browser verifiable compute would be awesome, IE: [Noir](https://medium.com/aztec-protocol/announcing-noirjs-privacy-preserving-zk-applications-in-your-browser-d7d78692f15a)

PR #374, opened 2024-06-04
feat: add WASM compatibility for verification functions

sebasti810:
This pull request introduces WebAssembly (WASM) compatibility for the verification process of Jolt. The main changes include:

- Extension of `jolt new` command with `--wasm` flag:
    - The `jolt new <project-name> --wasm` command modifies both the `root/Cargo.toml` and `guest/Cargo.toml` files to include the necessary dependencies for WASM. It also marks the project as a `cdylib` crate type, which is required for WASM compilation.
- Addition of `jolt build-wasm` command
   - The `build-wasm` command extracts all functions declared with `jolt::provable`. It then performs preprocessing steps to build and decode the program, storing the resulting data in binary form. An example `index.html` file is created to demonstrate how to use the functions generated by the macro extension. Finally, optimized WASM files are created using `wasm-pack` to make the project ready for use. 
 - Extension of macros:
    - The macros are enhanced to check if functions are marked with `jolt::provable(wasm)`. If so, necessary imports and helper functions are defined. The output of the preprocessing process is then used to build the WASM-compatible verify-functions. 

**Current Limitations**
- Dependencies used in `guest/src/lib.rs` need to be defined both in `guest/Cargo.toml` and `root/Cargo.toml`. This redundancy is known and could be optimized in future iterations.
- Currently, this only provides WASM support for proof verification. The proving itself is not yet WASM-compatible, bc it's unclear whether this is a desired requirement.

#314

### DIFF
--- a/jolt-sdk/macros/src/lib.rs
+++ b/jolt-sdk/macros/src/lib.rs
@@ -33,6 +33,7 @@ pub fn provable(attr: TokenStream, item: TokenStream) -> TokenStream {
 
     let mut token_stream = builder.build();
 
+    // Add wasm utilities and functions if the function is marked as wasm
     if builder.has_wasm_attr() {
         // wasm utilities should only be added once
         let mut module_added = WASM_IMPORTS_ADDED.lock().unwrap();
--- a/jolt-sdk/macros/src/lib.rs
+++ b/jolt-sdk/macros/src/lib.rs
@@ -23,7 +23,16 @@ use syn::{
 pub fn provable(attr: TokenStream, item: TokenStream) -> TokenStream {
     let attr = parse_macro_input!(attr as AttributeArgs);
     let func = parse_macro_input!(item as ItemFn);
-    MacroBuilder::new(attr, func).build()
+    let builder = MacroBuilder::new(attr, func);
+
+    let mut token_stream = builder.build();
+
+    if builder.has_wasm_attr() {
+        let wasm_token_stream: TokenStream = builder.make_wasm_function().into();
+        token_stream.extend(wasm_token_stream);
+    }
+
+    token_stream
 }
 
 struct MacroBuilder {
@@ -67,11 +76,17 @@ impl MacroBuilder {
         };
 
         quote! {
+            #[cfg(not(target_arch = "wasm32"))]
             #build_fn
+            #[cfg(not(target_arch = "wasm32"))]
             #execute_fn
+            #[cfg(not(target_arch = "wasm32"))]
             #analyze_fn
+            #[cfg(not(target_arch = "wasm32"))]
             #preprocess_fn
+            #[cfg(not(target_arch = "wasm32"))]
             #prove_fn
+            #[cfg(not(target_arch = "wasm32"))]
             #main_fn
         }
         .into()
@@ -445,6 +460,8 @@ impl MacroBuilder {
 
     fn parse_attributes(&self) -> Attributes {
         let mut attributes = HashMap::<_, u64>::new();
+        let mut wasm = false;
+
         for attr in &self.attr {
             match attr {
                 NestedMeta::Meta(Meta::NameValue(MetaNameValue { path, lit, .. })) => {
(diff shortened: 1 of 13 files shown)


jolt-sdk/macros/src/lib.rs
```diff
@@ -11,19 +11,42 @@ use common::{
...
 use proc_macro2::TokenStream as TokenStream2;
 use quote::quote;
+use std::sync::Mutex;
 use syn::{
     parse_macro_input, AttributeArgs, Ident, ItemFn, Lit, Meta, MetaNameValue, NestedMeta, PatType,
     ReturnType, Type,
 };
 
+lazy_static! {
+    static ref WASM_IMPORTS_ADDED: Mutex<bool> = Mutex::new(false);
```
ncitron, line 25: Why are we using a lazy static here? We have many other instances where we dynamically generate code based on an attribute in the provable macro and we don't need to use these.

src/main.rs
```diff
@@ -146,6 +160,240 @@ fn display_sysinfo() {
...
+#[derive(Serialize, Deserialize)]
+struct DecodedData {
+    bytecode: Vec<ELFInstruction>,
+    memory_init: Vec<(u64, u8)>,
+}
+
+fn preprocess_and_save(func_name: &str) -> Result<()> {
+    let mut program = Program::new("guest");
+    program.set_func(func_name);
+    program.set_std(false);
```
ncitron, line 172: Can we figure out how to extract these options from the macro? This is how we currently do it for normal usage from the macro generated functions, which I am sure we could extract if needed.

jolt-sdk/macros/src/lib.rs
sebasti810: I thought we could use lazy_static here because the macro code for WASM needs certain global imports and a helper structure for serialization to be initialized only once, regardless of how many functions are labelled with the `wasm` attribute..

Do you think the global initialization is a problem? If lazy_static is the issue, we could use Once from the std-library like this:
```
static WASM_IMPORTS_INIT: Once = Once::new();

#[proc_macro_attribute]
pub fn provable(attr: TokenStream, item: TokenStream) -> TokenStream {
    let attr = parse_macro_input!(attr as AttributeArgs);
    let func = parse_macro_input!(item as ItemFn);
    let mut builder = MacroBuilder::new(attr, func);

    let mut token_stream = builder.build();

    // Add wasm utilities and functions if the function is marked as wasm
    if builder.has_wasm_attr() {
        // wasm utilities should only be added once
        WASM_IMPORTS_INIT.call_once(|| {
            let wasm_utilities: TokenStream = builder.make_wasm_utilities().into();
            token_stream.extend(wasm_utilities);
        });
        let wasm_token_stream: TokenStream = builder.make_wasm_function().into();
        token_stream.extend(wasm_token_stream);
    }

    token_stream
}
```

Would this approach be more suitable for our use case?

src/main.rs
sebasti810: How would it be if we also read out whether the feature is set or not in the preprocessing, in which we modify Cargo.toml anyway?

sebasti810: Also, for the todo comment below I thought about extracting the parse_attributes function,  for example into "common". This way we could reuse it both in preprocessing and in the macro. What do you think of this idea?

ncitron: Yeah I think that is fine. It's just important that when we use this preprocess function it behaves the same way as it would in the non-wasm variants so that the prover/verifier are interoperable. Otherwise a proof generated would set the various attributes here but the wasm verifier would have these defaults instead.

jolt-sdk/macros/src/lib.rs
ncitron: Oh I see what you're doing here now. I think this is fine.

sebasti810:
@ncitron I have refactored the parse_attributes function so that we can use it in both the WASM preprocessing and within the macro. Additionally, I have moved the entire WASM preparation process to a separate file because it was getting very cluttered. Whether it uses std or no_std is determined by reading the features and setting it accordingly. lmk what you think

ncitron (approved):
LGTM!

ncitron:
Looks like CI is failing due to clippy. Can you resolve then we can merge?

sebasti810:
@ncitron should be done

PR #375, opened 2024-06-04
Implementation of section 5 of the Quarks Paper

aleph-v:
Implements a version of the grand product argument from the Quarks [paper](https://eprint.iacr.org/2020/1275), this version is the simpler version from the section 5 of the paper which directly trades off increased prover time for reducing proof size from log^2(n) to just log(n). Depending on the performance we may or may not followup with a hybrid approach from section 6.

Note - This just adds the sub protocol, it doesn't wire a default proof to use quarks.


moodlezoup (commented):
looking good!

PR #384, opened 2024-06-11
feat: Improve Aux computation

sragss:
Aux computation now deduplicates evaluation of Linear Combinations shared across the auxiliary variables. 28% of Linear Combinations are currently reused in `jolt_constraints.rs` and were previously computed from scratch. 35% speed up over existing on 2^22 trace length 50bps reduction e2e.

In the process removes `AuxComputation.{flat_vars, inputs_to_flat}`.

Separately speeds up + cleans up `CombinedUniformBuilder::compute_spartan_Az_Bz_Cz` by moving the batch LC evaluation logic onto LC.


sragss:
Wow, super!

PR #387, opened 2024-06-13
feat(jolt-sdk): Write ELF and Proof to project directory based on environment variable + refactors for over-wire proof verification.

PatStiles:
- Adds support for writing elf and proof files to project directory by setting the `"JOLT_SAVE"` env variable. 
- Changes tracer interface to accept span of elf bytes as input. This allows decoding the elf file at runtime without writing a file.
- Renames `Proof` -> `RV32IHyraxProof`and relocates it to `jolt-core/src/jolt/vm/rv32i_vm.rs` making it std. This was done as for my purposes having it be no_std lead to issues when compiling ffi bindings for rust -> go.


jolt-sdk/src/host_utils.rs
```diff
@@ -17,37 +10,7 @@ pub use jolt_core::host;
...
 };
 pub use tracer;
-
-pub type CommitmentScheme = HyraxScheme<G>;
-
-#[derive(CanonicalSerialize, CanonicalDeserialize)]
-pub struct Proof {
-    pub proof: RV32IJoltProof<F, CommitmentScheme>,
-    pub commitments: JoltCommitments<CommitmentScheme>,
-}
```
moodlezoup, line 31: can we leave this where it is? You can enable the `host` feature of this package to make it `std`

jolt-sdk/macros/src/lib.rs
```diff
@@ -443,6 +446,18 @@ impl MacroBuilder {
         }
     }
 
+    fn make_write_to_file(&self) -> TokenStream2 {
```
moodlezoup, line 449: I'm not a huge fan of using an env variable here; I'd rather just call `save_to_file` directly from the host

jolt-core/src/host/mod.rs
```diff
@@ -254,6 +257,16 @@ impl Program {
...
         format!("/tmp/jolt-guest-linkers/{}.ld", self.guest)
     }
+
+    /// Saves the proof to a file
+    pub fn save_elf(&self) {
+        if !PathBuf::from("elf/").exists() {
+            fs::create_dir_all("elf/").expect("could not create elf directory");
+        }
+        let elf = self.elf.as_ref().unwrap();
+        fs::copy(elf, format!("elf/{}.elf", self.guest))
```
moodlezoup, line 267: Could you explain the rationale of this function? If we already have the ELF at the `self.elf` path, why copy it to a new directory?

PatStiles: The elf file is saved to a temporary directory during proving and is then removed at the end of execution. The intention here was to have a method that could be called during execution to save the elf file to a non temporary directory. In retrospect there is not a need for a new directory to be created the project current directory is fine.

jolt-sdk/src/host_utils.rs
PatStiles: I moved this due to issues with compiling Jolt behind rust -> go ffi bindings with the host feature enabled.

jolt-sdk/macros/src/lib.rs
PatStiles: The issue I was hoping to address with the env var is in regards to saving the elf file when it is compiled to a temporary directory at run time. Since the program isn't directly accessible in the host I thought this was the least intrusive. Open to suggestions.

jolt-core/src/host/mod.rs
PatStiles: Running the default example it does not appear to persist.

Screenshot 2024-06-28 at 09 57 52

https://github.com/a16z/jolt/blob/main/jolt-core/src/host/mod.rs#L123

PatStiles:
@moodlezoup comments addressed

moodlezoup (approved):

moodlezoup:
@PatStiles could you resolve the merge conflict? Otherwise looks good to merge!

PatStiles:
@moodlezoup resolved!

moodlezoup:
@PatStiles could your run cargo fmt? last request 🙏

PatStiles:
Done! @moodlezoup

PR #390, opened 2024-06-14
Various Quark Optimizations

aleph-v:
Adds support for three optimizations: doing the sum opening as part of the sum check, reducing the commitment size and doing line reductions for points which are the same except for 1 point


moodlezoup (commented):
good stuff, just a few nits

moodlezoup (approved):

Issue #376, opened 2024-06-04
Fix CI tests


moodlezoup:
Tests are failing on CI due to `tar` complaining when trying to extract the jolt toolchain

PR #392, opened 2024-06-14
fib_e2e: Actually use the type parameters 🙃

imikushin:
`fib_e2e` is parameterized with `<F: JoltField, PCS: CommitmentScheme<Field = F>>`, but is currently using a concrete field and commitment scheme instead of them:
```rust
let (proof, commitments) =
    <RV32IJoltVM as Jolt<Fr, HyraxScheme<G1Projective>, C, M>>::prove(
        io_device,
        trace,
        circuit_flags,
        preprocessing.clone(),
    );
```

Given there are multiple tests calling `fib_e2e` with different implementations of JoltField and CommitmentScheme, this is probably not intended.


imikushin:
`cargo nextest run -p jolt-core` succeeded locally. Test fails due to #376

PR #413, opened 2024-07-16
Adds a hyperkzg subprotocol to the onchcain verifier

aleph-v:
Adds an onchain hyperkzg verifier subprotocol which can verify either single or batch openings.


moodlezoup (approved):
Great work, no notes 💯

PR #415, opened 2024-07-16
Multi-Offset-Eq Constraints

sragss:
Expands the non-uniform offset eq [constraints](https://jolt.a16zcrypto.com/how/r1cs_constraints.html) to support multiple constraints.


sragss:
Before merging: Make sure the allocated sizes of Az, Bz, Cz in `buillder::compute_spartan_Az_Bz_Cz` are correct.

jolt-core/src/r1cs/key.rs
```diff
@@ -299,63 +325,69 @@ impl<F: JoltField> UniformSpartanKey<F> {
...
+        for (i, constraint) in self.offset_eq_r1cs.constraints.iter().enumerate() {
+            let non_uni_a = compute_non_uniform(&constraint.eq);
+            let non_uni_b = compute_non_uniform(&constraint.condition);
+
+            let non_uni_constraint_index =
+                index_to_field_bitvector(self.uniform_r1cs.num_rows + i, constraint_rows_bits);
+            let row_constr_eq_non_uni =
+                EqPolynomial::new(r_row_constr.to_vec()).evaluate(&non_uni_constraint_index);
+
+            debug_assert_eq!(
```
moodlezoup, line 381: should this be a normal `assert_eq!`?

sragss: Shouldn't hurt perf much, switching.

PR #416, opened 2024-07-17
feat: m extension support in tracer

ncitron:
Adds ability to compile with the m extension and adds support to the tracer.

It currently crashes after tracing as it hits an `todo!` block related to converting the trace into the virtual instructions inside `host/mod.rs`.

Also adds an example that compiles down to include `MUL` and `DIVU`, but we may want to modify it at some point to ensure it triggers all of the m extension instructions for testing purposes.


moodlezoup (approved):
thanks! any idea why it's failing to install the toolchain in CI?

ncitron:
They were failing before because the rust toolchain had not finished building for x86, but now it as and everything seems to be passing.

Also are we sure we want to merge this into main? This will make main non functional, as it enabled the m extension compiler by default.

moodlezoup:
yep, merging it into a feature branch

Issue #328, opened 2024-04-25
EVM Friendly Transcript


sragss:
One of the quickest routes to on-chain verification is to switch the commitment scheme to [Zeromorph](https://eprint.iacr.org/2023/917.pdf) then verify those proofs directly on-chain. Currently our Fiat-shamir transcript uses the [Merlin](https://github.com/dalek-cryptography/merlin) library which depends on the [Strobe RNG](https://merlin.cool/transcript/rng.html). We'll want to swap this to Keccak to support cheaper EVM verification using the Keccak precompiles.

The Honk on-chain verifier has [an example](https://github.com/Maddiaa0/honk-verifier/blob/master/src/reference/Transcript.sol) of the transcript written in solidity.

Steps:
1) Make the transcript generic over the hash function
2) Support keccak and poseidon2
3) Reduce count and size of domain separators (in Merlin these are referred to as 'labels')
4) Integration test for parity with EVM version

MatteoMer:
I'd be happy to work on that!

 To make sure I understand correctly the idea is to completely remove merlin and write jolt own implementation of the transcript? 

Also, I'm not sure I understand exactly the 3rd step, do you mean removing the number of `append_message` in the code? 

I think I'll just work to be 1/1 with the EVM version 😄

Maddiaa0:
Looking into Merlin it appears that it is already using keccak as its hash function - see https://github.com/dalek-cryptography/merlin/blob/53535f32e6d6de421372d67f56176af0c0f55fd7/src/strobe.rs#L97

sragss:
That would be great @MatteoMer!

Good ponit @Maddiaa0 – new step 1 would be to determine if [RustCrypto/sponges](https://github.com/RustCrypto/sponges/blob/master/keccak/src/lib.rs) Keccak (used by Merlin) is the same as the EVM's Keccak256 and determine the [additional augmentations](https://github.com/dalek-cryptography/merlin/blob/53535f32e6d6de421372d67f56176af0c0f55fd7/src/strobe.rs#L94-L96) done by Merlin.

We'd like to append the same values to transcript in Rust and EVM and then sample the same bytes of randomness.

If we can do this and keep Merlin, great. If not, let's write our own.

MatteoMer:
Haven't took time to dive into determine if both keccak implementations are the same yet (will do soon)

[...]

PR #419, opened 2024-07-19
Work in progress onchain verifier

aleph-v:
This PR merges the onchain verifier branch into main so we can transition into doing PRs to finish the features. Currently there is still quite a lot of outstanding work before the verifier is functional.

We have working onchain protocols for:

- Fiat shamir transform
- Hyper KZG
-  GKR based grand products


jolt-evm-verifier/README.md
sragss, line 1: Given the forge submodule does this crate require `--recurse-submodules` or is that handled by foundry?

If so, could we add to the README?

jolt-evm-verifier/script/Counter.s.sol
sragss, line 1: rm

jolt-evm-verifier/script/src/bin/hyperkzg_batch_example.rs
```diff
@@ -0,0 +1,173 @@
...
+        VK_beta_g2: g2_beta_sol,
+    };
+
+    let mut com = vec![];
+    let mut w = vec![];
+    let ypos_scalar = proof.v[0].clone();
+    let yneg_scalar = proof.v[1].clone();
+    let y_scalar = proof.v[2].clone();
+
+    // Horrible type conversion here, possibly theres an easier way
```
sragss, line 115: Unfortunately the Bn254 points are stored in Montgomery form so they incur a field multiplication to get to BigInt form. I don't know how the EVM precompiles handle this but its possible we want to keep them in Montgomery form the whole time.

jolt-evm-verifier/script/src/main.rs
```diff
@@ -0,0 +1,123 @@
...
+fn fr_to_uint256(c: &Fr) -> U256 {
+    let mut serialize = vec![];
+    let _ = c.serialize_uncompressed(&mut serialize);
+    U256::from_le_slice(&serialize)
+}
+
+fn main() {
+    let _: Vec<_> = env::args().collect();
+
+    //initial test taken from https://github.com/a16z/jolt/blob/main/jolt-core/src/subprotocols/grand_product.rs#L1522-L1545
```
sragss, line 105: Let's switch to a permalink here https://github.com/a16z/jolt/blob/d5147f8d27bb4961f3d648b872b45ff99af860c0/jolt-core/src/subprotocols/grand_product.rs<LINE NUMBERS>

jolt-evm-verifier/src/reference/JoltVerifier.sol
```diff
@@ -0,0 +1,171 @@
...
+            // Sample random from the transcript
+            Fr r_i = Fr.wrap(transcript.challenge_scalar(MODULUS));
+            r[i] = r_i;
+
+            e = UniPolyLib.evaluate(poly, r_i);
+        }
+        return (e, r);
+    }
+
+    function buildEqEval(Fr[] memory rGrandProduct, Fr[] memory rSumcheck) internal pure returns (Fr eqEval) {
```
sragss, line 62: `evalEqMLE` might be a more apt name.

`buildEqEval` implies it may be building the `Eq` evaluation table.

jolt-evm-verifier/src/subprotocols/FiatShamirTranscript.sol
sragss, line 1: We can consider #328 closed.

jolt-evm-verifier/script/src/bin/hyperkzg_batch_example.rs
```diff
@@ -0,0 +1,173 @@
...
+        evals.push(eval);
+    }
+
+    for poly in polys.iter() {
+        borrowed.push(poly);
+    }
+
+    // prove an evaluation
+    let mut prover_transcript = ProofTranscript::new(b"TestEval");
+    let proof: HyperKZGProof<Bn254> = HyperKZG::batch_prove(
```
sragss, line 55: We likely will want to write this conversion logic directly on an `Into<HyperKZGSol>` trait eventually, but good for now.

sragss (approved):
Very nice. LGTM

jolt-evm-verifier/script/src/bin/hyperkzg_batch_example.rs
aleph-v: No sadly it assumes a affine non mongomery form

jolt-evm-verifier/script/src/bin/hyperkzg_batch_example.rs
aleph-v: Done in the next pr, but good note.

PR #420, opened 2024-07-19
Adds compressed polys to transcript instead normal ones

aleph-v:
We were computing compressed polys but not using them to reduce the committed transcript length despite that appearing to be their purpose.


moodlezoup (approved):
good catch, thanks!

PR #421, opened 2024-07-20
onchain-verifier: measuring gas usage for grand product argument

MatteoMer:
I got some troubles using foundry's `gas-report` to work with the code with internal functions, but will continue to investigate. 

 For now this is a new test that outputs some gas benchmarks for the internal functions of the verifier when `BENCHMARK_LOGGING=true` using a wrapper around the grand product argument library

You can run the gas report with `BENCHMARK_LOGGING=true forge test --via-ir --ffi -vv --match-test testGasGrandProductVerify`

I don't think it's the best, so would love some ideas on how to make it better!

### DIFF
--- a/jolt-evm-verifier/foundry.toml
+++ b/jolt-evm-verifier/foundry.toml
@@ -6,5 +6,4 @@ optimizer = true
 optimizer-runs = 1000
 via_ir = true
 
-
 # See more config options https://github.com/foundry-rs/foundry/blob/master/crates/config/README.md#all-options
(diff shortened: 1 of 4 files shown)


moodlezoup:
btw one pattern I've used in the past is a function modifier that prints the gas of the modified function:

```
modifier logGas(string memory label) {
    uint256 gasUsed = gasleft();
    _;
    gasUsed -= gasleft();
    console.log(label, gasUsed);
}
```

this can be used for internal functions, similar to the in-line logging you have now

Issue #406, opened 2024-06-25
Refactor tests for virtual instruction sequences


moodlezoup:
Write a test macro or util function for virtual sequences, similar to the `jolt_instruction_test` [macro](https://github.com/a16z/jolt/blob/main/jolt-core/src/jolt/instruction/test.rs)

Then refactor the tests for `DIV`, `DIVU`, `MUL`, `MULHSU`, `REM`, and `REMU` to use the new macro/function

PR #422, opened 2024-07-21
Refactor tests for virtual sequences

tumberger:
Quick refactor to address #406


moodlezoup (commented):
Looking good, thanks for taking this on! 

Let's add a `sequence_output` function to the `VirtualInstructionSequence` trait which computes the result from the operands x and y. This would allow us to move the remaining test logic (i.e. generating the random x and y values, generating the random registers, and computing the expected result) into the macro

moodlezoup:
Perfect, could you just fix the fmt and clippy errors?

tumberger:
Of course, I believe it's fixed now!

PR #424, opened 2024-07-22
fix: add deserialize function and change types to RV32IHyraxProof in wasm-related code

sebasti810:
- added serialize_to_bytes and deserialize_from_bytes methods to RV32IHyraxProof.
- modified imports to include RV32IHyraxProof in wasm.
- updated deserialization logic in verify_wasm_fn_name to use RV32IHyraxProof::deserialize_from_bytes.


sragss (approved):
LGTM. We should implement this logic more generically going forward. Issue here #425

PR #426, opened 2024-07-23
KZG, HyperKZG: speedup pairing checks

imikushin:
~20% speedup of KZG and HyperKZG verifiers due to multi-pairing vs 2 individual pairing operations.


moodlezoup (approved):
nice find, thank you!

PR #434, opened 2024-07-26
Finish M extension

moodlezoup:
Finishes wiring up the M extension to Jolt:
- tracer support h/t @ncitron 
- Adds new constraints
- Fixes various bugs (described in-line)


common/src/rv_trace.rs
```diff
@@ -355,6 +358,14 @@ impl ELFInstruction {
             RV32IM::VIRTUAL_ASSERT_VALID_DIV0,
         );
 
+        // All instructions in virtual sequence are mapped from the same
+        // ELF address. Thus if an instruction is virtual (and not the last one
+        // in its sequence), then we should *not* update the PC.
+        flags[11] = match self.virtual_sequence_remaining {
+            Some(i) => i != 0,
+            None => false
+        };
```
moodlezoup, line 367: We need this flag (and so we had to reverse the virtual sequence indexing) for the new non-uniform constraint, which ensures that virtual sequences are executed in full and in order

jolt-core/src/field/ark.rs
```diff
@@ -26,6 +26,23 @@ impl JoltField for ark_bn254::Fr {
         }
     }
 
+    fn to_u64(&self) -> Option<u64> {
```
moodlezoup, line 29: for debugging

jolt-core/src/jolt/instruction/div.rs
```diff
@@ -190,7 +201,26 @@ impl<const WORD_SIZE: usize> VirtualInstructionSequence for DIVInstruction<WORD_
             advice_value: None,
         });
 
-        virtual_sequence
+        virtual_trace.push(RVTraceRow {
```
moodlezoup, line 204: Previously, we had elided the MOVE instruction as described in the Jolt paper as an optimization, instead writing the advice directly to the destination register
But as it turns out, we actually did need the separate virtual register + MOVE for the edge case where `rd` is equal to `rs1` or `rs2`

jolt-core/src/jolt/instruction/lb.rs
```diff
@@ -55,6 +55,12 @@ impl JoltInstruction for LBInstruction {
                 Box::new(SignExtendSubtable::<F, 8>::new()),
                 SubtableIndices::from(C - 1),
             ),
+            (
```
moodlezoup, line 58: Fixes bug flagged by the Arithmic team

jolt-core/src/jolt/instruction/virtual_assert_lte.rs
moodlezoup, line 1: This was implemented as a signed ≤, but it should be an unsigned ≤

jolt-core/src/jolt/instruction/sh.rs
```diff
@@ -31,13 +30,20 @@ impl JoltInstruction for SHInstruction {
...
-            Box::new(IdentitySubtable::<F>::new()),
-            SubtableIndices::from(C - 1),
-        )]
+        vec![
+            (
+                Box::new(IdentitySubtable::<F>::new()),
+                SubtableIndices::from(C - 1),
+            ),
+            (
+                // Not used for lookup, but this implicitly range-checks
```
sragss, line 41: Let's make a utility function for this so it's self documenting and use across all the memory ops that need range-checking.

`fn RangeCheck(C) -> (Box<Subtable<...>>, SubtableIndices)`

sragss (approved):
LGTM

jolt-core/src/jolt/vm/bytecode.rs
```diff
@@ -287,8 +309,82 @@ impl<F: JoltField, C: CommitmentScheme<Field = F>> BytecodePolynomials<F, C> {
             DensePolynomial::new(rs2),
             DensePolynomial::new(imm),
         ];
-        let t_read = DensePolynomial::from_usize(&read_cts);
-        let t_final = DensePolynomial::from_usize(&final_cts);
+        let t_read: DensePolynomial<F> = DensePolynomial::from_usize(&read_cts);
+        let t_final: DensePolynomial<F> = DensePolynomial::from_usize(&final_cts);
+
+        #[cfg(test)]
```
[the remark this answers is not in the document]
moodlezoup: we need to declare the variables in the same scope as everything else

PR #435, opened 2024-07-29
Adds a partially complete spartan r1cs verifier to the onchain contracts

aleph-v:
Makes a number of changes in support of the spartan onchain verifier. The EVM contracts now have a verifier which is complete except for the final version of the second sumcheck because it is undergoing active refactoring at the same time, targeting main would result in an out of date version. Therefore we have tests with a frozen proof and which have constants at the point of the second sumcheck. This verifies the rest of the r1cs verifier is working, and a single additional PR focusing on the subsegment in the abc mle with testing should finish the system, once the version in main is finalized.


jolt-evm-verifier/src/subprotocols/R1CSMatrix.sol
```diff
@@ -0,0 +1,520 @@
...
+library R1CSMatrix {
+    using FrLib for Fr;
+
+    /// Evaluates the z mle for the outer sumcheck
+    /// @param r The r we are evaluating on
+    /// @param segmentEvals The segment evals from the proof, should be length 89
+    function eval_z_mle(Fr[] memory r, uint256[] memory segmentEvals) internal pure returns (Fr) {
+        require(segmentEvals.length == 89, "Segment Length must be 89");
+        // We calculate the evals based on a length 7 slice
+        Fr[] memory r_var_eqs = eq_poly_evals(r, 1, 7);
```
moodlezoup, line 17: let's pull 89 and 7 our into named constants

jolt-evm-verifier/src/subprotocols/R1CSMatrix.sol
```diff
@@ -0,0 +1,520 @@
...
+            row_bits + constraint_col_bits + 1,
+            col_bits - constraint_col_bits
+        );
+        Fr[] memory eq_poly_row = eq_poly_evals(r, 0, constraint_row_bits);
+        // Creates a 256 field element allocation but doesn't need it?
+        Fr[] memory eq_poly_col = eq_poly_evals(r, row_bits, constraint_col_bits + 1);
+
+        // Does the eval of the constraint row const via a bit vec
+        /// TODO this constant seems to always be zero
+        Fr col_eq_constant = Fr.wrap(0);
```
moodlezoup, line 66: the constant always seems to be zero in the Rust implementation? 🤔 
cc @sragss

jolt-evm-verifier/src/subprotocols/SpartanVerifier.sol
```diff
@@ -0,0 +1,132 @@
...
+        Fr aEval = Fr.wrap(0x0168ec8c28141fc3422b0ccee2fb350301b7a30900232c5c16ea8aaa5e48b63d);
+        Fr bEval = Fr.wrap(0x219d6c166058578e4e54e1819527d89d7f66d417c41c44733322d8f6204b581d);
+        Fr cEval = Fr.wrap(0x0b627be010723de7db4cac462721244d1f7e1dd84f8f60bc8334d3b86d67ee26);
+
+        Fr expected_left = aEval + r_inner_sumcheck_RLC * bEval + r_inner_sumcheck_RLC * r_inner_sumcheck_RLC * cEval;
+        require(claim_inner == expected_left * eval_Z, "SpartanError::InvalidInnerSumcheckClaim");
+
+        // We never use this memory again so we are ok to corrupt it like this
+        // TODO - (aleph) is it ok to refer to this as 'memory-safe'? We hit stack depth without it, but this does corrupt the memory in
+        //                r_y which should never be used again?
```
moodlezoup, line 113: pretty sure this is considered memory-safe

moodlezoup (commented):
Nice work! Just some nits/questions

moodlezoup:
looks like there's a solc build error

jolt-evm-verifier/src/subprotocols/R1CSMatrix.sol
aleph-v: These can be named but the code may break if you change them, so I generally don't prefer named constants which can't be changed.

jolt-evm-verifier/test/TestSpartanVerifier.sol
```diff
@@ -0,0 +1,59 @@
...
+
+        string[] memory cmds = new string[](2);
+        cmds[0] = "sh";
+        cmds[1] = "script/spartan_example.sh";
+        bytes memory result = vm.ffi(cmds);
+        (SpartanExample memory decodedProof) = abi.decode(result, (SpartanExample));
+
+        DeployableSpartan spartan = new DeployableSpartan(decodedProof.vk);
+
+        assert(spartan.verifySpartanR1CS(decodedProof.proof, decodedProof.commitments, transcript, 16, 17, 131072));
```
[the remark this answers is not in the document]
aleph-v: We can't use named parameters because this would require naming the unused parameter which would break the ci as it requires no warnings and the unused named param would trigger a compiler warning, and the unused named parameters are left in because the R1CS lib must be finished and will use the unnamed parameter.

PR #437, opened 2024-07-29
Properly uses compression in sumcheck verifier

aleph-v:
The compressed sumcheck uni polys are calculated as f(0) + f(1) = hint, currently we decompress, check this relationship and then evaluate, however because we decompress using this relation the check can never fail. Instead we should be taking advantage of the compression by computing the linear term implicitly the evaluating, as we do in this PR.

### DIFF
--- a/jolt-core/src/subprotocols/sumcheck.rs
+++ b/jolt-core/src/subprotocols/sumcheck.rs
@@ -520,18 +520,13 @@ impl<F: JoltField> SumcheckInstanceProof<F> {
         // verify that there is a univariate polynomial for each round
         assert_eq!(self.compressed_polys.len(), num_rounds);
         for i in 0..self.compressed_polys.len() {
-            let poly = self.compressed_polys[i].decompress(&e);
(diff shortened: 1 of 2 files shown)


jolt-core/src/subprotocols/sumcheck.rs
```diff
@@ -520,19 +520,14 @@ impl<F: JoltField> SumcheckInstanceProof<F> {
...
                 return Err(ProofVerifyError::InvalidInputLength(
                     degree_bound,
-                    poly.degree(),
+                    self.compressed_polys[i].degree(),
                 ));
             }
 
-            // check if G_k(0) + G_k(1) = e
-            assert_eq!(poly.eval_at_zero() + poly.eval_at_one(), e);
-
```
moodlezoup, line 535: we still need this check no? The verifier needs to check that the current round's polynomial is consistent with the claim from the previous round

moodlezoup (approved):
nice find!

PR #441, opened 2024-08-02
Make prove and verify closures Sync+Send

bmmoore:
This adds `Send` and `Sync` to the `Fn`s in the result of generated `build_<guest>` functions, so setup can be more easily shared in `#[test]` and other multithreaded code.


moodlezoup (approved):
Thanks!

PR #450, opened 2024-09-01
Add 64-bit version to all instructions

quangvdao:
This PR makes all current Jolt instructions have a 32-bit and a 64-bit version.

Importantly, it **does not change** the behavior of 32-bit instructions, nor change the list of subtables.

List of changes:
- Parametrize all instructions by word size
	- This is necessary so that the `random()` function correctly returns random `u64`s for 64-bit word sizes
- Add instruction behavior for 64-bit word size (or document that the behavior does not change for 64-bit word size)
- Add test cases for 64-bit word sizes (all tests passed)
- Add more detailed documentation to all `instruction_utils` functions


jolt-core/src/jolt/instruction/sw.rs
```diff
@@ -65,24 +71,59 @@ mod test {
...
+            SWInstruction::<WORD_SIZE>(1 << 8),
+            SWInstruction::<WORD_SIZE>(u32_max),
+            SWInstruction::<WORD_SIZE>(u32_max - 2),
+        ];
+        for instruction in instructions {
+            jolt_instruction_test!(instruction);
+        }
+    }
+
+    // Doesn't work for now
```
moodlezoup, line 102: is this comment still accurate?

jolt-core/src/jolt/instruction/sw.rs
```diff
@@ -65,24 +71,59 @@ mod test {
     use crate::{jolt::instruction::JoltInstruction, jolt_instruction_test};
 
     #[test]
-    fn sw_instruction_e2e() {
+    fn sw_instruction_32_e2e() {
         let mut rng = test_rng();
-        const C: usize = 4;
+        const C: usize = 2;
```
moodlezoup, line 76: Can we keep C = 4 for the 32-bit test and C = 8 for the 64-bit test? Those are probably the parameters we'd use in practice

moodlezoup (approved):
beautiful, just a couple of questions

jolt-core/src/jolt/instruction/sw.rs
quangvdao: Removed the comment. The test works now

jolt-core/src/jolt/instruction/sw.rs
quangvdao: Yes, that is totally fine. I think you're just wasting lookups though, since for 16-bit chunk length, you only need 2 chunks for 32-bit and 4 chunks for 64-bit.

quangvdao:
One final note for this PR: currently, I don't know how the rest of Jolt uses the range-checking part of the store operations. I'll keep this in mind as I work to integrate 64-bit instructions for the rest of Jolt, just to make sure there's no under-constrained part (esp. with regards to range-checking store values). The 32-bit version is fine since it's still identical to before.

Issue #395, opened 2024-06-17
Better errors for insufficient stack or memory size


moodlezoup:
https://jolt.a16zcrypto.com/usage/troubleshooting.html#insufficient-memory-or-stack-size

Currently, the tracer will error out with nondescript errors like `Unknown memory mapping 7FFFFE04.`

The tracer should detect when we have exhausted the stack and/or heap, and throw a descriptive error if so

ameya-deshmukh:
Hey! Can I take this up?

moodlezoup:
Yeah go for it!

moodlezoup:
@ameya-deshmukh Any update here?

ameya-deshmukh:
Hey @moodlezoup! Sorry, couldn't find time to work on this. Can I grind this out over the weekend and give you an update by early Monday ET?

moodlezoup:
Yeah that'd be great! Lmk if you need any pointers

4rgon4ut:
@ameya-deshmukh @moodlezoup I would like to have a try if you fine with that

ameya-deshmukh:
Hey @4rgon4ut sure! I didn't get the time to work on this issue, all yours to take

4rgon4ut:
@moodlezoup I have two questions:

1. Would you like it to be panics with specified messages (based on the case) or new trap type [here](https://github.com/a16z/jolt/blob/main/tracer/src/emulator/cpu.rs#L110)
2. Does `tracer/src/emulator/cpu.rs` is the right place to add related tests?

moodlezoup:
> @moodlezoup I have two questions:
> 
> 1. Would you like it to be panics with specified messages (based on the case) or new trap type [here](https://github.com/a16z/jolt/blob/main/tracer/src/emulator/cpu.rs#L110)
> 2. Does `tracer/src/emulator/cpu.rs` is the right place to add related tests?

Thanks for taking this! 
1. just a better panic message is fine
2. I think perhaps `tracer/src/emulator/mmu.rs` may be a better place for tests

moodlezoup:
> @moodlezoup does stack grows downwards, from higher stack addr down to the `DRAM_BASE` ? [ref](https://github.com/a16z/jolt/blob/main/tracer/src/emulator/mmu.rs#L3)
> 
> ```rust
> /// DRAM base address. Offset from this base address
> /// is the address in main memory.
> pub const DRAM_BASE: u64 = 0x80000000;
> ```

Yes, I believe so

PR #452, opened 2024-09-12
Chore(mmu): add overflow handling

4rgon4ut:
This PR add additional memory overflow detection(for both stack and heap).

Added:
- function to assert memory region where:
    1. zero padding region => stack overflow
    2. address higher than memory array length => heap overflow
    3. other not expected region => unknown mapping panic   
- unit tests
- example to trigger stack && heap overflow in runtime

Closes: https://github.com/a16z/jolt/issues/395


4rgon4ut:
I feel I need to get an acknowledgement for my direction as Im new to the codebase @moodlezoup. 

1. Memory model:
image

Thus every attempt to write to `zero_padding` addresses is stack overflow and all writes over `DRAM_BASE + memory.capacity` is heap overflow. Is that correct?

2. It would be also nice to have sort of example to face this cases in runtime. Is that something you would like to see?

3. `store_raw()` have no tracer call before actual operation, should I add checks in `*_raw()` functions?

moodlezoup:
> I feel I need to get an acknowledgement for my direction as Im new to the codebase @moodlezoup.
> 
> 1. Memory model:
> 
> image
> Thus every attempt to write to `zero_padding` addresses is stack overflow and all writes over `DRAM_BASE + memory.capacity` is heap overflow. Is that correct?

Yes this is correct 

> 
> 2. It would be also nice to have sort of example program to face this cases in runtime. Is that something you would like to see?

Yes, it would be great to have a test for this feature

> 3. [`store_raw()`](https://github.com/a16z/jolt/blob/40b8b4c5842cb4b000c3d79d484a1afe67ddad29/tracer/src/emulator/mmu.rs#L634) have no tracer call before actual operation, should I add checks in `*_raw()` functions?

I think so, but I haven't looked at that code in a while so if there's a better place to add the checks, go ahead

4rgon4ut:
@moodlezoup

moodlezoup (approved):
Really nice work, thanks for adding the test cases!

PR #462, opened 2024-09-27
Serializable

myceliaspy:
I think this was missing may be mistaken. also anyone getting 
image
when trying to build-wasm?


moodlezoup (approved):
thanks!

myceliaspy:
but seriously is this straight up not wasm compilable because of nix or am i missing something?

PR #471, opened 2024-09-29
(chore) Remove last `eq` lookup in comparison instructions

quangvdao:
For the SLT, SLTU, BGE, and BGEU instructions, we do not need the equality lookup on the least significant chunks. This is because the formulas for those instructions (such as SLTU) look like:

SLTU(x, y) = LTU(x_0, y_0) + LTU(x_1, y_1) * EQ(x_0, y_0) + ... + LTU(x_{C-1}, y_{C-1}) * EQ(x_0, y_0) * ... * EQ(x_{C-2}, y_{C-2})

and so the last equality EQ(x_{C-1}, y_{C-1}) is never needed.


moodlezoup (approved):
nice catch!

PR #473, opened 2024-10-03
feat: Improve Commitment cost for Quark Grand Producs

### LINKED ISSUE #444: Improve commitment cost in Quarks/hybrid grand product argument
About 93% of the factors arising the Lasso grand product arguments are 1. We are currently not leveraging this fact when committing to the partial products in the Hybrid or Quarks grand product argument. That is, the prover is currently paying one group operation to commit to each 1. These should actually be "free" to commit to. 

To achieve this, in pre-processing we should pre-compute the product of all group elements in the entire commitment key, obtaining a result G (we can, say, include G in the SRS, if we are using HyperKZG as the polynomial commitment). G is the commitment to the all-1s vector. Hence, G is like "pretending" that all gates in all the grand product circuits are 1. Then when we go to commit to a gate of value i, we instead commit to i-1, getting a result H. The final commitment is G*H.

This is a much faster way to compute the same commitment that we compute today. For Quarks, this modification should reduce commitment costs by a factor of 2 or more.

sagar-a16z:
### Changes: 
- [x] Preprocess a set of commitments to the group elements in the SRS for all supported lengths, L, where L is a power of 2
- [x] Optimize quark grand products commitments by reusing the preprocessed commitments as described in #444
- [x] Add a benchmark for Grand Products 
    - [x] Support HyperKZG
    - [x] Support  Zeromorph
### Pending:
- [ ] Benchmark is not representive of real usage, optimization does not get utilized.
- [ ] Implict nested usage of rayon inside of `batch_commit` in HyperKZG and then `grand_product_quark` is likely not optimal
- [ ] Address TODOs
- [ ] File an issue to add more commitment schemes to the benchmark

Fixes #444


sagar-a16z:
We need another issue to track the fact that we're currently limiting the size of leaves per layer in the quarks grand product

jolt-core/src/poly/commitment/kzg.rs
```diff
@@ -61,9 +63,22 @@ impl<P: Pairing> SRS<P> {
             || P::G2::normalize_batch(&g2_powers_projective),
         );
 
+        // Compute a commitment, G, to all the group elements in the SRS
```
moodlezoup, line 66:
```suggestion
        // Precompute a commitment to each power-of-two length vector of ones, which is just the sum of each power-of-two length prefix of the SRS```

File: jolt-core/src/poly/commitment/kzg.rs [hunk 2]
```diff
@@ -119,6 +134,15 @@ pub struct KZGVerifierKey<P: Pairing> {
     pub beta_g2: P::G2Affine,
 }
 
+#[derive(Clone, Copy, Debug)]
+pub enum CommitMode {
+    Default,
+    // We noticed that most (93%) of the coefficients arising from lasso grand products are 1.
+    // This mode uses a precomputed commitment, G, to save some compute.
+    // Where G is the commitment to the all the group elements in the SRS.
```
moodlezoup, line 142:
```suggestion
    // Where G is the commitment to the all-ones vector of length 2^k```

moodlezoup (approved):
Nice work, thanks for taking this on! Just a couple of inaccuracies in the comments to fix

PR #477, opened 2024-10-16
Transcript trait

imikushin:
ProofTranscript is now a type parameter everywhere. 

This is necessary for in-circuit Jolt verification (Keccak is very expensive in R1CS), also for anyone who might want to use their own transcript implementation.


imikushin:
@moodlezoup Would you be able to take a quick look? It'd help before investing effort into fixing the macros and onchain verifier tests.

imikushin:
@moodlezoup this is ready for review

moodlezoup (commented):
One nit: can we rename `DefaultTranscript` to `KeccakTranscript`?
Otherwise, looks good to me!

imikushin:
@moodlezoup Done.

PR #479, opened 2024-10-18
Fix `OutputSumcheckProof`

### LINKED ISSUE #478: `OutputSumcheckProof` does not work as expected
In `OutputSumcheckProof`, the `io_witness_range` is constructed incorrectly:

https://github.com/a16z/jolt/blob/2efbcaeb7a55434247cba055073e3864fe671b38/jolt-core/src/jolt/vm/read_write_memory.rs#L1186-L1196

The `memory_layout.input_start` represents a memory address in the VM but is compared with the index. This makes `io_witness_range` to be all zeros. This leads to the whole output check not working and arbitrary output will pass the check.

This can be reproduced by changing the output of a proof and finding that the proof is still valid.

I will make a PR to fix this.

randomsleep:
Closes #478 

This PR fixes two bugs:

1. The construction of `io_witness_range`  in prover and verifier are incorrect.

2. In the verifier, the `io_witness_range_eval` and `v_io_eval` are evaluated incorrectly. It's due to the first bug that we are not able to catch this bug in the test.

### DIFF
--- a/jolt-core/src/jolt/vm/read_write_memory.rs
+++ b/jolt-core/src/jolt/vm/read_write_memory.rs
@@ -1183,11 +1183,13 @@ where
         let r_eq = transcript.challenge_vector(num_rounds);
         let eq: DensePolynomial<F> = DensePolynomial::new(EqPolynomial::evals(&r_eq));
 
+        let input_start_index = memory_address_to_witness_index(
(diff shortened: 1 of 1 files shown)


moodlezoup (approved):
Thanks for fixing this! I could've sworn we had caught this before but I guess not 🤔

Issue #480, opened 2024-10-18
A truncated execution trace can still be valid


randomsleep:
The jolt verifier just checks if the execution trace follow specific restrictions. However, it does not check if the program really terminates. This means if we truncate a valid trace, it will still be valid but the output will be incorrect.

This can be reproduced by truncating the trace and set the output to 0 in `fib_e2e` test:

```rust
    fn fib_e2e<F: JoltField, PCS: CommitmentScheme<Field = F>>() {
        let artifact_guard = FIB_FILE_LOCK.lock().unwrap();
        let mut program = host::Program::new("fibonacci-guest");
        program.set_input(&9u32);
        let (bytecode, memory_init) = program.decode();
        let (mut io_device, mut trace) = program.trace();
        println!("origin trace length {}", trace.len());
        trace.truncate(100);
        println!("truncated trace length {}", trace.len());
        io_device.outputs[0] = 0; // change the output to 0
        drop(artifact_guard);

        let preprocessing =
            RV32IJoltVM::preprocess(bytecode.clone(), memory_init, 1 << 20, 1 << 20, 1 << 20);
        let (proof, commitments, debug_info) =
            <RV32IJoltVM as Jolt<F, PCS, C, M>>::prove(io_device, trace, preprocessing.clone());
        let verification_result =
            RV32IJoltVM::verify(preprocessing, proof, commitments, debug_info);
        assert!(
            verification_result.is_ok(),
            "Verification failed with error: {:?}",
            verification_result.err()
        );
    }
```

Run the test and it will still pass: 

```cargo test --package jolt-core --lib -- jolt::vm::rv32i_vm::tests::fib_e2e_hyperkzg --exact --show-output ```

We may need to check the program really terminates in the verifier.

moodlezoup:
Nice catch, opened the PR above to fix!

PR #482, opened 2024-10-21
Use termination bit to check that full trace is proven

moodlezoup:
Fixes https://github.com/a16z/jolt/issues/480


ncitron (approved):
LGTM!

PR #485, opened 2024-10-22
feat: add cargo machete to clean up unused deps

sagar-a16z:


sagar-a16z:
@moodlezoup I'm not sure 100% if this is a good idea, might result in false positives but ther was plenty of bloat that this helped me remove from the project. Should we try it?

moodlezoup:
> @moodlezoup I'm not sure 100% if this is a good idea, might result in false positives but ther was plenty of bloat that this helped me remove from the project. Should we try it?

I'm down to give it a try, I think the benefits outweigh the costs

moodlezoup (approved):

Issue #464, opened 2024-09-29
Toolchain uninstallation is missing


nazar-pc:
I expect Jolt to be great, but some people might want to not only install it, but also uninstall.

Just like https://github.com/risc0/risc0 I'd like to request symmetric uninstallation command to revert all the changes that were done.

Issue #463, opened 2024-09-29
`jolt install-toolchain` doesn't say what it does


nazar-pc:
There is zero indication as to what changes did `jolt install-toolchain` do to my machine.
I see it installed `riscv32im-jolt-zkvm-elf` toolchain, which is kind of expected, but it would be nice to see something like this:
```
riscv32im-jolt-zkvm-elf toolchain was successfully installed at /home/username/.rustup/toolchains/rve-nightly
```

And whatever else it did in the process.

And I see it stores both extracted toolchain and original `.tar.gz` for some reason, is that intentional? It just occupies another 434M of disk space for seemingly no good reason.

PR #487, opened 2024-10-22
feat: uninstall-toolchain command

sagar-a16z:
fix https://github.com/a16z/jolt/issues/464, https://github.com/a16z/jolt/issues/463


moodlezoup (approved):
thanks!

Issue #323, opened 2024-04-24
Fix multiple build target warnings on examples


sragss:
When running `cargo build -p jolt-core` the following build warnings appear:
```log
warning: /Users/sragsdale/Documents/Code/a16z/jolt/examples/sha3-ex/guest/Cargo.toml: file `/Users/sragsdale/Documents/Code/a16z/jolt/examples/sha3-ex/guest/src/lib.rs` found to be present in multiple build targets:
  * `lib` target `sha3-guest`
  * `bin` target `guest`
warning: /Users/sragsdale/Documents/Code/a16z/jolt/examples/sha3-chain/guest/Cargo.toml: file `/Users/sragsdale/Documents/Code/a16z/jolt/examples/sha3-chain/guest/src/lib.rs` found to be present in multiple build targets:
  * `lib` target `sha3-chain-guest`
  * `bin` target `guest`
warning: /Users/sragsdale/Documents/Code/a16z/jolt/examples/multi-function/guest/Cargo.toml: file `/Users/sragsdale/Documents/Code/a16z/jolt/examples/multi-function/guest/src/lib.rs` found to be present in multiple build targets:
  * `lib` target `multi-function-guest`
  * `bin` target `guest`
warning: /Users/sragsdale/Documents/Code/a16z/jolt/examples/sha2-ex/guest/Cargo.toml: file `/Users/sragsdale/Documents/Code/a16z/jolt/examples/sha2-ex/guest/src/lib.rs` found to be present in multiple build targets:
  * `lib` target `sha2-guest`
  * `bin` target `guest`
warning: /Users/sragsdale/Documents/Code/a16z/jolt/examples/sha2-chain/guest/Cargo.toml: file `/Users/sragsdale/Documents/Code/a16z/jolt/examples/sha2-chain/guest/src/lib.rs` found to be present in multiple build targets:
  * `lib` target `sha2-chain-guest`
  * `bin` target `guest`
warning: /Users/sragsdale/Documents/Code/a16z/jolt/examples/alloc/guest/Cargo.toml: file `/Users/sragsdale/Documents/Code/a16z/jolt/examples/alloc/guest/src/lib.rs` found to be present in multiple build targets:
  * `lib` target `alloc-guest`
  * `bin` target `guest`
[...]

PR #489, opened 2024-10-24
fix: multiple target cargo build warnings

sagar-a16z:
fixes https://github.com/a16z/jolt/issues/323


moodlezoup (approved):
nice, excited to no longer see these warnings in my terminal 😄

PR #492, opened 2024-10-27
Fix bug in default jolt fibonacci template

sashafrolov:
This import is broken in the file generated by jolt new [project] and the project fails to build as a result. I noticed this while doing the quickstart. Sorry if this was fixed already and has to do with an old version of Jolt on my computer.


moodlezoup (approved):
thanks!

Issue #490, opened 2024-10-24
Prover can use arbitrary memory layout


Merricx:
The Jolt verifier uses memory address of program I/O provided from the proof instead of constructing themselves. 
This allows malicious prover to be able to set the memory address of the output, panic, or termination bit to the same address as the inputs, for example, to fake these values.

It can also bypass #482  since the termination bit can be easily set to `true` via inputs.

This attack can be reproduced in the modified `truncated_trace` test below:

```rust
    #[test]
    fn truncated_trace() {
        let artifact_guard = FIB_FILE_LOCK.lock().unwrap();
        let mut program = host::Program::new("fibonacci-guest");
        program.set_input(&1u8); // change input to 1 so that termination bit equal true
        let (bytecode, memory_init) = program.decode();
        let (mut io_device, mut trace) = program.trace();
        trace.truncate(100);
        // change the output to the same as input to show that we can also forge the output value
        io_device.outputs[0] = 1;
        drop(artifact_guard);

        // change memory address of output & termination bit to the same address as input
        io_device.memory_layout.output_start = io_device.memory_layout.input_start;
        io_device.memory_layout.output_end = io_device.memory_layout.input_end;
        io_device.memory_layout.termination = io_device.memory_layout.input_start;

        let preprocessing =
            RV32IJoltVM::preprocess(bytecode.clone(), memory_init, 1 << 20, 1 << 20, 1 << 20);
        let (proof, commitments, debug_info) =
            <RV32IJoltVM as Jolt<Fr, HyperKZG<Bn254>, C, M>>::prove(
                io_device,
                trace,
                preprocessing.clone(),
            );
        let verification_result =
            RV32IJoltVM::verify(preprocessing, proof, commitments, debug_info);
        assert!(
            verification_result.is_ok(),
            "Verification failed with error: {:?}",
            verification_result.err()
        );
    }
```
Run the test and it should pass the verification.

`cargo test --package jolt-core --lib -- jolt::vm::rv32i_vm::tests::truncated_trace --exact --show-output `

PR #494, opened 2024-10-31
fix: add memory_layout to preprocessing so verifier doesn't rely on the prover

sagar-a16z:
Memory layout is a public input to the Jolt verifier and doesn't need to be trusted from the Proof

fixes https://github.com/a16z/jolt/issues/490

### DIFF
--- a/jolt-core/src/host/mod.rs
+++ b/jolt-core/src/host/mod.rs
@@ -181,9 +181,9 @@ impl Program {
 
     // TODO(moodlezoup): Make this generic over InstructionSet
     #[tracing::instrument(skip_all, name = "Program::trace")]
-    pub fn trace(mut self) -> (JoltDevice, Vec<JoltTraceStep<RV32I>>) {
+    pub fn trace(&mut self) -> (JoltDevice, Vec<JoltTraceStep<RV32I>>) {
         self.build();
-        let elf = self.elf.unwrap();
+        let elf = self.elf.clone().unwrap();
         let (raw_trace, io_device) =
             tracer::trace(&elf, &self.input, self.max_input_size, self.max_output_size);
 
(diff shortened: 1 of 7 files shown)


jolt-core/src/host/mod.rs
```diff
@@ -181,9 +181,9 @@ impl Program {
 
     // TODO(moodlezoup): Make this generic over InstructionSet
     #[tracing::instrument(skip_all, name = "Program::trace")]
-    pub fn trace(mut self) -> (JoltDevice, Vec<JoltTraceStep<RV32I>>) {
+    pub fn trace(&mut self) -> (JoltDevice, Vec<JoltTraceStep<RV32I>>) {
         self.build();
-        let elf = self.elf.unwrap();
+        let elf = self.elf.clone().unwrap();
```
moodlezoup, line 186: we shouldn't need this clone, we pass in `elf` by reference below

jolt-core/src/jolt/vm/rv32i_vm.rs
```diff
@@ -514,4 +550,50 @@ mod tests {
             verification_result.err()
         );
     }
+
+    #[test]
+    #[should_panic]
```
moodlezoup, line 555: why do we have both should_panic and the assertion at the end of this test?

jolt-core/src/host/mod.rs
sagar-a16z: Self.elf is an `Option<PathBuf>` the clone is required to unwrap it. Previously this API would consume itself `trace(mut self)`, which seemed weird to me so I changed it to `trace(&mut self)` and that means we can't just pull elf out of Self anymore since we just have a ref to Self.

Does that make sense?

Issue #496, opened 2024-11-05
Integrate Quarks grand product with sparse GKR implementation


moodlezoup:
We have implemented the "hybrid" grand product described in Section 6 of the Quarks paper, where the first k layers of the grand product tree are proven using a single sumcheck, and the remaining n-k layers are proven using the Thaler '13 grand product argument (based on GKR). https://github.com/a16z/jolt/blob/main/jolt-core/src/subprotocols/grand_product_quarks.rs

Our codebase has two different implementations of Thaler '13 grand products: one is a generic implementation (`BatchedDenseGrandProduct`), and the other is optimized for "sparse" grand products (`ToggledBatchedGrandProduct`), where many gate values are 1. 

Currently, our Quarks hybrid grand product is integrated with `BatchedDenseGrandProduct`, but not `ToggledBatchedGrandProduct`. We should support both.

PR #497, opened 2024-11-07
feat: toggled quark hybrid grand product

sagar-a16z:
fixes https://github.com/a16z/jolt/issues/496


moodlezoup (approved):
Nice work! it's not easy wrangling these different grand product implementations 😅

Issue #498, opened 2024-11-07
Divide-by-zero and division overflow handling


randomsleep:
According to the RISC-V spec, divide-by-zero and division overflow won't raise exception. Instead, it will return specific values.

![image](https://github.com/user-attachments/assets/3b9d2d41-cdee-4bae-8773-355e6b4e816a)

In Jolt `DIV`, `DIVU`, `REM`, and `REMU` instruction this is not handled. There is `AssertValidDiv0Instruction` but it seems it is not used.

randomsleep:
@moodlezoup Oh thanks! The code already handled that. The Jolt paper didn't mention these `AssertValidDiv0Instruction` instruction in `DIV`.

The `sequence_output` in `divu` and `remu` should be updated according to that?
https://github.com/a16z/jolt/blob/0220448cc8d515bbbaf2b9814ef02fd41d5d5261/jolt-core/src/jolt/instruction/divu.rs#L230-L232

moodlezoup:
Ah yes, nice catch. That method's only used in tests so not super urgent but will make a note to fix!

PR #506, opened 2024-11-17
fix(jolt-core): Fixes DIVU and REMU sequence_output

### LINKED ISSUE #499: DIVU and REMU `sequence_output` incorrectly handle 0 divisor
https://github.com/a16z/jolt/issues/498#issuecomment-2463692225

mw2000:
We return the correct output from sequence output when the divisor is zero, for divu and remu.

Should fix #499

### DIFF
--- a/jolt-core/src/jolt/instruction/remu.rs
+++ b/jolt-core/src/jolt/instruction/remu.rs
@@ -215,8 +215,12 @@ impl<const WORD_SIZE: usize> VirtualInstructionSequence for REMUInstruction<WORD
         } else {
             x / y
         };
-         
-        if y == 0 { x } else { x - quotient * y }
+
+        if y == 0 {
+            x
+        } else {
+            x - quotient * y
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


mw2000:
From #498 I believe we just need to fix the sequence output (which is used in the jolt virtual sequence test macro) to be output uint::max when the divisor is zero.

jolt-core/src/jolt/instruction/remu.rs
```diff
@@ -206,10 +206,20 @@ impl<const WORD_SIZE: usize> VirtualInstructionSequence for REMUInstruction<WORD
...
-            _ => panic!("Unsupported WORD_SIZE: {}", WORD_SIZE),
+        let quotient = if y == 0 {
+            match WORD_SIZE {
+                32 => u32::MAX as u64,
+                64 => u64::MAX,
+                _ => panic!("Unsupported WORD_SIZE: {}", WORD_SIZE),
+            }
+        } else {
+            x / y
+        };
```
moodlezoup, line 217: let's get rid of this and just move the existing `match` into the `else` block below

mw2000: Sounds good!

mw2000: Fix up

moodlezoup (approved):
thanks Mihir!

PR #507, opened 2024-11-20
fix(jolt-core): Fixes & lookup saving for `ValidRemainder` instrs

quangvdao:
This PR does two things:
1. It fixes the panic behavior in `AssertValidSignedRemainder` when either of the operands is the minimum `i32`/`i64` number. This is because `lookup_entry()` invokes `.abs()` which panics when this minimum value is passed in. The fix is to use the panic-free `unsigned_abs()` version instead.
2. It reduces one lookup in `AssertValidUnsignedRemainder`. This savings is similar to the one in `SLTU`, since the less-than unsigned function computation does not require the result of looking up the last chunk to `EqSubtable`.


moodlezoup (approved):
ty!

PR #514, opened 2024-11-29
chore: replace lazy_static with LazyLock

kevinji:
`LazyLock` does the same thing but is built-in as of Rust 1.80.0.


moodlezoup (approved):
thanks!

PR #525, opened 2024-12-15
fix: broken link risc-v.md

youyyytrok:
This pull request fixes a broken link in the `risc-v.md` file. The link to the Wagmi documentation was updated to point to the correct URL.

- Fixed `https://github.com/riscv/riscv-asm-manual/blob/master/riscv-asm.md` to `https://github.com/riscv-non-isa/riscv-asm-manual/blob/main/src/asm-manual.adoc`.


moodlezoup (approved):
ty!

Issue #512, opened 2024-11-25
Instruction operand chunks should be range-checked


moodlezoup:
Instruction operands are broken into `C = 4` [chunks](https://github.com/a16z/jolt/blob/main/jolt-core/src/r1cs/constraints.rs#L154-L171) of 8-bits each, and those chunks are (for some instructions) [interleaved](https://github.com/a16z/jolt/blob/main/jolt-core/src/r1cs/constraints.rs#L184-L188) to form the lookup query. 

Those operand chunks should be range-checked to [0, 2^8).

0xAndoroid:
outdated

PR #526, opened 2024-12-16
fix(r1cs): We add a range check to the chunk

mw2000:
Should solve #512


moodlezoup (commented):
thanks for looking into this Mihir! This approach works, but I think for performance reasons we'd prefer to use Lasso-style range checks (similar to what we do in `timestamp_range_check.rs`). Justin also has some ideas on how we could do them even more efficiently, so I was planning on tackling these range checks myself in the coming weeks

mw2000:
sounds good sir, will let you take this on 🫡

PR #528, opened 2024-12-16
fix: Fix incorrect method calls in sysinfo usage

famouswizard:
This update addresses an issue in how methods from the `sysinfo` library were being used. Previously, `name`, `os_version`, and `host_name` were incorrectly called as static methods on `System`.
However, these methods are instance-specific and must be invoked on a `System` object.  

The corrected implementation ensures proper instantiation of the `System` class before accessing these methods, aligning with the library's intended usage.

This fix improves compatibility and eliminates potential runtime errors.


moodlezoup (approved):

moodlezoup (changes requested):
looks like they are in fact static?

PR #530, opened 2024-12-16
The Return of the King

arasuarun:
Optimization crushing the second Spartan sumcheck. Instead of the inner sumcheck with `a * b` rounds where `a = num_vars_padded` and `b = num_steps_padded`, we now perform two sumchecks of `a` rounds and `b` rounds, respectively. 

There's a lot to document. I'm making the PR for now to unblock @moodlezoup's other optimizations. 

Edit: The below change is no longer needed. The optimization works with the trace as is. 
~~A *major* change in the code that we should discuss before merging:~~ 
- ~~There was an issue with how the cross-step (also called "non-uniform" constraints) were handled for the last step. Now, the changes in this PR assume that all witness elements belonging to the last step are 0, so the step is effectively ignored. This requires some changes to the trace polynomials (in `builder.rs`).~~
- ~~In particular, this means that if a program runs for *exactly* a power of 2 number of steps, we'd still need to pad it to the next power of 2. I don't think this is an issue for now but might pop up if we're sharding programs later for folding.~~

How ready is the PR? It's good enough to build on but needs cleanup before merging into main. 
- Needs code cleanup like variable names and so on (especially in `key.rs`). 
- I don't think there are any major wasteful clones. 
- Needs a sweep to see which parts can be parallelized better. Particularly a loop in the prover function that possibly runs into some of the old paging issues we faced before.

Also need to test with flamegraphs to make sure we're seeing the right performance improvements in the prover's sumchecks. 

Update: Performance improvement on Macbook M3.

| Trace Len | on Main  | with Opt | % improvement |
|-----------|------:|------------:|------:|
| 350k      |  1.3s |        1.0s |   23% |
| 1.3m      |  5.5s |        3.8s |   31% |
| 2.6m      | 10.5s |        7.7s |   27% |
| 10m       |   87s |         31s |   66% |

At 10m, total Jolt prove time sees a 33% improvement.

### DIFF
--- a/jolt-core/src/poly/eq_poly.rs
+++ b/jolt-core/src/poly/eq_poly.rs
@@ -101,33 +101,6 @@ impl<F: JoltField> EqPolynomial<F> {
     }
 }
 
-/* This MLE is 1 if y = x + 1 for x in the range [0... 2^l-2].
-That is, it ignores the case where x is all 1s, outputting 0.
-Assumes x and y are provided big-endian. */
-pub fn eq_plus_one<F: JoltField>(x: &[F], y: &[F], l: usize) -> F {
-    let one = F::one();
-
-    /* If y+1 = x, then the two bit vectors are of the following form.
-        Let k be the longest suffix of 1s in x.
-        In y, those k bits are 0.
-        Then, the next bit in x is 0 and the next bit in y is 1.
-        The remaining higher bits are the same in x and y.
-    */
-    (0..l)
-        .into_par_iter()
-        .map(|k| {
-            let lower_bits_product = (0..k)
-                .map(|i| x[l - 1 - i] * (F::one() - y[l - 1 - i]))
-                .product::<F>();
-            let kth_bit_product = (F::one() - x[l - 1 - k]) * y[l - 1 - k];
-            let higher_bits_product = ((k + 1)..l)
-                .map(|i| x[l - 1 - i] * y[l - 1 - i] + (one - x[l - 1 - i]) * (one - y[l - 1 - i]))
-                .product::<F>();
-            lower_bits_product * kth_bit_product * higher_bits_product
-        })
-        .sum()
-}
-
 impl<F: JoltField> EqPlusOnePolynomial<F> {
     pub fn new(x: Vec<F>) -> Self {
         EqPlusOnePolynomial { x }
--- a/jolt-core/src/poly/eq_poly.rs
+++ b/jolt-core/src/poly/eq_poly.rs
@@ -128,7 +128,7 @@ pub fn eq_plus_one<F: JoltField>(x: &[F], y: &[F], l: usize) -> F {
         .sum()
 }
 
-impl <F: JoltField> EqPlusOnePolynomial<F> {
+impl<F: JoltField> EqPlusOnePolynomial<F> {
     pub fn new(x: Vec<F>) -> Self {
         EqPlusOnePolynomial { x }
     }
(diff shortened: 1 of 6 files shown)


ncitron:
👑

jolt-core/src/r1cs/key.rs
```diff
@@ -5,7 +5,7 @@ use sha3::Sha3_256;
 
 use crate::{
     field::JoltField,
-    poly::eq_poly::{eq_plus_one, EqPolynomial},
```
moodlezoup, line 8: I think we can delete the `eq_plus_one` function from `eq_poly.rs` now

jolt-core/src/r1cs/spartan.rs
```diff
@@ -145,43 +152,177 @@ where
...
+            .collect_into_vec(&mut bind_z_ry_var);
+        drop(_guard);
+        drop(span);
+
+        let num_rounds_shift_sumcheck = num_steps_bits;
+        assert_eq!(bind_z_ry_var.len(), eq_plus_one_rx_step.len());
+
+        let mut shift_sumcheck_polys = vec![
+            MultilinearPolynomial::LargeScalars(DensePolynomial::new(bind_z_ry_var)),
+            MultilinearPolynomial::LargeScalars(DensePolynomial::new(eq_plus_one_rx_step)),
```
moodlezoup, line 274:
```suggestion
            MultilinearPolynomial::from(bind_z_ry_var),
            MultilinearPolynomial::from(eq_plus_one_rx_step),
```

jolt-core/src/poly/eq_poly.rs
```diff
@@ -123,3 +127,89 @@ pub fn eq_plus_one<F: JoltField>(x: &[F], y: &[F], l: usize) -> F {
...
+            if i == 0 {
+                return;
+            }
+            let ell = r.len();
+            let step = 1 << (ell - i); // step = (full / size)/2
+
+            let mut selected: Vec<_> = eq_evals.par_iter_mut().step_by(step).collect();
+
+            selected
+                .par_chunks_mut(2)
```
moodlezoup, line 184: I think this may save some allocations
```suggestion
            eq_evals
                .par_iter_mut()
                .step_by(step)
                .par_chunks(2)
```

jolt-core/src/poly/eq_poly.rs
```diff
@@ -123,3 +127,89 @@ pub fn eq_plus_one<F: JoltField>(x: &[F], y: &[F], l: usize) -> F {
...
+                .enumerate()
+                .for_each(|(_j, chunk)| {
+                    let (x, y) = chunk.split_at_mut(1);
+                    let x = &mut x[0];
+                    let y = &mut y[0];
+                    **y = **x * r[i - 1];
+                    **x *= F::one() - r[i - 1];
```
moodlezoup, line 191:
```suggestion
                    **x -= *y;
```

moodlezoup (approved):
great stuff, king
just a few nits

arasuarun:
Some numbers for `Spartan::prove` in main and with this optimization. All run on my M3 Macbook. 

| Trace Len | on Main  | with Opt | % improvement |
|-----------|------:|------------:|------:|
| 350k      |  1.3s |        1.0s |   23% |
| 1.3m      |  5.5s |        3.8s |   31% |
| 2.6m      | 10.5s |        7.7s |   27% |
| 10m       |   87s |         36s |   69% |

At 10m, total Jolt prove time sees a 25% improvement.

PR #554, opened 2025-01-09
Reorder R1CS constraints

jprider63:
This adds the flag `"reorder"` which reorders R1CS constraints so that all constraint rows for a given step are next to each other. This is necessary for streaming to minimize memory usage. 


jprider63:
Merged with #561.

PR #559, opened 2025-01-12
Implement Montgomery's batch inversion optimization

MozirDmitriy:
Description:
Added Montgomery's batch inversion trick to Fr.sol for efficient multiple field element inversions. This optimization:
- Reduces the number of expensive modular inversions from n to 1
- Replaces expensive inversions with cheaper multiplications
- Maintains the same mathematical correctness
- Improves gas efficiency for operations requiring multiple inversions


0xAndoroid:
We are planning to deprecate onchain verifier.

PR #561, opened 2025-01-13
Reorder R1CS constraints + CompactPolynomial refactor

moodlezoup:
This is a stack of two separate changes: 
1. https://github.com/a16z/jolt/pull/554 by @jprider63, which reorders Jolt's R1CS witnesses to be in "constraint-major" order (rather than "step-major" order). This is incremental progress toward a "streaming" Spartan prover algorithm. 
2. A significant overhaul/optimization of Jolt's data structures for multilinear polynomials, described in further detail below.

The main observation behind the optimizations/refactor is that most of Jolt's polynomials have small (relative to a 256-bit field) coefficients. Before, ~all polynomials were represented by the `DensePolynomial` struct, which uses a field element  per coefficient. With this PR we introduce the `CompactPolynomial` struct, which instead uses Rust's primitive integer types (`u8`, `u16`, etc.) to represent coefficients. 

This reduces prover memory usage and leads to some speed improvements as well. Note that once a CompactPolynomial is bound, its coefficients will need to be represented as field elements (but there will be half as many as there are unbound coefficients). 

A side effect of representing coefficients as primitive integers is that we need some tricks to avoid spending too much time converting the coefficients to Montgomery form when doing (field) arithmetic with them. These new tricks are described in the changes to `ark.rs`

The final main change in this PR is the introduction of `SpartanInterleavedPolynomial`, which is used to represent the Az, Bz, and Cz polynomials used in the first Spartan sumcheck. The R1CS reordering changes slowed down the computation of Az, Bz, and Cz; the `SpartanInterleavedPolynomial` refactor brings the time back to parity with main. The polynomial implementation itself is sort of a combination of the old `SparsePolynomial` and the `SparseInterleavedPolynomial` used sparse grand products.


jolt-core/src/msm/mod.rs
```diff
@@ -20,138 +23,122 @@ pub type GpuBaseType<G: ScalarMul> = G::MulBase;
...
-#[derive(Debug, Clone, Copy, Eq, PartialEq, Ord, PartialOrd)]
-pub enum MsmType {
-    Zero,
-    One,
-    Small(usize),
-    Medium(usize),
-    Large(usize),
-}
-
-impl MsmType {
```
sagar-a16z, line 32: I thought having this around made it easier to read/maintain.

moodlezoup: I removed it because it gets weird with `CompactPolynomial` –– the `Medium` size straddles the `u8` and `u16` size

jolt-core/src/poly/commitment/commitment_scheme.rs
```diff
@@ -82,18 +65,10 @@ pub trait CommitmentScheme<ProofTranscript: Transcript>: Clone + Sync + Send + '
...
         transcript: &mut ProofTranscript,
     ) -> Self::Proof;
-    fn batch_prove(
-        setup: &Self::Setup,
-        polynomials: &[&DensePolynomial<Self::Field>],
-        opening_point: &[Self::Field],
-        openings: &[Self::Field],
-        batch_type: BatchType,
-        transcript: &mut ProofTranscript,
-    ) -> Self::BatchedProof;
```
moodlezoup, line 96: Deleting `batch_prove` and `batch_verify` from trait because we now use the batched opening proof protocol in `opening_proof.rs` instead

PR #564, opened 2025-01-16
Add meaningful error message for modular inversion failure

VolodymyrBg:
Description:
Replace empty revert with descriptive error message in Fr.sol invert function.
The error message now clearly indicates when modular inversion operation fails,
making it easier to debug and understand failures during contract execution.


VolodymyrBg:
@huitseeker @mmaker @jeluard @joachimneu

0xAndoroid:
Appreciate the contribution, however we are planning to deprecate current oncahin verifier.

PR #569, opened 2025-01-24
fix: optimize opening proof gpu usage

sagar-a16z:
- Supports building `--features icicle` on MacOS (skips the CUDA backend from icicle-jolt)
- Ammortizes gpu copy costs for KZG Opening proofs 
- Reduce vector allocations by using `std::Borrow`
- Batched hyperKZG witness generation 
- Reintroduced GPU usage for trace commitments where scalars are  > 10bit. (needs review, potential memory regression)
- removed cloning of scalars before appending to transcript in HyperKZG

Brings a 1.4x(30%) speedup to the KZG opening proof. GPU opening proofs are ~2.8x(65%) faster than CPU now. 
Net improvement of 15% in prover time GPU vs CPU.


jolt-core/src/msm/mod.rs
```diff
@@ -219,45 +219,73 @@ where
...
+                        MultilinearPolynomial::U32Scalars(poly) => {
+                            Either::Right((i, max_num_bits, poly.coeffs_as_field_elements()))
+                        }
+                        MultilinearPolynomial::U64Scalars(poly) => {
+                            Either::Right((i, max_num_bits, poly.coeffs_as_field_elements()))
+                        }
+                        MultilinearPolynomial::I64Scalars(poly) => {
+                            Either::Right((i, max_num_bits, poly.coeffs_as_field_elements()))
+                        }
+                        _ => Either::Left((i, max_num_bits, poly)),
```
moodlezoup, line 283:
```suggestion
                        MultilinearPolynomial::U8Scalars(_) => unreachable!("MultilinearPolynomial::U8Scalars cannot have more than 10 bits")
```

moodlezoup (approved):
Nice work! Just a couple of nits but looks good to merge otherwise

PR #570, opened 2025-01-26
fix: Fix incorrect use of unsigned_abs() for i32 type in rv_trace.rs

fragchain:
`unsigned_abs()` was being used on an `i32` type, which doesn't exist.
I replaced it with `abs()` to correctly get the absolute value as `i32`, and then cast it to `u64` as needed.


moodlezoup:
? https://doc.rust-lang.org/std/primitive.i32.html#method.unsigned_abs

PR #572, opened 2025-01-29
Feature/ci plot benchmarks

zutshi:
**Summary**  
This PR adds a CI workflow that runs and plot metrics for the examples (in `./examples`), triggered by a new commit to the main branch. For now, the metrics include the wall time and peak memory required to run each of the example. The CI workflow uses [benchmark-action/github-action-benchmark@v1](https://github.com/benchmark-action/github-action-benchmark) to create the plots. The plots are then checked in as a webpage into the desired branch, ready to be published.  

**Example output**:   
https://galoisinc.github.io/jolt/dev/bench/

**Additional steps required to get this working:**  
1. Create an empty new branch (default name: `gh-pages`). This is where github-action-benchmark will push the webpages after every commit to main. The branch name can be changed in `ci-bench.yml`
2. Turn on github pages and automatic build and deploy workflow. Using the [github repo's settings](https://github.com/GaloisInc/jolt/settings/pages), configure the pages to be served from the desired branch. Select `deploy from branch` under `Build and deployment` and then select `gh-pages` under branch.

**Caveats:**
- The plots are triggered only when commits are pushed to the remote main branch.
- For now, the metrics are gathered using GNU time which collects the wall time and [Maximum Resident Size](https://en.wikipedia.org/wiki/Resident_set_size) for each example.
-  Some examples are excluded since they consume more memory than the usual runner limits and get the runner killed. Refer to `run_benchmarks.sh`: 
```bash
exclusion_list=("collatz" "overflow" "sha3-chain")
```

---


moodlezoup (commented):
Very cool! One thing to note –– the host `main` functions in the `examples` directory will (in addition to proving) compile the guest, compute the preprocessing, and verify the proof. For peak memory Ideally we would be able to isolate the prover time. 

One quick and dirty way to do so would be to measure the time from within the proc-macro-generated [`prove_*`](https://github.com/a16z/jolt/blob/main/jolt-sdk/macros/src/lib.rs#L248-L275) function and write it to stdout to be parsed by `run_benchmarks.sh`. Wdyt? Open to other suggestions as well, I know it's probably not the cleanest solution

zutshi:
Sure, looks possible. The reason to use a bash script is to enable support for compiling metrics generated from different sources. Since the examples already print stats upon completion, some of these stats can be read back in and plotted.

jprider63:
It seems straightforward to wrap the `prove_*` functions to record their runtime (I'm not sure it makes sense to modify the macro though since that would impact users). 

Isolating the prover's peak memory usage seems harder. Is it possible to do the preprocessing in a separate binary and then load the compiled guest/setup from disk?

moodlezoup:
> Isolating the prover's peak memory usage seems harder. 

Yeah, though in most cases I expect the peak memory of proving to be higher than the peak memory of preprocessing/guest compilation so it might not be an issue

> Is it possible to do the preprocessing in a separate binary and then load the compiled guest/setup from disk?

That said, this is definitely the right way to do things. We should add `CanonicalSerialize + CanonicalDeserialize` to the preprocessing and have the prover load it from disk. This has been on the backlog for a while, just haven't gotten around to it

jprider63:
In that case, for now should we modify the `main.rs` for each example to record and print prover time? Then when the ability to load the preprocessing from disk is complete, we can swap it over? 

We may have folks who can start working on preprocessing serialization. I'll check.

moodlezoup:
> In that case, for now should we modify the `main.rs` for each example to record and print prover time? Then when the ability to load the preprocessing from disk is complete, we can swap it over?
> 
> We may have folks who can start working on preprocessing serialization. I'll check.

Yes, that sounds like a good plan.

jprider63:
I've made the change to only benchmark prover times for the examples so this should be ready for review. 

Just a reminder that github pages need to be enabled from the `gh-pages` branch before merging as @zutshi mentioned. 

> That said, this is definitely the right way to do things. We should add CanonicalSerialize + CanonicalDeserialize to the preprocessing and have the prover load it from disk. This has been on the backlog for a while, just haven't gotten around to it

Is there an existing issue for this? I couldn't find one. @protoben has started looking into this.

moodlezoup:
> I've made the change to only benchmark prover times for the examples so this should be ready for review.
> 
> Just a reminder that github pages need to be enabled from the `gh-pages` branch before merging as @zutshi mentioned.

Thanks! 

> > That said, this is definitely the right way to do things. We should add CanonicalSerialize + CanonicalDeserialize to the preprocessing and have the prover load it from disk. This has been on the backlog for a while, just haven't gotten around to it
> 
> Is there an existing issue for this? I couldn't find one. @protoben has started looking into this.

No existing issue, will write one up and link here.

moodlezoup:
https://github.com/a16z/jolt/issues/585

PR #586, opened 2025-02-07
feat: add dory commitment scheme

g1684774:


moodlezoup:
Hi @g1684774 ! Thanks for taking this on! I haven't had a chance to thoroughly review your PR yet, but let me try to answer some of your questions first:

> * How should I populate the Transcript? Is it needed for Dory?
Our Transcript struct should be used for the "reduce" steps in the Dory code. Both are used to implement the [Fiat-Shamir](https://www.zkdocs.com/docs/zkdocs/protocol-primitives/fiat-shamir/) transform –– we have some running "digest" or "state", and whenever the prover makes a commitment of some sort, the commitment has to be "absorbed" into the digest (aka appended to the transcript) by making (in broad strokes) the following update:
state := hash(state, commitment)

> * How to deal with other types of MultilinearPolynomial;
See inline comment

> * Does Dory support a better type of batching other than just looping over;
What you have is fine for now!

> * Multiple operations might fail, how to handle them? Is it ok to panic?
Yes, panic is ok

> * Go implementation used Sha2 for digest, is it ok to use Sha3?
See above answer about the Transcript; note that we have a `KeccakTranscript` struct that implements the `Transcript` trait

> * How does the reduce part fits jolt?
See above answer about the Transcript

Btw, if you haven't already you might want to check out Section 15.4 of Justin's [book](https://people.cs.georgetown.edu/jthaler/ProofsArgsAndZK.pdf), which covers Dory. There may be some discrepancies between the description there and the Go implementation, but it should provide some high-level context for why things work the way they do.

jolt-core/src/poly/commitment/dory/scalar.rs
```diff
@@ -0,0 +1,101 @@
...
+            .zip(poly.iter())
+            .map(|(a, b)| *a * *b)
+            .collect::<Vec<G1<P>>>();
+
+        let v2 = params
+            .g2v
+            .iter()
+            .zip(poly.iter())
+            .map(|(a, b)| *a * *b)
+            .collect::<Vec<G2<P>>>();
```
moodlezoup, line 34: These are multi-scalar multiplications (MSMs), which we have our own optimized implementation for: https://github.com/a16z/jolt/blob/main/jolt-core/src/msm/mod.rs#L176-L220

Note that it supports all types of MultilinearPolynomials

g1684774:
Thanks for the response. I have a few more questions:

- Does that mean that the Go implementation already does the Fiat-Shamir transform by calling `ro()` whereas jolt wants to provide the points by calculating the points by itself sending digest produced by the transcript?

- And if the previous question is true, does it mean that I need to unwind the recursive part and let jolt call it for me? Will Jolt call it log(n) times before calling verify? Does jolt provide a state telling me if is step1 or step2?

- Something that I couldn't quite get it for Dory is about the Public Params. Are the Public Params actually a Vec<PublicParams>?

moodlezoup:
> Thanks for the response. I have a few more questions:
> 
> * Does that mean that the Go implementation already does the Fiat-Shamir transform by calling `ro()` whereas jolt wants to provide the points by calculating the points by itself sending digest produced by the transcript?

Where you see `β := step1Elements.RO()` in the Go code, for example, we'd instead want to append the `ReduceProverStep1Elements` struct to the transcript, then do `let beta = transcript.challenge_scalar()`
Hopefully that answers your question?

> * Something that I couldn't quite get it for Dory is about the Public Params. Are the Public Params actually a Vec?

The public params are the "preprocessing" required for Dory. They consist mainly of two vectors of group elements (from G1 and G2 respectively). This stuff should be output by `CommitmentScheme::setup`.

jolt-core/src/poly/commitment/dory/scalar.rs
g1684774: I took a deeper look and it seems that this is not MSM since I'm not taking the `.sum()`. This is a conversion of [ScalarField] -> [G1] to be used in the next steps. Does that make sense or am I missing anything?

jolt-core/src/poly/commitment/dory.rs
```diff
@@ -0,0 +1,148 @@
...
+
+fn append_gt<P: Pairing, ProofTranscript: Transcript>(transcript: &mut ProofTranscript, gt: Gt<P>) {
+    let mut buf = vec![];
+    gt.serialize_uncompressed(&mut buf).unwrap();
+    // Serialize uncompressed gives the scalar in LE byte order which is not
+    // a natural representation in the EVM for scalar math so we reverse
+    // to get an EVM compatible version.
+    buf = buf.into_iter().rev().collect();
+    transcript.append_bytes(&buf);
+}
```
g1684774, line 60: There's no `append_gt` in transcript, should I add it there?

PR #592, opened 2025-02-20
Remove unnecessary truncateOverflow subtable

benoitrazet:
The truncateOverflow subtable always returns 0 with WORD_SIZE = 32 or 64 bits. All the calls to this subtable are eliminated.


jprider63:
I added a link to the CI benchmark results in the README, but let me know if you don't want to include this.

moodlezoup (approved):
good catch, thanks!

PR #599, opened 2025-02-27
chore: fix issue with missing mode parameter in the function

hexcow:
I noticed that the `mode` parameter wasn’t being passed to the function, causing issues. To resolve this, I removed the reference to `mode`, which was unnecessary. Now the function works correctly and compiles without errors.

Here’s the updated line:

```rust
&format!("{} Commit(threshold: {}% Ones)", name, threshold),
```

### DIFF
--- a/jolt-core/benches/commit.rs
+++ b/jolt-core/benches/commit.rs
@@ -80,7 +80,7 @@ fn benchmark_commit<PCS, F, ProofTranscript>(
         .map(|layer| MultilinearPolynomial::from(layer))
         .collect::<Vec<_>>();
     c.bench_function(
-        &format!("{} Commit(mode:{:?}): {}% Ones", name, threshold),
+        &format!("{} Commit: {}% Ones", name, threshold),
         |b| {
             b.iter(|| {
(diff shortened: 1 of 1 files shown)


jolt-core/benches/commit.rs
```diff
@@ -80,7 +80,7 @@ fn benchmark_commit<PCS, F, ProofTranscript>(
         .map(|layer| MultilinearPolynomial::from(layer))
         .collect::<Vec<_>>();
     c.bench_function(
-        &format!("{} Commit(mode:{:?}): {}% Ones", name, mode, threshold),
+        &format!("{} Commit(mode:{:?}): {}% Ones", name, threshold),
```
moodlezoup, line 83:
```suggestion
        &format!("{} Commit: {}% Ones", name, threshold),
```

moodlezoup (commented):
ty, need to remove mode from the string as well

hexcow:
> moodlezoup

[moodlezoup](https://github.com/moodlezoup) I rushed a bit and clicked commit.

moodlezoup (approved):
thanks!

PR #606, opened 2025-03-05
 Update jolt-sdk dependency configuration with host feature

leopardracer:
Description:
Updates the jolt-sdk dependency configuration in src/main.rs to explicitly include the "host" feature flag. This change:

- Renames the dependency from `jolt` to `jolt-sdk`
- Adds the "host" feature flag to enable host-specific functionality
- Maintains the same git repository source

This modification ensures proper host feature functionality is available in the project.


leopardracer:
@moodlezoup

leopardracer:
@moodlezoup

PR #609, opened 2025-03-07
Rename non-uniform constraints to cross-step constraints

arasuarun:


arasuarun:
I combed through all occurrences of "`non-uni`", "`non_uni`", "`non uni`" and "`nonuni`" (did a case-insensitive search). Hopefully didn't miss anything.

We should keep the "uniform" constraints name cause that makes sense and it would be too much to change anyway.

moodlezoup (approved):
thanks!

PR #636, opened 2025-04-08
fix prover input in fibonacci

marsella:
I copied the input from the `analyze_fib` function instead of from `prove_fib`, so [the action that runs the examples failed](https://github.com/a16z/jolt/actions/runs/14324219897/job/40146699700). This should fix it!


moodlezoup:
thanks!

PR #643, opened 2025-04-26
Update actions/checkout to v4

leopardracer:
**Description:**
Updates the GitHub Actions checkout action from v3 to v4 across all workflow jobs.

This update brings the latest improvements and security fixes from the checkout action.

PR #644, opened 2025-04-28
add CI job to test the `jolt` SDK

marsella:
This adds a job to the Rust workflow that builds the `jolt` binary and makes sure that the code it generates is valid.

Note that it uses the `actions/checkout@v4`; I didn't change the rest of the checkouts to use the most updated version, but #643 does!

I think the clippy changes are unrelated but I can fix them here (or in a separate PR) if necessary. EDIT: Fixed in #646.


moodlezoup (approved):
thanks!

PR #646, opened 2025-04-28
Fix clippy errors

marsella:
This fixes a bunch of clippy errors that emerged in the last 3 weeks, using `clippy`'s automatic `--fix` flag.

It also removes a deprecated action from the CI. I can revert that change or put it in a separate PR since it is adjacent to the direct fixes.


moodlezoup (approved):
thanks!

PR #653, opened 2025-05-01
docs: update 404-error link 

gap-editor:
it was like that. now it is fixed. 
Zrzut ekranu 2025-05-01 231807


gap-editor:
hello. something wrong here?

PR #654, opened 2025-05-02
feat: update arkworks to v0.5.0, implement new optimized mul u64

quangvdao:
This PR bumps the (custom) arkworks version to v0.5.0, which includes the optimized multiplication of BN254 field elements with u64.

Not only is the new `mul_u64` method more efficient (x2.4 faster compared to field mult, versus only ~30% faster in the previous version), it also produces the **same** result as regular field mult. This means we can cut out lots of code that dealt with this mismatch.

Cutting out code also allows for further optimization. For instance, the binding process for `CompactPolynomial` is simpler now, and we can eliminate half of the number of multiplications! Essentially, we went from

`(mont_r2 * (1 - r)).mul_u64(a) + (mont_r2 * r).mul_u64(b)` to just `(1 - r) a + r b = a +- r.mul_u64(|a - b|)`

(well, I slightly lied. `a` still needs to be converted into Montgomery form which is 1 mult. But if `a` is small we have the precomputed table)

The v0.5.0 version seems to remove `FixedBase` MSMs in favor of `VariableBase` MSMs. The documentation claims potential efficiency improvement, but I haven't checked.


quangvdao:
@moodlezoup Should be ready to merge now. I haven't had the chance to bench yet, so if there are performance degradations I can fix that in the next PR.

jolt-core/src/field/ark.rs
```diff
@@ -209,18 +205,15 @@ mod tests {
...
+            assert_eq!(<Fr as JoltField>::from_u64(x), Fr::one().mul_u64(x));
         }
 
         for _ in 0..256 {
             let x = rng.next_u64();
             let y = Fr::random(&mut rng);
             assert_eq!(
                 y * <Fr as JoltField>::from_u64(x),
-                (y * Fr::montgomery_r2().unwrap()).mul_u64_unchecked(x)
+                (y * Fr::one()).mul_u64(x)
```
moodlezoup, line 216:
```suggestion
                y.mul_u64(x)
```

jolt-core/src/poly/compact_polynomial.rs
```diff
@@ -151,17 +180,28 @@ impl<T: SmallScalar, F: JoltField> PolynomialBinding<F> for CompactPolynomial<T,
...
             }
         } else {
-            let r_r2 = r * F::montgomery_r2().unwrap_or(F::one());
-            let one_minus_r_r2 = (F::one() - r) * F::montgomery_r2().unwrap_or(F::one());
+            // We want to compute `a * (1 - r) + b * r` where `a` and `b` are small scalars
+            // If `a == b`, we can just return `a`
+            // If `a < b`, we can compute `a + r * (b - a)`
+            // If `a > b`, we can compute `a - r * (a - b)`
+            // Note that we need to "promote" signed small scalars to unsigned ones with subtraction
+            // i.e. |a - b| can be computed as an u64
```
moodlezoup, line 188: nice

moodlezoup (approved):
this is awesome, thank you 🙏
just one nit

jolt-core/src/poly/compact_polynomial.rs
quangvdao: We'll see if it actually leads to a speedup

PR #655, opened 2025-05-03
fix doc comment for `prove_primary_sumcheck`

spiral-ladder:
Params seem to be outdated and no longer match the latest state of the function.


0xAndoroid:
To discourage PR farming minimal PRs are not accepted.

PR #657, opened 2025-05-03
feat: add Gruen's optimization for dense interleaved polynomial (draft)

quangvdao:
This **draft** PR implements Gruen's optimization for one particular protocol in `dense_interleaved_poly.rs`.

The point is to introduce the new `SplitEqPolynomial` type tailored to Gruen's optimization, and demonstrate the decrease in evaluation points needed (we only need evals at 0 and infinity, instead of at 0, 2, and 3).

The new changes are all marked as `new`. In particular, we have `compute_cubic_new` being the new version with Gruen's optimization.

There are three more targets for optimization: `sparse_interleaved_poly`, `sparse_grand_product`, and `spartan_interleaved_poly`. All three are more complicated than `dense_interleaved_poly`. I tried changing them in a different branch (https://github.com/quangvdao/jolt/tree/gruen-optimization), but was not successful in debugging to yield the same result as the old method.

Thus, I put the core of the changes in this branch so that people can look at it. I have tested the new implementation of `dense_interleaved_poly` to be the same as the old one, at least for certain cases (should definitely do more testing).

### DIFF
--- a/jolt-core/src/poly/dense_interleaved_poly.rs
+++ b/jolt-core/src/poly/dense_interleaved_poly.rs
@@ -14,7 +14,7 @@ use rayon::{prelude::*, slice::Chunks};
 #[cfg(test)]
 use super::dense_mlpoly::DensePolynomial;
 use super::{
-    split_eq_poly::{SplitEqPolynomial, NewSplitEqPolynomial},
+    split_eq_poly::{NewSplitEqPolynomial, SplitEqPolynomial},
     unipoly::{CompressedUniPoly, UniPoly},
 };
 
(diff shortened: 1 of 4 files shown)


quangvdao:
I’m working on spartan interleaved poly, including my small value optimization as well, so that won’t be necessary.

I guess the question is where else this could be relevant in Twist and Shout. Perhaps nowhere? In which case you don’t have to do anything :)

I haven’t benchmarked yet but I will.

jprider63:
Understood, we'll plan to work on those sumchecks relevant to Twist/Shout then.

PR #658, opened 2025-05-04
feat: add optimized `mul_i128`, change `from_u64`, revert `FixedBase` change

quangvdao:
WIP

This PR is downstream of new progress in [v0.5.0-optimize-mul-u64](https://github.com/a16z/arkworks-algebra/tree/v0.5.0-optimize-mul-u64), where we defined `mul_u128` and `mul_i128` using Barrett reduction, similar to `mul_u64`.

The new arkworks version also restores the `FixedBase` trait that is needed for `kzg.rs`. Without it, we got a bench timeout.

Finally, we also change the `from_u64` function for BN254 to call `from_u64` in arkworks instead (if the number fits in u16 or u32 then we call that). The reason is that doing 4 lookups + adding them together is slightly slower than just doing a single `Fr::one().mul_u64(val)`. But of course <=2 lookups is faster.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -226,7 +226,7 @@ dependencies = [
 [[package]]
 name = "ark-ec"
 version = "0.5.0"
-source = "git+https://github.com/a16z/arkworks-algebra?branch=v0.5.0-optimize-mul-u64#b926f92b4f655d8ea7980f09097d51589c1f707a"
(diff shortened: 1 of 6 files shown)


quangvdao:
@moodlezoup This should probably be merged soon to avoid failing CI bench

PR #665, opened 2025-05-06
fix: remove binius

sagar-a16z:
Binius is not used and not currently on the roadmap. 
Downloading binius from Gitlab is also very slow and makes our builds appear slow.


moodlezoup (approved):

wyattbenno777:
Thank you!!

sragss:
RIP

sagar-a16z:
> RIP

Don't worry! Binius is only on a temporary break.

PR #667, opened 2025-05-07
chore: add build script to capture Git commit information

zouguangxian:
This PR introduces a new `build.rs` file that retrieves the latest Git commit hash, short hash, and date, and sets them as environment variables for use in the application. Additionally, the `main.rs` file is updated to include a version function that formats the version string with the Git information.


moodlezoup (approved):
thanks!

PR #668, opened 2025-05-07
Fix broken link in GKR documentation

GarmashAlex:
Changed the link to memory-checking from .html to .md extension in the GKR background documentation to fix the broken reference and ensure proper navigation between documentation files.


emlazzarin:
Thank you! Now in #725

PR #671, opened 2025-05-08
Feat/cycle tracking

markosg04:
- Adds support for new feat: RISCV cycle tracking:
- Allows for the use of markers directly into guest rust code to track how many "real" and "virtual" (emulated) cycles a code span takes
- Adds ECALL (again?) into the emulator (this will be needed for "psuedo precompiles" / extensions)
- "zero" overhead profiling

Example usage:
```rust
use jolt::{end_cycle_tracking, start_cycle_tracking};

#[jolt::provable]
fn fib(n: u32) -> u128 {
    let mut a: u128 = 0;
    let mut b: u128 = 1;
    let mut sum: u128;

    start_cycle_tracking("fib_loop"); // Use `start_cycle_tracking("{name}")` to start a cycle span

    for _ in 1..n {
        sum = a + b;
        a = b;
        b = sum;
    }
    end_cycle_tracking("fib_loop"); // Use `end_cycle_tracking("{name}")` to end a cycle span
    b
}

```

### DIFF
--- a/tracer/src/emulator/cpu.rs
+++ b/tracer/src/emulator/cpu.rs
@@ -347,7 +347,7 @@ impl Cpu {
 
         // check if current instruction is real or not for cycle profiling
         if instr.is_real() {
-            self.executed_instrs = self.executed_instrs.wrapping_add(1);
+            self.executed_instrs += 1;
         }
         self.x[0] = 0; // hardwired zero
 
@@ -1446,6 +1446,16 @@ impl Cpu {
         match event {
             JOLT_CYCLE_MARKER_START => {
                 let label = self.read_c_string(ptr)?; // guest NUL-string
+
+                // Check if there's already an active marker with the same label
+                let duplicate = self
+                    .active_markers
+                    .values()
+                    .any(|marker| marker.label == label);
+                if duplicate {
+                    println!("Warning: Marker with label '{}' is already active", &label);
+                }
+
                 self.active_markers.insert(
                     ptr,
                     ActiveMarker {
@@ -1464,9 +1474,16 @@ impl Cpu {
                         "\"{}\": {} RV32IM cycles, {} virtual cycles",
                         mark.label, real, virt
                     );
+                } else {
+                    println!(
+                        "Warning: Attempt to end a marker (ptr: 0x{:x}) that was never started",
+                        ptr
+                    );
                 }
             }
-            _ => { /* ignore other opcodes */ }
+            _ => {
+                panic!("Unexpected event: event must match either start or end marker.")
+            }
         }
         Ok(())
     }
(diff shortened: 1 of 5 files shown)


examples/fibonacci/guest/src/lib.rs
```diff
@@ -1,15 +1,19 @@
 #![cfg_attr(feature = "guest", no_std)]
+use jolt::{end_cycle_tracking, start_cycle_tracking};
 
 #[jolt::provable]
 fn fib(n: u32) -> u128 {
     let mut a: u128 = 0;
     let mut b: u128 = 1;
     let mut sum: u128;
+
+    start_cycle_tracking("fib_loop"); // Use `start_cycle_tracking("{name}")` to start a cycle span
```
0xAndoroid, line 10: Would be nice to have a proc macro that would attribute a function or a block, 
```rust
#[cycle_track(label="optional_label")]
fn calculation() {...}
```
@moodlezoup what do you think?

0xAndoroid (commented):
A few suggestions

examples/fibonacci/guest/src/lib.rs
moodlezoup: good idea, it would be nice to closely match the `tracing` API. Could you write up a Linear issue for it?

tracer/src/emulator/cpu.rs
```diff
@@ -324,6 +344,11 @@ impl Cpu {
             .ok()
             .unwrap();
         instr.trace(self);
+
+        // check if current instruction is real or not for cycle profiling
+        if instr.is_real() {
+            self.executed_instrs = self.executed_instrs.wrapping_add(1);
```
moodlezoup, line 350: nit:
```suggestion
            self.executed_instrs += 1;
```

tracer/src/emulator/cpu.rs
```diff
@@ -1391,6 +1441,49 @@ impl Cpu {
...
+    /// Read a NUL-terminated guest string from memory.
+    fn read_c_string(&mut self, mut addr: u32) -> Result<String, Trap> {
+        let mut bytes = Vec::new();
+        loop {
+            let (b, _) = self.mmu.load(addr.into())?;
+            if b == 0 {
+                break;
+            }
+            bytes.push(b);
+            addr = addr.wrapping_add(1);
```
moodlezoup, line 1483: nit:
```suggestion
            addr += 1;
```

tracer/src/emulator/cpu.rs
```diff
@@ -1391,6 +1441,49 @@ impl Cpu {
...
+                if let Some(mark) = self.active_markers.remove(&ptr) {
+                    let real = self.executed_instrs - mark.start_instrs;
+                    let virt = self.trace.len() - mark.start_trace_len;
+                    println!(
+                        "\"{}\": {} RV32IM cycles, {} virtual cycles",
+                        mark.label, real, virt
+                    );
+                }
+            }
+            _ => { /* ignore other opcodes */ }
```
moodlezoup, line 1469: let's panic here with "unexpected event"

tracer/src/emulator/cpu.rs
```diff
@@ -1391,6 +1441,49 @@ impl Cpu {
     pub fn get_mut_terminal(&mut self) -> &mut Box<dyn Terminal> {
         self.mmu.get_mut_uart().get_mut_terminal()
     }
+
+    fn handle_jolt_cycle_marker(&mut self, ptr: u32, event: u32) -> Result<(), Trap> {
+        match event {
+            JOLT_CYCLE_MARKER_START => {
+                let label = self.read_c_string(ptr)?; // guest NUL-string
+                self.active_markers.insert(
```
moodlezoup, line 1449: let's emit a warning if there's already an active marker with the same label

moodlezoup (commented):
good stuff! just a few nits

moodlezoup (approved):

PR #672, opened 2025-05-08
fix: guest programs using stdlib

sagar-a16z:
Also fixes `./examples/run_benchmarks.sh` to work on MacOS as long as gtime is installed.


sagar-a16z:
@jprider63 this should fix the issue you're seeing

moodlezoup (approved):

jprider63:
Thanks @sagar-a16z!

PR #675, opened 2025-05-11
fix: use paste crate to resolve enum_dispatch inner scope issue

zouguangxian:
The error "cannot find value `inner` in this scope" occurs because enum_dispatch
creates an inner identifier during macro expansion, but it's not properly scoped.
The paste crate helps by creating hygienic identifiers during macro expansion,
ensuring the inner value is available in the correct scope.

This is a more robust solution than relying on trait definition order, as it
properly handles macro hygiene and identifier scoping.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2004,6 +2004,7 @@ dependencies = [
  "num-integer",
  "num-traits",
  "once_cell",
+ "paste",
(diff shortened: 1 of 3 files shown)


moodlezoup (approved):
thank you!

PR #676, opened 2025-05-12
Feat: Optimize SLLI, SRLI, SRAI

0xAndoroid:
Linear issue 104

Optimizes SLLI, SRLI, SRAI, by replacing this instruction by a virtual instruction with precalculated `imm` field.

Important for review, product constraint now accesses left and right instruction inputs instead of `rs1` and `rs2`. This is needed for MULI instruction.

TODOs (not in this PR):
- Support for 64 bits inside of `VirtualInstructionSequence` implementations
- We should consider removing the hardcoded virtual instruction enumeration (such as in load and store instructions).

### DIFF
--- a/tracer/src/instruction/format/format_virtual_i.rs
+++ b/tracer/src/instruction/format/format_virtual_i.rs
@@ -1,0 +1,73 @@
+use crate::emulator::cpu::Cpu;
+use common::constants::REGISTER_COUNT;
+use rand::rngs::StdRng;
+use rand::RngCore;
+use serde::{Deserialize, Serialize};
+use std::fmt::Debug;
+
+use super::{
+    normalize_register_value, InstructionFormat, InstructionRegisterState, NormalizedOperands,
+};
+
+#[derive(Default, Debug, Clone, Copy, PartialEq, Serialize, Deserialize)]
+pub struct FormatVirtualI {
+    pub rd: usize,
+    pub rs1: usize,
+    pub imm: u64,
+}
+
+#[derive(Default, Debug, Copy, Clone, Serialize, Deserialize)]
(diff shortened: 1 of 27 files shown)


tracer/src/instruction/format/format_virtual_i.rs
```diff
@@ -0,0 +1,73 @@
...
+use rand::RngCore;
+use serde::{Deserialize, Serialize};
+use std::fmt::Debug;
+
+use super::{
+    normalize_register_value, InstructionFormat, InstructionRegisterState, NormalizedOperands,
+};
+
+#[derive(Default, Debug, Clone, Copy, PartialEq, Serialize, Deserialize)]
+pub struct FormatVirtualI {
```
moodlezoup, line 13: If this is basically the exact same as `FormatI` let's just use that

moodlezoup (commented):
Nice work! Only a couple of nits.

> We should consider removing the hardcoded virtual instruction enumeration

Agreed. I think we don't even really need them enumerated –– the only information we need is (a) whether the instruction is part of a virtual sequence, and (2) if the instruction is the *last* one in a virtual sequence. 

So I guess instead of `Option<usize>` we can make use an `Option<bool>` to simplify this.

tracer/src/instruction/format/format_virtual_i.rs
0xAndoroid: I've replaced FormatVirtualI with FormatI in MULI instruction, however, there's still a need for a new format - `FormatVirtualRightShiftI` since `random` function has to have specific `imm`, same as in `FormatVirtualRightShift` (renamed to `FormatVirtualRightShiftR`)

moodlezoup (approved):
great work!

PR #677, opened 2025-05-12
Make the `common` crate `no_std`-compliant

amaloz:
This is Step 1 in a long line of steps towards making `jolt` `no_std` compliant.


moodlezoup (approved):
thank you @amaloz !

PR #680, opened 2025-05-14
Make `tracer` `no_std` compatible

amaloz:
This requires a slight change to the API that is exposed to `jolt-core`: The `trace` function no longer takes a `PathBuf` (since that's not available in a `no_std` environment) and instead takes the elf contents as a `Vec<u8>`. This requires a slight change to the `jolt-core` `host` module to make that API change compatible.


moodlezoup (approved):
thank you!

PR #681, opened 2025-05-15
feat: virtual sequence fuzzer

markosg04:
- Adds a fuzzer for testing the behavior of 'virtual sequences' to that of their 'real' counterpart in `tracer`
  - The fuzzer creates two cpu objects, one of which runs a virtual sequence and the other runs the real version of the sequence
  - The fuzzer then compares the real register states for both, with the invariant that they should be the same
- There is a known issue in which `DIV, DIVU, MULH, MULHSU, REM, REMU, SB, SH` may fail at random. This is due to some problems with the way that register data is handled throughout the `tracer` crate and many casting edge cases. There is a writeup in linear issue 121 that details the problem further.


tracer/src/instruction/test.rs
```diff
@@ -1,89 +1,246 @@
...
+    for _ in 0..10000 {
+        // Seed operands accordingly
+        let rs1_idx = rng.next_u64() % REG_COUNT as u64;
+        let rs2_or_rd_idx = rng.next_u64() % REG_COUNT as u64;
+        let addr_tag = rng.next_u64();
+
+        let (rs1_val, imm, _dest) = if is_memory {
+            generate_aligned_address(&mut rng)
+        } else {
+            (rng.next_u32() as u64, (rng.next_u32() & 0x0fff) as i64, 0)
```
0xAndoroid, line 141: Some instructions are expecting register values/imm in a certain format. For example, `VirtualSRL` expects second register to be in a certain format (spceifically `VirtualSRL` expects bitmask).
This should not be a problem currently, as all of the instructions that have virtual sequence do not have such issues, however, once we start working on pseudo precompiles, this might become an issue.
Take a look at `InstructionFormat::random()` and `InstructionRegisterState::random()` for generation of parameters.

tracer/src/instruction/test.rs
```diff
@@ -1,89 +1,246 @@
...
-//             }
-//         }
-//     }
-// }
+    macro_rules! suite {
+        (mem $($mem_instr:ty),* ; non $($non_instr:ty),* $(,)?) => {
+            $(
+                paste::paste! {
+                    #[test]
+                    fn [<test_ $mem_instr:snake _mem>]() {
```
0xAndoroid, line 223:
```suggestion
                    fn [<test_ $mem_instr:lower _mem>]() {
```
So that test names are `test_sb_mem`, and not `test_s_b_mem`.

tracer/src/instruction/test.rs
```diff
@@ -1,89 +1,246 @@
...
+                        virtual_sequence_fuzz::<$mem_instr>(true);
+                    }
+                }
+            )*
+            $(
+                paste::paste! {
+                    #[test]
+                    fn [<test_ $non_instr:snake _nonmem>]() {
```
0xAndoroid, line 231:
```suggestion
                    fn [<test_ $non_instr:lower _nonmem>]() {
```

tracer/src/instruction/test.rs
```diff
@@ -1,89 +1,246 @@
...
+use super::{RISCVCycle, RV32IMCycle};
 use crate::emulator::cpu::{Cpu, Xlen};
+use crate::emulator::terminal::DummyTerminal;
+use rand::{rngs::StdRng, RngCore, SeedableRng};
+
+const DRAM_BASE: u64 = 0x8000_0000;
+const DRAM_SIZE: u64 = 1024;
+
+// Implements a `from_operands` constructor for store instructions
+#[cfg(test)]
```
0xAndoroid, line 26: Entire test.rs file is behind `#[cfg(test)]` and these annotations can be removed in this file.
```suggestion

```

tracer/src/instruction/test.rs
```diff
@@ -1,89 +1,246 @@
...
+macro_rules! impl_from_operands_s {
+    ($($ty:ident),* $(,)?) => {
+        $(
+            #[cfg(test)]
+            impl $ty {
+                pub fn from_operands(rs1: usize, rs2: usize, imm: i64, address: u64) -> Self {
```
0xAndoroid, line 32: If you use `InstructionFormat::random()` and `InstructionRegisterState::random()`, as well as `normalize()` on instruction, these from_operands implementations can be removed.

tracer/src/instruction/test.rs
```diff
@@ -1,89 +1,246 @@
...
+            (rng.next_u32() as u64, (rng.next_u32() & 0x0fff) as i64, 0)
+        };
+
+        // Construct instruction
+        // Since we need control over operands we use a custom `from_operands` method
+        let instruction: I = if is_memory {
+            let ty = TypeId::of::<I>();
+            if ty == TypeId::of::<SB>() || ty == TypeId::of::<SH>() {
+                // Store → FormatS downcast
+                unsafe {
```
0xAndoroid, line 150: Maybe I'm wrong, but this downcasting should be removed as well if you use `random()` on the instructoin and state.

tracer/src/instruction/test.rs
```diff
@@ -1,89 +1,246 @@
...
+    RV32IMCycle: From<RISCVCycle<I>>,
+    <I as RISCVInstruction>::RAMAccess: Default,
+{
+    let mut rng = test_rng();
+    const REG_COUNT: usize = 32;
+
+    // Init CPUs
+    let mut cpu_exec = Cpu::new(Box::new(DummyTerminal::new()));
+    let mut cpu_trace = Cpu::new(Box::new(DummyTerminal::new()));
+    cpu_exec.update_xlen(Xlen::Bit32);
```
0xAndoroid, line 127: nit: put xlen as a parameter of a function for easy change once we support bit64

tracer/src/instruction/test.rs
```diff
@@ -1,89 +1,246 @@
...
+fn virtual_sequence_fuzz<
+    I: RISCVInstruction + VirtualInstructionSequence + RISCVTrace + Copy + 'static,
+>(
+    is_memory: bool,
+) where
+    RV32IMCycle: From<RISCVCycle<I>>,
+    <I as RISCVInstruction>::RAMAccess: Default,
+{
+    let mut rng = test_rng();
+    const REG_COUNT: usize = 32;
```
0xAndoroid, line 122: Should be able to make dependent on the xlen parameter.

0xAndoroid (changes requested):
Can we use `InstructionFormat::random()` and `InstructionRegisterState::random()` for generation of parameters, and remove these `from_operand` implementations?
And a couple other nits.

tracer/src/instruction/test.rs
```diff
@@ -1,89 +1,246 @@
...
+    };
+}
+
+impl_from_operands_s!(SB, SH);
+impl_from_operands_load!(LB, LBU, LH, LHU);
+impl_from_operands_r!(DIV, DIVU, REM, REMU);
+
+fn test_rng() -> StdRng {
+    // StdRng::from_seed([10u8; 32]) // deterministic
+    StdRng::from_entropy()
```
moodlezoup, line 88: prefer the seeded version here, so we can reproduce test failures

tracer/src/instruction/test.rs
moodlezoup: actually RV64 still has 32 registers

tracer/src/instruction/test.rs
```diff
@@ -1,89 +1,246 @@
...
+
+        // Execute the (random) seeded instructions
+        let mut ram_access = <I as RISCVInstruction>::RAMAccess::default();
+        instruction.execute(&mut cpu_exec, &mut ram_access);
+        instruction.trace(&mut cpu_trace);
+
+        // The first 32 registers should have the same state after execution
+        for i in 0..31 {
+            let real_reg = cpu_exec.x[i] as i32; // we cast due to sign extensions
+            let virtual_reg = cpu_trace.x[i] as i32;
```
moodlezoup, line 208: nit: let's call these `expected_value` and `virtual_sequence_value` or something to avoid confusion with the virtual registers `cpu.x[32..]`

PR #682, opened 2025-05-16
[JOLT-117] Feat/batched sumcheck

moodlezoup:
Adds an abstraction for running multiple sumchecks in parallel (aka batching). 
Currently we does this manually in a few places, but it quickly gets messy and verbose. 
That said, there are cases where we can gain some efficiency by a more bespoke batching implementation, e.g. when two sumcheck instances have some polynomials in common. In such a case, we would use a single `BatchableSumcheckInstance` for that pair of sumchecks, and then batch the *batched* sumcheck with other instances.
See, for example, the proof-of-concept usage of the new abtraction in `shout.rs`, where we use bespoke batching for the core PIOP sumcheck and the Hamming weight sumcheck, then use the new `BatchedSumcheck` to batch that with the booleanity sumcheck.

### DIFF
--- a/jolt-core/src/poly/eq_poly.rs
+++ b/jolt-core/src/poly/eq_poly.rs
@@ -79,6 +79,10 @@ impl<F: JoltField> EqPolynomial<F> {
 
         evals
     }
+
+    pub fn len(&self) -> usize {
+        self.r.len().pow2()
+    }
 }
 
 impl<F: JoltField> EqPlusOnePolynomial<F> {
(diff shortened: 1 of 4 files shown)


moodlezoup:
@0xAndoroid @markosg04 curious to get your thoughts on this abstraction

markosg04 (approved):
The abstraction itself makes sense to me. Still developing my intuition on these polynomials x sumcheck so I don't have any particular question or optimization suggestions or nits at this point, other than what we discussed in the meeting.

PR #683, opened 2025-05-17
feat: implement gruen's optimization & more for spartan

quangvdao:
This PR implements multiple optimizations for Spartan's first sum-check:

1. We implement Gruen's optimization 
2. We also implement the optimization that in the first sum-check round, there is no need to compute the evaluation at zero, since it is zero due to R1CS relation being satisfied
3. Finally, we change the output type of `unbound_coeffs` to be a vector of chunks of sparse coefficients, instead of a unified vector of coefficients. This eliminates the coalescing at the end of `new()`, and chunking at the beginning of `first_sumcheck_round()`. Instead, we now just pass these parallel chunks from one function directly to the other. In theory, I could also keep the chunks separate for the next few rounds of `subsequent_sumcheck_round()`, but not sure if it's worth it. When we switch to streaming, this would need to be done anyway.

Altogether, we may see up to 2x speedup for Spartan's first sum-check (half coming from 1+2 and half coming from 3). And this is before the additional small value optimization, which I'm still working on.

Re point 1: this is similar to a prior PR I made about Gruen's optimization for dense interleaved poly. The difference is that, in order to integrate the optimization into that, we would need to change the sparse interleave poly + sparse grand product at the same time, which I didn't have time to do.

@jprider63 you may want to take a look. This one is more fleshed out & tested to work.

TODOs: implement Gruen's optimization for the rest of the sub-protocols. Then we can delete the old `SplitEqPolynomial` and replace it fully with what's called `GruenSplitEqPolynomial` here.

### DIFF
--- a/jolt-core/src/poly/eq_poly.rs
+++ b/jolt-core/src/poly/eq_poly.rs
@@ -26,11 +26,23 @@ impl<F: JoltField> EqPolynomial<F> {
     }
 
     #[tracing::instrument(skip_all, name = "EqPolynomial::evals")]
+    /// Computes the table of coefficients: `{eq(r, x) for all x in {0, 1}^n}`
     pub fn evals(r: &[F]) -> Vec<F> {
         match r.len() {
             0..=PARALLEL_THRESHOLD => Self::evals_serial(r, None),
             _ => Self::evals_parallel(r, None),
         }
+    }
+
+    #[tracing::instrument(skip_all, name = "EqPolynomial::evals_cached")]
+    /// Computes the table of coefficients like `evals`, but also caches the intermediate results
+    ///
+    /// In other words, computes `{eq(r[i..], x) for all x in {0, 1}^{n - i}}` and for all `i in
+    /// 0..r.len()`.
+    pub fn evals_cached(r: &[F]) -> Vec<Vec<F>> {
+        // TODO: implement parallel version (it seems more difficult to parallelize all the
+        // intermediate results)
+        Self::evals_serial_cached(r, None)
     }
 
     /// Computes the table of coefficients:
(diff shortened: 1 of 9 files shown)


quangvdao:
Why do I need to fix unrelated errors lol

quangvdao:
@moodlezoup you should take a look at the `Arc` change I made. Testing & benchmarking works well. I wonder if you see the same opportunity for this kind of optimization elsewhere in Jolt (I briefly considered this but couldn't find another usage).

jolt-core/src/poly/spartan_interleaved_poly.rs
```diff
@@ -1,30 +1,27 @@
...
+use std::sync::Arc;
 
 #[derive(Default, Debug, Clone)]
 pub struct SpartanInterleavedPolynomial<F: JoltField> {
-    /// A sparse vector representing the (interleaved) coefficients in the Az, Bz, Cz
+    /// Shards of sparse vectors representing the (interleaved) coefficients in the Az, Bz, Cz
     /// polynomials used in the first Spartan sumcheck. Before the polynomial is bound
     /// the first time, all the coefficients can be represented by `i128`s.
-    pub(crate) unbound_coeffs: Vec<SparseCoefficient<i128>>,
+    pub(crate) unbound_coeffs_shards: Vec<Arc<Vec<SparseCoefficient<i128>>>>,
```
moodlezoup, line 24: I might be missing something but I don't we actually ned the `Arc` here?

jolt-core/src/poly/split_eq_poly.rs
```diff
@@ -13,6 +36,113 @@ pub struct SplitEqPolynomial<F> {
...
+impl<F: JoltField> GruenSplitEqPolynomial<F> {
+    #[tracing::instrument(skip_all, name = "GruenSplitEqPolynomial::new")]
+    pub fn new(w: &[F]) -> Self {
+        let m = w.len() / 2;
+        //   w = [w_out, w_in, w_last]
+        //        ↑   ↑    ↑
+        //        |   |    |
+        //        |   |    last element
+        //        |   second half of remaining elements (for E_in)
+        //        first half of remaining elements (for E_out)
```
moodlezoup, line 48: at least on Github's UI, the spacing of these arrows looks a little off

jolt-core/src/poly/split_eq_poly.rs
```diff
@@ -13,6 +36,113 @@ pub struct SplitEqPolynomial<F> {
...
+    pub fn E_out_current_len(&self) -> usize {
+        self.E_out_vec.last().unwrap().len()
+    }
+
+    /// Return the last vector from `E1` as a slice
+    pub fn E_in_current(&self) -> &[F] {
+        self.E_in_vec.last().unwrap()
+    }
+
+    pub fn to_E1_old(&self) -> Vec<F> {
```
moodlezoup, line 85: is this only used in tests? if so let's put it behind `#[cfg(test}]`

jolt-core/src/poly/eq_poly.rs
```diff
@@ -52,8 +64,58 @@ impl<F: JoltField> EqPolynomial<F> {
...
+            .map(|i| vec![scaling_factor.unwrap_or(F::one()); 1 << i])
+            .collect();
+        let mut size = 1;
+        for j in 0..r.len() {
+            size *= 2;
+            for i in (0..size).rev().step_by(2) {
+                let scalar = evals[j][i / 2];
+                evals[j + 1][i] = scalar * r[j];
+                evals[j + 1][i - 1] = scalar - evals[j + 1][i];
+            }
```
moodlezoup, line 84: regarding the TODO above about parallelizing; we can keep the outer loop serial and just parallelize this inner loop, right?

moodlezoup (commented):
nice work, I know this was gnarly to debug. Just a couple of minor comments

jolt-core/src/poly/spartan_interleaved_poly.rs
quangvdao: You're right. In the small value precomputation process, this seems to be necessary (I'm collecting a struct that has the vector of coefficients inside, along with some other accumulator data).

jolt-core/src/poly/eq_poly.rs
quangvdao: Yes, that would work. Unclear when it will lead to a speedup though - will refine the note

moodlezoup (approved):
thanks @quangvdao!

PR #690, opened 2025-05-24
feat: small value optimization for Spartan

quangvdao:
This PR overhauls the Spartan first sum-check to use small-value optimization (SVO). This will be described in detail in an upcoming paper by Bagad-Dao-Domb-Thaler, itself a merge of two prior works [BDT24] and [DT24].

We currently settle on 3 rounds of SVO as that leads to the most performance improvement (both estimated and demonstrated in benchmarks). When we do streaming, we may want to do 4 rounds of SVO instead, as the extra SVO round is slightly less expensive than one streaming round.

Some miscellaneous changes include:
- Update parts of the book to reflect the changes
- Update `bench.rs` which currently gives an error (`decode()` needs to be _after_ `trace()`, not before)
- Update README to have the correct command for profiling (`profile` instead of `trace`)

There was a prior performance degradation due to no pre-allocation of the `ab_unbound_coeffs` chunks. This has been fixed now (thanks @moodlezoup). We consistently see around a 1.3x speedup for Spartan first sum-check, with more when proving gets memory-bound (i.e. the new method uses 50% or fewer memory)

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ Jolt uses [tracing_chrome](https://crates.io/crates/tracing-chrome) for performa
 
 To generate a trace, run:
 
-```cargo run --release -p jolt-core trace --name sha3 --format chrome --pcs hyper-kzg```
+```cargo run --release -p jolt-core profile --name sha3 --format chrome --pcs hyper-kzg```
 
 Where `--name` can be `sha2`, `sha3`, `sha2-chain`, or `fibonacci`. The corresponding guest programs can be found in the [`examples`](./examples/) directory. The benchmark inputs are provided in [`bench.rs`](./jolt-core/src/benches/bench.rs).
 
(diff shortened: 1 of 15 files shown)


quangvdao:
@moodlezoup if you could profile examples like `sha2` or `sha3` yourself and confirm the lag, that would be great!

I don't really know how to fix it.

quangvdao:
@moodlezoup Looks ready to merge for me. Do give more comments (and clippy is buggy right now)

moodlezoup:
> Perhaps we could bring down this overhead via const-ifying the computation. We define macros that generate the R1CS constraints as a const, and then `evaluate_row` will adapt at compile-time based on these const. Might be better for formal verification too, as the constraints are more explicit than now (currently we require custom extraction code, in the future we can just read off the macro-generated constraints)

Interesting idea. Relatedly, Justin and I have talked about generating the (unbound) Az, Bz, Cz coefficients directly from the trace, avoiding the matrix-vector multiplication indirection entirely. This would likely bring down the overhead even further, but may not be as easily extractable for FV purposes.

quangvdao:
That would be great. For FV purposes it would just mean the "base" level for constraints will be the Jolt trace, not the flattened polys.

moodlezoup:
Amazing work as always, thanks @quangvdao

PR #692, opened 2025-05-25
feat: Dory

markosg04:
#### Features
- adds a complete generic and self-contained implementation of Dory PCS as an independent crate. The README of the crate details further what the exact API and features are.
- Is already optimized, performance matches the claims from the paper (~3 seconds for k=20 for eval proofs, ~50ms for verify).

#### Next steps
(1) I am (obviously) still not confident in soundness, although I've made a lot of progress in checking the implementation over the last week -- we are getting there. I left some TODOs for the parts that I am most concerned about. (2) Batching multiple evaluations is still left as TODOs, but there are clear next steps (hopefully easy?) for how to achieve this. The code should be able to support it with only minor changes. (3) We need more unit tests. 

I wanted to open the PR at this point so that I can start integration while I address the 3 next steps, and also to start getting some feedback. A future PR would address the above (3) and also complete the integration with our PCS traits + practical performance numbers.


markosg04:
Closing this PR in favor of a new PR which only brings the integration into the repo. All comments in this PR have been addressed or otherwise noted as appropriate.

The Dory-implementation itself will proceed with feedback & testing via the SXT repo. See the following branch: https://github.com/spaceandtimefdn/sxt-dory/tree/feat/impl. At this point, the interface should not change all that much w.r.t the integration in the Jolt repo.

PR #694, opened 2025-05-27
fix: improve emulator memory safety and error reporting

sagar-a16z:
- Improved error reporting for stack overflows and stack sizing
- Fixed bugs in Arithmetic Shift instructions  `SRA` and SRAW`


moodlezoup (approved):
thank you, this is much nicer

PR #695, opened 2025-05-29
[JOLT-123] Feat/sha256 precompile

0xAndoroid:
Adds a sha256 pseudo precompile - basically reduces number of instructions by utilizing two techniques:
- uses virtual registers so that loads/stores are not necessary all the time
- adds additional `VirtualROTRI` instruction that performs rotation to the right in one instruction instead of three, and sha uses quite a lot of these instructions.

## Benchmarking
The results show a performance gain of around **4x**, bringing a speed of hashing on a M3 Macbook Pro to 283 sha2/s.
Apart from that, thanks to the reduced trace, memory consumption has decreased also by the same factor.

These results are exceeding initial expectations slightly (initial expectations were estimating around 2x for loads/stores removal and another 1.5x for rotri).

Note that this PR doesn't introduce full sha256 test since currently virtual sequence instruction tests in the tracer are disabled.

### DIFF
--- a/tracer/src/instruction/inline_sha256/sha256init.rs
+++ b/tracer/src/instruction/inline_sha256/sha256init.rs
@@ -37,7 +37,7 @@ impl SHA256INIT {
         for (i, &word) in result.iter().enumerate() {
             cpu.mmu
                 .store_word(
-                    cpu.x[self.operands.rs1].wrapping_add(((i + 16) * 4) as i64) as u64,
+                    cpu.x[self.operands.rs2].wrapping_add((i * 4) as i64) as u64,
                     word,
                 )
                 .expect("SHA256INIT: Failed to store result");


jolt-sdk/src/sha256.rs
```diff
@@ -0,0 +1,415 @@
...
+    #[inline(always)]
+    pub fn new() -> Self {
+        Self {
+            state: unsafe { core::mem::MaybeUninit::uninit().assume_init() },
+            buffer: unsafe { core::mem::MaybeUninit::uninit().assume_init() },
+            buffer_len: 0,
+            total_len: 0,
+            initial: true,
+        }
+    }
```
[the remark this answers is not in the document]
0xAndoroid: This is intentional. To use initial state we use `SHA256COMPRESSI` instruction, that's around 300 instructions cheaper by utilizing precomputation on imm values.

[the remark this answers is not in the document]
0xAndoroid: For the initial round the `SHA256COMPRESSIONI` should be used, it reduces another couple hundred cycles.

tracer/src/instruction/mod.rs
```diff
@@ -52,6 +52,8 @@ use sw::SW;
 use xor::XOR;
 use xori::XORI;
 
+use precompile_sha256::sha256compress::SHA256COMPRESS;
```
markosg04: i don't feel strongly about this but just making a note of all the naming: 'inline-precompile' vs 'precompile' for future

tracer/src/instruction/mod.rs
```diff
@@ -544,9 +561,23 @@ impl RV32IMInstruction {
...
+            0b0001011 => {
+                // Custom-0 opcode: SHA256 compression instructions
+                let funct3 = (instr >> 12) & 0x7;
+                let funct7 = (instr >> 25) & 0x7f;
+                if funct7 == 0x00 {
+                    match funct3 {
+                        0x0 => Ok(SHA256COMPRESS::new(instr, address, true).into()),
+                        0x1 => Ok(SHA256COMPRESSI::new(instr, address, true).into()),
+                        _ => Err("Unknown funct3 for custom SHA256 instruction"),
+                    }
```
markosg04, line 584: so we are choosing to use '0' as SHA2 identifier, and then '0' in funct3 as the standard function and '1' as the immediate version? Maybe we could just add a comment to clarify how we are reserving the namings here (unless it's somewhere else that i haven't seen yet). Or, better yet, we might want to (now or later?) update the Jolt book with information like this -- it would also be a nice opportunity to put the write-up actually as well in the docs now that im thinking. @moodlezoup @0xAndoroid thoughts? Like a precompiles section (we will need it sooner or later)

tracer/src/instruction/precompile_sha256/sha256compress.rs
```diff
@@ -0,0 +1,88 @@
+use common::constants::virtual_register_index;
+use serde::{Deserialize, Serialize};
+
```
markosg04: 'I' here is for "Initial", not "Immediate"? Should we clarify this? Or we care about reserving the 'i' at end for immediate? (If we do care, we could use "in' for 'initial" as convention)?

tracer/src/instruction/precompile_sha256/mod.rs
```diff
@@ -0,0 +1,512 @@
...
+use crate::instruction::format::format_s::FormatS;
+use crate::instruction::format::format_virtual_right_shift_i::FormatVirtualRightShiftI;
+use crate::instruction::lw::LW;
+use crate::instruction::srli::SRLI;
+use crate::instruction::sw::SW;
+use crate::instruction::virtual_rotri::VirtualROTRI;
+use crate::instruction::xor::XOR;
+use crate::instruction::xori::XORI;
+use crate::instruction::RV32IMInstruction;
+
```
markosg04: nit: perhaps add a link to some page that describes the SHA2 function. It's prolly consistent across most descriptions I would imagine but could just add the wiki so reader can reference for the notation and stuff

markosg04: or later we could have a succinct description in jolt book

0xAndoroid: I consider the function in the mod.rs file (after builder implementation) to be the reference implementation as it's extremely clear. What do you think?

tracer/src/instruction/precompile_sha256/sha256compress.rs
0xAndoroid: Will rename SHA256C and SHA256CINIT

tracer/src/instruction/precompile_sha256/mod.rs
```diff
@@ -0,0 +1,512 @@
...
+/// Builds assembly sequence for SHA256 compression
+/// Expects input words to be in RAM at location rs1..rs1+16
+/// Expects A..H to be in RAM at location rs2..rs2+8
+/// Output will be written to rs2..rs2+8
+struct Sha256SequenceBuilder {
+    address: u64,
+    sequence: Vec<RV32IMInstruction>,
+    /// Round id
+    rid: i32,
```
markosg04: nit: I would prefer `round_id` or something a little more verbose than `rid` because these short variable that start with r (ex. rs1 rs2 rd) are usually referring to registers, and this is not

tracer/src/instruction/precompile_sha256/mod.rs
```diff
@@ -0,0 +1,512 @@
...
+        // Round 2: A,B,C,E,F,G are available
+        // Round 3+: All values are in registers
+        if self.initial
+            && (self.rid == 0
+                || (self.rid == 1 && !['A', 'E'].contains(&shift))
+                || (self.rid == 2 && !['A', 'B', 'E', 'F'].contains(&shift))
+                || (self.rid == 3 && !['A', 'B', 'C', 'E', 'F', 'G'].contains(&shift)))
+        {
+            return Imm(
+                BLOCK[((shift as usize - 'A' as usize) as i32 - self.rid).rem_euclid(8) as usize]
```
markosg04: I thought this line was a little hard to unpack what it is doing

tracer/src/instruction/precompile_sha256/mod.rs
markosg04: I think is fine, we will prolly want a page in jolt book dedicated to stuff like this anyways

tracer/src/instruction/precompile_sha256/mod.rs
```diff
@@ -0,0 +1,512 @@
...
+enum Value {
+    Imm(u64),
+    Reg(usize),
+}
+
+pub fn execute_sha256_compression_initial(input: [u32; 16]) -> [u32; 8] {
+    execute_sha256_compression(BLOCK.map(|x| x as u32), input)
+}
+
+pub fn execute_sha256_compression(initial_state: [u32; 8], input: [u32; 16]) -> [u32; 8] {
```
markosg04: maybe consider annotating the stuff in this function some more, especially if we consider this the main 'reference' implementation.

jolt-sdk/src/sha256.rs
```diff
@@ -0,0 +1,403 @@
+//! SHA-256 hash function implementation optimized for Jolt zkVM.
+//!
+//! This module provides an API similar to the `sha2` crate.
+
+use core::mem::MaybeUninit;
+
+/// SHA-256 hasher state.
+pub struct Sha256 {
+    /// Current hash state (8 x 32-bit words)
+    ///
```
markosg04, line 10: is it worth mentioning here as a 1-liner that we avoid the initialization just as a cycle optimization?

jolt-sdk/src/sha256.rs
```diff
@@ -0,0 +1,403 @@
...
+                let buf = unsafe { self.buffer_as_u32_mut() };
+                Sha256::swap_bytes(buf);
+            }
+
+            unsafe {
+                self.sha256_compress();
+            }
+
+            // Second block: all zeros except length at the end
+            self.buffer[0].write(0);
```
markosg04, line 208: maybe mention once in this file that you unrolled all these loops, just so people are not confused that this is intentional and also because it is cool

tracer/src/instruction/mod.rs
0xAndoroid: I renamed module as `inline_sha256` to avoid long names. Imo, it makes sense for these types of instructions.

tracer/src/instruction/mod.rs
0xAndoroid: So, my idea was to have `funct7` responsible for operation type (SHA256 in this case), and `funct3` be responsible for all the necessary instructions for such operation.
We can fit 128 different operations this way, and each can have up to 8 instructions.
Obviously, we're wasting space, but I think this should be good. I added comment for clarity.

tracer/src/instruction/precompile_sha256/mod.rs
0xAndoroid: Added comment

tracer/src/instruction/precompile_sha256/mod.rs
0xAndoroid: Not quite sure what to annotate

tracer/src/instruction/precompile_sha256/sha256compress.rs
0xAndoroid: Actually decided to rename `SHA256` and `SHA256INIT`, as C seemed out of place

tracer/src/instruction/mod.rs
moodlezoup: I'm good with this approach. Agree that we should add a section to the docs on this at some point.

tracer/src/instruction/inline_sha256/mod.rs
```diff
@@ -0,0 +1,516 @@
...
+                rs1,
+                imm: 10,
+            },
+            virtual_sequence_remaining: Some(0),
+        };
+        self.sequence.push(srli.into());
+        self.xor(Reg(rd), Reg(ss), rd);
+    }
+
+    fn lw(&mut self, rs1: usize, imm: i64, rd: usize) {
```
moodlezoup: to avoid confusion with the actual `imm`
```suggestion
    fn lw(&mut self, rs1: usize, offset: i64, rd: usize) {
```

tracer/src/instruction/inline_sha256/mod.rs
```diff
@@ -0,0 +1,516 @@
...
+                rd,
+                rs1,
+                imm: imm * 4,
+            },
+            virtual_sequence_remaining: Some(0),
+        };
+        self.sequence.push(lw.into());
+    }
+
+    fn sw(&mut self, rs1: usize, rs2: usize, imm: i64) {
```
moodlezoup: same here
```suggestion
    fn sw(&mut self, rs1: usize, rs2: usize, offset: i64) {
```

moodlezoup (approved):
good stuff, it's awesome to see this idea coming to fruition

tracer/src/instruction/inline_sha256/mod.rs
0xAndoroid: Fixed

PR #697, opened 2025-06-01
Added trace snapshotting for streaming

### LINKED ISSUE #696: Add Checkpointing and Replay support for trace generation
Corresponding PR #697 

# Description - Current Implementation

To improve memory efficiency for streaming, implement:

1. Checkpointing during trace generation, and
2. Replaying a trace from a checkpoint.

### Generating Checkpoints
Modified the trace function to support checkpointing emulator state every N steps (configurable via a checkpoint_interval parameter). A _step_ here corresponds to an emulator `tick`.

```rust
pub fn trace(
    elf_contents: Vec<u8>,
    inputs: &[u8],
    memory_config: &MemoryConfig,
    checkpoint_interval: Option<usize>,
) -> (Vec<RV32IMCycle>, JoltDevice, Option<Vec<Checkpoint>>) {
    let mut emulator = setup_emulator(...);
    // checkpoints are emulator states for now
    let checkpoints = run_and_get_checkpoints(&mut emulator, checkpoint_interval);
    let execution_trace = std::mem::take(&mut emulator.get_mut_cpu().trace);
    let device = std::mem::take(&mut emulator.get_mut_cpu().get_mut_mmu().jolt_device);
    return (execution_trace, device, checkpoints);
}
```
### Saving Checkpoints
The emulator saves checkpoints at specified intervals using `run_and_get_checkpoints`:

```rust
pub fn run_and_get_checkpoints(
    emulator: &mut Emulator,
    checkpoint_interval: Option<usize>,
) -> Option<Vec<Checkpoint>> {
   ...
    match checkpoint_interval {
        Some(interval) => {
            let mut count = 0;
            loop {
                if count % interval == 0 {
                    checkpoints.push(emulator.save_state());
                }
                count += 1;
                if step_emulator(emulator, &mut prev_pc, true) {
                    break;
                }
            }
        }
        ...
```

### Key aspects of checkpoints:
- Checkpoints are full snapshots of emulator state and include complete state of CPU, memory, devices, and more. 
- Hence, checkpoints are memory intensive. 
- Profiling on a few benchmarks shows that the processor memory makes up more than 90% of the emulator state. 
- Compressing memory contents and/or using other clever encoding schemes can increase the memory efficiency of checkpoints.

## 2. Replaying a Trace from a Checkpoint

Introduced an iterator to support trace regeneration from a given checkpoint, without restarting from the beginning.

### Lazy Trace Iterator
The `LazyTraceIterator` provides efficient trace generation:
- Implements an iterator that generates instruction traces on-demand
- `next` returns a single trace element, even though an `emulator.tick()` call can generate more than one element. The extra elements are buffered internally. 
- Buffers traces from each emulator tick for FIFO ordering

````rust
pub struct LazyTraceIterator {
    emulator: Emulator,      // Reconstructed from checkpoint
    prev_pc: u64,           // For termination detection
    current_traces: Vec<RV32IMCycle>, // Buffer for multi-trace instructions
}
````

### Trace From a Checkpoint

   ```rust
pub fn trace_from_checkpoint(checkpoint: &Checkpoint) -> LazyTraceIterator {
    LazyTraceIterator {
        emulator: Emulator::from_state(checkpoint),
        prev_pc: 0,
        current_traces: Vec::new(),
    }
}
```

## TODOs (?)
The below are some tentative TODOs that can be taken care in this issue.
 
- [ ] Turn off trace collection during checkpoint generation.
- [ ] Implement checkpoint compression.
- [ ] Check if returning an empty trace cause issues downstream.
- [ ] Add usage example

## Additional Changes
The emulator execution loop as been slightly modified while addressing this issue:
- The existing emulator execution loop would execute the last instruction twice before detecting the end of program.
- This has been modified so the last instruction is only executed once.

## Further Discussion: Trade-off between saving checkpoints vs traces
- Each checkpoint takes substantial amount of memory, primarily dictated by the amount of allocated processor memory by the guest program. This will constraint the number of checkpoints that one might need to store at the same time. 
- Compressing checkpoints will increase computation time.
- Delta encoding checkpoints seem another alternative for compressing, but will additionally enforce a compute penalty.

zutshi:
Closes #696.


tracer/src/lib.rs
```diff
@@ -38,26 +281,112 @@ pub fn trace(
...
+/// * `prev_pc` - Previous program counter value, used for termination detection
+/// * `current_traces` - Buffer of trace entries from the most recent emulator tick
+/// * `length` - Length of the iterator. This length is interpreted as the number of
+///              RV32IM Cycles.
+#[derive(Clone)]
+pub struct LazyTraceIterator {
+    emulator_state: EmulatorState,
+    prev_pc: u64,
+    current_traces: Vec<RV32IMCycle>,
+    length: Option<usize>, // number of cycles to execute before stopping iteration
```
jprider63: Is length still needed? It might not be now since your using `take`?

tracer/src/instruction/mod.rs
```diff
@@ -222,7 +222,7 @@ pub trait RISCVTrace: RISCVInstruction
 where
     RISCVCycle<Self>: Into<RV32IMCycle>,
 {
-    fn trace(&self, cpu: &mut Cpu) {
+    fn trace(&self, cpu: &mut Cpu) -> Vec<RV32IMCycle> {
```
jprider63: Should we tweak this to take a mutable reference to the `Vec<RV32IMCycle>`? Otherwise we're constantly allocating and concating vectors. 

We may even want to make it an `Option<&mut Vec<RV32IMCycle>>` if we want to option to turn disable tracing.

tracer/src/lib.rs
zutshi: I am not using the length. Wanted to keep in there since it allows for creating a fixed length iterator, which might be useful when resuming from checkpoints - allowing for trace generation from a checkpoint without knowing the `checkpoint_interval`.
It's related, but not directly connected to the code using `take`.

tracer/src/instruction/mod.rs
zutshi: Yes, it is something I did think of since the existing code followed that pattern. I wasn't sure of how much of an overall performance hit it is to and if its worth it to muddle the code.

zutshi: I wanted to make the vector `Option<&mut Vec<RV32IMCycle>>` to have optional tracing while reducing allocations. But I run into borrow checker problems as there is no Copy trait implementation for `Option<T>`. What does work however, is `&mut Option<&mut Vec<RV32IMCycle>>`. I haven't usually seen `&mut Option<&mut T>` in the wild and was hoping all we need here is `Option<&mut T>`.

zutshi: What's your rust opinion on this? Are the double references the cleanest way to work with the borrow checker?

jprider63:
Skipping the ‘Option’ might be the best approach. Based on our conversation earlier, I don’t think it’ll end up being used. On Jun 25, 2025, at 6:35 PM, Aditya Zutshi ***@***.***> wrote:﻿
@zutshi commented on this pull request.



In tracer/src/instruction/mod.rs:
> @@ -222,7 +222,7 @@ pub trait RISCVTrace: RISCVInstruction
 where
     RISCVCycle<Self>: Into<RV32IMCycle>,
 {
-    fn trace(&self, cpu: &mut Cpu) {
+    fn trace(&self, cpu: &mut Cpu) -> Vec<RV32IMCycle> {

I tried to make the vector Option<&mut Vec<RV32IMCycle>>, but I run into borrow checker problems since there is no Copy trait implemented. What does work however, is &mut Option<&mut Vec<RV32IMCycle>>. I haven't usually seen &mut Option<T> in the wild and was hoping all we need here is Option<&mut T>. What's your rust opinion on this?

zutshi:
Strongly recommend to squash the commits during merge. In hindsight, the commit history is more confusing than helpful.

tracer/src/emulator/cpu.rs
```diff
@@ -95,7 +96,8 @@ pub struct Cpu {
     is_reservation_set: bool,
     _dump_flag: bool,
     unsigned_data_mask: u64,
-    pub trace: Vec<RV32IMCycle>,
+    // pub trace: Vec<RV32IMCycle>,
+    pub trace_len: usize,
```
moodlezoup, line 100: Might be missing something but I don't see where this is getting updated

tracer/src/lib.rs
```diff
@@ -23,12 +25,528 @@ pub mod instruction;
...
+    };
+
+    let mut final_emulator_state = emulator_trace_iter.get_emulator_state();
+    let mut_jolt_device = &mut final_emulator_state.get_mut_cpu().get_mut_mmu().jolt_device;
+    let device = std::mem::take(mut_jolt_device);
+    (trace, device, checkpoint_interval.map(|_| checkpoints))
+}
+
+#[cfg(test)]
+mod test {
```
moodlezoup, line 292: nit: move tests to the bottom of the file

tracer/src/lib.rs
```diff
@@ -38,26 +556,89 @@ pub fn trace(
...
+///
+/// * `emulator` - Clone of the checkpoint emulator state to execute from
+/// * `prev_pc` - Previous program counter value, used for termination detection
+/// * `current_traces` - Buffer of trace entries from the most recent emulator tick
+#[derive(Clone)]
+pub struct LazyTraceIterator {
+    emulator_state: EmulatorState,
+    prev_pc: u64,
+    current_traces: Vec<RV32IMCycle>,
+    // If length is None, the iterator will run until the program ends.
```
moodlezoup: this looks like an outdated comment, but I do think it'd be nice to have a `length: Option<usize>` field here

tracer/src/lib.rs
```diff
@@ -23,12 +25,528 @@ pub mod instruction;
...
+/// assert!(trace_from_checkpoint_2 == execution_trace[2*n..])
+///
 #[tracing::instrument(skip_all)]
 pub fn trace(
     elf_contents: Vec<u8>,
     inputs: &[u8],
     memory_config: &MemoryConfig,
-) -> (Vec<RV32IMCycle>, JoltDevice) {
+    checkpoint_interval: Option<usize>,
+) -> (Vec<RV32IMCycle>, JoltDevice, Option<Vec<LazyTraceIterator>>) {
```
moodlezoup: Let's split this function into two versions –– one that returns the full `Vec<RV32IMCycle>` trace, and one that returns checkpoints `Vec<LazyTraceIterator>`. 
For all the use cases that I can think of, we'd want one or the other (but not both; keeping the full trace in memory defeats the purpose of streaming the trace from checkpoints).

moodlezoup (commented):
Great work, just a couple of minor comments

tracer/src/lib.rs
zutshi: I can add it back!

tracer/src/lib.rs
zutshi: I can add two function that call this function for those two use-cases.

jprider63: You probably want to split it up into the two separate functions Michael recommended. Otherwise you'll pay the cost of the entire trace allocation when you don't use it.

tracer/src/lib.rs
jprider63: I believe @zutshi switched to using `Take` which keeps track of the length internally. @zutshi, if you switch back to `length: Option<usize>`, we probably want to drop the `Take` so it's not duplicated. I somewhat prefered `Take` since it's not always used, but either way is fine.

tracer/src/lib.rs
jprider63: Another thought: For the one that returns a `Vec<RV32IMCycle>`, should it return a `LazyTraceIterator` instead?

tracer/src/lib.rs
moodlezoup: Ah I missed the `Take` –– let's leave it as is

tracer/src/lib.rs
moodlezoup: ^yeah let's do that

PR #698, opened 2025-06-02
feat: Dory integration

markosg04:
- This new PR (initially) integrates the Dory implementation into Jolt (the implementation itself has been moved to the Space and Time repo. See: https://github.com/a16z/jolt/pull/692). This helps to unblock the e2e completion of twist and shout.

Some initial Benchmarks:
```
Testing Dory PCS with 22 variables, 4194304 coefficients, sigma = 11
Setup time: 1.314285666s
Commit time: 6.305086167s
Prove time: 8.623711417s
Verify time: 80.691667ms
Total time: 16.323774917s
```

The numbers were obtained via 4 core M1Pro, with standard arkworks MSM/multipairing with Rayon.

There are quite a few TODOs leftover, which are listed below:
1. Optimize the polynomial coefficient <> DoryField conversions (right now it uses full arkworks Fr which can be costly for small poly coefficient cases)
2. This integration is coupled to arkworks bn254. if we want other implementations, we will need some refactors
3. Possibly consider removing labels being appended to the transcript for performance reasons
4. Implement batching (the "standard" linear combination trick) for multiple polynomial evals at the same point


jolt-core/src/poly/commitment/dory.rs
```diff
@@ -0,0 +1,568 @@
...
+        // Convert opening point to arkworks Fr
+        let point: Vec<Fr> = opening_point
+            .iter()
+            .map(|&p| unsafe { std::mem::transmute_copy(&p) })
+            .collect();
+
+        let num_vars = poly.get_num_vars();
+        let sigma = (num_vars + 1) / 2;
+
+        let dory_transcript_prover = JoltToDoryTranscript::new(transcript.clone());
```
moodlezoup: Let's not clone the transcript and instead pass by mutable reference. Technically, if there's only one opening proof and it's the last thing that happens in Jolt I think this would be fine, but it's pretty fragile and I think could have some dire security consequences if we mess it up

jolt-core/src/poly/commitment/dory.rs
```diff
@@ -0,0 +1,568 @@
...
+    unsafe fn transmute_label(label: &[u8]) -> &'static [u8] {
+        std::mem::transmute(label)
+    }
+}
+
+impl<T: Transcript> DoryTranscript for JoltToDoryTranscript<T> {
+    type Scalar = Fr;
+
+    fn append_bytes(&mut self, label: &[u8], bytes: &[u8]) {
+        let label_str = unsafe { Self::transmute_label(label) };
```
0xAndoroid: Instead of this transmuting maybe use to avoid unsafe code, and potential UB?
```rust
let label_str: &'static [u8] = Box::leak(label.to_vec().into_boxed_slice());
```
This also applies to all other occurrences.

jolt-core/src/poly/commitment/dory.rs
```diff
@@ -0,0 +1,568 @@
...
+pub struct DoryCommitmentScheme<F: JoltField, ProofTranscript: Transcript> {
+    _phantom_field: PhantomData<F>,
+    _phantom_transcript: PhantomData<ProofTranscript>,
+}
+
+/// Setup structure containing both prover and verifier parameters.
+/// Dory API exposes an SRS to disk option, but is not (yet) utilized here.
+#[derive(Clone, CanonicalSerialize, CanonicalDeserialize)]
+pub struct DorySetup {
+    pub prover_setup: ProverSetup<ArkBn254Pairing>,
```
0xAndoroid: I can see that other PCS in the codebase use generic `ark_ec::pairing::Pairing` trait. Should we use it here as well? This applies everywhere else.

jolt-core/src/poly/commitment/dory.rs
```diff
@@ -0,0 +1,568 @@
...
+            max_log_n,
+        }
+    }
+
+    fn commit(poly: &MultilinearPolynomial<Self::Field>, setup: &Self::Setup) -> Self::Commitment {
+        // Extract polynomial coefficients
+        let field_coeffs = extract_field_coefficients(poly);
+
+        // Convert JoltField coefficients to Fr
+        // This assumes F is Fr or can be safely transmuted to Fr
```
0xAndoroid: Is this always guaranteed?

jolt-core/src/poly/commitment/dory.rs
```diff
@@ -0,0 +1,568 @@
...
+                &point,
+                verifier_builder,
+                proof.sigma,
+                &setup.verifier_setup,
+                dory_transcript,
+            );
+
+        match verify_result {
+            Ok(()) => Ok(()),
+            Err(_) => Err(ProofVerifyError::InternalError),
```
0xAndoroid: Maybe add more context for this error, or a custom one

jolt-core/src/poly/commitment/dory.rs
```diff
@@ -0,0 +1,568 @@
...
+
+    // @TODO: Implement linear combination batching
+    fn batch_commit<U>(polys: &[U], setup: &Self::Setup) -> Vec<Self::Commitment>
+    where
+        U: Borrow<MultilinearPolynomial<Self::Field>> + Sync,
+    {
+        todo!()
+    }
+
+    // Note that Dory implementation sometimes uses the term 'evaluation'/'evaluate' -- this is same as 'opening'/'open'
```
0xAndoroid: should we just stick with opening everywhere in jolt?

jolt-core/src/poly/commitment/dory.rs
```diff
@@ -0,0 +1,568 @@
...
+    use ark_bn254::Fr;
+    use ark_std::rand::thread_rng;
+
+    #[test]
+    fn test_dory_commitment_scheme() {
+        use ark_std::UniformRand;
+        use std::time::Instant;
+
+        // Create a random polynomial and other related preliminaries
+        let max_log_n = 22; // This will support polynomials up to 2^22 coefficients
```
0xAndoroid: any merit to use `max_log_n = 20` or less to speed up tests? Or this would lose coverage?

0xAndoroid (changes requested):
Great implementation. I'm pretty worried about unsafe transmutes, but everything else is great except for some nits.

jolt-core/src/poly/commitment/dory.rs
```diff
@@ -0,0 +1,568 @@
...
+    F: JoltField,
+    ProofTranscript: Transcript,
+{
+    type Field = F;
+    type Setup = DorySetup;
+    type Commitment = DoryCommitment;
+    type Proof = DoryProof<F>;
+    type BatchedProof = DoryBatchedProof<F>;
+
+    fn setup(max_log_n: usize) -> Self::Setup {
```
moodlezoup: The `CommitmentScheme` trait (and the existing implementations thereof) has this parameter as `max_len`, which is `1 << max_log_n`. That said, I'm down to change it everywhere to this (but maybe rename `max_log_n` to `max_num_vars`)

jolt-core/src/poly/commitment/dory.rs
moodlezoup: yeah let's dial it down to 18

jolt-core/src/poly/commitment/dory.rs
```diff
@@ -0,0 +1,568 @@
...
+            .iter()
+            .map(|&coeff| unsafe { std::mem::transmute_copy(&coeff) })
+            .collect();
+
+        // For Dory, sigma is the number of columns in the matrix representation
+        // we use sigma = ceil(num_vars / 2) so we have a nice square matrix.
+        let num_vars = poly.get_num_vars();
+        let sigma = (num_vars + 1) / 2;
+        let commitment =
+            commit::<ArkBn254Pairing, OptimizedMsmG1>(&coeffs, 0, sigma, &setup.prover_setup);
```
moodlezoup: this is really a comment for the other repo, but let's change the output type of `commit` to `PairingOutput<Bn254>`. The (unwrapped) target field type doesn't implement `Mul<P::ScalarField>`, which we need to homomorphically combine Dory commitments

jolt-core/src/poly/commitment/dory.rs
markosg04: I beleive we do use 'opening' within jolt but we make a call to `evaluate()` and make use of `evaluation` which come from the Dory repo (and the Dory paper)

jolt-core/src/poly/commitment/dory.rs
```diff
@@ -0,0 +1,993 @@
...
+    poly::{
+        compact_polynomial::CompactPolynomial, dense_mlpoly::DensePolynomial,
+        multilinear_polynomial::MultilinearPolynomial,
+    },
+    utils::{errors::ProofVerifyError, transcript::Transcript},
+};
+
+use std::{borrow::Borrow, marker::PhantomData};
+
+// Arkworks
```
0xAndoroid: Useless comments.

I generally think there's an overuse of comments in some places here. Especially `// ===== Fieold Wrapper =====` I'd remove.

jolt-core/src/poly/commitment/dory.rs
```diff
@@ -0,0 +1,993 @@
...
+    }
+}
+
+// ===== Group Wrappers =====
+
+/// Wrapper for elliptic curve groups (G1, G2)
+#[repr(transparent)]
+#[derive(Debug, Clone, PartialEq, CanonicalSerialize, CanonicalDeserialize)]
+pub struct JoltGroupWrapper<G: CurveGroup> {
+    pub inner: G,
```
0xAndoroid: Why do we have `inner` here but just a `(pub F)` in the field wrapper?

I think that having the implementation using tuple is looks cleaner.

jolt-core/src/poly/commitment/dory.rs
```diff
@@ -0,0 +1,993 @@
...
+    msm::{Icicle, VariableBaseMSM},
+    poly::{
+        compact_polynomial::CompactPolynomial, dense_mlpoly::DensePolynomial,
+        multilinear_polynomial::MultilinearPolynomial,
+    },
+    utils::{errors::ProofVerifyError, transcript::Transcript},
+};
+
+use std::{borrow::Borrow, marker::PhantomData};
+
```
0xAndoroid: run the `cargo fmt`

jolt-core/src/poly/commitment/dory.rs
```diff
@@ -0,0 +1,993 @@
...
+impl<P> DoryMultiScalarMul<JoltGTWrapper<P>> for JoltGTCustomMSM<P>
+where
+    P: ArkPairing,
+    P::ScalarField: JoltField,
+{
+    fn msm(
+        bases: &[JoltGTWrapper<P>],
+        scalars: &DoryMultilinearPolynomial<JoltFieldWrapper<P::ScalarField>>,
+    ) -> JoltGTWrapper<P> {
+        use rayon::prelude::*;
```
0xAndoroid: Is there a reason to not put this import in the top level?

jolt-core/src/poly/commitment/dory.rs
```diff
@@ -0,0 +1,993 @@
...
+                    .map(|(base_chunk, coeff_chunk)| {
+                        base_chunk
+                            .iter()
+                            .zip(coeff_chunk.iter())
+                            .filter(|(_, coeff)| !coeff.0.is_zero())
+                            .fold(JoltGTWrapper::<P>::identity(), |acc, (base, coeff)| {
+                                acc.add(&base.scale(coeff))
+                            })
+                    })
+                    .reduce(JoltGTWrapper::<P>::identity, |a, b| a.add(&b))
```
0xAndoroid: I think reduce method docs say that you should call `.sum()` if you are summing. 

Speaking of, is `DoryGroup` implementing `std::ops::Add` and other traits, or is it just the `.add` method on the trait? Implementing `std::ops` should be preferred

jolt-core/src/poly/commitment/dory.rs
```diff
@@ -0,0 +1,993 @@
...
+
+// ===== Type Aliases =====
+
+pub type JoltG1Wrapper = JoltGroupWrapper<G1Projective>;
+pub type JoltG2Wrapper = JoltGroupWrapper<G2Projective>;
+pub type JoltGTBn254 = JoltGTWrapper<Bn254>;
+
+pub type JoltG1MSM = JoltCustomMSM<G1Projective>;
+pub type JoltG2MSM = JoltCustomMSM<G2Projective>;
+pub type JoltGTMSM = JoltGTCustomMSM<Bn254>;
```
0xAndoroid: You don't actually need these type aliases.
You can just have `pub struct JoltMSM;` instead of these `JoltCustomMSM<...>` and `JoltGTCustomMSM`, and write the implementations based on the generics:
```rust
impl<G> DoryMultiScalarMul<JoltGroupWrapper<G>> for JoltMSM
```
```rust
impl<P> DoryMultiScalarMul<JoltGTWrapper<P>> for JoltMSM
```
And rust will pick everything up automatically.

jolt-core/src/poly/commitment/dory.rs
```diff
@@ -0,0 +1,993 @@
...
+}
+
+/// Proof of polynomial evaluation
+#[derive(Clone, Debug, CanonicalSerialize, CanonicalDeserialize)]
+pub struct DoryProofData<F: JoltField> {
+    pub evaluation: F,
+    pub opening_point: Vec<F>,
+    pub sigma: usize,
+    pub dory_proof_data: DoryProof<JoltG1Wrapper, JoltG2Wrapper, JoltGTBn254>,
+    _phantom_f_marker: PhantomData<F>,
```
0xAndoroid: that's not needed

jolt-core/src/poly/commitment/dory.rs
```diff
@@ -0,0 +1,993 @@
...
+        }
+    }
+}
+
+// ===== Transcript Adapter =====
+
+/// NewType wrapper for Jolt transcript to Dory transcript
+#[derive(Default)]
+pub struct JoltToDoryTranscriptRef<'a, F: JoltField, T: Transcript> {
+    transcript: Option<&'a mut T>,
```
0xAndoroid, line 680: Why is this an `Option`? Seems like it is only initialized with `Some` in `new` method.

jolt-core/src/poly/commitment/dory.rs
0xAndoroid: This point still holds, would help debugging to know which of three options of DoryError it is

0xAndoroid (commented):
Love this generic solution!
Added a note to remove some verbosity, but overall great work!

jolt-core/src/poly/commitment/dory.rs
markosg04: added a comment for it but the short answer is we needed to derive Default

jolt-core/src/poly/commitment/dory.rs
```diff
@@ -0,0 +1,904 @@
...
+
+    fn len(&self) -> usize {
+        match self {
+            MultilinearPolynomial::LargeScalars(dense) => dense.Z.len(),
+            MultilinearPolynomial::U8Scalars(compact) => compact.coeffs.len(),
+            MultilinearPolynomial::U16Scalars(compact) => compact.coeffs.len(),
+            MultilinearPolynomial::U32Scalars(compact) => compact.coeffs.len(),
+            MultilinearPolynomial::U64Scalars(compact) => compact.coeffs.len(),
+            MultilinearPolynomial::I64Scalars(compact) => compact.coeffs.len(),
+        }
```
moodlezoup: can use `MultilinearPolynomial::len` here

jolt-core/src/poly/commitment/dory.rs
```diff
@@ -0,0 +1,904 @@
...
+// Dory's Poly trait: right now it uses default implementations for the poly utilities. We will want to override them
+// for the sparse case or for further optimizations.
+// We implement get() and len() since they do not have default implementations.
+
+impl<F, G> DoryPolynomial<JoltFieldWrapper<F>, JoltGroupWrapper<G>> for MultilinearPolynomial<F>
+where
+    F: JoltField + PrimeField,
+    G: CurveGroup<ScalarField = F> + VariableBaseMSM,
+{
+    fn get(&self, index: usize) -> JoltFieldWrapper<F> {
```
moodlezoup, line 662: can use `MultilinearPolynomial::get_coeff` here

moodlezoup: Where do we use this in Dory? I ask because this could be problematic for the sparse polynomials

jolt-core/src/poly/commitment/dory.rs
```diff
@@ -0,0 +1,904 @@
...
+impl<'a, F: JoltField, T: Transcript> DoryTranscript for JoltToDoryTranscriptRef<'a, F, T> {
+    type Scalar = JoltFieldWrapper<F>;
+
+    fn append_bytes(&mut self, label: &[u8], bytes: &[u8]) {
+        let transcript = self
+            .transcript
+            .as_mut()
+            .expect("Transcript not initialized");
+        let label_str: &'static [u8] = Box::leak(label.to_vec().into_boxed_slice());
+        transcript.append_message(label_str);
```
moodlezoup: let's just not append the label for these methods
```suggestion
```

jolt-core/src/poly/commitment/dory.rs
```diff
@@ -0,0 +1,904 @@
...
+
+#[derive(Clone, Debug)]
+pub struct DoryCommitmentScheme<ProofTranscript: Transcript>(PhantomData<ProofTranscript>);
+
+#[derive(Clone, CanonicalSerialize, CanonicalDeserialize)]
+pub struct DorySetup {
+    pub prover_setup: ProverSetup<JoltBn254>,
+    pub verifier_setup: VerifierSetup<JoltBn254>,
+    pub max_num_vars: usize,
+}
```
moodlezoup, line 757:
```suggestion
pub struct DorySetup {
    pub prover_setup: ProverSetup<JoltBn254>,
    pub verifier_setup: VerifierSetup<JoltBn254>,
    pub max_num_vars: usize,
    pub sigma: Option<usize>,
}
```
Somewhat janky, but in order to batch polynomials of different lengths together (i.e. dense and sparse polynomials) I think we'll need to make sure they all use the same # columns (i.e. `sigma`). 
So we can leave this `Option` as `None` in preprocessing, then have the prover populate the field before it commits to all the polynomials

jolt-core/src/poly/commitment/dory.rs
```diff
@@ -0,0 +1,904 @@
...
+
+    fn verify(
+        proof: &Self::Proof,
+        setup: &Self::Setup,
+        transcript: &mut ProofTranscript,
+        _opening_point: &[Self::Field],
+        _opening: &Self::Field,
+        commitment: &Self::Commitment,
+    ) -> Result<(), ProofVerifyError> {
+        let opening_point: Vec<JoltFieldWrapper<Self::Field>> = proof
```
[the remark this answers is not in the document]
moodlezoup: we should remove `evaluation` and `opening_point` from `DoryProofData` and use these function parameters instead

moodlezoup (commented):
Looking good after the refactor! A few minor comments/questions

jolt-core/src/poly/commitment/dory.rs
markosg04: only for default implementations

jolt-core/src/poly/commitment/dory.rs
markosg04: addressed via the v = M*R trick

pre_opt1.json
moodlezoup, line 1: let's remove these json files

markosg04:
https://github.com/a16z/jolt/pull/712 closing in favour of this

PR #709, opened 2025-06-17
Draft: Use Gruen's optimization when computing univariate polynomials

protoben:
This PR changes the computation of the univariate polynomials in each sumcheck round to use both the Dao-Thaler and Gruen optimizations (see Sec. 3 of https://eprint.iacr.org/2024/1210.pdf). To do so, we use the `GruenSplitEqPolynomial` data structure, which allows us to avoid computing the cubic evaluations at `{0, 1, 2, 3}` directly, and instead compute evaluations of quadratic polynomials at 0 and infinity.

Changes:
* Change `BatchedGrandProductProof::compute_cubic` to use `GruenSplitEqPolynomial`.

Planned changes:
* Use `GruenSplitEqPolynomial` in the following places in T&S:
  * [The RAM-check polynomials](https://github.com/a16z/jolt/blob/feat/twist-shout-integration/jolt-core/src/jolt/vm/ram.rs#L635)
  * [The register polynomials](https://github.com/a16z/jolt/blob/feat/twist-shout-integration/jolt-core/src/jolt/vm/registers.rs#L469)
  * [The bytecode polynomials](https://github.com/a16z/jolt/blob/feat/twist-shout-integration/jolt-core/src/jolt/vm/bytecode.rs#L366)

### DIFF
--- a/jolt-core/src/poly/dense_interleaved_poly.rs
+++ b/jolt-core/src/poly/dense_interleaved_poly.rs
@@ -345,64 +345,6 @@ impl<F: JoltField, ProofTranscript: Transcript> BatchedCubicSumcheck<F, ProofTra
             previous_round_claim,
         );
 
-        #[cfg(test)]
-        {
-            let eq_merged = eq_poly.merge();
-            let naive_cubic_evals = self
-                .coeffs
-                .par_chunks(4)
-                .zip(eq_merged.evals().par_chunks(2))
-                .map(|(self_chunk, eq_chunk)| {
-                    let eq_evals = {
-                        let eval_point_0 = eq_chunk[0];
-                        let m_eq = eq_chunk[1] - eq_chunk[0];
-                        let eval_point_2 = eq_chunk[1] + m_eq;
-                        let eval_point_3 = eval_point_2 + m_eq;
-                        (eval_point_0, eval_point_2, eval_point_3)
-                    };
-                    let left = (
-                        *self_chunk.get(0).unwrap_or(&F::zero()),
-                        *self_chunk.get(2).unwrap_or(&F::zero()),
-                    );
-                    let right = (
-                        *self_chunk.get(1).unwrap_or(&F::zero()),
-                        *self_chunk.get(3).unwrap_or(&F::zero()),
-                    );
-
-                    let m_left = left.1 - left.0;
-                    let m_right = right.1 - right.0;
-
-                    let left_eval_2 = left.1 + m_left;
-                    let left_eval_3 = left_eval_2 + m_left;
-
-                    let right_eval_2 = right.1 + m_right;
-                    let right_eval_3 = right_eval_2 + m_right;
-
-                    (
-                        eq_evals.0 * left.0 * right.0,
-                        eq_evals.1 * left_eval_2 * right_eval_2,
-                        eq_evals.2 * left_eval_3 * right_eval_3,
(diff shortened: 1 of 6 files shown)


jolt-core/src/poly/dense_interleaved_poly.rs
```diff
@@ -299,40 +273,121 @@ impl<F: JoltField, ProofTranscript: Transcript> BatchedCubicSumcheck<F, ProofTra
...
+                        *E_out_eval * inner_sum.0,
+                        *E_out_eval * inner_sum.1,
                     )
                 })
                 .reduce(
-                    || (F::zero(), F::zero(), F::zero()),
-                    |sum, evals| (sum.0 + evals.0, sum.1 + evals.1, sum.2 + evals.2),
+                    || (F::zero(), F::zero()),
+                    |sum, evals| (sum.0 + evals.0, sum.1 + evals.1),
                 )
```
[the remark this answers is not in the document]
protoben: Fixed.

jolt-core/src/subprotocols/sparse_grand_product.rs
```diff
@@ -884,10 +788,11 @@ impl<F: JoltField, ProofTranscript: Transcript> BatchedCubicSumcheck<F, ProofTra
                 dense_cubic_evals.1,
                 dense_cubic_evals.2,
             ];
-            assert_eq!(dense_cubic_evals, cubic_evals);
+            let dense_cubic = UniPoly::from_evals(&dense_cubic_evals);
+            assert_eq!(dense_cubic, cubic);
```
protoben, line 792: Assert currently failing.

jolt-core/src/poly/sparse_interleaved_poly.rs
```diff
@@ -656,54 +601,41 @@ impl<F: JoltField, ProofTranscript: Transcript> BatchedCubicSumcheck<F, ProofTra
...
+        let cubic = UniPoly::from_linear_times_quadratic_with_hint(
+            // The coefficients of `eq(w[(n - i)..], r[..i]) * eq(w[n - i - 1], X)`
+            [
+                eq_poly.current_scalar - scalar_times_w_i,
+                scalar_times_w_i + scalar_times_w_i - eq_poly.current_scalar,
+            ],
+            quadratic_evals.0,
+            quadratic_evals.1,
+            previous_round_claim,
+        );
```
protoben, line 638: Assert below currently failing.

protoben:
Porting all implementors of `BatchedGrandProductProof` to use `GruenSplitEqPolynomial` is complete, however, I'm not getting the correct output in some of the cases. `DenseInterleavedPolynomial` appears to be producing the correct output, but `SparseInterleavedPolynomial` and `SparseGrandProduct` are not. The failing asserts are linked in comments [here](https://github.com/a16z/jolt/pull/709#discussion_r2155458714) and [here](https://github.com/a16z/jolt/pull/709#discussion_r2155459557).

@quangvdao If you happen to have a chance to take a look and see if there's anything obvious that I'm doing wrong, it would be greatly appreciated. I'll continue trying to debug it, myself.

quangvdao:
This is because the `evals_assuming_all_ones` distort the decomposition. You will need to derive the correct formula first, then fix that part. I was aware of this mismatch but did not have the time to dive in, could do so if you would like.

PR #710, opened 2025-06-17
[JOLT-131] RAM Raf Sumcheck

0xAndoroid:


jolt-core/src/poly/identity_poly.rs
```diff
@@ -82,12 +82,139 @@ impl<F: JoltField> PolynomialEvaluation<F> for IdentityPolynomial<F> {
...
+    fn bind(&mut self, r: F, order: BindingOrder) {
+        debug_assert!(self.num_bound_vars < self.num_vars);
+        debug_assert_eq!(
+            order,
+            BindingOrder::LowToHigh,
+            "UnmapAddressPolynomial only supports low-to-high binding"
+        );
+
+        // Update the linear term: add 4 * 2^i * r_i
+        self.linear_term += F::from_u64(4 * (1u64 << self.num_bound_vars)) * r;
```
moodlezoup: makes ~zero difference for the changes in this PR, but just fyi [mul_u64](https://github.com/a16z/jolt/blob/feat/twist-shout-integration/jolt-core/src/field/ark.rs#L155) is faster than `F::from_u64` + multiply

jolt-core/src/poly/identity_poly.rs
```diff
@@ -82,12 +82,139 @@ impl<F: JoltField> PolynomialEvaluation<F> for IdentityPolynomial<F> {
...
+        for i in 1..degree {
+            // Evaluate at point i+1
+            linear_eval += m;
+            linear_evals[i] = linear_eval;
+            product_evals[i] = if index == 0 {
+                self.product_term * (-F::from_u32(i as u32))
+            } else {
+                F::zero()
+            };
+        }
```
moodlezoup: microoptimization
```suggestion
        let m = F::from_u32(4 * (1 << self.num_bound_vars));
        let mut linear_eval = linear_evals[0] + m;
        let mut product_eval = F::zero();
        for i in 1..degree {
            // Evaluate at point i+1
            linear_eval += m;
            linear_evals[i] = linear_eval;
            product_evals[i] = if index == 0 {
                product_eval -= self.product_term;
                product_eval
            } else {
                F::zero()
            };
        }
```

moodlezoup (approved):
lgtm. just some tiny perf suggestions

jolt-core/src/poly/identity_poly.rs
0xAndoroid: Nice, didn't know about this

PR #712, opened 2025-06-18
fix: split PCS setup into prover and verifier setup

sagar-a16z:


jolt-core/src/poly/commitment/commitment_scheme.rs
```diff
@@ -23,9 +24,14 @@ pub trait CommitmentScheme<ProofTranscript: Transcript>: Clone + Sync + Send + '
     type Proof: Sync + Send + CanonicalSerialize + CanonicalDeserialize;
     type BatchedProof: Sync + Send + CanonicalSerialize + CanonicalDeserialize;
 
-    fn setup(max_len: usize) -> Self::Setup;
-    fn commit(poly: &MultilinearPolynomial<Self::Field>, setup: &Self::Setup) -> Self::Commitment;
-    fn batch_commit<U>(polys: &[U], gens: &Self::Setup) -> Vec<Self::Commitment>
+    fn setup_prover(max_len: usize) -> Self::ProverSetup;
+    fn setup_verifier(setup: &Self::ProverSetup) -> Self::VerifierSetup;
+    fn srs_size(setup: &Self::ProverSetup) -> usize;
```
moodlezoup, line 29: do we need this function? imo the "size" of an SRS is kind of ambiguous, so I'd prefer to delete it if it's unneeded

sagar-a16z: It was added in: https://github.com/a16z/jolt/pull/704 on `main`. I can remove it from this PR.
It's being used for better warnings https://github.com/a16z/jolt/blob/e8c245607a4ab881f40713bae7cc4119dc117ae4/jolt-core/src/jolt/vm/mod.rs#L478-L482

sagar-a16z: Alternatively I can rename it on both branches.

moodlezoup: ah right –– let's leave it then

PR #726, opened 2025-06-24
[2/2] Twist/Shout PCS integration

moodlezoup:
Follow-up to https://github.com/a16z/jolt/pull/699 adding support for one-hot and Inc polynomials

With this PR, every committed polynomial has _at least one_ of its openings proven using Dory, but not necessarily _all_ of its openings. This is because the exact openings are still in flux, as we further batch sumchecks together.

### DIFF
--- a/jolt-core/src/jolt/vm/mod.rs
+++ b/jolt-core/src/jolt/vm/mod.rs
@@ -324,8 +324,7 @@ where
             .map(|poly| poly.generate_witness(&preprocessing, &trace))
             .collect();
         let commitments: Vec<_> = committed_polys
-            // .par_iter()
-            .iter()
+            .par_iter()
             .map(|poly| PCS::commit(poly, &preprocessing.generators))
             .collect();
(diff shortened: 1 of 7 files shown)


jolt-core/src/subprotocols/sparse_grand_product.rs
```diff
@@ -1334,6 +1338,7 @@ mod tests {
     }
 
     #[test]
+    #[ignore]
```
moodlezoup, line 1341: Ignoring this for now, will delete this file in a separate PR

jolt-core/src/subprotocols/grand_product_quarks.rs
```diff
@@ -722,11 +725,13 @@ mod quark_grand_product_tests {
     }
 
     #[test]
+    #[ignore]
```
moodlezoup, line 728: Ignoring these tests for now, will delete this file in a separate PR

PR #727, opened 2025-06-25
Feat/ra virtualization

markosg04:
- Adds a generic implementation of RA virtualization sumcheck
- Integrates this new sumcheck into the RAM twist instance

### DIFF
--- a/jolt-core/src/subprotocols/ra_virtual.rs
+++ b/jolt-core/src/subprotocols/ra_virtual.rs
@@ -197,7 +197,7 @@ impl<F: JoltField, ProofTranscript: Transcript, const D: usize>
     }
 
     fn input_claim(&self) -> F {
-        self.ra_claim.clone()
+        self.ra_claim
     }
 
     fn cache_openings(&mut self) {
--- a/jolt-core/src/subprotocols/sumcheck.rs
+++ b/jolt-core/src/subprotocols/sumcheck.rs
@@ -333,7 +333,6 @@ impl BatchedSumcheck {
 
         Ok(r_sumcheck)
     }
-
 }
 
 impl<F: JoltField, ProofTranscript: Transcript> SumcheckInstanceProof<F, ProofTranscript> {


jolt-core/src/subprotocols/ra_virtual.rs
```diff
@@ -0,0 +1,204 @@
...
+        let eq_evals = EqPolynomial::evals(&r_cycle);
+
+        let eq_poly = MultilinearPolynomial::from(eq_evals);
+
+        // We do pre-binding. The sumcheck is in variable j hence we can pre-bind the r_address chunks:
+        for (poly, chunk) in ra_i_polys.iter_mut().zip(r_address_chunks.iter()) {
+            for &r_value in chunk.iter() {
+                poly.bind_parallel(r_value, BindingOrder::LowToHigh);
+            }
+        }
```
moodlezoup: Instead of "pre-binding", which requires that we fully materialize each ra_i polynomial (which have length K^(1/d) * T), we should do the following:
- Compute the table of evals for EQ(k_i, r_address_chunks[i]) for each `i in 0..D`. Each table is size K^(1/d), which is small and quick to compute. 
- Then compute each pre-bound ra_i polynomial (i.e. ra_i(r_address_chunks[i], i)) with a single pass over the original addresses. We sort of do this in prefix-suffix Shout [here](https://github.com/a16z/jolt/blob/feat/twist-shout-integration/jolt-core/src/subprotocols/sparse_dense_shout.rs#L503-L510) –– `v` is the EQ table described above, just constructed incrementally over the course of the prefix-suffix phase. This takes time O(d * T)

jolt-core/src/subprotocols/ra_virtual.rs
```diff
@@ -0,0 +1,204 @@
...
+use crate::poly::multilinear_polynomial::PolynomialEvaluation;
+use rayon::iter::IndexedParallelIterator;
+use rayon::iter::{IntoParallelRefIterator, IntoParallelRefMutIterator, ParallelIterator};
+pub struct RAProof<F: JoltField, ProofTranscript: Transcript> {
+    pub ra_i_claims: Vec<F>,
+    pub sumcheck_proof: SumcheckInstanceProof<F, ProofTranscript>,
+}
+
+/// Virtual RA sumcheck for d-way chunked addresses
+pub struct RASumcheck<F: JoltField, const D: usize> {
```
moodlezoup: We probably want to have explicit separation here between the prover's sumcheck state and the verifier's sumcheck state (see [here](https://github.com/a16z/jolt/blob/feat/twist-shout-integration/jolt-core/src/subprotocols/shout.rs#L471-L475) for example) because this struct will be used by both prover and verifier.

jolt-core/src/subprotocols/ra_virtual.rs
```diff
@@ -0,0 +1,204 @@
...
+        let eval_points: Vec<usize> = (0..=degree)
+            .filter(|&i| i != 1)
+            .collect();
+        
+        let mut evals = vec![F::zero(); eval_points.len()];
+
+        let remaining_vars = <Self as BatchableSumcheckInstance<F, ProofTranscript>>::num_rounds(&self) - round - 1;
+        
+        for (eval_idx, &point) in eval_points.iter().enumerate() {
+            for j in 0..(1 << remaining_vars) {
```
moodlezoup: I would invert these loops; see [here](https://github.com/a16z/jolt/blob/feat/twist-shout-integration/jolt-core/src/subprotocols/shout.rs#L171-L186) for the idiomatic way to implement this function

jolt-core/src/subprotocols/ra_virtual.rs
```diff
@@ -0,0 +1,404 @@
...
+            .into_par_iter()
+            .map(|i| {
+                let eq_evals = eq_poly.sumcheck_evals(i, degree, BindingOrder::LowToHigh);
+
+                let mut evals = vec![F::zero(); degree];
+
+                for eval_point in 0..degree {
+                    let mut result = eq_evals[eval_point];
+
+                    for ra_i_poly in ra_i_polys.iter() {
```
moodlezoup: we can invert these two for loops to avoid calling `ra_i_poly.sumcheck_evals` unnecessarily

jolt-core/src/subprotocols/sumcheck.rs
```diff
@@ -310,6 +310,21 @@ impl BatchedSumcheck {
...
+        Self::prove(vec![sumcheck_instance], transcript)
+    }
+
+    pub fn verify_single<F: JoltField, ProofTranscript: Transcript>(
+        proof: &SumcheckInstanceProof<F, ProofTranscript>,
+        sumcheck_instance: &dyn BatchableSumcheckInstance<F, ProofTranscript>,
+        transcript: &mut ProofTranscript,
+    ) -> Result<Vec<F>, ProofVerifyError> {
+        Self::verify(proof, vec![sumcheck_instance], transcript)
+    }
```
moodlezoup: nit: let's move these to the `BatchableSumcheckInstance` trait

moodlezoup (commented):
nice work, just a couple of nits

PR #738, opened 2025-07-02
Instruction execution RAF sumcheck, prefix-suffix abstraction

0xAndoroid:


jolt-core/src/poly/prefix_suffix.rs
```diff
@@ -0,0 +1,473 @@
...
+                .get_or_init(|| self.sumcheck_evals(index, degree, order))
+                .clone()
+        } else {
+            self.sumcheck_evals(index, degree, order)
+        }
+    }
+
+    pub fn clear_cache(&mut self, use_cache: bool) {
+        if use_cache {
+            // OnceCell doesn't support clearing, so we need to replace with new ones
```
moodlezoup: you can use `OnceCell::take`

jolt-core/src/poly/prefix_suffix.rs
```diff
@@ -0,0 +1,474 @@
...
+    fn index(&self, index: Prefix) -> &Self::Output {
+        &self.polys[index]
+    }
+}
+
+impl<F: JoltField> IndexMut<Prefix> for PrefixRegistry<F> {
+    fn index_mut(&mut self, index: Prefix) -> &mut Self::Output {
+        &mut self.polys[index]
+    }
+}
```
moodlezoup, line 59: seems a bit unnecessary

jolt-core/src/poly/prefix_suffix.rs
```diff
@@ -0,0 +1,474 @@
...
+        phase: usize,
+        prefix_registry: &mut PrefixRegistry<F>,
+    ) -> [Option<Arc<RwLock<CachedPolynomial<F>>>>; ORDER];
+    fn suffixes(&self) -> [Box<dyn SuffixPolynomial<F> + Sync>; ORDER];
+}
+
+pub struct PrefixSuffixDecomposition<F: JoltField, const ORDER: usize> {
+    poly: Box<dyn PrefixSuffixPolynomial<F, ORDER> + Send + Sync>,
+    P: Vec<Option<Arc<RwLock<CachedPolynomial<F>>>>>,
+    Q: Vec<DensePolynomial<F>>,
```
moodlezoup: any reason not to use arrays?
```suggestion
    P: [Option<Arc<RwLock<CachedPolynomial<F>>>>; ORDER],
    Q: [DensePolynomial<F>; ORDER],
```

jolt-core/src/poly/prefix_suffix.rs
```diff
@@ -0,0 +1,474 @@
...
+        if self.phase != 0 {
+            self.reset_Q();
+        }
+        let suffix_len = self.suffix_len();
+        let suffixes = self.poly.suffixes();
+        self.Q
+            .par_iter_mut()
+            .zip(suffixes.par_iter())
+            .for_each(|(poly, suffix)| {
+                for (j, k) in indices.clone() {
```
moodlezoup, line 267: I think you're right, the reason this is relatively slow is that `indices` is dense for `lookup_indices_uninterleave` and `lookup_indices_identity`. 
Instead of passing in `indices` as an iterator, let's pass it in as a `&[(usize, &LookupBits)]` and do the following:
- if `indices.len() < SOME_THRESHOLD`, iterate serially as we do now
- if `indices.len() >= SOME_THRESHOLD`, iterate using `indices.par_chunks()`, allocating an intermediate result vector of length `self.chunk_len.pow2()` per chunk, then adding up the intermediate results

jolt-core/src/poly/prefix_suffix.rs
```diff
@@ -0,0 +1,474 @@
...
+}
+
+pub trait CacheablePolynomial<F: JoltField>:
+    PolynomialEvaluation<F> + PolynomialBinding<F> + Send + Sync
+{
+}
+
+pub struct CachedPolynomial<F: JoltField> {
+    pub inner: Box<dyn CacheablePolynomial<F>>,
+    pub sumcheck_evals_cache: Vec<OnceCell<Vec<F>>>,
```
moodlezoup: we already assume degree=2 elsewhere in this file, perhaps we should just have an explicit `const DEGREE: usize = 2` and change this to
```suggestion
    pub sumcheck_evals_cache: Vec<OnceCell<[F; DEGREE]>>,
```

jolt-core/src/poly/identity_poly.rs
```diff
@@ -25,15 +44,205 @@ impl<F: JoltField> PolynomialBinding<F> for IdentityPolynomial<F> {
...
             order,
-            BindingOrder::LowToHigh,
-            "IdentityPolynomial only supports low-to-high binding"
+            BindingOrder::HighToLow,
+            "OperandPolynomial only supports high-to-low binding"
         );
 
-        self.bound_value += F::from_u32(1u32 << self.num_bound_vars) * r;
+        if (self.num_bound_vars.is_multiple_of(2) && self.side == OperandSide::Right)
+            || (self.num_bound_vars % 2 == 1 && self.side == OperandSide::Left)
```
moodlezoup: nit: let's use `is_even()` and `is_odd()` throughout this file

jolt-core/src/poly/prefix_suffix.rs
```diff
@@ -0,0 +1,474 @@
...
+    }
+}
+
+pub trait CacheablePolynomial<F: JoltField>:
+    PolynomialEvaluation<F> + PolynomialBinding<F> + Send + Sync
+{
+}
+
+pub struct CachedPolynomial<F: JoltField> {
+    pub inner: Box<dyn CacheablePolynomial<F>>,
```
moodlezoup, line 81: I'm slightly tempted to simplify this and eliminate the `CacheablePolynomial` trait
```suggestion
    pub inner: MultilinearPolynomial<F>,
```
So we'd make `prefix_polynomial` return a `MultilinearPolynomial<F>`, constructing by computing all 2^16 coefficients for the current phase. 
We'd sacrifice some performance in doing so (e.g. IdentityPolynomial can be constructed/bound in constant time) but it would be a low-order cost

jolt-core/src/poly/prefix_suffix.rs
```diff
@@ -0,0 +1,474 @@
...
+    /// Assumes binding_order to be HighToLow
+    fn prefix_polynomial(
+        &self,
+        checkpoints: &PrefixCheckpoints<F>,
+        chunk_len: usize,
+        phase: usize,
+    ) -> CachedPolynomial<F>;
+
+    /// Sets the checkpoint value
+    fn set_bound_value(&mut self, bound_value: F, bound_vars: usize);
```
moodlezoup: let's just roll `set_bound_value` into `prefix_polynomial`

jolt-core/src/subprotocols/sparse_dense_shout.rs
```diff
@@ -247,42 +253,77 @@ fn compute_sumcheck_prover_message<const WORD_SIZE: usize, F: JoltField>(
...
-                    let suffixes_left: Vec<_> =
-                        suffixes.iter().map(|suffix| suffix[b.into()]).collect();
-                    let suffixes_right: Vec<_> = suffixes
-                        .iter()
-                        .map(|suffix| suffix[usize::from(b) + len / 2])
+    let gamma_squared = gamma.square();
+    let mut result0 = (F::zero(), F::zero());
+    let mut result2 = (F::zero(), F::zero());
+
+    rayon::join(
```
moodlezoup: instead of this `rayon::join` let's split `compute_sumcheck_prover_message` into two functions

moodlezoup (commented):
dumping some thoughts here, no need to take all my suggestions

jolt-core/src/poly/prefix_suffix.rs
0xAndoroid: Is there any reason to do this? I had it done like this before, but decided to optimize it, I don't remember if it gave me a reasonable performance gain.

jolt-core/src/poly/prefix_suffix.rs
0xAndoroid: I kind of like this syntactic sugar

jolt-core/src/poly/prefix_suffix.rs
0xAndoroid: Another reason why I like `CachablePolynomial` is that we can just implement it on any polynomial that implements `PolynomialBinding` and `PolynomialEvaluation`, compared to if we need to return `MultilinearPolynomial` from `prefix` method, it would me yet another function (apart from `evaluate`, `bind`, and `sumcheck_evals`) that we need to maintain identical evaluation of.

moodlezoup: yeah I guess we can leave it as is

jolt-core/src/poly/prefix_suffix.rs
0xAndoroid: I've been trying to get any performance out of this, but it doesn't seem to get any benefit. I've been testing different chunk sizes as well (2^10, 2^12, 2^14), as well as tried to return a sparse vector of updates.
It's either same performance, or much worse due to allocation.

PR #749, opened 2025-07-04
fix: Virtual Sequence for rv32imac and rv64imac

zouguangxian:


tracer/src/lib.rs
```diff
@@ -250,8 +250,12 @@ impl Iterator for LazyTraceIterator {
 }
 
 #[tracing::instrument(skip_all)]
-pub fn decode(elf: &[u8]) -> (Vec<RV32IMInstruction>, Vec<(u64, u8)>) {
+pub fn decode(elf: &[u8]) -> (Vec<RV32IMInstruction>, Vec<(u64, u8)>, bool) {
```
moodlezoup:
```suggestion
pub fn decode(elf: &[u8]) -> (Vec<RV32IMInstruction>, Vec<(u64, u8)>, Xlen) {
```

tracer/src/instruction/amomaxd.rs
```diff
@@ -51,4 +69,114 @@ impl AMOMAXD {
...
+        sequence.push(add.into());
+
+        let sd = SD {
+            address: self.address,
+            operands: FormatS {
+                rs1: self.operands.rs1,
+                rs2: v_rs2,
+                imm: 0,
+            },
+            virtual_sequence_remaining: Some(1),
```
moodlezoup: both of these have `virtual_sequence_remaining: Some(1)`

to prevent mistakes like this, I think we should
- add a utility function `fn enumerate_virtual_sequence(mut sequence: Vec<RV32IMInstruction>) -> Vec<RV32IMInstruction>` that populates the `virtual_sequence_remaining` fields for each instruction in the sequence, and/or
- add a test that randomly generates an instruction and checks that its virtual sequence has the correct `virtual_sequence_remaining` values

tracer/src/instruction/mod.rs
```diff
@@ -327,19 +345,19 @@ where
...
-        self.operands()
-            .capture_post_execution_state(&mut cycle.register_state, cpu);
-        if let Some(trace_vec) = trace {
-            trace_vec.push(cycle.into());
-        }
+        //self.operands()
+        //    .capture_post_execution_state(&mut cycle.register_state, cpu);
+        //if let Some(trace_vec) = trace {
+        //    trace_vec.push(cycle.into());
+        //}
```
moodlezoup: why was this commented out?

zouguangxian: @absolute0kelvin

tracer/src/instruction/amomaxd.rs
zouguangxian: I added a test_remaining function in test.rs to check the ordering.

PR #759, opened 2025-07-08
DAG stage 1

markosg04:
Introduces the initial DAG + state manager pattern and the stage 1 integration (spartan propagation)

### DIFF
--- a/jolt-core/src/dag/jolt_dag.rs
+++ b/jolt-core/src/dag/jolt_dag.rs
@@ -1,30 +1,48 @@
+use crate::dag::stage::{StageResult, SumcheckStages};
 use crate::dag::state_manager::StateManager;
-// use crate::dag::stage::SumcheckStages; // TODO: Use when implementing prove/verify
 use crate::field::JoltField;
 use crate::poly::commitment::commitment_scheme::CommitmentScheme;
+use crate::r1cs::spartan::UniformSpartanProof;
 use crate::utils::transcript::Transcript;
 
-pub struct JoltDAG<'a, F: JoltField, ProofTranscript: Transcript, PCS: CommitmentScheme<ProofTranscript, Field = F>> {
+pub struct JoltDAG<
+    'a,
+    F: JoltField,
+    ProofTranscript: Transcript,
+    PCS: CommitmentScheme<ProofTranscript, Field = F>,
+> {
     state_manager: StateManager<'a, F, ProofTranscript, PCS>,
 }
 
-impl<'a, F: JoltField, ProofTranscript: Transcript, PCS: CommitmentScheme<ProofTranscript, Field = F>> JoltDAG<'a, F, ProofTranscript, PCS> {
-    pub fn new(state_manager: StateManager<'a, F, ProofTranscript, PCS>, _transcript: ProofTranscript) -> Self {
-        Self {
-            state_manager
+impl<
+        'a,
+        F: JoltField,
+        ProofTranscript: Transcript,
+        PCS: CommitmentScheme<ProofTranscript, Field = F>,
+    > JoltDAG<'a, F, ProofTranscript, PCS>
+{
+    pub fn new(
+        state_manager: StateManager<'a, F, ProofTranscript, PCS>,
+        _transcript: ProofTranscript,
+    ) -> Self {
+        Self { state_manager }
+    }
+
+    pub fn prove(&mut self) -> Result<(), String> {
+        // Stage 1: Run Spartan's outer sumcheck proof
+        let spartan_proof = UniformSpartanProof::<F, ProofTranscript>::default();
+        match spartan_proof.stage1_prove(&mut self.state_manager) {
+            StageResult::Success => Ok(()),
+            StageResult::Failed(reason) => Err(format!("Stage 1 prove failed: {}", reason)),
         }
     }
 
-
-    pub fn prove(&mut self) {
-
-        // Stage 1: 
-
-        
-        todo!()
-    }
-
-    pub fn verify(&mut self) -> Result<(), crate::utils::errors::ProofVerifyError> {
-        todo!()
+    pub fn verify(&mut self) -> Result<(), String> {
+        // Stage 1: Verify Spartan's outer sumcheck proof
+        let spartan_proof = UniformSpartanProof::<F, ProofTranscript>::default();
+        match spartan_proof.stage1_verify(&mut self.state_manager) {
+            StageResult::Success => Ok(()),
+            StageResult::Failed(reason) => Err(format!("Stage 1 verify failed: {}", reason)),
+        }
     }
 }
--- a/jolt-core/src/dag/jolt_dag.rs
+++ b/jolt-core/src/dag/jolt_dag.rs
@@ -1,4 +1,4 @@
-use crate::dag::stage::{StageResult, SumcheckStages};
+use crate::dag::stage::SumcheckStages;
 use crate::dag::state_manager::StateManager;
 use crate::field::JoltField;
 use crate::poly::commitment::commitment_scheme::CommitmentScheme;
@@ -23,26 +23,21 @@ impl<
 {
     pub fn new(
         state_manager: StateManager<'a, F, ProofTranscript, PCS>,
-        _transcript: ProofTranscript,
     ) -> Self {
         Self { state_manager }
     }
 
-    pub fn prove(&mut self) -> Result<(), String> {
+    pub fn prove(&mut self) -> anyhow::Result<()> {
         // Stage 1: Run Spartan's outer sumcheck proof
         let spartan_proof = UniformSpartanProof::<F, ProofTranscript>::default();
-        match spartan_proof.stage1_prove(&mut self.state_manager) {
-            StageResult::Success => Ok(()),
-            StageResult::Failed(reason) => Err(format!("Stage 1 prove failed: {reason}")),
-        }
+        spartan_proof.stage1_prove(&mut self.state_manager)?;
+        Ok(())
     }
 
-    pub fn verify(&mut self) -> Result<(), String> {
+    pub fn verify(&mut self) -> anyhow::Result<()> {
         // Stage 1: Verify Spartan's outer sumcheck proof
         let spartan_proof = UniformSpartanProof::<F, ProofTranscript>::default();
-        match spartan_proof.stage1_verify(&mut self.state_manager) {
-            StageResult::Success => Ok(()),
-            StageResult::Failed(reason) => Err(format!("Stage 1 verify failed: {reason}")),
-        }
+        spartan_proof.stage1_verify(&mut self.state_manager)?;
+        Ok(())
     }
 }
--- a/jolt-core/src/dag/jolt_dag.rs
+++ b/jolt-core/src/dag/jolt_dag.rs
@@ -21,20 +21,18 @@ impl<
         PCS: CommitmentScheme<ProofTranscript, Field = F>,
     > JoltDAG<'a, F, ProofTranscript, PCS>
 {
-    pub fn new(
-        state_manager: StateManager<'a, F, ProofTranscript, PCS>,
-    ) -> Self {
+    pub fn new(state_manager: StateManager<'a, F, ProofTranscript, PCS>) -> Self {
         Self { state_manager }
     }
 
-    pub fn prove(&mut self) -> anyhow::Result<()> {
+    pub fn prove(&mut self) -> Result<(), anyhow::Error> {
         // Stage 1: Run Spartan's outer sumcheck proof
         let spartan_proof = UniformSpartanProof::<F, ProofTranscript>::default();
         spartan_proof.stage1_prove(&mut self.state_manager)?;
         Ok(())
     }
 
-    pub fn verify(&mut self) -> anyhow::Result<()> {
+    pub fn verify(&mut self) -> Result<(), anyhow::Error> {
         // Stage 1: Verify Spartan's outer sumcheck proof
         let spartan_proof = UniformSpartanProof::<F, ProofTranscript>::default();
         spartan_proof.stage1_verify(&mut self.state_manager)?;
--- a/jolt-core/src/dag/jolt_dag.rs
+++ b/jolt-core/src/dag/jolt_dag.rs
@@ -17,6 +17,10 @@ impl<'a, F: JoltField, ProofTranscript: Transcript, PCS: CommitmentScheme<ProofT
 
 
     pub fn prove(&mut self) {
+
+        // Stage 1: 
+
+        
         todo!()
     }
 
--- a/jolt-core/src/dag/jolt_dag.rs
+++ b/jolt-core/src/dag/jolt_dag.rs
@@ -1,0 +1,81 @@
+use crate::dag::stage::SumcheckStages;
+use crate::dag::state_manager::StateManager;
+use crate::field::JoltField;
+use crate::subprotocols::sumcheck::BatchedSumcheck;
+use crate::utils::transcript::Transcript;
+
+pub struct JoltDAG<'a, F: JoltField, ProofTranscript: Transcript> {
+    state_manager: StateManager<'a, F, ProofTranscript>,
+    transcript: ProofTranscript,
+    registry: Vec<Box<dyn SumcheckStages<F, ProofTranscript> + 'a>>,
+}
+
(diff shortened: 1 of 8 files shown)


jolt-core/src/dag/jolt_dag.rs
```diff
@@ -0,0 +1,48 @@
...
+    ) -> Self {
+        Self { state_manager }
+    }
+
+    pub fn prove(&mut self) -> Result<(), String> {
+        // Stage 1: Run Spartan's outer sumcheck proof
+        let spartan_proof = UniformSpartanProof::<F, ProofTranscript>::default();
+        match spartan_proof.stage1_prove(&mut self.state_manager) {
+            StageResult::Success => Ok(()),
+            StageResult::Failed(reason) => Err(format!("Stage 1 prove failed: {reason}")),
```
0xAndoroid: nit: don't use string as error, maybe we should consider using anyhow, or panic right away since you can't really recover from this

jolt-core/src/dag/stage.rs
```diff
@@ -0,0 +1,110 @@
+use crate::dag::state_manager::StateManager;
+use crate::field::JoltField;
+use crate::poly::commitment::commitment_scheme::CommitmentScheme;
+use crate::subprotocols::sumcheck::BatchableSumcheckInstance;
+use crate::utils::transcript::Transcript;
+
+#[derive(Debug, Clone)]
+pub enum StageResult {
```
0xAndoroid: Why can't we use Result<(), anyhow::Error>?

jolt-core/src/dag/state_manager.rs
```diff
@@ -0,0 +1,195 @@
...
+        &self[&OpeningsKeys::SpartanZ(index)].1
+    }
+}
+
+#[derive(Hash, PartialEq, Eq, Copy, Clone, Debug)]
+pub enum OpeningsKeys {
+    SpartanZ(JoltR1CSInputs),
+    InstructionTypeFlag(usize),
+    InstructionRa(usize),
+    OuterSumcheckClaims, // (Az, Bz, Cz)
```
0xAndoroid: Does this one contain the RLC of the Az Bz Cz for the inner sumcheck? Which one contains the Az, Bz, Cz of the outer sumcheck?

jolt-core/src/dag/state_manager.rs
```diff
@@ -0,0 +1,195 @@
...
+    OuterSumcheckR,      // r_cycle from outer sumcheck
```
0xAndoroid: Outer sumcheck r_cycle should be in the outer_sumcheck_claims as opening point

jolt-core/src/dag/state_manager.rs
```diff
@@ -0,0 +1,195 @@
...
+    OuterSumcheckRxVar,  // rx_var from outer sumcheck
```
0xAndoroid: I don't remember what rx_var is, but it's also an opening point for some claim.

jolt-core/src/dag/state_manager.rs
```diff
@@ -0,0 +1,195 @@
...
+pub type Openings<F> = HashMap<OpeningsKeys, (OpeningPoint<LITTLE_ENDIAN, F>, Option<F>)>;
+
+#[derive(Hash, PartialEq, Eq, Clone, Debug)]
+pub enum ProofKeys {
+    SpartanOuterSumcheck,
+    SpartanInnerSumcheck,
+    SpartanShiftSumcheck,
+    RegistersReadWrite(usize),
+    RamReadWrite(usize),
+    InstructionLookups(usize),
```
0xAndoroid: I don't like the enumeration of sumchecks by usize.

jolt-core/src/dag/state_manager.rs
```diff
@@ -0,0 +1,195 @@
...
+}
+
+pub enum ProofData<F: JoltField, ProofTranscript: Transcript> {
+    Spartan(UniformSpartanProof<F, ProofTranscript>),
+    SpartanSumcheck(SumcheckInstanceProof<F, ProofTranscript>, Vec<F>, [F; 3]),
```
0xAndoroid: What is the `Vec<F>` and `[F; 3]` here? Isn't this the opening points? They should be in the `openings`

jolt-core/src/dag/state_manager.rs
```diff
@@ -0,0 +1,195 @@
...
+}
+
+pub struct StateManager<
+    'a,
+    F: JoltField,
+    ProofTranscript: Transcript,
+    PCS: CommitmentScheme<ProofTranscript, Field = F>,
+> {
+    pub openings: Arc<Mutex<Openings<F>>>,
+    pub prover_accumulator: &'a mut ProverOpeningAccumulator<F, PCS, ProofTranscript>,
```
0xAndoroid: `Option<Arc<Mutex<...>>>`

jolt-core/src/dag/state_manager.rs
```diff
@@ -0,0 +1,195 @@
...
+    pub verifier_accumulator: &'a mut VerifierOpeningAccumulator<F, PCS, ProofTranscript>,
+    pub prover_transcript: RefCell<&'a mut ProofTranscript>,
+    pub verifier_transcript: RefCell<&'a mut ProofTranscript>,
```
0xAndoroid: why do we need verifier_transcript if its the same as prover_transcript?

jolt-core/src/dag/state_manager.rs
```diff
@@ -0,0 +1,195 @@
...
+pub struct StateManager<
+    'a,
+    F: JoltField,
+    ProofTranscript: Transcript,
+    PCS: CommitmentScheme<ProofTranscript, Field = F>,
+> {
+    pub openings: Arc<Mutex<Openings<F>>>,
+    pub prover_accumulator: &'a mut ProverOpeningAccumulator<F, PCS, ProofTranscript>,
+    pub verifier_accumulator: &'a mut VerifierOpeningAccumulator<F, PCS, ProofTranscript>,
+    pub prover_transcript: RefCell<&'a mut ProofTranscript>,
```
0xAndoroid: Should the state manager own the transcript? Or I guess it's just a temp placeholder to not refactor half the codebase.

jolt-core/src/dag/state_manager.rs
```diff
@@ -0,0 +1,195 @@
...
+pub enum OpeningsKeys {
+    SpartanZ(JoltR1CSInputs),
+    InstructionTypeFlag(usize),
+    InstructionRa(usize),
+    OuterSumcheckClaims, // (Az, Bz, Cz)
+    OuterSumcheckR,      // r_cycle from outer sumcheck
+    OuterSumcheckRxVar,  // rx_var from outer sumcheck
+}
+
+pub type Openings<F> = HashMap<OpeningsKeys, (OpeningPoint<LITTLE_ENDIAN, F>, Option<F>)>;
```
0xAndoroid: Why is it `Option<F>`?

jolt-core/src/r1cs/spartan.rs
```diff
@@ -262,6 +267,25 @@ where
         })
     }
 
+    #[tracing::instrument(skip_all)]
```
0xAndoroid, line 272: `prove_spartan_small_value` is already wrapped in `tracnig::instrument`. Why do we need this function at all?

jolt-core/src/r1cs/spartan.rs
```diff
@@ -836,3 +860,141 @@ impl<F: JoltField, ProofTranscript: Transcript> BatchableSumcheckInstance<F, Pro
...
+
+        ProofTranscript::append_scalars(
+            *state_manager.prover_transcript.borrow_mut(),
+            &outer_sumcheck_claims,
+        );
+
+        let opening_point = OpeningPoint::<LITTLE_ENDIAN, F>::new(outer_sumcheck_r.clone());
+        state_manager.openings.lock().unwrap().insert(
+            OpeningsKeys::OuterSumcheckClaims,
+            (opening_point, None), // to be proven using PCS
```
0xAndoroid: Inner sumcheck proves outer sumcheck claim, not PCS

jolt-core/src/r1cs/spartan.rs
```diff
@@ -836,3 +860,141 @@ impl<F: JoltField, ProofTranscript: Transcript> BatchableSumcheckInstance<F, Pro
...
+
+    fn stage1_verify(
+        &self,
+        state_manager: &mut StateManager<'_, F, ProofTranscript, PCS>,
+    ) -> StageResult {
+        let (key, _, _) = state_manager.get_spartan_data();
+
+        let num_rounds_x = key.num_cons_total.log_2();
+
+        let tau: Vec<F> = state_manager
```
0xAndoroid, line 987: We should standardize `gamma` as the name for bespoke batching challenge

jolt-core/src/r1cs/spartan.rs
markosg04: `prove_outer_sumcheck` will die soon anyways it was just help me organize some stuff as i was doing it

jolt-core/src/dag/state_manager.rs
markosg04: I think that it can.. if everything else is going in there then transcript might as well, too. I don't think it matters that much though -- we could split it, it does create some annoying borrow scenarios hence the RefCell

jolt-core/src/dag/stage.rs
```diff
@@ -0,0 +1,94 @@
...
+use crate::subprotocols::sumcheck::BatchableSumcheckInstance;
+use crate::utils::transcript::Transcript;
+
+pub trait SumcheckStages<
+    F: JoltField,
+    ProofTranscript: Transcript,
+    PCS: CommitmentScheme<ProofTranscript, Field = F>,
+>: Send + Sync
+{
+    // Stage 1 is special case of outer sumchec from spartan
```
moodlezoup:
```suggestion
    // Stage 1 is special case of outer sumcheck from spartan
```

jolt-core/src/dag/stage.rs
```diff
@@ -0,0 +1,94 @@
...
+    fn stage1_prove(
+        &self,
+        state_manager: &mut StateManager<'_, F, ProofTranscript, PCS>,
+    ) -> Result<(), anyhow::Error> {
+        let _ = state_manager;
+        Ok(())
+    }
```
moodlezoup: Is this just to appease clippy? if so prefer to do this
```suggestion
    fn stage1_prove(
        &self,
        _state_manager: &mut StateManager<'_, F, ProofTranscript, PCS>,
    ) -> Result<(), anyhow::Error> {
        Ok(())
    }
```

jolt-core/src/dag/state_manager.rs
moodlezoup: I think just a single transcript here would suffice

jolt-core/src/dag/mod.rs
```diff
@@ -0,0 +1,140 @@
...
+            padded_trace_length,
+        );
+
+        // Create input polynomials from trace
+        let mut input_polys = Vec::new();
+        for i in 0..JoltR1CSInputs::num_inputs() {
+            let input = JoltR1CSInputs::from_index(i);
+            let poly = input.generate_witness(&trace, &preprocessing);
+            input_polys.push(poly);
+        }
```
moodlezoup: reading this, I feel like this Spartan state should be handled within Spartan, and instead the state manager would own the `preprocessing`, `trace`, `program_io`, and `final_memory_state`. 
Every stage is going to need at least the preprocessing and trace, and idk where they'd get it from if not the state manager, wdyt?

jolt-core/src/r1cs/spartan.rs
```diff
@@ -836,3 +860,209 @@ impl<F: JoltField, ProofTranscript: Transcript> BatchableSumcheckInstance<F, Pro
...
+            _marker: PhantomData,
+        }
+    }
+}
+
+impl<
+        F: JoltField,
+        ProofTranscript: Transcript,
+        PCS: CommitmentScheme<ProofTranscript, Field = F>,
+    > SumcheckStages<F, ProofTranscript, PCS> for UniformSpartanProof<F, ProofTranscript>
```
moodlezoup: nit: would prefer to impl SumcheckStages on a separate empty struct rather than on the proof struct

jolt-core/src/r1cs/spartan.rs
```diff
@@ -836,3 +860,209 @@ impl<F: JoltField, ProofTranscript: Transcript> BatchableSumcheckInstance<F, Pro
...
+           Proves: \sum_x eq(gamma, x) * (Az(x) * Bz(x) - Cz(x)) = 0
+
+           The matrices A, B, C have a block-diagonal structure with repeated blocks
+           A_small, B_small, C_small corresponding to the uniform constraints.
+        */
+        let (key, constraint_builder, input_polys) = state_manager.get_spartan_data();
+
+        let num_rounds_x = key.num_cons_total.log_2();
+
+        let gamma: Vec<F> = state_manager
```
moodlezoup: nit: we were using `tau` for this variable to match the Spartan paper

jolt-core/src/r1cs/spartan.rs
```diff
@@ -836,3 +860,209 @@ impl<F: JoltField, ProofTranscript: Transcript> BatchableSumcheckInstance<F, Pro
...
+    fn stage1_verify(
+        &self,
+        state_manager: &mut StateManager<'_, F, ProofTranscript, PCS>,
+    ) -> Result<(), anyhow::Error> {
+        // Get the spartan-related data:
+        let (key, _, _) = state_manager.get_spartan_data();
+
+        let num_rounds_x = key.num_cons_total.log_2();
+
+        let gamma: Vec<F> = state_manager
```
moodlezoup: tau

jolt-core/src/r1cs/spartan.rs
```diff
@@ -836,3 +860,209 @@ impl<F: JoltField, ProofTranscript: Transcript> BatchableSumcheckInstance<F, Pro
...
+                Ok(result) => result,
+                Err(_) => return Err(anyhow::anyhow!("Outer sumcheck verification failed")),
+            }
+        };
+
+        // Outer sumcheck is bound from the top, reverse the challenge TODO(markosg04 make use of Endianness here ?)
+        let outer_sumcheck_r_reversed: Vec<F> =
+            outer_sumcheck_r_original.into_iter().rev().collect();
+
+        let gamma_bound_rx = EqPolynomial::mle(&gamma, &outer_sumcheck_r_reversed);
```
moodlezoup:
```suggestion
        let eq_tau_rx = EqPolynomial::mle(&gamma, &outer_sumcheck_r_reversed);
```

jolt-core/src/r1cs/spartan.rs
markosg04: i thought we standardized gamma 😢

jolt-core/src/dag/mod.rs
markosg04: it is probably cleaner than this type of setup, yes.

jolt-core/src/r1cs/spartan.rs
moodlezoup: we decided on gamma for taking a random linear combination of multiple sumcheck instances; this is different

jolt-core/src/dag/jolt_dag.rs
```diff
@@ -0,0 +1,45 @@
...
+use crate::r1cs::spartan::SpartanDag;
+use crate::utils::transcript::Transcript;
+pub struct JoltDAG<
+    'a,
+    F: JoltField,
+    ProofTranscript: Transcript,
+    PCS: CommitmentScheme<ProofTranscript, Field = F>,
+> {
+    prover_state_manager: StateManager<'a, F, ProofTranscript, PCS>,
+    verifier_state_manager: StateManager<'a, F, ProofTranscript, PCS>,
```
moodlezoup, line 14: I think either we have a separate ProverStateManager and VerifierStateManager and these fields are Options, or these should be a single field (in practice you'd never need both at the same time)

I'm leaning towards the former, given that we already have `prover_accumulator` and `verifier_accumulator`, plus `ProgramState` is prover-only (though the verifier will need `program_io` and `JoltVerifierPreprocessing`)

moodlezoup (approved):

PR #763, opened 2025-07-10
Feat/batchable opening accumulation

markosg04:
wip....


markosg04:
we went with this approach instead: https://github.com/a16z/jolt/pull/763/files

PR #768, opened 2025-07-11
stage 2 and 3 spartan

markosg04:
- adds the stage 2 and stage 3 of the DAG for spartan
- refactors to the state manager API for claim/opening handling

### DIFF
--- a/jolt-core/src/jolt/vm/instruction_lookups/read_raf_checking.rs
+++ b/jolt-core/src/jolt/vm/instruction_lookups/read_raf_checking.rs
@@ -7,7 +7,7 @@ use tracer::instruction::RV32IMCycle;
 use super::{D, K_CHUNK, LOG_K, LOG_K_CHUNK, LOG_M, M, PHASES, RA_PER_LOG_M, WORD_SIZE};
 
 use crate::{
-    dag::state_manager::{Openings, OpeningsKeys, StateManager},
+    dag::state_manager::StateManager,
     field::JoltField,
     jolt::{
         instruction::{InstructionFlags, InstructionLookup, InterleavedBitsMarker, LookupQuery},
@@ -24,7 +24,7 @@ use crate::{
         multilinear_polynomial::{
             BindingOrder, MultilinearPolynomial, PolynomialBinding, PolynomialEvaluation,
         },
-        opening_proof::ProverOpeningAccumulator,
+        opening_proof::{Openings, OpeningsExt, OpeningsKeys, ProverOpeningAccumulator},
         prefix_suffix::{Prefix, PrefixRegistry, PrefixSuffixDecomposition},
     },
     r1cs::inputs::JoltR1CSInputs,
@@ -95,8 +95,12 @@ impl<'a, F: JoltField> ReadRafSumcheck<'a, F> {
         let mut ps = ReadRafProverState::new(trace, eq_r_cycle, unbound_ra_polys);
         ps.init_phase(0);
         let r_cycle = sm
-            .openings_point(OpeningsKeys::SpartanZ(JoltR1CSInputs::LookupOutput))
-            .r;
+            .get_prover_accumulator()
+            .borrow()
+            .get_opening_point(OpeningsKeys::SpartanZ(JoltR1CSInputs::LookupOutput))
+            .unwrap()
+            .r
+            .clone();
         Self {
             gamma,
             gamma_squared: gamma.square(),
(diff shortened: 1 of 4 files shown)


jolt-core/src/dag/state_manager.rs
```diff
@@ -1,120 +1,25 @@
...
-            .unwrap_or(F::zero())
-    }
-}
-
 #[derive(Hash, PartialEq, Eq, Clone, Debug)]
 pub enum ProofKeys {
     SpartanOuterSumcheck,
     SpartanInnerSumcheck,
     SpartanShiftSumcheck,
+    SpartanStage2Sumcheck,
```
0xAndoroid: Shouldn't we have just five proof keys - one per stage?

jolt-core/src/dag/jolt_dag.rs
```diff
@@ -24,14 +30,214 @@ impl<'a, F: JoltField, ProofTranscript: Transcript, PCS: CommitmentScheme<Field
...
+        let padded_trace_length = trace.len().next_power_of_two();
+        let spartan_dag = SpartanDag::<F>::new::<ProofTranscript>(padded_trace_length);
         spartan_dag.stage1_prove(&mut self.prover_state_manager)?;
+
+        // Stage 2:
+        let mut stage2_instances =
+            spartan_dag.stage2_prover_instances(&mut self.prover_state_manager);
+        let stage2_instances_mut: Vec<&mut dyn BatchableSumcheckInstance<F>> = stage2_instances
+            .iter_mut()
+            .map(|instance| &mut **instance as &mut dyn BatchableSumcheckInstance<F>)
```
moodlezoup, line 54: can we not do this?
```suggestion
            .map(|instance| *instance as &mut dyn BatchableSumcheckInstance<F>)
```

jolt-core/src/dag/jolt_dag.rs
```diff
@@ -24,14 +30,214 @@ impl<'a, F: JoltField, ProofTranscript: Transcript, PCS: CommitmentScheme<Field
...
+        let K = [
+            preprocessing.shared.bytecode.code_size,
+            ram_K,
+            1 << 16, // K for instruction lookups Shout
+        ]
+        .into_iter()
+        .max()
+        .unwrap();
+
+        let _guard = crate::poly::commitment::dory::DoryGlobals::initialize(K, padded_trace_length);
```
moodlezoup, line 223: we'll want to move this to the top of `prove`, because we tear down the statics when `_guard` is dropped

jolt-core/src/dag/jolt_dag.rs
```diff
@@ -24,14 +30,214 @@ impl<'a, F: JoltField, ProofTranscript: Transcript, PCS: CommitmentScheme<Field
...
+
+        let accumulator = self.verifier_state_manager.get_verifier_accumulator();
+        for instance in stage3_instances.iter_mut() {
+            instance.cache_openings_verifier(Some(accumulator.clone()));
+        }
+
+        Ok(())
+    }
+
+    fn receive_claims(&mut self) -> Result<(), anyhow::Error> {
```
moodlezoup, line 176: We can't send the whole prover opening accumulator to the verifier because it's way too big. We'll need the prover to serialize just the claims (as part of the `JoltProof` struct) and the verifier can populate its opening accumulator from that

jolt-core/src/jolt/vm/instruction_lookups/read_raf_checking.rs
```diff
@@ -412,27 +454,30 @@ impl<F: JoltField, PCS: CommitmentScheme<Field = F>> CacheSumcheckOpenings<F, PC
...
+                claim,
             );
         });
         ra_claims.iter().enumerate().for_each(|(i, claim)| {
-            openings.insert(
+            accumulator.borrow_mut().append_virtual(
                 OpeningsKeys::InstructionRa(i),
-                (ps.r.to_vec().into(), *claim),
+                ps.r.to_vec(),
+                *claim,
```
moodlezoup, line 469: note that these are not virtual

jolt-core/src/r1cs/spartan.rs
```diff
@@ -1057,12 +1123,152 @@ impl<F: JoltField, ProofTranscript: Transcript, PCS: CommitmentScheme<Field = F>
...
+            .borrow()
+            .get_opening_point(OuterSumcheckAz)
+            .unwrap();
+        let (r_cycle, _rx_var) = outer_sumcheck_r.split_at(num_cycles_bits);
+
+        // Evaluate all witness polynomials P_i at r_cycle for the verifier
+        // Verifier computes: z(r_inner, r_cycle) = Σ_i eq(r_inner, i) * P_i(r_cycle)
+        let flattened_polys_ref: Vec<_> = input_polys.iter().collect();
+        let (claimed_witness_evals, _) =
+            MultilinearPolynomial::batch_evaluate(&flattened_polys_ref, r_cycle);
```
moodlezoup, line 1201: the PC sumcheck really only needs two of the input polys (PC and UnexpandedPC) and two of the claims (NextPC and NextUnexpandedPC), which were added to the opening accumulator in stage 1

jolt-core/src/jolt/vm/instruction_lookups/read_raf_checking.rs
0xAndoroid: I'll fix this when integrating instrucitons

jolt-core/src/dag/jolt_dag.rs
markosg04: compiler complains about a non-primitive cast in the expression

jolt-core/src/dag/jolt_dag.rs
markosg04: I updated the implementation such that it borrows the accumulator and only copies the claim values over, so it shouldn't be too big unless I misunderstand. I know eventually we will want to do the serialization of the `JoltProof` but for now I thought this is equivalent in cost

PR #770, opened 2025-07-12
Draft: Apply Gruens to phase 2 of RAM and register sumchecks

protoben:
This PR is an attempt to create a version of `GruenSplitEqPolynomial` that uses high-to-low binding order and apply it to phase 2 of the RAM and register sumchecks. **Note:** I'm not currently getting an equivalent result.

You currently need to pick a binding order in the constructor, meaning that we would need to maintain two split polynomials in the RAM and register sumchecks, which isn't ideal. However, my hope is that implementing it this way will make it easy to swap in a version of `GruenSplitEqPolynomial` with switchable binding order, if we can implement it.

Current progress:
- Changing the binding order of `GruenSplitEqPolynomial` is implemented. There's a (passing) test for equivalence to binding `DensePolynomial` in high-to-low order [here](https://github.com/GaloisInc/jolt/blob/0788f79ee4eef589d3414d99fb45b850e0d9c252/jolt-core/src/poly/split_eq_poly.rs#L411).
- I've attempted to apply the high-to-low poly to the RAM sumcheck phase 2 [here](https://github.com/GaloisInc/jolt/blob/0788f79ee4eef589d3414d99fb45b850e0d9c252/jolt-core/src/jolt/vm/ram_read_write_checking.rs#L981). This isn't currently working, however.

### DIFF
--- a/jolt-core/src/jolt/vm/ram_read_write_checking.rs
+++ b/jolt-core/src/jolt/vm/ram_read_write_checking.rs
@@ -50,37 +50,37 @@ struct DataBuffers<F: JoltField> {
     dirty_indices: Vec<usize>,
 }
 
-enum MaybeSplitEq<F: JoltField> {
-    Normal(MultilinearPolynomial<F>),
-    Split(GruenSplitEqPolynomial<F>),
+enum HighOrLowSplitEq<F: JoltField> {
+    HighToLow(MultilinearPolynomial<F>),
+    LowToHigh(GruenSplitEqPolynomial<F>),
 }
 
-impl<F: JoltField> MaybeSplitEq<F> {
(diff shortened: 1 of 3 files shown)


protoben:
OK. I see at least two thing wrong with what I did here:
- I was binding the high-to-low polynomial high-to-low from the start, rather than starting with the paritally low-to-high bound coefficients, which is clearly wrong.
- Looking closer at the `GruenSplitEqPolynomial::merge` function, it looks like the `bind_high_to_low` test I added doesn't test the `E_in` and `E_out` vectors, just the `w` vector, so I suspect I'm doing the binding wrong.

I'll go ahead and try to fix those.

markosg04:
Thank you for the initial work on the HighToLow, it later got completed in https://github.com/a16z/jolt/pull/802

PR #772, opened 2025-07-13
Wire spartan, registers, and more fixes

markosg04:
- addresses comments from https://github.com/a16z/jolt/pull/768
- fully wires the registers to spartan


jolt-core/src/jolt/vm/registers_read_write_checking.rs
```diff
@@ -978,20 +1084,219 @@ where
...
+            inc_claim,
         });
+
+        // Append claims to accumulator if provided
+        if let Some(accumulator) = accumulator {
+            // Get the sumcheck opening point from the accumulator
+            // We can use any of the spartan z openings since they all share the same opening point
+            let r_sumcheck: Vec<F> = accumulator
+                .borrow()
+                .get_opening_point(OpeningsKeys::SpartanZ(JoltR1CSInputs::Rs1Value))
```
moodlezoup: I don't think this is the bug we're looking for but this is wrong

jolt-core/src/jolt/vm/registers_read_write_checking.rs
```diff
@@ -359,6 +401,36 @@ impl<F: JoltField> RegistersReadWriteChecking<F> {
...
+        value_claims: ReadWriteValueClaims<F>,
+        trace_length: usize,
+        chunk_size: usize,
+        sumcheck_claims: ReadWriteSumcheckClaims<F>,
+    ) -> Self {
+        let T = trace_length;
+        let z = transcript.challenge_scalar();
+
+        let verifier_state = ReadWriteCheckingVerifierState {
+            sumcheck_switch_index: chunk_size.log_2(),
```
moodlezoup, line 416: Note that the prover will need to send the verifier the `sumcheck_switch_index` in the proof. What you have now works in tests because the prover and verifier are running in the same environment, but in practice there's no guarantee that `rayon::current_num_threads()` will be the same between prover and verifier

PR #777, opened 2025-07-14
DAG Proof + Serialization

markosg04:
- Adds a jolt DAG proof, From/Into traits, and all serialization components

### DIFF
--- a/jolt-core/src/dag/jolt_dag.rs
+++ b/jolt-core/src/dag/jolt_dag.rs
@@ -1,113 +1,170 @@
 use crate::dag::stage::SumcheckStages;
-use crate::dag::state_manager::{ProofData, ProofKeys, StateManager};
+use crate::dag::state_manager::{ProofData, ProofKeys, Proofs, StateManager};
 use crate::field::JoltField;
 use crate::jolt::vm::instruction_lookups::LookupsDag;
 use crate::jolt::vm::registers::RegistersDag;
 use crate::jolt::vm::JoltCommitments;
 use crate::jolt::witness::ALL_COMMITTED_POLYNOMIALS;
 use crate::poly::commitment::commitment_scheme::CommitmentScheme;
-use crate::poly::opening_proof::OpeningPoint;
-use crate::poly::opening_proof::LITTLE_ENDIAN;
 use crate::r1cs::spartan::SpartanDag;
 use crate::subprotocols::sumcheck::{BatchableSumcheckInstance, BatchedSumcheck};
 use crate::utils::transcript::Transcript;
+use ark_serialize::{CanonicalDeserialize, CanonicalSerialize};
 use rayon::prelude::*;
 pub struct JoltDAG<'a, F: JoltField, ProofTranscript: Transcript, PCS: CommitmentScheme<Field = F>>
 {
-    prover_state_manager: StateManager<'a, F, ProofTranscript, PCS>,
-    verifier_state_manager: StateManager<'a, F, ProofTranscript, PCS>,
+    prover_state_manager: Option<StateManager<'a, F, ProofTranscript, PCS>>,
+    verifier_state_manager: Option<StateManager<'a, F, ProofTranscript, PCS>>,
+}
+
+#[derive(CanonicalSerialize, CanonicalDeserialize, Debug, Clone)]
+pub struct JoltDagProof<const WORD_SIZE: usize, F, PCS, ProofTranscript>
+where
+    F: JoltField,
+    PCS: CommitmentScheme<Field = F>,
+    ProofTranscript: Transcript
+{
+    pub verifier_preprocessing: crate::jolt::vm::JoltVerifierPreprocessing<F, PCS>,
+    pub program_io: tracer::JoltDevice,
+    pub trace_length: usize,
+    pub sumcheck_switch_index_registers: usize,
+    pub sumcheck_switch_index_ram: usize,
+    
+    pub commitments: JoltCommitments<F, PCS>,
+    pub dag_proofs: Proofs<F, ProofTranscript>,
+    pub claims: crate::dag::state_manager::Claims<F>,
 }
 
 impl<'a, F: JoltField, ProofTranscript: Transcript, PCS: CommitmentScheme<Field = F>>
     JoltDAG<'a, F, ProofTranscript, PCS>
 {
-    pub fn new(
-        prover_state_manager: StateManager<'a, F, ProofTranscript, PCS>,
-        verifier_state_manager: StateManager<'a, F, ProofTranscript, PCS>,
-    ) -> Self {
+    pub fn new() -> Self {
         Self {
-            prover_state_manager,
-            verifier_state_manager,
+            prover_state_manager: None,
+            verifier_state_manager: None,
         }
     }
 
-    pub fn prove(&mut self) -> Result<(), anyhow::Error> {
-        // Initialize DoryGlobals at the beginning to keep it alive for the entire proof
-        let (preprocessing, trace, _, _) = self.prover_state_manager.get_prover_data();
-        let trace_length = trace.len();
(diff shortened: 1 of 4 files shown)


jolt-core/src/dag/jolt_dag.rs
```diff
@@ -1,79 +1,124 @@
...
 use crate::utils::transcript::Transcript;
+use ark_serialize::{CanonicalDeserialize, CanonicalSerialize};
 use rayon::prelude::*;
-pub struct JoltDAG<'a, F: JoltField, ProofTranscript: Transcript, PCS: CommitmentScheme<Field = F>>
+use std::sync::Arc;
+#[derive(Default)]
+pub struct JoltDAG;
+
+#[derive(CanonicalSerialize, CanonicalDeserialize, Debug, Clone)]
+pub struct JoltDagProof<const WORD_SIZE: usize, F, PCS, ProofTranscript>
```
moodlezoup, line 23:
```suggestion
pub struct JoltDagProof<F, PCS, ProofTranscript>
```

jolt-core/src/dag/jolt_dag.rs
```diff
@@ -1,79 +1,124 @@
...
+    pub generators: PCS::ProverSetup,
+}
 
-    pub fn prove(&mut self) -> Result<(), anyhow::Error> {
-        // Initialize DoryGlobals at the beginning to keep it alive for the entire proof
-        let (preprocessing, trace, _, _) = self.prover_state_manager.get_prover_data();
+impl JoltDAG {
+    pub fn prove<
+        'a,
+        const WORD_SIZE: usize,
```
moodlezoup, line 54:
```suggestion
```

jolt-core/src/dag/jolt_dag.rs
```diff
@@ -1,79 +1,124 @@
...
+                        &preprocessing.shared.memory_layout,
+                    ) as usize
+                })
+                .collect();
+            let ram_K = ram_addresses.par_iter().max().unwrap().next_power_of_two();
+
+            let K = [
+                preprocessing.shared.bytecode.code_size,
+                ram_K,
+                1 << 16, // K for instruction lookups
```
moodlezoup, line 81:
```suggestion
                1 << 8, // K for instruction lookups
```

jolt-core/src/dag/jolt_dag.rs
```diff
@@ -1,79 +1,124 @@
...
-            .par_iter()
-            .map(|cycle| {
-                crate::jolt::vm::ram::remap_address(
-                    cycle.ram_access().address() as u64,
-                    &preprocessing.shared.memory_layout,
-                ) as usize
-            })
-            .collect();
-        let ram_K = ram_addresses.par_iter().max().unwrap().next_power_of_two();
+        let _guard = {
```
moodlezoup, line 66: I don't think we need to create a new scope here

jolt-core/src/dag/jolt_dag.rs
```diff
@@ -1,79 +1,124 @@
...
-        let K = [
-            // preprocessing.shared.bytecode.code_size,
-            // ram_K,
-            1 << 8, // K for instruction lookups
-        ]
-        .into_iter()
-        .max()
-        .unwrap();
+        // Generate and commit to all witness polynomials
+        {
```
moodlezoup, line 100: same here

jolt-core/src/dag/jolt_dag.rs
```diff
@@ -189,71 +233,75 @@ impl<'a, F: JoltField, ProofTranscript: Transcript, PCS: CommitmentScheme<Field
...
         );
 
-        Ok(())
+        // Convert the state manager to a proof
+        Ok(JoltDagProof::from(&*prover_state_manager))
     }
 
-    pub fn verify(&mut self) -> Result<(), anyhow::Error> {
+    pub fn verify<
+        const WORD_SIZE: usize,
```
moodlezoup, line 252:
```suggestion
```

jolt-core/src/dag/jolt_dag.rs
```diff
@@ -189,71 +233,75 @@ impl<'a, F: JoltField, ProofTranscript: Transcript, PCS: CommitmentScheme<Field
...
-                .prover_state_manager
-                .get_prover_accumulator()
-                .borrow()
-                .clone();
-            let (prover_preprocessing, _, _, _) = self.prover_state_manager.get_prover_data();
-            self.verifier_state_manager
+            // Compare transcripts
+            let verifier_transcript = verifier_state_manager.get_transcript().borrow().clone();
+            let mut prover_transcript = prover_data.transcript;
+            prover_transcript.compare_to(verifier_transcript);
```
moodlezoup, line 272: This is flipped, should be `verifier_transcript.compare_to(prover_transcript)`
Also we need to move this to be before we append the commitments to the transcript

jolt-core/src/dag/proof_serialization.rs
```diff
@@ -0,0 +1,823 @@
...
+        )
+    }
+
+    fn deserialize_uncompressed_unchecked<R: Read>(reader: R) -> Result<Self, SerializationError> {
+        Self::deserialize_with_mode(
+            reader,
+            ark_serialize::Compress::No,
+            ark_serialize::Validate::No,
+        )
+    }
```
moodlezoup, line 158: aren't there default implementations for these?

jolt-core/src/dag/proof_serialization.rs
```diff
@@ -0,0 +1,823 @@
...
+
+        for (key, (_, value)) in prover_acc_borrow.evaluation_openings().iter() {
+            claims_map.insert(*key, *value);
+        }
+
+        let claims = Claims(claims_map);
+
+        // TODO(markosg04) get the real values for these and update ram and registers to use these
+        let sumcheck_switch_index_registers = 0;
+        let sumcheck_switch_index_ram = 0;
```
moodlezoup, line 759: let's just do this now before we forget

jolt-core/src/dag/proof_serialization.rs
```diff
@@ -0,0 +1,823 @@
...
+
+// Conversion from JoltDagProof to StateManager (with verifier state)
+impl<'a, const WORD_SIZE: usize, F, PCS, ProofTranscript>
+    Into<StateManager<'a, F, ProofTranscript, PCS>>
+    for JoltDagProof<WORD_SIZE, F, PCS, ProofTranscript>
+where
+    F: JoltField,
+    PCS: CommitmentScheme<Field = F>,
+    ProofTranscript: Transcript,
+{
```
moodlezoup, line 782: Let's remove `verifier_preprocessing` from the proof struct and add an `from_proof(proof: JoltDagProof, preprocessing: JoltVerifierPreprocessing)` method on `StateManager` instead of implementing `Into` here

jolt-core/src/dag/proof_serialization.rs
```diff
@@ -0,0 +1,823 @@
...
+// Conversion from StateManager (with prover state) to JoltDagProof
+impl<'a, const WORD_SIZE: usize, F, PCS, ProofTranscript>
+    From<&StateManager<'a, F, ProofTranscript, PCS>>
+    for JoltDagProof<WORD_SIZE, F, PCS, ProofTranscript>
+where
+    F: JoltField,
+    PCS: CommitmentScheme<Field = F>,
+    ProofTranscript: Transcript,
+{
+    fn from(state_manager: &StateManager<'a, F, ProofTranscript, PCS>) -> Self {
```
moodlezoup, line 734: Should we do this conversion by value instead of by reference? afaict this serialization should be the last thing that happens for the prover, so we shouldn't need the state manager after that. would allow us to replace some of the clones below with `take` (though it's a low-order cost)
```suggestion
    fn from(state_manager: StateManager<'a, F, ProofTranscript, PCS>) -> Self {
```

PR #782, opened 2025-07-16
Various minor optimizations

moodlezoup:
See comments below

### DIFF
--- a/jolt-core/src/dag/jolt_dag.rs
+++ b/jolt-core/src/dag/jolt_dag.rs
@@ -394,7 +394,7 @@ impl<'a, F: JoltField, ProofTranscript: Transcript, PCS: CommitmentScheme<Field 
             .collect();
 
         let commitments: Vec<_> = committed_polys
-            .par_iter()
+            .iter()
             .map(|poly| PCS::commit(poly, &preprocessing.generators))
             .collect();
 
--- a/jolt-core/src/poly/multilinear_polynomial.rs
+++ b/jolt-core/src/poly/multilinear_polynomial.rs
@@ -377,6 +377,45 @@ impl<F: JoltField> MultilinearPolynomial<F> {
             _ => unimplemented!("Unexpected MultilinearPolynomial variant"),
         }
     }
+
+    #[inline]
+    pub fn sumcheck_evals_array<const DEGREE: usize>(
+        &self,
+        index: usize,
+        order: BindingOrder,
+    ) -> [F; DEGREE] {
+        debug_assert!(DEGREE > 0);
+        debug_assert!(index < self.len() / 2);
+
+        let mut evals = [F::zero(); DEGREE];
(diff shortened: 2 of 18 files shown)


jolt-core/src/dag/jolt_dag.rs
```diff
@@ -394,7 +394,7 @@ impl<'a, F: JoltField, ProofTranscript: Transcript, PCS: CommitmentScheme<Field
             .collect();
 
         let commitments: Vec<_> = committed_polys
-            .par_iter()
+            .iter()
```
moodlezoup, line 397: this is slightly slower but it makes the flamegraphs more interpretable

jolt-core/src/poly/opening_proof.rs
```diff
@@ -203,27 +204,32 @@ pub struct DensePolynomialProverOpening<F: JoltField> {
...
+        let eq_poly = &self.eq_poly.borrow().eq_poly;
         let polynomial = &self.polynomial;
         let mle_half = polynomial.len() / 2;
         let eval_0: F = (0..mle_half)
-            .map(|i| polynomial.get_bound_coeff(i) * shared_eq.eq_poly[i])
+            .into_par_iter()
+            .map(|i| polynomial.get_bound_coeff(i) * eq_poly[i])
             .sum();
         let eval_2: F = (0..mle_half)
+            .into_par_iter()
```
moodlezoup, line 220: optimization 1: missing `into_par_iter` here

jolt-core/src/poly/multilinear_polynomial.rs
```diff
@@ -377,6 +377,45 @@ impl<F: JoltField> MultilinearPolynomial<F> {
             _ => unimplemented!("Unexpected MultilinearPolynomial variant"),
         }
     }
+
+    #[inline]
+    pub fn sumcheck_evals_array<const DEGREE: usize>(
```
moodlezoup, line 382: optimization 2: array version of `sumcheck_evals`

jolt-core/src/poly/one_hot_polynomial.rs
```diff
@@ -307,34 +307,63 @@ impl<F: JoltField> OneHotPolynomial<F> {
         let row_len = DoryGlobals::get_num_columns();
         let T = DoryGlobals::get_T();
 
-        let num_chunks = rayon::current_num_threads().next_power_of_two();
-        let chunk_size = std::cmp::max(1, num_rows / num_chunks);
+        let rows_per_k = T / row_len;
+        if rows_per_k >= rayon::current_num_threads() {
+            // This is the typical case (T >> K)
```
moodlezoup, line 312: optimization 3: speed up row commitment for one-hot polynomials by optimizing for the common case where T >> K. In this case, we compute the row commitments with just a single parallelized pass over the trace

PR #786, opened 2025-07-22
Refactor/append ~all openings

moodlezoup:
Refactors opening accumulator to defer the computation of any random linear combinations (of polynomials, of commitments) until the the very end (i.e. `reduce_and_prove` or `reduce_and_verify`). This allows us to remove `PCS` as a type parameter on the accumulator structs, which in turn allows us to clean up a bunch of stuff. 

This PR also includes the following:
- implements `normalize_opening_point` everywhere 
- Adds the verifier opening accumulator as a parameter of the `expected_output_claim` method (TODO: do the same for `input_claim` too?)
- Fixes some issues related to dynamically handling the d parameter for RAM
- Correctly(?) appends ~all openings to the opening accumulators in the relevant `cache_openings_*` methods. One exception is that we are not proving what is currently called the `raf_flag` claim, which is really a sum of claims for `CircuitFlags::AddOperands`, `CircuitFlags::AddOperands`, and `CircuitFlags::SubtractOperands` (cc @0xAndoroid)


jolt-core/src/jolt/vm/bytecode/read_checking.rs
```diff
@@ -365,60 +412,61 @@ impl<F: JoltField> BatchableSumcheckInstance<F> for ReadCheckingSumcheck<F> {
...
+
         let r: Vec<_> = r.iter().rev().copied().collect();
 
         // Verify sumcheck_claim = ra_claim * val_eval
-        *ra_claim * self.val_poly.evaluate(&r)
+        ra_claim * self.val_poly.evaluate(&r)
+    }
+
+    fn normalize_opening_point(&self, opening_point: &[F]) -> OpeningPoint<BIG_ENDIAN, F> {
+        OpeningPoint::new(opening_point.iter().copied().rev().collect())
```
0xAndoroid: Should only the r_cycle be reversed since we are binding it LowToHigh, but r_address is HighToLow?

0xAndoroid: Nvm, it's LowToHigh in both if d=1

Issue #741, opened 2025-07-02
Prover and/or Verifier Fail on Correct Program for Specific Arguments


DanielHoffmann91:
### Description

I have split this issue into 2 problems, as the second was revealed while debugging the first.

#### Problem 1

I am unable to verify the rust program listed in the source section below for certain arguments. For small arguments, e.g. `1`, `2` or `3`, everything behaves as expected. However, for bigger values with the LSB set, like `4000000001`, the prover succeeds, but the verifier fails, which could be a completeness issue.

```
$> cargo run --release --a 4000000001
 ...
Trace length: 193 (2^8)
output = 1

thread 'main' panicked at src/main.rs:27:9:
verifier failed!
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

#### Problem 2

When replacing the inline assembly with the "equivalent" rust code, ` (a & 1) % a`, I do not even get a proof and it fails with following error:

```
$> cargo run --release --a 4000000000
 ...
thread 'main' panicked at /root/jolt/jolt-core/src/jolt/vm/read_write_memory.rs:283:19:
index out of bounds: the len is 4096 but the index is 4096
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Note how in this case it is no longer relevant if the LSB is set or not. 😕 

### Sources

**guest/src/lib.rs** -- Version 1
```rust
#![cfg_attr(feature = "guest", no_std)]

#[jolt::provable]
fn test(a: u32) -> u32 {
    let result: u32;
    unsafe {
        core::arch::asm!(
            "andi {r}, {a}, 1",
            "rem {r}, {r}, {a}",
            r = out(reg) result,
            a = in(reg) a,
        );
    }
    result
}
```

**guest/src/lib.rs** -- Version 2
```rust
#![cfg_attr(feature = "guest", no_std)]

#[jolt::provable]
fn test(a: u32) -> u32 {
    (a & 1) % a
}
```

**src/main.rs**
```rust
use clap::Parser;

#[derive(Parser, Debug)]
#[clap(author, version, about, long_about = None)]
struct Args {
    #[clap(long)]
    a: u32,
}

pub fn main() {
    let args = Args::parse();
    let a = args.a;

    let target_dir = "/tmp/jolt-guest-targets";
    let program = guest::compile_test(target_dir);

    let prover_preprocessing = guest::preprocess_prover_test(&program);
    let verifier_preprocessing = guest::preprocess_verifier_test(&program);

[...]

PR #790, opened 2025-07-23
fix: prevent bounds error when bytecode is at higher address

SadPencil:
I also encountered the 2nd problem mentioned in https://github.com/a16z/jolt/issues/741 today.

Explanation:
> @moodlezoup We're running into the index out of bounds error when the guest bytecode is at a higher memory address than is ever accessed during the execution. Should be an easy fix.

-----

Before this PR:

(some debug outputs here:)
> max_trace_address: **3589**
> memory_size: **4096**
> REGISTER_COUNT: 64
> preprocessing.min_bytecode_address: 2147483648
> preprocessing.max_bytecode_address: 2147484086
> memory_layout.input_start: 2147467520
> v_init_index starts from: **4096** <---- out of bounds here!

error message here:
```
thread 'main' panicked at /home/ubuntu/jolt/jolt-core/src/jolt/vm/read_write_memory.rs:292:19:
index out of bounds: the len is 4096 but the index is 4096
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

After this PR:

(some debug outputs here:)
> max_trace_address: 4205
> memory_size: 8192
> REGISTER_COUNT: 64
> preprocessing.min_bytecode_address: 2147483648
> preprocessing.max_bytecode_address: 2147484086
> memory_layout.input_start: 2147467520
> v_init_index starts from: 4096

success message:
```
Prover runtime: 1.769510363 s
output: 3
valid: true
```

### DIFF
--- a/jolt-core/src/jolt/vm/read_write_memory.rs
+++ b/jolt-core/src/jolt/vm/read_write_memory.rs
@@ -37,6 +37,7 @@ use super::{JoltPolynomials, JoltStuff, JoltTraceStep};
 #[derive(Clone, CanonicalSerialize, CanonicalDeserialize)]
 pub struct ReadWriteMemoryPreprocessing {
     min_bytecode_address: u64,
+    max_bytecode_address: u64,
     bytecode_words: Vec<u32>,
     // HACK: The verifier will populate this field by copying inputs/outputs from the
     // `ReadWriteMemoryProof` and the memory layout from preprocessing.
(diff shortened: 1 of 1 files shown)


SadPencil:
@moodlezoup Sorry to bother but how can I request a code review for this PR?

moodlezoup:
lgtm, thanks!

PR #794, opened 2025-07-24
Optimize/reuse dory row commitments

moodlezoup:
In a Dory opening proof, we need the row commitments for the polynomial being opened. Prior to this PR, our batched opening proof protocol would provide the RLC polynomial to Dory and we'd compute the row commitments for this RLC polynomial from scratch. However, it's more efficient to cache the row commitments to the individual polynomials (which we need to compute anyway) and compute the RLC of the row commitments.

### DIFF
--- a/jolt-core/src/poly/commitment/commitment_scheme.rs
+++ b/jolt-core/src/poly/commitment/commitment_scheme.rs
@@ -24,6 +24,9 @@ pub trait CommitmentScheme: Clone + Sync + Send + 'static {
         + Clone;
     type Proof: Sync + Send + CanonicalSerialize + CanonicalDeserialize + Clone + Debug;
     type BatchedProof: Sync + Send + CanonicalSerialize + CanonicalDeserialize;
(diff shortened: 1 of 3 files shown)


jolt-core/src/poly/one_hot_polynomial.rs
```diff
@@ -148,7 +148,7 @@ impl<F: JoltField> OneHotPolynomialProverOpening<F> {
             let univariate_poly_evals: [F; 2] = (0..B.len() / 2)
                 .into_par_iter()
                 .map(|k_prime| {
-                    let B_evals = B.sumcheck_evals_array::<2>(k_prime, BindingOrder::HighToLow);
+                    let B_evals = B.sumcheck_evals(k_prime, 2, BindingOrder::HighToLow);
```
moodlezoup, line 151: I was getting intermitted stack overflows during the batch opening proof sumcheck, so hopefully this fixes that

PR #796, opened 2025-07-24
[JOLT-135] Feat/ram dummy reads optimization

0xAndoroid:
Flagging changes to DAG: in stage 3 there's now hamming booleanity sumcheck that checks hamming weight for booleanity.


jolt-core/src/jolt/vm/ram/hamming_booleanity.rs
```diff
@@ -0,0 +1,194 @@
...
+        self.log_T
+    }
+
+    fn input_claim(&self) -> F {
+        F::zero()
+    }
+
+    fn compute_prover_message(&mut self, _round: usize) -> Vec<F> {
+        let p = self.prover_state.as_ref().unwrap();
+        const DEGREE: usize = 3;
```
moodlezoup: `DEGREE` is already declared above

PR #800, opened 2025-07-28
fix: prevent index out of bounds in UniPoly operations

DeVikingMark:
Description:

- fix potential panics due to empty coeffs in eval_at_zero, degree

- add safe checks for division by zero polynomials


markosg04:
Airdrop farming

PR #802, opened 2025-07-28
Feat/sparse batching gruen

markosg04:
- Adds gruen x Dao-Thaler high to low support
- Adds Dao-Thaler x Gruen optimization for the log(T) rounds of the sparse batching sumcheck
- removes some old code


jolt-core/src/poly/eq_poly.rs
```diff
@@ -95,6 +100,25 @@ impl<F: JoltField> EqPolynomial<F> {
         }
         evals
     }
+    /// evals_serial_cached but for "high to low" ordering, used specifically in the Gruen x Dao Thaler optimization.
+    fn evals_serial_cached_rev(r: &[F], scaling_factor: Option<F>) -> Vec<Vec<F>> {
+        let rev_r = r.iter().rev().collect::<Vec<_>>();
+        // let rev_r = r.iter().collect::<Vec<_>>();
```
moodlezoup: nit: delete

jolt-core/src/poly/one_hot_polynomial.rs
```diff
@@ -197,27 +199,45 @@ impl<F: JoltField> OneHotPolynomialProverOpening<F> {
...
 
-            let univariate_poly_evals: [F; 2] = (0..n)
-                .into_par_iter()
-                .map(|j| {
+            let mut gruen_eval_0 = F::zero();
+
+            if d_gruen.E_in_current_len() == 1 {
+                // E_in is fully bound
+                for j in 0..d_gruen.len() / 2 {
                     let H_evals = H.sumcheck_evals(j, 2, BindingOrder::HighToLow);
```
moodlezoup: we'll want to parallelize this loop. also, since the effective degree here is 1, we only need `H[j]`, we don't need to compute `H.sumcheck_evals`

jolt-core/src/poly/one_hot_polynomial.rs
```diff
@@ -197,27 +199,45 @@ impl<F: JoltField> OneHotPolynomialProverOpening<F> {
...
+                    let h_eval = H_evals[0];
+                    gruen_eval_0 += d_eval * h_eval;
+                }
+            } else {
+                // E_in has not been fully bound
+                let num_x_in_bits = d_gruen.E_out_current_len().log_2();
+                let x_bitmask = (1 << num_x_in_bits) - 1;
+
+                for j in 0..d_gruen.len() / 2 {
+                    let H_evals = H.sumcheck_evals(j, 2, BindingOrder::HighToLow);
```
moodlezoup: same here

jolt-core/src/poly/one_hot_polynomial.rs
```diff
@@ -197,27 +199,45 @@ impl<F: JoltField> OneHotPolynomialProverOpening<F> {
...
+                    let x_in = j >> num_x_in_bits;
+                    let x_out = j & x_bitmask;
+                    let d_e_out_eval = d_gruen.E_out_current()[x_out];
+                    let d_e_in_eval = d_gruen.E_in_current()[x_in];
+                    let h_eval = H_evals[0];
+
+                    gruen_eval_0 += d_e_out_eval * d_e_in_eval * h_eval;
```
moodlezoup: we should have a nested for loop instead of multiplying by both `d_e_out_eval` and `d_e_in_eval` here

jolt-core/src/poly/split_eq_poly.rs
```diff
@@ -1,93 +1,81 @@
...
+                //         |      second half of remaining elements (for E_in)
+                //         first half of remaining elements (for E_out)
+                let (_, wprime) = w.split_last().unwrap();
+                let (w_out, w_in) = wprime.split_at(m);
+                let (E_out_vec, E_in_vec) = rayon::join(
+                    || EqPolynomial::evals_cached(w_out),
+                    || EqPolynomial::evals_cached(w_in),
+                );
+                Self {
+                    current_index: w.len(),
```
moodlezoup, line 49: I know this was already like this, but imo this makes more sense
```suggestion
                    current_index: w.len() - 1,
```

jolt-core/src/poly/one_hot_polynomial.rs
```diff
@@ -103,6 +102,7 @@ impl<F: JoltField> OneHotPolynomialProverOpening<F> {
         let chunk_size = (T / num_chunks).max(1);
         let D = &self.eq_state.lock().unwrap().D;
 
+        let D_full_table = D.merge();
```
moodlezoup: we can't do this here, it's too slow. we'll need to compute G directly from `D`

jolt-core/src/poly/one_hot_polynomial.rs
```diff
@@ -197,27 +197,51 @@ impl<F: JoltField> OneHotPolynomialProverOpening<F> {
...
+                        let mut inner_sum = F::zero();
+                        for x_in in 0..num_x_in {
+                            let j = (x_in << num_x_out_bits) | x_out;
+                            if j < d_gruen.len() / 2 {
+                                let d_e_in_eval = d_gruen.E_in_current()[x_in];
+                                let h_eval = H.Z[j];
+                                inner_sum += d_e_in_eval * h_eval;
+                            }
+                        }
+                        inner_sum * d_gruen.E_out_current()[x_out]
```
moodlezoup: we should parallelize this inner loop too, otherwise threads won't be saturated for some rounds. Also, we should loop over `0..d_gruen.len() / 2` instead of `0..num_x_in`

jolt-core/src/poly/split_eq_poly.rs
```diff
@@ -306,116 +280,67 @@ impl<F: JoltField> GruenSplitEqPolynomial<F> {
...
-    }
+    /// Compute the quadratic sumcheck evaluations (i.e., the evaluations at {0, 2, 3}) of a
+    /// polynomial s(X) = l(X) * q(X), where l(X) is the current (linear) Dao-Thaler eq polynomial and
+    /// q(X) = c + dx
+    /// - c, the constant term of q
+    /// - the previous round claim, s(0) + s(1)
+    ///
+    /// important: This assumes `HighToLow` ordering (as used in the stage 5 batching sumcheck)
+    pub fn gruen_evals_deg_2(&self, q_0: F, previous_claim: F) -> [F; 2] {
+        // We want to compute the evaluations of the quadratic polynomial s(X) = l(X) * q(X), where
```
0xAndoroid, line 294: add an assert of binding order. Same for evals deg 3.

jolt-core/src/poly/split_eq_poly.rs
```diff
@@ -306,116 +283,68 @@ impl<F: JoltField> GruenSplitEqPolynomial<F> {
...
-                    merged.push(self.E2[i] * self.E1[j])
-                }
+        let evals = match self.binding_order {
+            BindingOrder::LowToHigh => {
+                // For low-to-high, current_index tracks how many variables remain unbound
+                // We want eq(w[0..current_index], x)
+                EqPolynomial::evals(&self.w[..self.current_index])
+                    .par_iter()
+                    .map(|x| *x * self.current_scalar)
+                    .collect()
```
moodlezoup: oh actually we can optimize this further using this:
```suggestion
                EqPolynomial::evals_parallel(&self.w[..self.current_index], Some(self.current_scalar))
```

jolt-core/src/poly/split_eq_poly.rs
```diff
@@ -306,116 +283,68 @@ impl<F: JoltField> GruenSplitEqPolynomial<F> {
...
             }
-            DensePolynomial::new(merged)
-        }
+            BindingOrder::HighToLow => {
+                // For high-to-low, current_index tracks how many variables have been bound
+                // We want eq(w[current_index..], x)
+                EqPolynomial::evals(&self.w[self.current_index..])
+                    .par_iter()
+                    .map(|x| *x * self.current_scalar)
+                    .collect()
```
moodlezoup: same here
```suggestion
                EqPolynomial::evals_parallel(&self.w[self.current_index..], Some(self.current_scalar))
```

PR #805, opened 2025-07-29
[JOLT-166] RV64

### LINKED ISSUE #439: RV64 exploration
Jolt can theoretically support 64-bit word sizes with relative ease. We would like to evaluate the extent to which RV64 would be more efficient than RV32 for certain programs. For example, we expect RV64 to generate smaller traces than RV32 for programs using a lot of large-field arithmetic. 

As an initial exploration, we should compare the trace lengths of programs compiled to RV32IM vs RV64IM.

### LINKED ISSUE #892: Sumcheck verification fails for inline assembly `div`
The verification stage fails for me when executing an inline assembly `div` with constants `45855881` and `-1884026147`.

**guest/src/lib.rs**
```rust

#[jolt::provable(guest_only, memory_size = 10240, max_trace_length = 65536)]
fn test() -> i32 {
    let result: i32;
    unsafe {
        core::arch::asm!(
            "div {result}, {a}, {b}",
            result = out(reg) result,
            a = in(reg) 45855881,
            b = in(reg) -1884026147,
        );
    }
    result
}
```

**src/main.rs**
```rust
pub fn main() {
    let target_dir = "/tmp/jolt-guest-targets";
    let mut program = guest::compile_test(target_dir);

    let prover_preprocessing = guest::preprocess_prover_test(&mut program);
    let verifier_preprocessing = guest::verifier_preprocessing_from_prover_test(&prover_preprocessing);

    let prove_circuits = guest::build_prover_test(program, prover_preprocessing);
    let verify_circuits = guest::build_verifier_test(verifier_preprocessing);

    let (output, proof, program_io) = prove_test();
    let is_valid = verify_test(output, program_io.panic, proof);
    println!("valid: {:?}", is_valid);
}
```

**Panic**
```
$> cargo run --release
...
thread 'main' panicked at /root/.cargo/git/checkouts/jolt-bc4943ecdf5f6930/f23b955/jolt-core/src/zkvm/mod.rs:295:40:
Verification failed: Stage 2

Caused by:
    Sumcheck verification failed
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
---

* commit: f23b95527683f3a935afba18f4368bf1cdac12ed

0xAndoroid:
Fixes #439
Fixes #892 

This PR contains other development initiatives: inlines, performance and memory optimizations, etc.
The footprint of the RV64 is not huge.


0xAndoroid:
Now rv64 is plausibly feature complete.
Next things on the list:
- Fix SVO
- Fix SHA2 inline
- Investigate and optimize memory usage
- Large d optimization in instruction execution shout

jolt-core/src/zkvm/instruction/virtual_pow2w.rs
```diff
@@ -0,0 +1,62 @@
...
+    fn lookup_table(&self) -> Option<LookupTables<XLEN>> {
+        Some(Pow2WTable.into())
+    }
+}
+
+impl InstructionFlags for VirtualPow2W {
+    fn circuit_flags(&self) -> [bool; NUM_CIRCUIT_FLAGS] {
+        let mut flags = [false; NUM_CIRCUIT_FLAGS];
+        flags[CircuitFlags::LeftOperandIsRs1Value as usize] = true;
+        flags[CircuitFlags::RightOperandIsImm as usize] = true;
```
moodlezoup:
```suggestion
```

jolt-core/src/zkvm/instruction/virtual_rotriw.rs
```diff
@@ -0,0 +1,62 @@
...
+        )
+    }
+
+    fn to_lookup_output(&self) -> u64 {
+        let (x, y) = LookupQuery::<XLEN>::to_instruction_inputs(self);
+        match XLEN {
+            #[cfg(test)]
+            8 => {
+                let (x, y) = (x as u8, (y as u8).trailing_zeros());
+                (((x & 0x0F) >> (y % 4)) | (((x & 0x0F) << (4 - (y % 4))) & 0x0F)) as u64
```
moodlezoup, line 42: any reason not to use `rotate_right` here?

jolt-core/src/zkvm/instruction/virtual_extend.rs
```diff
@@ -0,0 +1,61 @@
+use tracer::instruction::{virtual_extend::VirtualExtend, RISCVCycle};
+
+use crate::zkvm::lookup_table::{lower_half_word::LowerHalfWordTable, LookupTables};
+
+use super::{CircuitFlags, InstructionFlags, InstructionLookup, LookupQuery, NUM_CIRCUIT_FLAGS};
+
+impl<const XLEN: usize> InstructionLookup<XLEN> for VirtualExtend {
```
moodlezoup: we should call this instruction `VirtualZeroExtendHalfWord` imo

jolt-core/src/zkvm/instruction/virtual_sign_extend.rs
```diff
@@ -0,0 +1,70 @@
+use tracer::instruction::{virtual_sign_extend::VirtualSignExtend, RISCVCycle};
+
+use crate::zkvm::lookup_table::{sign_extend_half_word::SignExtendHalfWordTable, LookupTables};
+
+use super::{CircuitFlags, InstructionFlags, InstructionLookup, LookupQuery, NUM_CIRCUIT_FLAGS};
+
+impl<const XLEN: usize> InstructionLookup<XLEN> for VirtualSignExtend {
```
moodlezoup: we should call this instruction `VirtualZeroExtendHalfWord` imo

jolt-core/src/zkvm/lookup_table/movsign.rs
```diff
@@ -7,37 +7,42 @@ use super::PrefixSuffixDecomposition;
...
+    fn materialize_entry(&self, index: u128) -> u64 {
+        let sign_bit_pos = 2 * XLEN - 1;
+        let sign_bit = 1u128 << sign_bit_pos;
         if index & sign_bit != 0 {
-            (1 << WORD_SIZE) - 1
+            if XLEN == 64 {
+                u64::MAX
+            } else {
+                (1u64 << XLEN) - 1
+            }
```
moodlezoup: nit:
```suggestion
           ((1u128 << XLEN) - 1) as u64
```

tracer/src/utils/virtual_registers.rs
```diff
@@ -0,0 +1,110 @@
+use common::constants::{RISCV_REGISTER_COUNT, VIRTUAL_REGISTER_COUNT};
+use core::cell::RefCell;
+use std::ops::Deref;
+
+const NUM_VIRTUAL_REGISTERS: usize = VIRTUAL_REGISTER_COUNT as usize;
+const RISCV_REGISTER_BASE: u8 = RISCV_REGISTER_COUNT;
+
+thread_local! {
```
moodlezoup: why is this `thread_local`?

jolt-core/src/zkvm/lookup_table/prefixes/change_divisor.rs
```diff
@@ -0,0 +1,58 @@
+use crate::{field::JoltField, utils::lookup_bits::LookupBits};
+
+use super::{PrefixCheckpoint, Prefixes, SparseDensePrefix};
+
+pub enum ChangeDivisorPrefix<const XLEN: usize> {}
```
moodlezoup, line 5: Can you add a doc comment describing what this does?

jolt-core/src/zkvm/lookup_table/prefixes/change_divisor_w.rs
```diff
@@ -0,0 +1,62 @@
+use crate::{field::JoltField, utils::lookup_bits::LookupBits};
+
+use super::{PrefixCheckpoint, Prefixes, SparseDensePrefix};
+
+pub enum ChangeDivisorWPrefix<const XLEN: usize> {}
```
moodlezoup, line 5: same here

jolt-core/src/zkvm/lookup_table/suffixes/lower_word.rs
```diff
@@ -2,12 +2,16 @@ use crate::utils::lookup_bits::LookupBits;
...
-impl<const WORD_SIZE: usize> SparseDenseSuffix for LowerWordSuffix<WORD_SIZE> {
-    fn suffix_mle(b: LookupBits) -> u32 {
-        (u64::from(b) % (1 << WORD_SIZE)) as u32
+impl<const XLEN: usize> SparseDenseSuffix for LowerWordSuffix<XLEN> {
+    fn suffix_mle(b: LookupBits) -> u64 {
+        if XLEN == 64 {
+            u128::from(b) as u64
+        } else {
+            (u128::from(b) % (1 << XLEN)) as u64
+        }
```
moodlezoup:
```suggestion
        (u128::from(b) % (1 << XLEN)) as u64
```

jolt-core/src/zkvm/lookup_table/suffixes/sign_extension_right_operand.rs
moodlezoup, line 1: comments in this file are incorrect

jolt-core/src/zkvm/r1cs/constraints.rs
```diff
@@ -1,268 +1,346 @@
...
+    //     } else {
+    //         assert!(NextUnexpandedPC == UnexpandedPC + 4)
+    //     }
+    // }
+    // Note that ShouldBranch and Jump instructions are mutually exclusive
+    r1cs_prod!(
+        ({ JoltR1CSInputs::OpFlags(CircuitFlags::IsCompressed) })
+            * ({ JoltR1CSInputs::OpFlags(CircuitFlags::DoNotUpdateUnexpandedPC) })
+            == ({ JoltR1CSInputs::CompressedDoNotUpdateUnexpPC })
+    ),
```
moodlezoup: is it possible to make `IsCompressed` and `DoNotUpdateUnexpandedPC` mutually exclusive? afaict only the last instruction in a virtual sequence would need to have `IsCompressed` set

jolt-core/src/zkvm/lookup_table/virtual_rotrw.rs
```diff
@@ -0,0 +1,112 @@
...
+        (0..XLEN).rev().skip(XLEN / 2).for_each(|i| {
+            let x = x_bits >> i & 1;
+            let y = y_bits >> i & 1;
+            first_sum *= 1 + y;
+            first_sum += x * y;
+            second_sum += x * (1 - y) * prod_one_plus_y * (1 << i);
+            prod_one_plus_y *= 1 + y;
+        });
+
+        first_sum + second_sum
```
moodlezoup, line 31: can we use `u32::rotate_right` instead of doing this?

jolt-core/src/zkvm/lookup_table/virtual_change_divisor_w.rs
```diff
@@ -0,0 +1,122 @@
...
+        let mut y_product = F::one();
+        for i in XLEN / 2..XLEN {
+            y_product *= r[2 * i + 1];
+        }
+
+        let sign_extension = F::from_u128((1u128 << XLEN) - (1u128 << (XLEN / 2))) * sign_bit;
+
+        let adjustment = F::from_u64(2) - F::from_u128(1u128 << XLEN);
+
+        divisor_value + adjustment * x_product * y_product + sign_extension
```
moodlezoup, line 72: I'm probably missing something here but doesn't this give `1 + sign_extension` if `remainder == i32::MIN && divisor == -1`?

moodlezoup (commented):
First wave of comments –– still need to review inlines and tracer instructions

tracer/src/lib.rs
```diff
@@ -295,25 +297,68 @@ pub fn decode(elf: &[u8]) -> (Vec<RV32IMInstruction>, Vec<(u64, u8)>, u64) {
...
+                let first_halfword = u16::from_le_bytes([raw_data[offset], raw_data[offset + 1]]);
+
+                // Check if it's a compressed instruction (lowest 2 bits != 11)
+                if (first_halfword & 0b11) != 0b11 {
+                    // Compressed 16-bit instruction
+                    let compressed_inst = first_halfword;
+                    if compressed_inst == 0x0000 {
+                        offset += 2;
+                        continue;
+                    }
```
moodlezoup, line 367: do we need a special case for this?

tracer/src/utils/virtual_registers.rs
0xAndoroid: Tests run in a single environment, but in multiple threads. So, tests of inline sequences would use all the registers, since all are spawned in parallel.

jolt-core/src/zkvm/instruction/virtual_rotriw.rs
0xAndoroid: We have to rotate in the lower 4 bit space only. It's the `w` type instruction. And we don't have `u4` datatype.

jolt-core/src/zkvm/instruction/virtual_extend.rs
0xAndoroid: Half word is 16bits, since word is 32bits. Renamed to `VirtualZeroExtendWord`.

jolt-core/src/zkvm/instruction/virtual_sign_extend.rs
0xAndoroid: Half word is 16bits, since word is 32bits. Renamed to `VirtualSignExtendWord`.

jolt-core/src/zkvm/lookup_table/virtual_rotrw.rs
0xAndoroid: No, this calculation matches `x.rotate_right(y.trailing_zeros())` iff `y` is of the format all ones followed by all zeros. However, it diverges on other (irrelevant) values of y. That's why we need to have it like this to match the `evaluate_mle` on the hypercube.

jolt-core/src/zkvm/lookup_table/virtual_change_divisor_w.rs
0xAndoroid: It does match `materialize_entry` on the entire hypercube, and that one works as expected. I think sign extension cancels out with the adjustment

tracer/src/instruction/format/format_b.rs
```diff
@@ -13,7 +10,7 @@ use super::{
 pub struct FormatB {
     pub rs1: u8,
     pub rs2: u8,
-    pub imm: i64,
+    pub imm: i128,
```
moodlezoup, line 13: maybe wrong here, but I think we should be able to leave this as `i64`

tracer/src/instruction/format/mod.rs
```diff
@@ -19,23 +18,26 @@ pub struct NormalizedOperands {
     pub rs1: u8,
     pub rs2: u8,
     pub rd: u8,
-    pub imm: i64,
+    pub imm: i128,
```
moodlezoup, line 22: same here, afaict we can keep this as `i64`

jolt-core/src/zkvm/r1cs/constraints.rs
0xAndoroid: I think we can do this. Removed this committed polynomial.

tracer/src/utils/virtual_registers.rs
moodlezoup: I've been using `#[serial]` for this purpose, e.g. https://github.com/a16z/jolt/blob/main/jolt-core/src/zkvm/dag/mod.rs#L19

tracer/src/instruction/virtual_lw.rs
moodlezoup, line 1: why don't `VirtualSW` and `VirtualLW` appear in `zkvm/instruction/`?

tracer/src/instruction/virtual_assert_word_alignment.rs
```diff
@@ -0,0 +1,32 @@
...
+
+use super::{
+    format::format_virtual_halfword_alignment::HalfwordAlignFormat, RISCVInstruction, RISCVTrace,
+};
+
+declare_riscv_instr!(
+    name = VirtualAssertWordAlignment,
+    mask = 0,
+    match = 0,
+    format = HalfwordAlignFormat,
```
moodlezoup: we should change this format name to `AssertAlignFormat`

tracer/src/instruction/test.rs
```diff
@@ -1,83 +1,142 @@
...
+            original_cpu.x[instr.operands.rs2 as usize] = register_state.rs2_value() as i64;
+            virtual_cpu.x[instr.operands.rs2 as usize] = register_state.rs2_value() as i64;
+        }
+
+        let mut ram_access = Default::default();
+
+        let res = panic::catch_unwind(AssertUnwindSafe(|| {
+            instruction.execute(&mut original_cpu, &mut ram_access);
+        }));
+        if res.is_err() {
```
moodlezoup, line 120: maybe emit a warning here

tracer/src/instruction/mod.rs
```diff
@@ -322,10 +353,14 @@ where
...
-}
-
-pub trait VirtualInstructionSequence: RISCVInstruction {
-    fn virtual_sequence(&self) -> Vec<RV32IMInstruction>;
+    // Default implementation. Instructions with inline sequences will override this.
+    // This allows other modules (e.g. inline_helpers) to call this method on all instructions.
+    fn inline_sequence(&self, _xlen: Xlen) -> Vec<RV32IMInstruction>
+// where
+    //     Self: Into<RV32IMInstruction>,
+    {
```
moodlezoup:
```suggestion
    {
```

jolt-inlines/keccak256/src/trace_generator.rs
```diff
@@ -0,0 +1,462 @@
...
+];
+
+// Numb
+/// Layout of the 96 virtual registers (`vr`).
+///
+/// Jolt requires the total number of registers (physical + virtual) to be a power of two.
+/// With 32 physical registers, we need 96 virtual registers to reach 128.
+///
+/// While only 67 registers are actively used by this builder, we allocate 96
+/// to satisfy the system requirement.
```
moodlezoup: this comment is incorrect: `NEEDED_REGISTERS` below is 80, and we don't need to allocate more registers than we need

tracer/src/instruction/amoxorw.rs
```diff
@@ -38,4 +45,25 @@ impl AMOXORW {
...
+        let v_rs2 = allocate_virtual_register();
+        let v_rd = allocate_virtual_register();
+
+        let mut asm = InstrAssembler::new(self.address, self.is_compressed, xlen);
+        asm.emit_ld::<LW>(*v_rd, self.operands.rs1, 0);
+        asm.emit_r::<XOR>(*v_rs2, *v_rd, self.operands.rs2);
+        asm.emit_s::<SW>(self.operands.rs1, *v_rs2, 0);
+        asm.emit_i::<VirtualMove>(self.operands.rd, *v_rd, 0);
+        asm.finalize()
+    }
```
moodlezoup: why doesn't this use `amo_pre*` and `amo_post*`?

jolt-core/src/zkvm/r1cs/constraints.rs
moodlezoup: let's also add an assert somewhere that these two flags are in fact mutually exclusive

tracer/src/instruction/virtual_lw.rs
0xAndoroid: Because loading/storing word is done by loading/storing doubleword and shifting stripping upper bits, I believe

tracer/src/instruction/format/format_b.rs
0xAndoroid: Some instructions (such as ADDI) require value to be in the bitspace of 0..u64::max, and some require negative values - requiring a signed i65

tracer/src/lib.rs
0xAndoroid: No. I already investigated this. This doesn't mess up the PC

tracer/src/instruction/format/mod.rs
0xAndoroid: Some instructions (such as ADDI) require IMM to be in 0..u64::MAX, and some require negative values - needing at least 65 bits.

tracer/src/instruction/format/format_b.rs
0xAndoroid: Specifically format B I believe can be i64, but why not have everything imm as i128

tracer/src/instruction/test.rs
0xAndoroid: Panicking is expected behavior - on the assert instructions.

tracer/src/instruction/virtual_lw.rs
moodlezoup: I don't follow. `VirtualLW` is used in several virtual sequences, so afaict it should implement the `InstructionLookup`, `InstructionFlags` and `LookupQuery` traits. In particular, `CircuitFlags::Load` should be set for `VirtualLW` (respectively, `CircuitFlags::Store` and `VirtualSW`) otherwise we would be underconstrained

0xAndoroid: ViltualLW/SW is only used in RV32 inline sequences. Do we still want to have it implemented? I believe consensus was to support RV32 in tracer, but keep jolt-core RV64 only.

moodlezoup: hmm we can leave it unimplemented, but let's add an assert that xlen = 32 in the `exec` implementations for `VirtualLW` and `VirtualSW`

jolt-core/src/zkvm/r1cs/constraints.rs
0xAndoroid: added in the bytecode read-raf

tracer/src/utils/virtual_registers.rs
0xAndoroid: allocator is now passed as argument to `inline_sequence` funciton

tracer/src/instruction/amoxorw.rs
0xAndoroid: Answer from GX: we need it only for W type atomic instructions.

PR #814, opened 2025-07-30
Drop sumcheck data structures

markosg04:
- drops various data structures when they are not needed
- Optimizes batching sumcheck sparse case to use F for the first round of T variables to avoid materializing H to save memory
- Otherwise cuts most of the memory across the board by half and we observe more or less the theoretical 1 GiB / 2^20 cycles (other than spartan which is known, a constant spike in instructions and some overhead still in batching sumcheck)

### DIFF
--- a/jolt-core/src/poly/one_hot_polynomial.rs
+++ b/jolt-core/src/poly/one_hot_polynomial.rs
@@ -165,28 +165,15 @@ impl<F: JoltField> OneHotPolynomialProverOpening<F> {
                         .skip(k_prime)
                         .step_by(B.len() / 2)
                         .map(|(k, &G_k)| {
-                            // k_m is the bit corresponding to the variable we'll be binding next
                             let k_m = (k >> (num_unbound_address_variables - 1)) & 1;
-                            // We then index into F using the high order bits of k
                             let F_k = F[k >> num_unbound_address_variables];
-
-                            // G_times_F := G[k] * F[k_1, ...., k_{m-1}]
                             let G_times_F = G_k * F_k;
 
-                            // For c \in {0, 2} compute:
-                            //    G[k] * F[k_1, ...., k_{m-1}, c]
-                            //    = G[k] * F[k_1, ...., k_{m-1}] * eq(k_m, c)
-                            //    = G_times_F * eq(k_m, c)
-                            let eval_c0 = match k_m {
-                                0 => G_times_F, // eq(0, 0) = 0 * 0 + (1 - 0) * (1 - 0) = 1,
-                                1 => F::zero(), // eq(1, 0) = 1 * 0 + (1 - 1) * (1 - 0) = 0
-                                _ => unreachable!(),
-                            };
-
-                            let eval_c2 = match k_m {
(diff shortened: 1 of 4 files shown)


jolt-core/src/poly/one_hot_polynomial.rs
```diff
@@ -396,7 +432,13 @@ impl<F: JoltField> OneHotPolynomial<F> {
                     let min_row_index = chunk_index * chunk_size;
                     let max_row_index = min_row_index + chunk_size;
 
-                    for (t, k) in self.nonzero_indices.iter().enumerate() {
+                    for (t, k) in self
+                        .nonzero_indices
+                        .as_ref()
+                        .expect("nonzero_indices not initialized")
```
0xAndoroid: if we replace expect with unwrap, does it stack into one line? Maybe it's better to have it as one line. Applicable everywhere, since expect shouldn't be triggered regardless.

jolt-core/src/zkvm/ram/booleanity.rs
```diff
@@ -248,28 +256,42 @@ impl<F: JoltField> SumcheckInstance<F> for BooleanitySumcheck<F> {
...
+                }
+                if let Some(f) = prover_state.F.take() {
+                    drop_in_background_thread(f);
+                }
+
+                // Drop addresses and memory_layout as they're no longer needed in phase 2
+                if let Some(addresses) = self.addresses.take() {
+                    drop_in_background_thread(addresses);
+                }
+                if let Some(memory_layout) = self.memory_layout.take() {
```
0xAndoroid: memory layout should be like a couple of integers, probably not worth bloating the codebase

jolt-core/src/zkvm/instruction_lookups/booleanity.rs
```diff
@@ -30,10 +34,10 @@ const DEGREE: usize = 3;
 struct BooleanityProverState<F: JoltField> {
```
0xAndoroid: Maybe there's a better way to handle this? For exmaple, what if we have phase 1 prover state, and phase 2 prover state? I just don't like everything being an option.

jolt-core/src/zkvm/ram/booleanity.rs
moodlezoup: +1

jolt-core/src/zkvm/instruction_lookups/booleanity.rs
moodlezoup: yeah I was kinda thinking this as well

markosg04: I made a ticket: [JOLT-169]. I think it would be best to get these memory savings in now and then refactor all of the prover states

moodlezoup: I think we should just address this now. If you don't want to do the separate struct approach, we can still avoid the Options by using `std::mem::take` + `drop_in_background_thread` on the underlying Vecs

PR #815, opened 2025-07-30
fix: Assumptions about stack and heap locations in MemoryLayout

sagar-a16z:
- aligned linker and emulator memory region sizes
- ensure bytecode size is known before constructing MemoryLayout so that bounds checks can be enforced
- aligned DRAM_BASE and RAM_START_ADDRESS
- made JoltDevice optional in mmu instead of initializing it with a dummy memory layout

### DIFF
--- a/jolt-core/src/host/mod.rs
+++ b/jolt-core/src/host/mod.rs
@@ -8,6 +8,7 @@ use std::{
     process::Command,
 };
 
+use common::constants::{EMULATOR_MEMORY_CAPACITY, RAM_START_ADDRESS};
 use common::{
     constants::{
         DEFAULT_MAX_INPUT_SIZE, DEFAULT_MAX_OUTPUT_SIZE, DEFAULT_MEMORY_SIZE, DEFAULT_STACK_SIZE,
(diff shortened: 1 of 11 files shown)


jolt-core/src/host/mod.rs
```diff
@@ -203,17 +209,23 @@ impl Program {
             File::open(elf).unwrap_or_else(|_| panic!("could not open elf file: {elf:?}"));
         let mut elf_contents = Vec::new();
         elf_file.read_to_end(&mut elf_contents).unwrap();
+        let (_, raw_bytes) = tracer::decode(&elf_contents);
```
sagar-a16z, line 212: This part is slightly wasteful. Maybe there's a better way to get the bytecode size only instead of decoding the whole thing?

tracer/src/emulator/mmu.rs
```diff
@@ -736,12 +769,30 @@ impl Mmu {
...
+                    self.jolt_device
+                        .as_mut()
+                        .expect("JoltDevice not set")
+                        .store(effective_address, value);
                 }
             },
         };
     }
 
+    pub fn setup_bytecode(&mut self, p_address: u64, value: u8) {
```
sagar-a16z, line 781: added a new function so that we can load the bytecode into memory without triggering safety checks

jolt-core/src/host/mod.rs
moodlezoup: maybe a naive question, but can we just use `elf_contents.len()` as the bytecode size?

moodlezoup (approved):
lgtm

jolt-core/src/host/mod.rs
sagar-a16z: I will try that. It seemed less safe to do since there might be other stuff in the elf.

0xAndoroid: I'd prefer not doing that, since when risc64imac is introduced, c type instructions are 2 bytes and not 4 bytes.
(not applicable)

PR #818, opened 2025-07-31
Misceallaneous cleanup

moodlezoup:
- Makes the usage of the `CommitmentScheme::setup_prover` parameter consistent throughout (prior to this PR, Dory interpreted the parameter as max # variables, while other PCS interpreted it as max # coefficients)
- Removes unused code
- Various other cosmetic improvements

### DIFF
--- a/jolt-core/benches/commit.rs
+++ b/jolt-core/benches/commit.rs
@@ -9,7 +9,8 @@ use jolt_core::utils::transcript::{KeccakTranscript, Transcript};
 use rand_chacha::ChaCha20Rng;
 use rand_core::{RngCore, SeedableRng};
 
-const SRS_SIZE: usize = 1 << 10;
+const NUM_VARS: usize = 10;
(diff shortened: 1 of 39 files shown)


jolt-core/benches/commit.rs
```diff
@@ -9,7 +9,8 @@ use jolt_core::utils::transcript::{KeccakTranscript, Transcript};
 use rand_chacha::ChaCha20Rng;
 use rand_core::{RngCore, SeedableRng};
 
-const SRS_SIZE: usize = 1 << 10;
+const NUM_VARS: usize = 10;
+const SRS_SIZE: usize = 1 << NUM_VARS;
```
0xAndoroid, line 13: URS, applicable everywhere. We should probably change it now

0xAndoroid: Oh, already merged. I'll create a linear issue for it

PR #828, opened 2025-08-02
Omid/dispatch

omibo:
# Overview
This PR extracts inline instructions from the `tracer` crate so that external crates using Jolt can define their own inline instructions—without any changes to Jolt itself.

# Key Features
1. Up to 1024 custom instructions (7 bits from funct7 + 3 bits from funct3) can be supported with zero Jolt modifications.
2. Adds a single instruction to the jolt code-base, INLINE, whose execution and trace behavior are injected at runtime by the external crate.
3. Keeps Jolt-side changes minimal for an easy review and merge.

# Implementation Details
**tracer module:**
1. tracer/instructions/mod.rs – adds the `INLINE` variant to the instruction enum.
2. tracer/instructions/mod.rs::decode() – maps opcode 0b0101011 (0x2B) to `INLINE`; validation is disabled because funct3 and funct7 vary.
3. tracer/instructions/inline.rs – defines the `INLINE` struct (captures funct3, funct7) and a registry keyed by those fields that stores custom `exec` and `virtual-sequence-generator` functions.

**jolt-sdk module:**
1. The compression function now emits `INLINE` instead of SHA256 (still SHA-256 logic).

**jolt-core module:**
1. jolt-core/src/bench.rs registers sha256 and sha256_init in the `INLINE` registry (this serves as our external crate for now).

**examples module:**
1. precompile-example shows SHA-256 implemented via `INLINE`, with tests covering both execution and tracing paths.

# Open Questions
1- Will Jolt support any built-in inline instruction?
2- How should we allocate opcodes—are any reserved?


tracer/src/instruction/mod.rs
```diff
@@ -354,6 +353,9 @@ macro_rules! define_rv32im_enums {
             $(
                 $instr($instr),
             )*
+            /// Inline instruction from external crates
+            #[serde(skip)]
```
0xAndoroid: I'm worried that this will cause issues. We are using serde in `jolt-core/host/analyze.rs` only. @moodlezoup

omibo: I removed #[serde(skip)] and works fine.

jolt-inlines/sha2/src/trace_generator.rs
```diff
@@ -446,75 +450,37 @@ impl Sha256SequenceBuilder {
...
-        initial_state[0].wrapping_add(a),
-        initial_state[1].wrapping_add(b),
-        initial_state[2].wrapping_add(c),
-        initial_state[3].wrapping_add(d),
-        initial_state[4].wrapping_add(e),
-        initial_state[5].wrapping_add(f),
-        initial_state[6].wrapping_add(g),
-        initial_state[7].wrapping_add(h),
-    ]
+// Virtual instructions builder for sha256_init - returns empty sequence as XOR is atomic
```
0xAndoroid: What does this mean "XOR is atomic"?

jolt-inlines/sha2/src/trace_generator.rs
```diff
@@ -446,75 +450,37 @@ impl Sha256SequenceBuilder {
...
+pub fn sha2_init_virtual_sequence_builder(
+    _address: u64,
```
0xAndoroid: It seems like these variables are used in the function, but prefixed with `_` as unused variables.

jolt-inlines/sha2/src/trace_generator.rs
```diff
@@ -446,75 +450,37 @@ impl Sha256SequenceBuilder {
...
-    Reg(u8),
+const VIRTUAL_REGISTER_COUNT: u64 = 32; //  see Section 6.1 of Jolt paper
+pub const fn virtual_register_index(index: u64) -> u64 {
+    index + VIRTUAL_REGISTER_COUNT
 }
 
-pub fn execute_sha256_compression_initial(input: [u32; 16]) -> [u32; 8] {
-    execute_sha256_compression(BLOCK.map(|x| x as u32), input)
+// Virtual instructions builder for sha256 - returns empty sequence as XOR is atomic
+pub fn sha2_virtual_sequence_builder(_address: u64, _rs1: u8, _rs2: u8) -> Vec<RV32IMInstruction> {
```
0xAndoroid: Same here, variables are prefixed with `_`.

jolt-inlines/sha2/src/trace_generator.rs
```diff
@@ -446,75 +450,37 @@ impl Sha256SequenceBuilder {
     }
 }
 
-#[derive(Clone, Copy)]
-enum Value {
-    Imm(u64),
-    Reg(u8),
+const VIRTUAL_REGISTER_COUNT: u64 = 32; //  see Section 6.1 of Jolt paper
+pub const fn virtual_register_index(index: u64) -> u64 {
```
0xAndoroid: I think that this function's signature should change as now register id is `u8`.

.gitignore
```diff
@@ -1,6 +1,7 @@
 # Generated by Cargo
 # will have compiled files and executables
 /target/
+inlines/**/target/
```
0xAndoroid: this seems to be redundant since now inlines are part of top workspace.

PR #829, opened 2025-08-04
Add Keccak256 inline, ANDN op to tracer

danielwlz:
- Keccak256 (SHA-3) inline added to tracer
- Helper Keccak256 added to jolt-sdk (similar interface to sha3 crate)
- `Keccak256SequenceBuilder` builds the virtual sequence from primitive instructions. Using `VirtualROTRI` and  `ANDN` reduces sequence length to ≈3.3k RV64 instructions (approximately 25 cycles per byte)
- Introduced `ANDN` (AND-with-NOT) instruction with test coverage;
- Migrated `sha3-ex` and `sha3-chain` examples to use the inline
- Added `CpuTestHarness` to streamline unit tests for instructions and inlines
- Fixed VirtualROTRI for RvV64 (Omid)

### Test plan
```bash
# unit + algorithm tests (passing)
cargo test --package tracer --lib -- instruction::inline_keccak256

# sdk integration (passing)
cargo test --package jolt-sdk --features host,keccak256 --lib -- keccak256::tests
```

Tests verify against known input/outputs and verify trace and exec paths match.

### Next steps 
- Add lookup table and prefix-suffix decompositions for `ANDN` to jolt-core
- (Daniel) Move shared emitters from `Keccak256SequenceBuilder` into `inline_helpers` to benefit SHA-2 and future inlines
- (Daniel) Amend SHA2 inline to use added code and instructions

**Request:** please squash this PR when merging.

PR #838, opened 2025-08-05
Add inline_helpers module and SHA2 tests

danielwlz:
Eliminate duplicated emitter logic with `InstrAssembler`; migrate Keccak & SHA-256 builders, add RV32 helpers, fix decode for ANDN

- Adds `InstrAssembler` to deduplicate instruction emitter helpers in inliners. Makes it easier to build inlines.
- Add SHA256 unit tests
- Renames virtual sequences to inline sequences
- Removes VirtualInstructionSequence trait, folding it into RISCVTrace
- Code impact: removes ~500 LOC of duplicated code
- Not tackled in this PR: failing SHA2 inline tests

### Test plan

```
cargo test --package tracer --lib -- instruction::
cargo test --package jolt-sdk --features host,sha256 --lib -- sha256::tests
cargo test --package jolt-sdk --features host,keccak256 --lib -- keccak256::tests
```

All relevant tests pass locally.

```
running 24 tests
test instruction::andn::tests::decode_constant ... ok
test instruction::inline_keccak256::tests::test_execute_chi ... ok
test instruction::inline_keccak256::tests::test_execute_iota ... ok
test instruction::inline_keccak256::tests::test_execute_keccak_f ... ok
test instruction::inline_keccak256::tests::test_execute_rho_and_pi ... ok
test instruction::inline_keccak256::tests::test_execute_keccak256 ... ok
test instruction::andn::tests::test_andn ... ok
test instruction::inline_keccak256::tests::test_execute_theta ... ok
test instruction::inline_keccak256::tests::test_rotl64 ... ok
test instruction::inline_keccak256::tests::test_step_by_step_round_1 ... ok
test instruction::inline_sha256::sha256::tests::measure_sha256_length ... ok
test instruction::inline_keccak256::keccak256::tests::test_keccak256_direct_execution ... ok
test instruction::inline_keccak256::keccak256::tests::test_keccak_against_reference ... ok
test instruction::inline_keccak256::keccak256::tests::measure_keccak_length ... ok
test instruction::inline_sha256::sha256::tests::test_sha256_direct_execution ... ok
test instruction::inline_sha256::sha256init::tests::test_sha256init_direct_execution ... ok
test instruction::test::tests::test_remaining ... ok
test instruction::inline_sha256::sha256::tests::test_sha256_exec_trace_equal ... ok
test instruction::virtual_rotri::tests::test_manual_rotation ... ok
test instruction::virtual_rotri::tests::test_virtual_rotri ... ok
test instruction::inline_keccak256::keccak256::tests::test_keccak_exec_trace_equal ... ok
test instruction::inline_sha256::sha256init::tests::test_sha256init_exec_trace_equal ... ok
test instruction::virtual_rotri::tests::test_exec_vs_trace ... ok
test instruction::inline_keccak256::keccak256::tests::test_keccak_exec_trace_intermediate_vr_equal ... ok

test result: ok. 24 passed; 0 failed; 0 ignored; 0 measured; 18 filtered out; finished in 0.07s

running 1 test
test sha256::tests::test_sha256_empty_string ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

running 1 test
test keccak256::tests::test_keccak256_empty ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```


tracer/src/inline_helpers.rs
```diff
@@ -0,0 +1,353 @@
...
+                rs2,
+                imm: offset_lanes * 8,
+            },
+            virtual_sequence_remaining: Some(0),
+            is_compressed: self.is_compressed,
+        };
+        self.sequence.push(inst.into());
+    }
+
+    // ---------------------------------------------------------------------
```
0xAndoroid: not a fan of these kind of comments

danielwlz:
Ready for review @moodlezoup @0xAndoroid . This is follow-up work from #829

tracer/src/inline_helpers.rs
```diff
@@ -0,0 +1,353 @@
...
+            address: self.address,
+            operands: FormatLoad {
+                rd,
+                rs1,
+                imm: offset_words * 4,
+            },
+            virtual_sequence_remaining: Some(0),
+            is_compressed: self.is_compressed,
+        };
+        self.sequence.push(inst.into());
```
0xAndoroid: I think we don't have recursive expansion of inline sequences. We'd either need to implement recursive expansion, or update this to be `sequence.extend(instr.virtual_sequence())`. This changed because `LW` is now implemented as Inline sequence.

tracer/src/inline_helpers.rs
```diff
@@ -0,0 +1,353 @@
...
+            address: self.address,
+            operands: FormatS {
+                rs1,
+                rs2,
+                imm: offset_words * 4,
+            },
+            virtual_sequence_remaining: Some(0),
+            is_compressed: self.is_compressed,
+        };
+        self.sequence.push(inst.into());
```
0xAndoroid: Same as above.

tracer/src/inline_helpers.rs
```diff
@@ -0,0 +1,353 @@
...
+    }
+
+    /// 32-bit wrapping add (ADD/ADDI) with constant folding.
```
0xAndoroid: Here regular `ADD` is used, not 32bit add. We should probably clarify this in the comment.
Ultimately, SHA should use 32bit add, `ADDW`.

0xAndoroid: Ultimately the plan is to support both architectures in tracer, but I'm okay to assume 64bit architecture for inlines.

tracer/src/inline_helpers.rs
```diff
@@ -0,0 +1,353 @@
...
+    }
+
+    /// Rotate-right by amount on 32-bit word.
+    pub fn rotri32(&mut self, rs1: Value, shamt: u32, rd: usize) -> Value {
+        if shamt == 0 {
+            return self.xor(rs1, Imm(0), rd);
+        }
+        let ones = (1u64 << (32 - shamt)) - 1;
+        let mask = ones << shamt;
+        self.rotri(rs1, mask, rd)
```
0xAndoroid, line 407: Is this right? We use regular rotri? Or the 32bit rotri

tracer/src/instruction/inline_sha256/test_constants.rs
```diff
@@ -0,0 +1,89 @@
+#[cfg(test)]
```
0xAndoroid: This is unnecessary since entire module is behind #cfg(test) in mod.rs.

tracer/src/instruction/inline_sha256/test_utils.rs
```diff
@@ -0,0 +1,177 @@
...
+        let mut vr = [0; NEEDED_REGISTERS];
+        for i in 0..NEEDED_REGISTERS {
+            vr[i] = common::constants::virtual_register_index(i as u64) as usize;
+        }
+
+        Self {
+            // RV32.
+            harness: CpuTestHarness::new_32(),
+            // TODO: Fix RV64 failure (below)
+            // harness: CpuTestHarness::new(),
```
0xAndoroid, line 45: The RV64 failure is most likely due to things mentioned before. Some instructions still operate in 64bit space. Say ROTRI, etc. Please check with Omid, since he's also been fixing such problems.

tracer/src/instruction/mod.rs
```diff
@@ -708,6 +708,7 @@ impl RV32IMInstruction {
                     (0b101, 0b0100000) => Ok(SRA::new(instr, address, true, compressed).into()),
                     (0b110, 0b0000000) => Ok(OR::new(instr, address, true, compressed).into()),
                     (0b111, 0b0000000) => Ok(AND::new(instr, address, true, compressed).into()),
+                    (0b111, 0b0100000) => Ok(ANDN::new(instr, address, true, compressed).into()),
```
0xAndoroid: Why is this the opcode for `ADDN`? I don't think that `ADDN` is part of RISCV64IMAC, so we probably shouldn't put it in this decode at all.

tracer/src/inline_helpers.rs
danielwlz: Agreed, resolved!

tracer/src/instruction/inline_sha256/test_constants.rs
danielwlz: Resolved

tracer/src/inline_helpers.rs
danielwlz: My first attempt to deal with this ended in flames, will step back and think more about it. One option that could work would be to implement a default virtual_sequence method in the RISCVInstructionTrait and kill the VirtualInstructionSequence trait, but that seems drastic and I am hesitant about doing it (unless you think it's the way to go).

```
    // Default implementation for non-virtual instructions. Virtual instructions
    // will override this. This allows other modules (e.g. inline_helpers) to call this method on all instructions.
    fn virtual_sequence(&self, _xlen: Xlen) -> Vec<RV32IMInstruction>
    where
        Self: Into<RV32IMInstruction>,
    {
        vec![(*self).into()]
    }
```

tracer/src/instruction/mod.rs
danielwlz: Right, it's not in RISCV64IMAC, although it is part of the B extension (zbb specifically). Removed.

tracer/src/inline_helpers.rs
danielwlz: Thanks! Does this mean I should switch this to ADDW/ADDWI?

danielwlz: If I switch to ADDW/ADDWI, I get the error `VirtualSignExtend is not supported for 32-bit mode`

0xAndoroid: Because you are testing SHA256 in the XLen 32 in your test. We want to get it working in RV64. You can also make this conditional, i.e. ADD in Bit32, and ADDW in Bit64.

tracer/src/inline_helpers.rs
danielwlz: I think so? I did try different ways of implementing this and my conclusion was that this is the correct way with 60-70% certainty (supported by the passing tests) - otherwise it would fail

tracer/src/instruction/inline_sha256/test_utils.rs
danielwlz: As spoken, we'll leave fixing the sha2 inline for another PR.

tracer/src/inline_helpers.rs
0xAndoroid: Why doesn't
```rust
        self.sequence.extend(lw.virtual_sequence(Xlen::Bit64));
```
work?

tracer/src/inline_helpers.rs
0xAndoroid: Postponing fixing sha256 to later

0xAndoroid:
And noting to fix clippy

tracer/src/inline_helpers.rs
danielwlz: That was the first thing I tried (see my [patch](https://gist.github.com/danielwlz/0e9560518ad9443958519ee558ebdcc1)); but I ran into compiler errors because virtual_sequence does not exist on RISCVInstruction. But adding it there doesn't work either because it conflicts with the definition of the VirtualInstructionSequence trait (can't define it twice). I could do this, though, if we remove the virtual_sequence method from the VirtualInstructionSequence such that there is no conflict. Do you think that's an acceptable solution?

0xAndoroid: You'd just put it in those instructions that implement `VirtualInstructionSequence`, specifically `LW`, and `SW` here. Everything else would stay the same.

tracer/src/instruction/format/format_i.rs
```diff
@@ -73,6 +73,15 @@ impl InstructionFormat for FormatI {
             rs1: (rng.next_u64() % REGISTER_COUNT) as usize,
         }
     }
+
+    fn from_normalized(operands: NormalizedOperands) -> Self {
```
0xAndoroid: It's also probably better to implement `From<NormalizedOperands>` instead of having this function.

danielwlz: Good suggestion, will implement. Thanks!

tracer/src/inline_helpers.rs
danielwlz: I folded VirtualInstructionSequence::virtual_sequence into RISCVTrace and have eliminated VirtualInstructionSequence as discussed. Changes incoming, will let you know when they are in.

tracer/src/inline_helpers.rs
danielwlz: Made it conditional (but the tests still fail)

0xAndoroid (approved):
nice. thanks for refactoring the virtual sequence function.

PR #843, opened 2025-08-07
Optimisations/polynomial evaluations

abiswas3:
See [Blog](https://randomwalks.xyz/publish/fast_polynomial_evaluation.html) exact details.

Contributions:

## Tracked Data Types

We now have the ability to track actual arithmetic operations see `field/tracked_ark.rs`

Blog has more details, but this allows us to test and and check for optimality.

## Faster Polynomial Evaluations

+ Split Eq polynomial evaluation speeds up single polynomial evaluation for sparse and non sparse cases (old dot product code had an inefficiency, see blog, not that it matters now)

single_poly_evals_num_vars

single_poly_eval_sparsity

+ Batched Split Eq polynomial evaluation also speeds up batched polynomial evaluation but not as much as batch size * sparsity domin
batch_poly_eval_sparsity
ates

batch_poly_evals_num_vars

### DIFF
--- a/jolt-core/examples/polynomial_evaluation.rs
+++ b/jolt-core/examples/polynomial_evaluation.rs
@@ -73,8 +73,8 @@ fn benchmark_batch_polynomial_evaluation(batch_size: usize) {
         "exp,num_vars,c,algorithm,time_ms,mults,trial,batch_size"
     )
     .unwrap();
-    let num_trials = 1;
-    for exp in [14, 16, 18, 20] {
+    let num_trials = 10;
+    for exp in [14, 16, 18, 20, 22] {
         let num_evals = 1 << exp;
 
         for c in [0.005, 0.20, 0.50, 0.75] {
--- a/jolt-core/examples/polynomial_evaluation.rs
+++ b/jolt-core/examples/polynomial_evaluation.rs
@@ -92,7 +92,7 @@ fn benchmark_batch_polynomial_evaluation(batch_size: usize) {
                 // --- Algorithm 1: Dot Product ---
                 reset_mult_count();
                 let start = Instant::now();
-                MultilinearPolynomial::batch_evaluate(&poly_refs, &eval_point);
+                MultilinearPolynomial::batch_evaluate_with_eq(&poly_refs, &eval_point);
                 let time_ms = start.elapsed().as_millis();
                 let mults = get_mult_count();
                 writeln!(
@@ -116,10 +116,7 @@ fn benchmark_batch_polynomial_evaluation(batch_size: usize) {
                 // --- Algorithm 3: Sparse Dot Product ---
                 reset_mult_count();
                 let start = Instant::now();
-                MultilinearPolynomial::batch_evaluate_optimised_for_sparse_polynomials(
-                    &poly_refs,
-                    &eval_point,
-                );
+                MultilinearPolynomial::batch_evaluate(&poly_refs, &eval_point);
                 let time_ms = start.elapsed().as_millis();
                 let mults = get_mult_count();
                 writeln!(
@@ -174,7 +171,7 @@ fn benchmark_single_polynomial_evaluation() {
                 // --- Algorithm 2: Inside-Out Product ---
                 reset_mult_count();
                 let start = Instant::now();
-                let inside_out_prod = poly.evaluate(&eval_point);
+                let inside_out_prod = poly.evaluate_inside_out(&eval_point);
                 let time_ms = start.elapsed().as_millis();
                 let mults = get_mult_count();
                 writeln!(
@@ -187,7 +184,7 @@ fn benchmark_single_polynomial_evaluation() {
                 // --- Algorithm 3: Sparse Dot Product ---
                 reset_mult_count();
                 let start = Instant::now();
-                let sparse_prod = poly.evaluate_sparse_dot_product(&eval_point);
+                let sparse_prod = poly.evaluate(&eval_point);
                 let time_ms = start.elapsed().as_millis();
                 let mults = get_mult_count();
                 writeln!(
--- a/jolt-core/examples/polynomial_evaluation.rs
+++ b/jolt-core/examples/polynomial_evaluation.rs
@@ -1,0 +1,195 @@
+#![allow(clippy::uninlined_format_args)]
+use ark_ff::Zero;
+use ark_std::rand::{rngs::StdRng, Rng, SeedableRng};
+use jolt_core::field::tracked_ark::TrackedFr as Fr;
+use jolt_core::field::JoltField;
+use jolt_core::utils::counters::{get_mult_count, reset_mult_count};
+use jolt_core::{
+    poly::multilinear_polynomial::{MultilinearPolynomial, PolynomialEvaluation},
+    utils::math::Math,
+};
+
+use std::fs::OpenOptions;
+use std::io::Write;
+use std::time::Instant;
+
+fn setup_batch_inputs(
+    n: usize,
+    batch_size: usize,
+    sparsity: f64,
+) -> (Vec<MultilinearPolynomial<Fr>>, Vec<Fr>) {
+    let mut rng = StdRng::seed_from_u64(123);
+    let eval_loc: Vec<Fr> = (0..n.log_2())
+        .map(|_| Fr::random(&mut rng))
+        .collect::<Vec<Fr>>();
+
+    // Get many dense polynomials
+    let polys: Vec<MultilinearPolynomial<Fr>> = (0..batch_size)
+        .map(|_| {
+            let (poly, _) = sparse_inputs(n, sparsity);
+            poly
+        })
+        .collect();
+
+    (polys, eval_loc)
+}
+
+fn sparse_inputs(n: usize, c: f64) -> (MultilinearPolynomial<Fr>, Vec<Fr>) {
+    assert!(n.is_power_of_two(), "n must be a power of 2");
+
+    let mut rng = StdRng::seed_from_u64(123);
+    // Compute number of zeros
+    // Each position independently: zero with prob c, random otherwise
+    let values: Vec<Fr> = (0..n)
+        .map(|_| {
+            if rng.gen::<f64>() < c {
+                Fr::zero()
+            } else {
+                Fr::random(&mut rng)
+            }
+        })
+        .collect();
+
+    let poly = MultilinearPolynomial::from(values);
+
+    // Random evaluation point remains unchanged
+    let eval_point = (0..n.log_2())
+        .map(|_| Fr::random(&mut rng))
+        .collect::<Vec<_>>();
+
+    (poly, eval_point)
+}
+fn benchmark_batch_polynomial_evaluation() {
+    let mut file = OpenOptions::new()
+        .create(true)
+        .write(true)
+        .truncate(true)
+        .open("batch_results.csv")
+        .expect("Unable to open file");
+
+    let batch_size = 10;
+    // Write CSV header
+    writeln!(
+        file,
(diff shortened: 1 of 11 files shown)


jolt-core/src/poly/multilinear_polynomial.rs
```diff
@@ -583,34 +595,216 @@ impl<F: JoltField> PolynomialBinding<F> for MultilinearPolynomial<F> {
...
+                    vec![F::zero(); num_polys]
+                } else {
+                    // computing agg[x1]
+                    let inner_sums = (0..eq_two.len())
+                        .into_par_iter()
+                        .filter_map(|x2| {
+                            let eq2_val = eq_two[x2];
+                            if eq2_val.is_zero() {
+                                return None;
+                            }
```
moodlezoup: nit: `eq2_val` will never be 0 (assuming `r` is random)

jolt-core/src/poly/multilinear_polynomial.rs
```diff
@@ -583,34 +595,216 @@ impl<F: JoltField> PolynomialBinding<F> for MultilinearPolynomial<F> {
...
+                                            eq2_val * z
+                                        }
+                                    }
+                                    MultilinearPolynomial::U8Scalars(poly) => {
+                                        let z = poly.coeffs[idx];
+                                        if z == 0 {
+                                            F::zero()
+                                        } else {
+                                            z.field_mul(eq2_val)
+                                        }
```
moodlezoup: Looking at this again, I think we should add a check in the various `field_mul` implementations that returns 0 if `self` is zero. That would allow us to make this function less verbose

jolt-core/examples/polynomial_evaluation.rs
moodlezoup, line 1: could you move this file into [jolt-core/benches](https://github.com/a16z/jolt/tree/main/jolt-core/benches)?

jolt-core/src/poly/multilinear_polynomial.rs
```diff
@@ -583,34 +595,216 @@ impl<F: JoltField> PolynomialBinding<F> for MultilinearPolynomial<F> {
...
+            MultilinearPolynomial::RLC(_) => {
+                // TODO(moodlezoup): This case is only hit in the Dory opening proof,
+                // which doesn't actually do anything with this value. We should
+                // remove that call from Dory.
+                F::zero()
+            }
+            _ => {
+                let chis = EqPolynomial::evals(r);
+                self.dot_product(&chis)
+            }
```
moodlezoup:
```suggestion
            MultilinearPolynomial::OneHot(poly) => poly.evaluate(r),
            _ => unimplemented!("Unsupported MultilinearPolynomial variant"),
```

jolt-core/src/poly/multilinear_polynomial.rs
```diff
@@ -310,6 +310,18 @@ impl<F: JoltField> MultilinearPolynomial<F> {
...
+            MultilinearPolynomial::LargeScalars(poly) => poly.inside_out_evaluate(r),
+            MultilinearPolynomial::U8Scalars(poly) => poly.inside_out_evaluate(r),
+            MultilinearPolynomial::U16Scalars(poly) => poly.inside_out_evaluate(r),
+            MultilinearPolynomial::U32Scalars(poly) => poly.inside_out_evaluate(r),
+            MultilinearPolynomial::U64Scalars(poly) => poly.inside_out_evaluate(r),
+            MultilinearPolynomial::I64Scalars(poly) => poly.inside_out_evaluate(r),
+            MultilinearPolynomial::OneHot(poly) => poly.evaluate(r),
+            _ => unimplemented!("Unsupported MultilinearPolynomial variant"),
+        }
+    }
```
moodlezoup: since this is only used in the benchmark, let's move it into the benchmark file

jolt-core/src/poly/multilinear_polynomial.rs
```diff
@@ -583,34 +595,216 @@ impl<F: JoltField> PolynomialBinding<F> for MultilinearPolynomial<F> {
...
+                MultilinearPolynomial::I64Scalars(poly) => poly.inside_out_evaluate(r),
+                _ => {
+                    let eq = EqPolynomial::evals(r);
+                    poly.dot_product(&eq)
+                }
+            })
+            .collect();
+        evals
+    }
+}
```
moodlezoup: same here

jolt-core/src/poly/dense_mlpoly.rs
```diff
@@ -250,11 +250,58 @@ impl<F: JoltField> DensePolynomial<F> {
...
+        } else {
+            self.evaluate_split_eq_parallel(eq_one, eq_two)
+        }
+    }
+    fn evaluate_split_eq_parallel(&self, eq_one: &[F], eq_two: &[F]) -> F {
+        let eval: F = (0..eq_one.len())
+            .into_par_iter()
+            .map(|x1| {
+                if eq_one[x1].is_zero() {
+                    F::zero()
```
moodlezoup: Can remove this check, `eq_one` won't have any zeros (assuming that we're evaluating on a random point)

jolt-core/src/poly/dense_mlpoly.rs
```diff
@@ -250,11 +250,58 @@ impl<F: JoltField> DensePolynomial<F> {
...
+            })
+            .reduce(|| F::zero(), |acc, val| acc + val);
+        eval
+    }
+
+    fn evaluate_split_eq_serial(&self, eq_one: &[F], eq_two: &[F]) -> F {
+        let eval: F = (0..eq_one.len())
+            .map(|x1| {
+                if eq_one[x1].is_zero() {
+                    F::zero()
```
moodlezoup: same here

jolt-core/src/utils/counters.rs
moodlezoup, line 1: let's just move this into the `tracked_ark` file

jolt-core/src/field/tracked_ark.rs
```diff
@@ -0,0 +1,329 @@
...
+}
+
+impl SubAssign for TrackedFr {
+    fn sub_assign(&mut self, other: Self) {
+        self.0 -= other.0;
+    }
+}
+
+impl MulAssign for TrackedFr {
+    fn mul_assign(&mut self, other: Self) {
```
moodlezoup, line 168: this should also increment the counter

jolt-core/src/field/tracked_ark.rs
```diff
@@ -0,0 +1,329 @@
...
+    fn product<I: Iterator<Item = Self>>(iter: I) -> Self {
+        iter.fold(Self::one(), |a, b| Self(a.0 * b.0))
+    }
+}
+
+impl<'a> Product<&'a Self> for TrackedFr {
+    fn product<I: Iterator<Item = &'a Self>>(iter: I) -> Self {
+        iter.fold(Self::one(), |a, b| Self(a.0 * b.0))
+    }
+}
```
moodlezoup: increment counter

jolt-core/src/field/tracked_ark.rs
```diff
@@ -0,0 +1,329 @@
...
+        self.0.to_u64()
+    }
+
+    fn square(&self) -> Self {
+        MULT_COUNT.fetch_add(1, Ordering::Relaxed);
+        TrackedFr(self.0.square())
+    }
+
+    fn inverse(&self) -> Option<Self> {
+        self.0.inverse().map(TrackedFr)
```
moodlezoup, line 317: increment counter

jolt-core/src/poly/multilinear_polynomial.rs
0xAndoroid: Maybe this trait in relevant.
```rust
pub trait OptimizedMul<Rhs, Output>: Sized + Mul<Rhs, Output = Output> {
    fn mul_0_optimized(self, other: Rhs) -> Self::Output;
    fn mul_1_optimized(self, other: Rhs) -> Self::Output;
    fn mul_01_optimized(self, other: Rhs) -> Self::Output;
}
```

jolt-core/src/poly/multilinear_polynomial.rs
abiswas3: You mean it will not be 0 with overwhelming probability right?

jolt-core/examples/polynomial_evaluation.rs
abiswas3: Yes sure.

jolt-core/src/poly/multilinear_polynomial.rs
abiswas3: Yes, thanks. I've used this trait but have not been consistent with its use. I'll tidy this up when this is easy.

jolt-core/src/field/tracked_ark.rs
abiswas3: The counter should be incremented for every a.0 * b.0, i believe, will double check and get back to you

jolt-core/src/field/tracked_ark.rs
abiswas3: Mul_assign will eventually call self*other so we do not want to double count.

jolt-core/src/field/tracked_ark.rs
abiswas3: Once again we need to make sure we don't double count. I'll get back to you on this one.

jolt-core/src/poly/multilinear_polynomial.rs
moodlezoup: yes

jolt-core/src/field/tracked_ark.rs
abiswas3: Actually, you're right -- the inner mul_assign mult happens using ark_bn254::Fr. 
The change should be reflected in the next commit.

jolt-core/src/field/tracked_ark.rs
abiswas3: Inversion is a good deal more expensive compared to multiplication.
I've added an inverse counter for this.

jolt-core/src/field/tracked_ark.rs
abiswas3: This should be resolve in the next commit.

jolt-core/examples/tracked_ark.rs
moodlezoup, line 1: this file feels a bit unnecessary, I think the `tracked_ark` API is intuitive enough

jolt-core/src/poly/multilinear_polynomial.rs
moodlezoup: @abiswas3 let's move these functions into the benchmark file if they aren't used elsewhere

jolt-core/src/field/mod.rs
```diff
@@ -77,11 +77,27 @@ pub trait JoltField:
...
-        *self * Self::from_i128(n)
+        if n == 0 || self.is_zero() {
+            Self::zero()
+        } else if n == 1 {
+            *self
+        } else if self.is_one() {
+            Self::from_i128(n)
+        } else {
+            *self * Self::from_i128(n)
+        }
```
moodlezoup: These methods get overridden in `ark.rs`, so we should make the same optimizations there

jolt-core/examples/tracked_ark.rs
abiswas3: This accidentally got added in the commit -- let me remove this file. There are tests for tracked_ark that are sufficient.

jolt-core/src/field/mod.rs
abiswas3: Sorted them out in ark.rs

PR #845, opened 2025-08-07
Better batching sumcheck prep parallelism

markosg04:
- better parallelism for the sumcheck prep in batching, 2x improvement
- makes CI on main happy by fixing typo


moodlezoup:
Looks like you need to update the verifier text fixtures (presumably just `fib_proof.bin`)

PR #853, opened 2025-08-08
prepare sumcheck parallelism (fixed)

markosg04:
- Reintroduces changes from https://github.com/a16z/jolt/pull/845
- fixes the deadlock by changing mutex to rwlock for the shared eq poly


jolt-core/src/poly/one_hot_polynomial.rs
```diff
@@ -280,7 +280,7 @@ impl<F: JoltField> OneHotPolynomialProverOpening<F> {
             let g = mem::take(&mut polynomial.G);
             drop_in_background_thread(g);
             drop(shared_eq);
-            let mut eq_state = self.eq_state.lock().unwrap();
+            let mut eq_state = self.eq_state.write().unwrap();
```
moodlezoup: nit: I just noticed we're dropping `shared_eq` and then immediately calling `write()` again

PR #859, opened 2025-08-11
Feat: transcript optimizations

markosg04:
- Adds method for 128 bit scalars and makes default scalars 128 bits under the hood
- Adds some tests to make sure the 128 bit scalars work as intended
- Immediately benefits performance to all EC scalar muls from transcript
- Adds `challenge_u128` method to pull a u128 for future optimal field mul optimizations
- Implements a Blake2b transcript
- Swaps all KeccakTranscript with Blake2bTranscript
- Refactors the transcript module slightly to be better organized
- Reworks the challenge pulling before `prepare_sumcheck` so that we maintain 128 bit values (as opposed to taking powers which gives us random non-128-bit values)


jolt-core/src/field/ark.rs
```diff
@@ -147,6 +147,11 @@ impl JoltField for ark_bn254::Fr {
         ark_bn254::Fr::from_le_bytes_mod_order(bytes)
     }
 
+    fn from_bytes_128_bits(bytes: &[u8]) -> Self {
+        assert_eq!(bytes.len(), 16);
+        ark_bn254::Fr::from_le_bytes_mod_order(bytes)
+    }
+
```
moodlezoup: eh let's just get rid of the `assert_eq` in `from_bytes` and get rid of this function

PR #861, opened 2025-08-12
Feat/allocative

moodlezoup:


0xAndoroid (approved):

jolt-core/src/utils/profiling.rs
```diff
@@ -1,9 +1,15 @@
...
+use std::path::Path;
 use std::{
     collections::HashMap,
     sync::{LazyLock, Mutex},
 };
+#[cfg(not(target_arch = "wasm32"))]
+use tracing::debug;
 
 static MEMORY_USAGE_MAP: LazyLock<Mutex<HashMap<&'static str, f64>>> =
     LazyLock::new(|| Mutex::new(HashMap::new()));
```
markosg04, line 15: do we still want to keep these now that we have allocative (these maps and their functions)?

jolt-core/src/utils/profiling.rs
```diff
@@ -56,15 +62,47 @@ pub fn print_current_memory_usage(label: &str) {
     if let Some(usage) = memory_stats() {
         let memory_usage_gb = usage.physical_mem as f64 / 1_000_000_000.0;
         if memory_usage_gb >= 1.0 {
-            println!("\"{label}\" current memory usage: {memory_usage_gb} GB");
+            debug!("\"{label}\" current memory usage: {memory_usage_gb:.2} GB");
```
markosg04, line 65: we should probably make a ticket to transition all println! -> debug! and add some info! and warn! around

jolt-core/src/subprotocols/sumcheck.rs
```diff
@@ -69,6 +76,9 @@ pub trait SumcheckInstance<F: JoltField>: Send + Sync {
         accumulator: Rc<RefCell<VerifierOpeningAccumulator<F>>>,
         opening_point: OpeningPoint<BIG_ENDIAN, F>,
     );
+
+    #[cfg(feature = "allocative")]
+    fn update_flamegraph(&self, flamegraph: &mut FlameGraphBuilder);
```
markosg04, line 81: I noticed that pretty much all the `SumcheckInstance` have the same implementation of this. Is there a reason why we can't just have a default implementation, then?

jolt-core/src/utils/profiling.rs
moodlezoup: can't hurt

jolt-core/src/subprotocols/sumcheck.rs
moodlezoup: dyn-compatibility reasons :/

PR #865, opened 2025-08-12
feat: support both nightly and stable guest toolchains

sagar-a16z:
* added support for print statements inside guest programs
* added support for getrandom
* upgrade to rust 1.89.0
* support for nightly toolchain via the "nightly" #provable attribute


jolt-sdk/src/random.rs
```diff
@@ -0,0 +1,39 @@
...
+// Warning: This is a deterministic PRNG implementation. We could allow the prover/verifier to agree on a seed in the future.
+// No atomics support so we're using a static mutable variable
+#[allow(clippy::missing_safety_doc)]
+#[no_mangle]
+pub unsafe fn sys_rand(dest: *mut u8, len: usize) {
+    let rng_ptr: *mut Option<StdRng> = &raw mut RNG;
+
+    unsafe {
+        if (*rng_ptr).is_none() {
+            *rng_ptr = Some(StdRng::seed_from_u64(SEED));
```
moodlezoup, line 21: maybe we should a print here warning about the fact that this is deterministic?

moodlezoup (approved):
lgtm, just one nit –– feel free to merge when addressed

jolt-sdk/src/random.rs
sagar-a16z: yeah that is a good idea. done!

PR #877, opened 2025-08-17
feat: support resumable downloads when installing toolchain

zouguangxian:
This PR add support for resumable download.

* Added functionality to resume downloads if interrupted.
* Implemented checks for partial downloads and adjusted file handling accordingly.

### DIFF
--- a/jolt-core/src/host/toolchain.rs
+++ b/jolt-core/src/host/toolchain.rs
@@ -128,25 +128,76 @@ fn unpack_toolchain(channel: &str) -> Result<()> {
 
 #[cfg(not(target_arch = "wasm32"))]
 async fn download_toolchain(client: &Client, url: &str) -> Result<()> {
+    use tracing::info;
+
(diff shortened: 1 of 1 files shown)


jolt-core/src/host/toolchain.rs
```diff
@@ -130,23 +130,72 @@ fn unpack_toolchain(channel: &str) -> Result<()> {
 async fn download_toolchain(client: &Client, url: &str) -> Result<()> {
     let jolt_dir = jolt_dir();
     let output_path = jolt_dir.join("rust-toolchain.tar.gz");
+    println!("Downloading toolchain to {output_path:?}");
```
0xAndoroid: We've decided to use `info!`, `debug!`, `warn!`, and `error!` macros from `tracing` crate.
Let's just use `info!` here.

0xAndoroid (approved):
lgtm, just one note

PR #879, opened 2025-08-17
feat: make R1CS constraints compile-time constants

quangvdao:


jolt-core/src/poly/spartan_interleaved_poly.rs
```diff
@@ -255,56 +255,44 @@ impl<const NUM_SVO_ROUNDS: usize, F: JoltField> SpartanInterleavedPolynomial<NUM
...
+                                //         .evaluate_row(flattened_polynomials, current_step_idx);
+                                //     let cz = constraint
+                                //         .c
+                                //         .evaluate_row(flattened_polynomials, current_step_idx);
+                                //     if az * bz != cz {
+                                //         panic!(
+                                //             "Constraint violated at step {current_step_idx}",
+                                //         );
+                                //     }
+                                // }
```
0xAndoroid: Can we uncomment and bring back this inline test? It was very useful for debugging.

jolt-core/src/zkvm/r1cs/inputs.rs
```diff
@@ -201,6 +141,107 @@ impl JoltR1CSInputs {
         }
     }
 
+    /// Const version of to_index() for use in const contexts.
+    /// Must be kept in sync with ALL_R1CS_INPUTS ordering.
+    pub const fn to_index_const(&self) -> usize {
```
0xAndoroid: For this function, can we just loop over the array? Also, I think we can just replace the `to_index` function itself.

jolt-core/src/zkvm/r1cs/inputs.rs
```diff
@@ -201,6 +141,107 @@ impl JoltR1CSInputs {
...
+            JoltR1CSInputs::OpFlags(CircuitFlags::Assert) => 37,
+            JoltR1CSInputs::OpFlags(CircuitFlags::DoNotUpdateUnexpandedPC) => 38,
+            JoltR1CSInputs::OpFlags(CircuitFlags::Advice) => 39,
+            JoltR1CSInputs::OpFlags(CircuitFlags::IsNoop) => 40,
+            JoltR1CSInputs::OpFlags(CircuitFlags::IsCompressed) => 41,
+        }
+    }
+
+    /// Convert to CommittedPolynomial if this input represents a committed polynomial
+    pub fn to_committed_polynomial(&self) -> Result<CommittedPolynomial, &'static str> {
```
0xAndoroid: Can we have this function implemented as `TryFrom<&JoltR1CSInputs> for CommittedPolynomial`.
Same for `to_virtual_polynomial` and for `to_opening_id`.

jolt-core/src/zkvm/r1cs/key.rs
```diff
@@ -29,7 +28,7 @@ pub struct UniformSpartanKey<F: JoltField> {
 /// (row, col, value)
 pub type Coeff<F> = (usize, usize, F);
 
-/// Sparse representation of a single R1CS matrix.
+/// (Deprecated) Sparse representation of a single R1CS matrix.
```
0xAndoroid, line 31: Can we just remove these deprecated structs, or are they to be removed in subsequent PRs?

quangvdao: The next PR will remove it.

jolt-core/src/zkvm/r1cs/inputs.rs
quangvdao: I think looping over the array is difficult to make `const`? Let me try again.

quangvdao: So the solution I came to is that the pattern-matching version should be the default, cuz it's the simplest & most efficient. There is the issue of compatibility, which is solved through test.

0xAndoroid: Sounds good.

PR #890, opened 2025-08-19
Feat/inline sequence fuzzer

0xAndoroid:


tracer/src/instruction/test.rs
```diff
@@ -1,295 +1,109 @@
...
+        let register_state =
+            <<I::Format as InstructionFormat>::RegisterState as InstructionRegisterState>::random(
+                &mut rng,
+            );
 
-    // Get the type name at runtime
-    fn get_instruction_name<T: RISCVInstruction>(_instance: &T) -> &'static str {
-        type_name::<T>().split("::").last().unwrap_or("Unknown")
-    }
+        let mut original_cpu = Cpu::new(Box::new(DummyTerminal::default()));
```
danielwlz, line 98: Optional: using CpuTestHarness would save a few lines and make it clearer, especially if one adds a write_register method

Example: https://github.com/a16z/jolt/commit/7262c3f6d68c72ad3d322377d02141ffe30ddf99

0xAndoroid: I think this will be useful in the future, once we also add load and store instructions to the fuzzer. But I'm going to merge this bare bone one for now.

PR #891, opened 2025-08-20
feat: Split host/guest responsibilities and add recursion support with guest module

zouguangxian:
This PR introduces an improvement by splitting host and guest responsibilities, enabling clean separation between program building and execution phases. The refactoring adds a new guest module for recursion support.

### 1. Split Host/Guest Responsibilities
**Host Program (`jolt-core/src/host/program.rs`)** - Building Phase:
- Compilation and ELF generation (`build()`, `build_with_channel()`)
- Configuration management (`set_memory_config()`, `set_func()`)
- Toolchain installation and linker script generation
- ELF content extraction (`get_elf_contents()`)

**Guest Program (`jolt-core/src/guest/program.rs`)** - Execution Phase:
- Pure runtime operations with pre-built ELF contents
- Program decoding and instruction parsing (`decode()`)
- Execution tracing with inputs (`trace()`)
- No compilation dependencies - lightweight and embeddable

### 2. Refactor zkvm/mod.rs Dependencies
- Remove dependencies on `Program` struct in zkvm module
- Prove functions now only require `elf_contents` instead of full Program objects

### 3. Add Guest Module for Recursion Support
- **New guest module structure**: `jolt-core/src/guest/`
  - `program.rs`: Lightweight guest program representation
  - `prover.rs`: Prover functions for guest program execution
  - `verifier.rs`: Verifier functions for guest program validation
- **Clean abstraction**: Guest programs work directly with ELF contents

### 4. Update jolt-sdk for Guest Verifier Support
- Add `ark_bn254::Fr` dependency for guest verifier functionality

### 5. Add Memory Configuration Export Function
- **`make_memory_config_fn`**: Exports function to get memory configuration

### 6. Add Comprehensive Recursion Example
- **New recursion example**: `examples/recursion/`
  - Demonstrates the new host/guest architecture in action
  - Supports both embedded bytes and parameter-based proof passing

### Modular Architecture
- **Clear separation**: Build-time vs runtime responsibilities
- **Dependency reduction**: Guest programs only need ELF contents
- **Reusability**: Build once, run many times with different inputs
- **Embeddability**: Lightweight guest programs suitable for constrained environments


examples/recursion/src/main.rs
```diff
@@ -0,0 +1,484 @@
...
+    program.set_std(true);
+    program.set_memory_config(memory_config);
+    program.build(target_dir);
+    let elf_contents = program.get_elf_contents().unwrap();
+    let recursion = jolt_sdk::guest::program::Program::new(&elf_contents, &memory_config);
+
+    let recursion_prover_preprocessing =
+        jolt_sdk::guest::prover::preprocess(&recursion, max_trace_length);
+    let recursion_verifier_preprocessing =
+        jolt_sdk::guest::verifier::preprocess(&recursion, max_trace_length);
```
sagar-a16z, line 371: you can construct verifier preprocessing form the prover preprocessing object directly instead of tracing twice.

examples/recursion/src/main.rs
```diff
@@ -0,0 +1,484 @@
...
+    let elf_contents = program.get_elf_contents().unwrap();
+    println!("Creating guest program...");
+    let guest_prog = jolt_sdk::guest::program::Program::new(&elf_contents, &memory_config);
+
+    println!("Preprocessing guest prover...");
+    let guest_prover_preprocessing =
+        jolt_sdk::guest::prover::preprocess(&guest_prog, max_trace_length);
+    println!("Preprocessing guest verifier...");
+    let guest_verifier_preprocessing =
+        jolt_sdk::guest::verifier::preprocess(&guest_prog, max_trace_length);
```
sagar-a16z, line 249: Same here, get the verifier preprocessing from the prover's preprocessing obj

jolt-sdk/Cargo.toml
```diff
@@ -29,8 +29,10 @@ guest-std = [
 ]
 guest-verifier = [
     "guest-std",
+    "dep:tracer",
     "dep:common",
-    "jolt-core/parallel"
+    "jolt-core/prover",
```
sagar-a16z, line 34: Need to review this. The intention of the "guest-verifier" feature was to skip the prover bits from jolt-core but allow building the verifier components. 

Some things inside of jolt-core/prover might not be compatible with riscv compilation (I know it works for now - but better to have this separation). 

Maybe some of this naming has to change or you need to use `guest-verifier` strictly for guest builds and use `host` for host builds.

jolt-sdk/src/host_utils.rs
```diff
@@ -1,4 +1,4 @@
-#[cfg(feature = "host")]
+#[cfg(any(feature = "host", feature = "guest-verifier"))]
```
sagar-a16z, line 1: consider removing this line altogether. This file only included if this condition is true. 
In lib.rs
```
#[cfg(any(feature = "host", feature = "guest-verifier"))]
pub mod host_utils;
```

examples/recursion/guest/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
```
sagar-a16z, line 1: Does this example make the `examples/verifier` redundant? Should we remove that from the code base? 

Maybe this example can do what the verifier example does already.

examples/recursion/guest/src/lib.rs
```diff
@@ -0,0 +1,62 @@
...
+    let use_embedded = !embedded_bytes::EMBEDDED_BYTES.is_empty();
+    let data_bytes = if use_embedded {
+        embedded_bytes::EMBEDDED_BYTES
+    } else {
+        bytes
+    };
+
+    let mut cursor = std::io::Cursor::new(data_bytes);
+    start_cycle_tracking("deserialize");
+    let _ = u32::deserialize_compressed(&mut cursor).unwrap();
```
sagar-a16z, line 30: Why is this first u32 deserialize needed?

examples/recursion/guest/src/lib.rs
```diff
@@ -0,0 +1,62 @@
...
+    end_cycle_tracking("deserialize");
+
+    start_cycle_tracking("preprocessing");
+    let verifier_preprocessing: JoltVerifierPreprocessing<F, PCS> =
+        JoltVerifierPreprocessing::<F, PCS>::deserialize_compressed(&mut cursor).unwrap();
+    end_cycle_tracking("preprocessing");
+
+    start_cycle_tracking("n");
+    let n: u32 = u32::deserialize_compressed(&mut cursor).unwrap();
```
sagar-a16z, line 39: and this one, maybe add a comment

jolt-sdk/Cargo.toml
```diff
@@ -29,8 +29,10 @@ guest-std = [
...
+    "dep:ark-bn254",
```
sagar-a16z, line 35: Why is the guest-verifier directly dependent on ark? it should be sufficient to have a transitive dependency via jolt-core/parallel (no prover)

sagar-a16z (changes requested):
I think this example looks really good. Great work. 

Might be helpful to have a README.md in there that explains what it does and how to use it?

jolt-core/src/guest/prover.rs
```diff
@@ -0,0 +1,48 @@
...
+    let mut memory_config = guest.memory_config;
+    memory_config.program_size = Some(program_size);
+    let memory_layout = MemoryLayout::new(&memory_config);
+
+    JoltRV32IM::prover_preprocess(bytecode, memory_layout, memory_init, max_trace_length)
+}
+
+#[allow(clippy::type_complexity)]
+#[cfg(feature = "prover")]
+pub fn prove<F, PCS, FS>(
```
sagar-a16z, line 28: Why does the guest have "prove" ? Maybe gate this module on the cfg feature = prover

examples/recursion/guest/src/lib.rs
zouguangxian: Just for checking the bytes is correct, it can be removed.

jolt-core/src/guest/prover.rs
zouguangxian: "guest" here refers to functionalities or features that serve the guest program at **runtime**, not to code that is executed within the guest program itself.
"host" here refers to functionalities at **build-time**.

But it can be organized better to avoid misunderstanding.

jolt-sdk/Cargo.toml
zouguangxian: If deserialize preprocessing only and without using it in JoltRV32IM::verify function
```rust
let verifier_preprocessing =
    JoltVerifierPreprocessing::deserialize_compressed(&mut cursor).unwrap();
```
Then it will throw errors when compiling:
```error[E0283]: type annotations needed
  --> examples/recursion/guest/src/lib.rs:37:5
   |
37 |     JoltVerifierPreprocessing::deserialize_compressed(&mut cursor).unwrap();
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^ cannot infer type for type parameter `F` declared on the struct `JoltVerifierPreprocessing`
```

PR #904, opened 2025-08-23
fix: trim trailing newline from toolchain tag in download URL

GarmashAlex:
TOOLCHAIN_TAG in URL: include_str! almost certainly picks up the final translation of the string from guest-toolchain-tag. In toolchain_url, this will end up in the URL and result in a 404 error.

The fix trims the TOOLCHAIN_TAG constant before using it in the URL formation,
ensuring clean URLs while maintaining the original tag content for local
toolchain verification.

this pr fixes toolchain download failures caused by malformed URLs.


0xAndoroid:
`guest-toolchain-tag` doesn't have a new line at the end of the file.

PR #905, opened 2025-08-24
fix: correct uninterleave length calculation for odd bit lengths

VolodymyrBg:
**Problem:**
LookupBits::uninterleave() incorrectly calculated bit lengths for odd lengths.
For len=5, it returned x.len=2, y.len=3, but uninterleave_bits() actually
distributes bits as 3 even + 2 odd positions.

**Solution:**
Fixed length calculation:
- x.len = (self.len + 1) / 2  (even positions)
- y.len = self.len / 2         (odd positions)

**Why:**
- Prevents bugs in lookup table operations
- Ensures correct bit distribution for all lengths
- Maintains x.len + y.len = self.len invariant

**Testing:**
Added tests for even/odd lengths, edge cases, and full range validation.

### DIFF
--- a/jolt-core/src/utils/lookup_bits.rs
+++ b/jolt-core/src/utils/lookup_bits.rs
@@ -22,8 +22,13 @@ impl LookupBits {
 
     pub fn uninterleave(&self) -> (Self, Self) {
         let (x_bits, y_bits) = uninterleave_bits(self.bits);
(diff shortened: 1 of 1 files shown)


0xAndoroid:
To discourage PR farming, minimal PRs are not accepted.

PR #906, opened 2025-08-25
[JOLT-129] Omid/inline bigint

omibo:
This PR implements inline multiplication for 256-bit integers, producing a 512-bit output. Each integer is represented as 4 limbs, with each limb being 64 bits.


danielwlz:
LGTM. We should work in fuzzer coverage to inlines like this (#890) for greater assurance of correctness -- but the fuzzer doesn't yet support instructions where we write back the result to memory. If you have cycles you should check that out (I don't think it's a huge change to fuzzer), but I don't think it's a must-have to merge this PR.

jolt-inlines/bigint/src/lib.rs
```diff
@@ -0,0 +1,33 @@
...
+        "BIGINT256_MUL",
+        std::boxed::Box::new(exec::bigint_mul_exec),
+        std::boxed::Box::new(trace_generator::bigint_mul_sequence_builder),
+    )?;
+    Ok(())
+}
+
+// Automatic registration when the library is loaded
+#[cfg(feature = "host")]
+#[ctor::ctor]
```
0xAndoroid, line 28: Put this behind `not(target = wasm)`, since `ctor` doesn't work on wasm.

jolt-inlines/bigint/src/lib.rs
```diff
@@ -0,0 +1,33 @@
...
+pub mod multiplication;
+pub use multiplication::*;
+
+#[cfg(feature = "host")]
+use tracer::register_inline;
+
+// Initialize and register inlines
+#[cfg(feature = "host")]
+pub fn init_inlines() -> Result<(), String> {
+    // Register 256-bit Int multiplication with funct3=0x00 and funct7=0x01
```
0xAndoroid:
```suggestion
    // Register 256-bit Int multiplication with funct3=0x00 and funct7=0x02
```
Seems to be what is being registered.

0xAndoroid (commented):
Thanks, just a couple of nits.
Bot's comments also seem to be relevant.

PR #913, opened 2025-08-26
feat: enhance RISC-V decoder and cargo execution logging

zouguangxian:
This PR includes two improvements to enhance the development experience and fix a decoding issue:

### Changes
1. **Enhanced Cargo Execution Logging** 
   - Added logging of the composed command line for the `cargo` build process
   - Improves visibility during execution for better debugging
   - Modified: `jolt-core/src/host/program.rs`

2. **RISC-V Decoder Zero Padding Fix** 
   - Added a check to skip processing when the compressed instruction is zero
   - Ensures correct decoding of subsequent instructions
   - Modified: `tracer/src/lib.rs`

### DIFF
--- a/tracer/src/lib.rs
+++ b/tracer/src/lib.rs
@@ -313,6 +313,10 @@ pub fn decode(elf: &[u8]) -> (Vec<RV32IMInstruction>, Vec<(u64, u8)>, u64, Xlen)
                 if (first_halfword & 0b11) != 0b11 {
                     // Compressed 16-bit instruction
                     let compressed_inst = first_halfword;
+                    if compressed_inst == 0x0000 {
+                        offset += 2;
+                        continue;
+                    }
 
                     if let Ok(inst) = RV32IMInstruction::decode(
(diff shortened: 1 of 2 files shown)


tracer/src/lib.rs
```diff
@@ -313,6 +313,10 @@ pub fn decode(elf: &[u8]) -> (Vec<RV32IMInstruction>, Vec<(u64, u8)>, u64, Xlen)
                 if (first_halfword & 0b11) != 0b11 {
                     // Compressed 16-bit instruction
                     let compressed_inst = first_halfword;
+                    if compressed_inst == 0x0000 {
+                        offset += 2;
+                        continue;
+                    }
```
0xAndoroid, line 319: Does such thing ever happen in practice?
I'm pretty sure if we just skip it here our verification will fail. This is basically a compresed noop instruction, and we can't have noops in the bytecode/trace at all (except for padding).

0xAndoroid: I verified that this is okay. Specifically, bytecode can contain 0x0000, but this instruction would never be executed.

PR #914, opened 2025-08-26
[JOLT-199] fix: `commit_rows` stack overflow

moodlezoup:
still not 100% what was taking up so much stack space, but this seems to fix it
Uses functions added in https://github.com/a16z/arkworks-algebra/pull/14


0xAndoroid (approved):
I'm not sure what could be causing an issue either. Everything seems to be heap allocated, maybe just some bad parallelism.
Thanks for taking a look.

PR #915, opened 2025-08-26
Implement streaming Dory

jprider63:
This draft PR implements the streaming version of Dory commitments. The commitments and proof hints match the non-streaming versions. 

@zutshi and I still need to:
- Add parallelization
- Benchmark and optimize
- Cleanup code

One thing to note is that we were trying to avoid padding the trace length to a power of two when computing the commitments. It turns out the padding is necessary since the one hot polynomials treat the padded values differently. Some of the one hots were padding with `Some(0)` while others were padding with `None`. 

---

### DIFF
--- a/jolt-core/src/zkvm/dag/jolt_dag.rs
+++ b/jolt-core/src/zkvm/dag/jolt_dag.rs
@@ -969,7 +969,7 @@ impl JoltDAG {
     >(
         state_manager: &mut StateManager<'a, F, ProofTranscript, PCS>,
     ) -> Option<PCS::OpeningProofHint> {
-        let (preprocessing, _, program_io, _) = state_manager.get_prover_data();
+        let (preprocessing, _, _, program_io, _) = state_manager.get_prover_data();
 
         if program_io.untrusted_advice.is_empty() {
             return None;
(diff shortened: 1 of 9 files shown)


jprider63:
I'm not sure if it makes sense since I haven't looked into the details, but I just tried naively parallelizing the pairing operations. I'm seeing ~7s faster prover times. Here's the flamegraph. 
[trace-881a720b4c66008232b4bdd609a511ff492e2a43_streaming_par_pairing.json](https://github.com/user-attachments/files/22649808/trace-881a720b4c66008232b4bdd609a511ff492e2a43_streaming_par_pairing.json)

jolt-core/src/zkvm/dag/jolt_dag.rs
```diff
@@ -478,36 +482,108 @@ impl JoltDAG {
         'a,
         F: JoltField,
         ProofTranscript: Transcript,
-        PCS: CommitmentScheme<Field = F>,
+        PCS: StreamingCommitmentScheme<Field = F>,
```
markosg04: would it be possible to feature gate streaming vs non streaming?

PR #917, opened 2025-08-27
fix(msm): avoid i64::MIN overflow in I64Scalars MSM

radik878:
1. Replace unsafe (-scalar) as u64 with scalar.unsigned_abs() to safely handle i64::MIN.
2. Move length check before parallel partitioning to avoid unnecessary work on mismatched inputs.
3. Keeps existing MSM structure (pos - neg) while removing UB risk.

### DIFF
--- a/jolt-core/src/msm/mod.rs
+++ b/jolt-core/src/msm/mod.rs
@@ -67,7 +67,6 @@ where
 
             // TODO: Check if this is the fastest way forward.
             MultilinearPolynomial::I64Scalars(poly) => {
-                // Validate input sizes before doing any heavy work.
                 if bases.len() != poly.coeffs.len() {
                     return Err(ProofVerifyError::KeyLengthError(
                         bases.len(),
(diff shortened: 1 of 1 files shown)


jolt-core/src/msm/mod.rs
```diff
@@ -67,6 +67,14 @@ where
 
             // TODO: Check if this is the fastest way forward.
             MultilinearPolynomial::I64Scalars(poly) => {
+                // Validate input sizes before doing any heavy work.
```
markosg04: delete this comment

jolt-core/src/msm/mod.rs
```diff
@@ -83,7 +91,8 @@ where
                                 pos_s.push(scalar as u64);
                                 pos_b.push(*base);
                             } else if scalar < 0 {
-                                neg_s.push((-scalar) as u64);
+                                // Use unsigned_abs to safely handle i64::MIN without overflow.
```
markosg04: and this comment

radik878: > and this comment

deleted

PR #922, opened 2025-09-02
Omid/refactor inlines

omibo:
This PR refactors inlines and their related functionalities. The changes are as follows:

1. Introduce a new format, `FormatInline`, which is specific to inlines and uses `rs3` instead of `rd`.
2. Add a common crate to the inlines folder that includes constants defining the `opcode`, `funct3`, and `funct7` of each inline.
3. Add a `write_to_file()` function, enabling inlines to store their trace in a text file.


tracer/src/utils/inline_helpers.rs
```diff
@@ -77,22 +79,74 @@ pub struct InstrAssembler {
...
     /// and return ownership of the underlying `Vec`.
+    ///
+    /// For FormatInline instructions, this also adds instructions to zero out
+    /// all virtual registers (32-127).
     pub fn finalize(mut self) -> Vec<RV32IMInstruction> {
+        // If this is a FormatInline sequence, add instructions to zero out all virtual registers
+        if self.is_format_inline {
+            // Zero out all virtual registers from 32 to 127
+            // ADDI rd, x0, 0 sets rd = 0
+            for rd in RISCV_REGISTER_COUNT..REGISTER_COUNT {
```
0xAndoroid: We should zero out only registers that were touched by the inline, not everything.
Additionally, some virtual registers will be used as native ones, so it's best to not rely on these constants.

tracer/src/utils/inline_helpers.rs
```diff
@@ -77,22 +79,74 @@ pub struct InstrAssembler {
...
     /// Accumulated instruction buffer.
     sequence: Vec<RV32IMInstruction>,
+    /// Indicates whether this instruction uses the FormatInline instruction format.
+    ///
+    /// FormatInline is a custom format used for inline operations that guarantees
+    /// no RISC-V registers are modified. When `is_format_inline` is true:
+    /// - We enforce that no RISC-V registers can be written to during execution
+    /// - All virtual register values are reset to 0 in the finalize step to ensure
+    ///   the no-register-modification invariant is maintained
+    is_format_inline: bool,
```
0xAndoroid: We should handle zeroing of registers through allocator. I.e. in the allocator we would have a tracker that is enabled at the beginning of virtual sequence through the first call of allocate_virtual_register.
And then we would have a function that returns virtual registers that were used.

tracer/src/utils/inline_helpers.rs
```diff
@@ -77,22 +79,74 @@ pub struct InstrAssembler {
...
+    ///
+    /// FormatInline instructions must not modify RISC-V registers, so rd must be
+    /// a virtual register (>= 32) if it's used at all.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `is_format_inline` is true and `rd` is in the range [0, 32).
+    #[inline]
+    fn validate_rd_for_inline(&self, rd: u8) {
+        if self.is_format_inline && rd < RISCV_REGISTER_COUNT {
```
0xAndoroid: We should have the 32 base risc-v registers, then 8 (later will be reduced) registers for the inline sequences, and the remaining virtual registers for inline precompiles. Those 8 registers also can't be used. 
Allocation of these registers would also go in the allocator.
I think the best way to do it is to have a `pub(crate)` allocate function that allocates a register for inline sequence, and `pub` function for inline precompile allocation.

tracer/src/utils/inline_helpers.rs
```diff
@@ -77,22 +79,74 @@ pub struct InstrAssembler {
...
+    ///
+    /// For FormatInline instructions, this also adds instructions to zero out
+    /// all virtual registers (32-127).
     pub fn finalize(mut self) -> Vec<RV32IMInstruction> {
+        // If this is a FormatInline sequence, add instructions to zero out all virtual registers
+        if self.is_format_inline {
+            // Zero out all virtual registers from 32 to 127
+            // ADDI rd, x0, 0 sets rd = 0
+            for rd in RISCV_REGISTER_COUNT..REGISTER_COUNT {
+                let zero_instr = ADDI::from(NormalizedInstruction {
```
0xAndoroid: We can use the `emit_i::<ADDI>()` here

tracer/src/utils/inline_helpers.rs
```diff
@@ -77,22 +79,74 @@ pub struct InstrAssembler {
...
+    /// - We enforce that no RISC-V registers can be written to during execution
+    /// - All virtual register values are reset to 0 in the finalize step to ensure
+    ///   the no-register-modification invariant is maintained
+    is_format_inline: bool,
 }
 
 impl InstrAssembler {
     /// Create a new assembler with an empty instruction buffer.
-    pub fn new(address: u64, is_compressed: bool, xlen: Xlen) -> Self {
+    pub fn new(address: u64, is_compressed: bool, xlen: Xlen, is_format_inline: bool) -> Self {
```
0xAndoroid: The bool argument isn't the best thing for public APIs.
Let's add a `pub(crate) new_inline_sequence()` function that constructs thing for inline sequences, and this `pub new()` will construct for inline precompiles.

jolt-inlines/bigint/src/lib.rs
```diff
@@ -5,21 +5,57 @@
...
+    let instructions = trace_generator::bigint_mul_sequence_builder(
+        sequence_inputs.address,
+        sequence_inputs.is_compressed,
+        Xlen::Bit64,
+        sequence_inputs.rs1,
+        sequence_inputs.rs2,
+        sequence_inputs.rd,
+    );
+    write_inline_trace(
+        "bigint_mul256_trace.txt",
```
0xAndoroid:
```suggestion
        "bigint_mul256_trace.joltinline",
```

jolt-inlines/common/src/trace_writer.rs
```diff
@@ -0,0 +1,150 @@
...
+use std::path::Path;
+use tracer::instruction::RV32IMInstruction;
+
+// Default constants for SequenceInputs
+pub const DEFAULT_RAM_START_ADDRESS: u64 = 0x80000000;
+pub const DEFAULT_XLEN: u32 = 64;
+pub const DEFAULT_RS1: u8 = 10;
+pub const DEFAULT_RS2: u8 = 11;
+pub const DEFAULT_RS3: u8 = 12;
+pub const DEFAULT_IS_COMPRESSED: bool = false;
```
0xAndoroid: Why are we having default values here?

jolt-inlines/common/src/trace_writer.rs
```diff
@@ -0,0 +1,150 @@
...
+
+/// Input parameters for instruction sequence generation
+#[derive(Debug, Clone)]
+pub struct SequenceInputs {
+    /// Memory address of the instruction
+    pub address: u64,
+    /// Whether the instruction is compressed
+    pub is_compressed: bool,
+    /// CPU architecture width (32 or 64)
+    pub xlen: u32,
```
0xAndoroid: XLen enum from tracer

jolt-inlines/common/src/trace_writer.rs
```diff
@@ -0,0 +1,150 @@
...
+    writeln!(
+        file,
+        "{:#010x}, {}, {}, {}, {}, {}",
+        sequence_inputs.address,
+        sequence_inputs.is_compressed,
+        sequence_inputs.xlen,
+        sequence_inputs.rs1,
+        sequence_inputs.rs2,
+        sequence_inputs.rs3
+    )?;
```
0xAndoroid: I think there's more coordination required on the format that inlines are being stored in. 
Generally, I don't like storing arbitrary default values. We should at least store placeholders.
But I'm not too sure how this inline file storage thing should work.

jolt-inlines/sha2/src/host.rs
```diff
@@ -2,32 +2,101 @@
...
+    );
+    write_inline_trace(
+        "sha256_trace.txt",
+        &inline_info,
+        &sequence_inputs,
+        &instructions,
+        false, // Don't append for the first one
+    )?;
+
+    // Store SHA256 init inline trace (append to the same file)
```
0xAndoroid, line 58: Why are we appending to the same file?

jolt-inlines/common/src/trace_writer.rs
```diff
@@ -0,0 +1,150 @@
...
+/// * `inline_info` - Descriptor containing inline instruction information
+/// * `sequence_inputs` - Input parameters for the instruction sequence
+/// * `instructions` - Slice of RV32IMInstruction to write
+/// * `append` - If true, append to existing file; if false, overwrite
+pub fn write_inline_trace(
+    file_path: impl AsRef<Path>,
+    inline_info: &InlineDescriptor,
+    sequence_inputs: &SequenceInputs,
+    instructions: &[RV32IMInstruction],
+    append: bool,
```
0xAndoroid, line 104:
```suggestion
```

jolt-inlines/keccak256/src/host.rs
```diff
@@ -2,24 +2,67 @@
 pub use crate::exec;
 pub use crate::trace_generator;
 
+#[cfg(feature = "host")]
```
0xAndoroid: I think `host.rs` is already behind the feature = "host"

jolt-inlines/keccak256/src/host.rs
```diff
@@ -2,24 +2,67 @@
...
+    let instructions = trace_generator::keccak256_inline_sequence_builder(
+        sequence_inputs.address,
+        sequence_inputs.is_compressed,
+        Xlen::Bit64,
+        sequence_inputs.rs1,
+        sequence_inputs.rs2,
+        sequence_inputs.rs3,
+    );
+    write_inline_trace(
+        "keccak256_trace.txt",
```
0xAndoroid:
```suggestion
        "keccak256_trace.joltinline",
```

jolt-inlines/keccak256/Cargo.toml
```diff
@@ -6,11 +6,13 @@ description = "Keccak-256 inline implementations for Jolt VM"
 
 [features]
 default = []
-host = ["dep:ctor", "dep:tracer"]
+host = ["dep:ctor", "dep:tracer", "dep:jolt-inlines-common"]
+save_trace = ["host"]
```
0xAndoroid: I'm not a fan of this feature flag. I think it's okay to not have it and always save the inline to the file, since that's the way we are moving forward with. I believe we should remove the other way we initialize inlines as well.

jolt-inlines/sha2/src/trace_generator.rs
```diff
@@ -368,3 +368,20 @@ pub fn sha2_init_inline_sequence_builder(
     );
     builder.build()
 }
+
+#[cfg(test)]
+mod test {
+    use crate::trace_generator::sha2_inline_sequence_builder;
+
+    #[test]
+    fn print_sequence() {
```
0xAndoroid: What is this test testing?

tracer/src/instruction/format/format_inline.rs
```diff
@@ -0,0 +1,127 @@
...
+//!
+//! ## Relationship to FormatR
+//!
+//! FormatInline is very similar to FormatR instructions, with one key difference: we use `rs3`
+//! instead of `rd`. This reflects the fact that inline operations don't have a destination
+//! register in the traditional sense.
+//!
+//! In program SDKs where inline assembly is called, to avoid introducing a new format to the
+//! `asm!` macro, FormatR is still used for inline instructions in the assembly code. However,
+//! these instructions are parsed as FormatInline instructions by the tracer.
```
0xAndoroid: Too verbose

0xAndoroid (changes requested):
First pass over the PR. 
Ultimately, I think we need to agree on the format of the `.joltinline` file, I'll create a thread in slack.

jolt-inlines/sha2/src/host.rs
omibo: We generate fewer files for inlines that have many modes. One example is Blake2b, for which we plan to support at least 5 modes.

jolt-inlines/common/src/trace_writer.rs
omibo: Each inline requires the following inputs to generate the sequence:
```
pub fn keccak256_inline_sequence_builder(
    address: u64,
    is_compressed: bool,
    xlen: Xlen,
    operand_rs1: u8,
    operand_rs2: u8,
    _operand_rs3: u8,
)
```
To generate the sequence, I need to pass actual values for these inputs. Later, when storing the files, we can replace these values with placeholders. However, each register should have a unique placeholder so that we can distinguish them afterward.

jolt-inlines/common/src/trace_writer.rs
omibo: The main part of writing to the file is:
`
for instruction in instructions {
      writeln!(file, "{instruction:?}")?;
  }
`
  
We can define a custom placeholder for each register and replace the registers with their corresponding placeholders. Personally I prefer above default values as they are more readable when looking in the trace. Though I can use placeholders, if you prefer.

jolt-inlines/common/src/trace_writer.rs
omibo: Needed for appending to a file.

tracer/src/utils/inline_helpers.rs
omibo: In the recent commit, I only zeroed out the virtual registers allocated by the inline.

tracer/src/utils/inline_helpers.rs
omibo: Since virtual registers for inlines are allocated consecutively and we only allocate NEEDED_REGISTERS, we can zero them out easily without needing a tracker.

tracer/src/utils/inline_helpers.rs
omibo: Addressed in the recent commit.

jolt-inlines/common/src/trace_writer.rs
omibo: I used $RS1, $RS2, $RS3, and $ADDR as placeholders to keep the code readable while still making use of placeholders.

jolt-inlines/common/src/trace_writer.rs
```diff
@@ -0,0 +1,279 @@
...
+        // Path to the keccak256 trace file
+        let trace_file_path = Path::new("keccak256_trace.joltinline");
+
+        // Ensure the file is deleted at the end of the test
+        let _cleanup = FileCleanup(trace_file_path);
+
+        // Now read and validate the file
+        let file = fs::File::open(trace_file_path).expect("Failed to open test trace file");
+        let reader = BufReader::new(file);
+        let mut lines = reader.lines();
```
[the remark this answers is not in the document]
omibo: The file is created automatically when the crate is loaded.

jolt-inlines/sha2/src/trace_generator.rs
```diff
@@ -103,7 +103,7 @@ impl Sha256SequenceBuilder {
             let src = self.vr(*ch);
             self.asm.emit_s::<SW>(self.operand_rs2, src, (i as i64) * 4);
         }
-        self.asm.finalize()
+        self.asm.finalize_inline(NEEDED_REGISTERS)
```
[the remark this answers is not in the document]
omibo: NEEDED_REGISTERS is u8

tracer/src/utils/inline_helpers.rs
```diff
@@ -104,11 +122,47 @@ impl InstrAssembler {
...
+        // Zero inline virtual registers using ADDI rd, x0, 0
+        for reg_offset in 0..num_inline_virtual_regs {
+            self.emit_i::<ADDI>(INLINE_REGISTER_BASE + reg_offset, 0, 0);
+        }
+        self.finalize()
+    }
+
+    /// Validates that rd is an inline virtual register (not RISC-V or reserved virtual registers).
+    #[inline]
+    fn is_valide_virtual_rd(virtual_rd: u8) -> bool {
```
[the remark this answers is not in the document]
omibo: Fixed

PR #928, opened 2025-09-08
BLAKE2b and BLAKE3 Inlines

omibo:
This code adds BLAKE2b and BLAKE3 inlines, along with some inline refactoring.

# Updates in Detail
- Implemented a BLAKE2b inline with a full SDK that supports arbitrary input length (no keyed hashing support yet).
- Implemented a BLAKE3 inline with an SDK that supports inputs up to 64 bytes, including keyed hashing.
- Removed `exec()` blocks from inline instructions, as they were never invoked; updated the tests accordingly.
- Made inlines/common `no_std` and reused its constants across other inline SDKs.


examples/blake2-ex/guest/src/lib.rs
```diff
@@ -0,0 +1,21 @@
...
+    // Split the 64-byte hash into two 32-byte arrays
+    let mut first_half = [0u8; 32];
+    // let mut second_half = [0u8; 32];
+
+    first_half.copy_from_slice(&hash[0..32]);
+    // second_half.copy_from_slice(&hash[32..64]);
+
+    // (first_half, second_half)
+    first_half
+}
```
0xAndoroid: why not just return `[u8; 64]`?

jolt-inlines/bigint/src/multiplication/exec.rs
```diff
@@ -49,42 +44,3 @@ pub fn bigint_mul(lhs: [u64; INPUT_LIMBS], rhs: [u64; INPUT_LIMBS]) -> [u64; OUT
     }
     result
 }
-
```
0xAndoroid, line 52: Can we remove the `bigint_mul` function as well?

jolt-inlines/bigint/src/multiplication/test_utils.rs
```diff
@@ -78,10 +79,9 @@ impl BigIntCpuHarness {
                 rs2: Self::RS2,
                 rs3: Self::RS3,
             },
-            // BIGINT256_MUL has opcode 0x0B, funct3 0x00, funct7 0x02
-            opcode: 0x0B,
-            funct3: 0x00,
-            funct7: 0x02,
+            opcode: constants::INLINE_OPCODE,
+            funct3: constants::bigint::mul256::FUNCT3,
+            funct7: constants::bigint::mul256::FUNCT7, // Updated to match new constant value (was 0x02)
```
0xAndoroid:
```suggestion
            funct7: constants::bigint::mul256::FUNCT7,
```

jolt-inlines/blake2/src/exec.rs
```diff
@@ -0,0 +1,108 @@
+use crate::{IV, SIGMA};
+
+/// Rust implementation of BLAKE2 compression on the host.
+pub fn execute_blake2b_compression(state: &mut [u64; 8], message_words: &[u64; 18]) {
```
0xAndoroid, line 4: didn't we remove the exec functions?

jolt-inlines/blake3/src/exec.rs
```diff
@@ -0,0 +1,75 @@
+use crate::IV;
+
+/// Rust implementation of BLAKE3 compression on the host.
+/// The following code is obtained from reference BLAKE3 implementation (https://github.com/BLAKE3-team/BLAKE3/blob/master/reference_impl/reference_impl.rs)
+pub fn execute_blake3_compression(
```
0xAndoroid, line 5: exec function

jolt-inlines/common/src/constants.rs
```diff
@@ -30,12 +30,24 @@ pub mod keccak256 {
...
+    pub const NAME: &str = "BLAKE3_INLINE";
+}
+
 /// BigInt operations inline instruction constants
 pub mod bigint {
     /// 256-bit multiplication
     pub mod mul256 {
         pub const FUNCT3: u32 = 0x00;
-        pub const FUNCT7: u32 = 0x02;
+        pub const FUNCT7: u32 = 0x04; // Changed from 0x02 to avoid collision with blake2
```
0xAndoroid:
```suggestion
        pub const FUNCT7: u32 = 0x04;
```

jolt-inlines/sha2/src/tests.rs
```diff
@@ -44,39 +43,10 @@ mod exec {
     }
 }
 
-mod exec_trace_equivalence {
-    use crate::test_constants::TestVectors;
-    use crate::test_utils::{sverify, Sha256CpuHarness};
-
+mod cpb {
```
0xAndoroid: What's cpb?

tracer/src/instruction/mod.rs
```diff
@@ -573,6 +575,8 @@ define_rv32im_enums! {
         LRW, SCW, AMOSWAPW, AMOADDW, AMOANDW, AMOORW, AMOXORW, AMOMINW, AMOMAXW, AMOMINUW, AMOMAXUW,
         // RV64A (Atomic Memory Operations)
         LRD, SCD, AMOSWAPD, AMOADDD, AMOANDD, AMOORD, AMOXORD, AMOMIND, AMOMAXD, AMOMINUD, AMOMAXUD,
+        // Virtual instructions required for Blake2
+        VirtualXORROT32, VirtualXORROT24, VirtualXORROT16, VirtualXORROT63,
```
0xAndoroid: Let's comment these out for now, before we have the lookup tables for them

tracer/src/instruction/virtual_xor_rot.rs
```diff
@@ -0,0 +1,180 @@
...
+                cpu.x[self.operands.rd as usize] = rotated as i64;
+            }
+        }
+
+        impl RISCVTrace for $name {}
+    };
+}
+
+// Generate the 4 virtual XOR-rotate instructions for 64-bit values
+declare_virtual_xor_rot_64!(VirtualXORROT32, 32);
```
0xAndoroid: I think it might be better to have these as `VirtualXORROT<const ROT_AMOUNT>`, though not sure.

tracer/src/instruction/virtual_xor_rot.rs
```diff
@@ -0,0 +1,180 @@
...
+declare_virtual_xor_rot_64!(VirtualXORROT24, 24);
+declare_virtual_xor_rot_64!(VirtualXORROT16, 16);
+declare_virtual_xor_rot_64!(VirtualXORROT63, 63);
+
+#[cfg(test)]
+mod tests {
```
0xAndoroid: these tests are not needed since we'll have the tests in jolt-core for this instruction

tracer/src/utils/inline_helpers.rs
```diff
@@ -507,6 +507,10 @@ impl InstrAssembler {
         }
     }
 
+    pub fn rotr64(&mut self, rs1: Value, amount: u32, rd: u8) -> Value {
+        self.rotl64(rs1, 64 - amount, rd)
```
0xAndoroid: `rotl64` does rotri with 64-n. It's better to change this one to call `rotri`, but `rotl64` to use `64-n`

jolt-inlines/blake2/src/exec.rs
0xAndoroid: disregard

jolt-inlines/blake3/src/exec.rs
0xAndoroid: disregard

examples/blake2-ex/guest/src/lib.rs
omibo: I get the following error:
`the trait bound `[u8; 64]: serde::de::Deserialize<'_>` is not satisfied`
Fixed it by returning to [u8; 32] values.

tracer/src/instruction/mod.rs
omibo: Removed for now to keep the code-base clean. Will add later together with lookup tables.

PR #930, opened 2025-09-09
More parallelism in compute_claimed_witness_evals

moodlezoup:
Previously, we would only parallelize over the (41) R1CS inputs, which led to CPU underutilization on big machines


0xAndoroid (approved):
nice, lgtm

PR #931, opened 2025-09-10
feat: optimize parameter passing for &[u8] 

zouguangxian:
When a function parameter is of type &[u8], pass the input slice directly between host 
and guest, bypassing postcard serialization/deserialization.


moodlezoup (approved):
lgtm, can you just run fmt?

zouguangxian:
> @zouguangxian Looking at the CI error, it seems like somehow your changes have caused an _increase_ in the trace length for the sha2 guest program, crossing the boundary of 2^16 cycles (which is the max trace length we provide to `prover_preprocess` in that test). This is the opposite of the expected behavior, right?

```
$ uname -a
Linux ip-172-31-62-238 6.14.0-1011-aws #11~24.04.1-Ubuntu SMP Fri Aug  1 02:07:25 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux

$ git rev-parse HEAD
cf5ce10fa1b3652a8526346bfc6d228080aa803c

$ cargo nextest run --release -p jolt-core
```

It reports:
```
──────────── 
Summary [ 148.755s] 267 tests run: 265 passed (38 slow), 2 failed, 0 skipped
 FAIL [ 97.406s] jolt-core zkvm::tests::sha2_e2e_dory 
 FAIL [ 96.769s] jolt-core zkvm::tests::sha3_e2e_dory 
 error: test run failed
```
When first run, but it succeeds the second time.

0xAndoroid:
@zouguangxian It seems like in the sha2 example, when &[u8] being passed as argument, in the guest a slice of length 32 is being interpreted as length 4096 (therefore more hashing, and more cycles).

PR #938, opened 2025-09-12
Omid/inline blake2 rv64

omibo:
Implement XORROT and XORROTW instructions specific to BLAKE2 and BLAKE3.

# Details
- VirtualXORROT takes two operands a and b, xors them, then rotates by a constant value. I added four instructions; VirtualXORROT32, VirtualXORROT24, VirtualXORROT16, and VirtualXORROT63; specific to BLAKE2.
- VirtualXORROTW takes two operands a and b, xors them, then rotates by a constant value. This instruction operates on the lower 32-bit values of each operand, and the rotation applies only to those lower 32 bits. I added four instructions; VirtualXORROTW16, VirtualXORROTW12, VirtualXORROTW8, and VirtualXORROTW7; specific to BLAKE3.


jolt-core/src/zkvm/instruction/virtual_xor_rotw.rs
```diff
@@ -0,0 +1,106 @@
...
+                }
+            }
+
+            fn to_lookup_output(&self) -> u64 {
+                let (x, y) = LookupQuery::<XLEN>::to_instruction_inputs(self);
+                match XLEN {
+                    #[cfg(test)]
+                    8 => {
+                        let xor_result = (x as u8) ^ (y as u8);
+                        xor_result.rotate_right($rotation).into()
```
0xAndoroid: I believe this is incorrect because rotation should be performed only in 4 lower bits. Take a look at `ROTRIW` `to_lookup_output` function.

jolt-core/src/zkvm/instruction/virtual_xor_rotw.rs
```diff
@@ -0,0 +1,106 @@
...
+                    }
+                    32 => {
+                        let xor_result = (x as u32) ^ (y as u32);
+                        xor_result.rotate_right($rotation).into()
```
0xAndoroid: Same as above. Here we can just cast to `u16` and rotate it.

jolt-core/src/zkvm/lookup_table/prefixes/xor_rot.rs
```diff
@@ -0,0 +1,84 @@
...
+        let suffix_len = current_suffix_len(j);
+
+        let shift = if suffix_len as i32 / 2 - ROTATION as i32 >= 0 {
+            suffix_len / 2 - ROTATION as usize
+        } else {
+            (XLEN as i32 + (suffix_len as i32 / 2 - ROTATION as i32)) as usize
+        };
+
+        // Rotate left to position the XOR result bits correctly in the final output.
+        result += F::from_u64((u64::from(x) ^ u64::from(y)).rotate_left(shift as u32));
```
0xAndoroid, line 61: I believe this works only for `XLEN` 64. This is okay (as such problem exists in half of our prefixes and MLEs), but can you just put a comment about it?

jolt-core/src/zkvm/lookup_table/test.rs
```diff
@@ -38,6 +38,17 @@ pub fn lookup_table_mle_full_hypercube_test<F: JoltField, T: JoltLookupTable + D
     }
 }
 
+pub fn lookup_table_mle_full_hypercube_64_xlen_test<F: JoltField, T: JoltLookupTable + Default>() {
+    let materialized = T::default().materialize();
```
0xAndoroid: `materialize` function only generates hypercube of size 2^16. 
I believe you wanted to test over entire hypercube here, but then we would have to materialize array of size 2^128.

omibo: No. XORROT functions are only implemented for XLEN=64, so I cannot test them with `lookup_table_mle_full_hypercube_test`(), which works for XLEN=8.
I will refactor to use a better naming.

omibo: Changed the name to 
`lookup_table_mle_128_length_bitvector_test()`

jolt-core/src/zkvm/instruction/virtual_xor_rotw.rs
omibo: Removed the implementation for cases that XLEN is not 64.

jolt-core/src/zkvm/instruction/virtual_xor_rotw.rs
omibo: Removed the implementation for cases that XLEN is not 64.

jolt-core/src/zkvm/lookup_table/test.rs
0xAndoroid: It would be great to get XLEN=8 working as well to get the hypercube test working. For the XLEN=64 we have the random fuzzing test. We can't test on the entire 2^128 hypercube because it's too large.

jolt-core/src/zkvm/instruction/virtual_xor_rotw.rs
0xAndoroid: We require XLEN=8 for entire hypercube testing.

omibo: But the instructions does not work for such XLEN. For XORROT, we have rotations 32, 24, 16, 63. None-of-these make sense for XLEN=8. Moreover, these are Blake2 specific instructions that only work on 64-bit platform.

That's the reason I added a new test in tests.rs.

jolt-core/src/zkvm/lookup_table/test.rs
omibo: The issue with `lookup_table_mle_full_hypercube_test()` comes from the following line:
`T::default().evaluate_mle(&index_to_field_bitvector(i as u128, 16)), `
This fails because these instructions are specifically designed for XLEN=64, not a generic XLEN.

In the new test, I changed it to:
`T::default().evaluate_mle(&index_to_field_bitvector(i as u128, 128)),`
which works correctly with XLEN=64.

These instructions do not make sense for XLEN=8, since a rotation by 63 reduces to rotation by (63 mod 8 = 7).

0xAndoroid: Just to be clear, only the evaluate_mle, and the materialize_entry need to support XLEN=8, prefix-suffix can stay XLEN 64 only.

PR #944, opened 2025-09-17
feat: Add custom getrandom backend for RISC-V bare-metal targets

zouguangxian:
This PR introduces a custom implementation of the `getrandom` backend tailored for RISC-V bare-metal environments, where standard OS-level randomness sources (e.g. `/dev/urandom`, system calls) are unavailable.

#### Why this is needed:
- The `getrandom` crate expects platform-specific entropy sources, which do not exist in bare-metal targets.
- Without a custom backend, any crate relying on `getrandom` (e.g. `rand`, `secp256k1`, cryptographic protocols) will fail to compile or panic at runtime.
- RISC-V guest environments require deterministic or syscall-based entropy generation.

#### What this PR adds:
- A simple linear congruential generator (LCG) seeded with a fixed value, exposed via a `sys_rand` syscall.
- Implementations for both `getrandom_v02` and `getrandom_v03` interfaces.
- Integration via `getrandom_v02::register_custom_getrandom!` and a `#[no_mangle]` hook for v0.3.

This enables cryptographic crates to function correctly in guest-mode RISC-V environments, while maintaining compatibility with the `getrandom` ecosystem.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1817,6 +1817,7 @@ dependencies = [
  "ark-ec",
  "common",
  "getrandom 0.2.16",
+ "getrandom 0.3.3",
  "hex-literal",
  "jolt-core",
  "jolt-platform",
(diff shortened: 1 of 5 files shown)


sagar-a16z:
There is already an implementation of a custom getRandom backend in jolt-platform. Please reconcile this with that implementation.

PR #947, opened 2025-09-17
feat: decreasing Number of Virtual Registers for DIV, MUL, REM

omibo:
This PR improves the virtual register assignment for MULH, DIV, and REM instructions, ensuring that each now uses at most 7 virtual registers. This allows us to cap the number of virtual registers used by virtual instructions at 7.

Additionally, a new instruction, `VirtualAssertMulNoOverflow`, is added to the `DIVU` instruction to assert that there is no overflow in quotient \times divisor.


tracer/src/instruction/div.rs
```diff
@@ -119,21 +116,25 @@ impl RISCVTrace for DIV {
         asm.emit_b::<VirtualAssertValidDiv0>(a1, *a2, 0);
         asm.emit_r::<VirtualChangeDivisor>(*t0, a0, a1);
         // check that quotient * divisor don't overflow
-        asm.emit_r::<MUL>(*t1, *a2, *t0);
-        asm.emit_r::<MULH>(*t2, *a2, *t0);
-        asm.emit_i::<SRAI>(*t3, *t1, shmat);
-        asm.emit_b::<VirtualAssertEQ>(*t2, *t3, 0);
+        asm.emit_r::<MULH>(*t1, *a2, *t0);
+        // TODO(Omid): When MULH is done, its allocated virtual registers are freed, then we can allocate registers for the next instructions of the division.
+        // We probably need to refactor to allocate registers on dereference.
```
0xAndoroid: I thought about this a bit, I don't think this is a way. Let's just explicitly have allocations as they are. We can remove this TODO, and just leave it as a note.

0xAndoroid (approved):

PR #950, opened 2025-09-19
chore: support custom getrandom backend for RISC-V bare-metal targets

zouguangxian:
This PR is based on #944, and adds an example to use rand to get a random value. It supports the no_std environment.


jolt-platform/src/random.rs
```diff
@@ -12,8 +12,6 @@ static mut RNG: Option<StdRng> = None;
 #[allow(clippy::missing_safety_doc)]
 #[no_mangle]
 pub unsafe fn sys_rand(dest: *mut u8, len: usize) {
-    jolt_println!("Warning: sys_rand is a deterministic PRNG, not a cryptographically secure RNG. Use with caution.");
```
sagar-a16z, line 15: Please keep this warning in. Determistic prng is not secure

zouguangxian: Ok, added it back.

PR #951, opened 2025-09-19
Omid/inline blake3 keyed rv64

omibo:
This PR implements BLAKE3_KEYED64 inline and its corresponding SDKs.
This inline is specific to 64-byte input hashing with a given key.


jolt-inlines/blake3/src/sdk.rs
```diff
@@ -210,6 +255,14 @@ mod tests {
         parts
     }
 
+    #[test]
+    fn check_output() {
+        let input = super::Aligned64ByteInput([5u8; 64]);
+        let key: [u32; 8] = [0, 1, 2, 3, 4, 5, 6, 7];
+        let hash = Blake3::keyed_hash(&input, key);
+        println!("Hash is: {hash:02x?}");
+    }
```
0xAndoroid: Seems like a useless test

jolt-inlines/blake3/src/sdk.rs
```diff
@@ -224,16 +277,16 @@ mod tests {
     #[test]
     fn test_keyed_digest_random_keys_match_standard() {
         for _ in 0..1000 {
-            let len = rand::random::<usize>() % (BLOCK_INPUT_SIZE_IN_BYTES);
-            let input = generate_random_bytes(len);
+            let input = super::Aligned64ByteInput(generate_random_bytes(64).try_into().unwrap());
```
0xAndoroid, line 279: Can we make this `generate_random_bytes` function to use seeded rng to make tests deterministic?

PR #962, opened 2025-09-24
[JOLT-114] Feat/tracing logs

0xAndoroid:


markosg04 (approved):
only nitty comment: is using fully qualified paths here `tracing::macro!` fine?

I am pretty indifferent but maybe if we planned to introduce more logging (particularly in `jolt-core` then we can just import. For small files / examples is fine i think.

PR #965, opened 2025-09-25
Use split Eq poly optimization for MLEs product sumcheck

andrewmilson:


jolt-core/src/subprotocols/mles_product_sum.rs
```diff
@@ -9,60 +9,57 @@ use crate::{
...
+    r_prime: &[F],
 ) -> UniPoly<F> {
-    let sum_n_terms = 1 << log_sum_n_terms;
-
-    let chunk_size = 512;
+    // Split Eq poly optimization.
+    // See https://eprint.iacr.org/2025/1117.pdf section 5.2.
+    let w = &r[r_prime.len() + 1..];
+    let (wr, wl) = w.split_at(w.len() / 2);
+    let eq_constant_factor = EqPolynomial::mle(r_prime, &r[..r_prime.len()]);
```
markosg04, line 26: is there a reason we don't use `GruenSplitEqPolynomial` abstraction here?

andrewmilson: Discussed offline. Added todo

PR #969, opened 2025-09-26
Small Scalar Binding via Associated Types

abiswas3:
In Jolt we now use 125 bit random challenges -- It makes a few types of multiplication operations 1.5-1.8x faster. 

Run 

```zsh
cargo bench -p jolt-core --bench mul
```

to see the results.

These multiplications are particularly critical during the binding phase of sum-check. To see this empirically, run

```zsh
cargo bench -p jolt-core --bench binding
```

To see how much we've improved swap out the challenge the following lines in `jolt-core/src/fields/ark.rs` 

```rust
    type SmallValueLookupTables = [Vec<Self>; 2];
    type Challenge = MontU128Challenge<ark_bn254::Fr>; 
    // uncomment below and comment out above 
    //type Challenge = TrivialChallenge<ark_bn254::Fr>; 
```

and re-run the `bench` command
You should see that `MontU128Challenge` was 1.4-1.6 times faster.

```zsh
Compiling jolt-core v0.1.0 (/Users/francis/Work-With-A16z/jolt/jolt-core)

warning: unused import: `MontU128Challenge`

--> jolt-core/src/field/ark.rs:3:24

  |

3 |     field::challenge::{MontU128Challenge, TrivialChallenge},

  |                        ^^^^^^^^^^^^^^^^^

  |

  = note: `#[warn(unused_imports)]` on by default

  

warning: `jolt-core` (lib) generated 1 warning (run `cargo fix --lib -p jolt-core` to apply 1 suggestion)

    Finished `bench` profile [optimized + debuginfo] target(s) in 1m 40s

     Running benches/binding.rs (/Users/francis/Work-With-A16z/jolt/target/release/deps/binding-19f4c8703b3096e4)

Gnuplot not found, using plotters backend

DensePolynomial::bind 20 variables

                        time:   [16.606 ms 16.666 ms 16.775 ms]

                        change: [+39.158% +39.729% +40.658%] (p = 0.00 < 0.05)

                        Performance has regressed.

Found 8 outliers among 100 measurements (8.00%)

  5 (5.00%) high mild

  3 (3.00%) high severe

  

Benchmarking DensePolynomial::bind 22 variables: Warming up for 5.0000 s

Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 19.7s, or reduce sample count to 20.

DensePolynomial::bind 22 variables

                        time:   [66.335 ms 66.354 ms 66.374 ms]

                        change: [+39.396% +39.511% +39.626%] (p = 0.00 < 0.05)

                        Performance has regressed.
```
  
  
> [!remark] 
> These benchmarks were run on an M4 Macbook Air with 24GB of memory.

  **Conclusion**: The binding phase, or more particularly whenever we multiply a Field element F with a challenge, it is 1.5-1.8x faster than just multiplying two field elements. 
  
> [!remark] 
> I'll update the patches in `jolt-core/Cargo.toml` shortly so we have to manage a single arkworks PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,8 @@ trace-**.json
 
 # SRS files
 *.srs
+# URS files
+*.urs
 
 temp_file
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,18 +2,18 @@
 name = "jolt"
 version = "0.1.0"
 authors = [
-  # author of original Spartan paper and code base
-  "Srinath Setty <srinath@microsoft.com>",
-  # authors who contributed to the Arkworks Spartan fork
-  "Zhenfei Zhang <zhenfei.zhang@hotmail.com>",
-  # authors who contributed to Lasso/Jolt
-  "Michael Zhu <mzhu@a16z.com>",
-  "Sam Ragsdale <sragsdale@a16z.com>",
-  "Noah Citron <ncitron@a16z.com>",
-  "Arasu Arun <arasu.arun5@gmail.com>",
-  "Quang Dao <dvquang85@gmail.com>",
-  "Andrew Tretyakov <atretyakov@a16z.com>",
-  "Markos Georghiades <mgeorghiades@a16z.com>",
+    # author of original Spartan paper and code base
+    "Srinath Setty <srinath@microsoft.com>",
+    # authors who contributed to the Arkworks Spartan fork
+    "Zhenfei Zhang <zhenfei.zhang@hotmail.com>",
+    # authors who contributed to Lasso/Jolt
+    "Michael Zhu <mzhu@a16z.com>",
+    "Sam Ragsdale <sragsdale@a16z.com>",
+    "Noah Citron <ncitron@a16z.com>",
+    "Arasu Arun <arasu.arun5@gmail.com>",
+    "Quang Dao <dvquang85@gmail.com>",
+    "Andrew Tretyakov <atretyakov@a16z.com>",
+    "Markos Georghiades <mgeorghiades@a16z.com>",
 ]
 edition = "2021"
 description = "The lookup singularity. Based on Spartan; built on Arkworks."
@@ -24,55 +24,55 @@ keywords = ["SNARK", "cryptography", "proofs"]
 
 [workspace]
 members = [
-  "jolt-core",
-  "tracer",
-  "common",
-  "jolt-platform",
-  "jolt-sdk",
-  "jolt-sdk/macros",
-  "jolt-inlines/sha2",
-  "jolt-inlines/keccak256",
-  "jolt-inlines/blake2",
-  "jolt-inlines/blake3",
-  "jolt-inlines/bigint",
-  "examples/btreemap/host",
-  "examples/btreemap/guest",
-  "examples/collatz",
-  "examples/collatz/guest",
-  "examples/fibonacci",
-  "examples/fibonacci/guest",
-  "examples/sha2-ex",
-  "examples/sha2-ex/guest",
-  "examples/blake2-ex",
-  "examples/blake2-ex/guest",
-  "examples/blake3-ex",
-  "examples/blake3-ex/guest",
-  "examples/sha3-ex",
-  "examples/sha3-ex/guest",
-  "examples/sha2-chain",
-  "examples/sha2-chain/guest",
-  "examples/sha3-chain",
-  "examples/sha3-chain/guest",
-  "examples/multi-function",
-  "examples/multi-function/guest",
-  "examples/alloc",
-  "examples/alloc/guest",
-  "examples/stdlib",
-  "examples/stdlib/guest",
-  "examples/muldiv",
-  "examples/muldiv/guest",
-  "examples/overflow",
-  "examples/overflow/guest",
-  "examples/memory-ops",
-  "examples/memory-ops/guest",
-  "examples/recursion",
-  "examples/recursion/guest",
-  "examples/recover-ecdsa",
-  "examples/recover-ecdsa/guest",
-  "examples/random",
-  "examples/random/guest",
-  "examples/malloc",
-  "examples/malloc/guest",
+    "jolt-core",
+    "tracer",
+    "common",
+    "jolt-platform",
+    "jolt-sdk",
+    "jolt-sdk/macros",
+    "jolt-inlines/sha2",
+    "jolt-inlines/keccak256",
+    "jolt-inlines/blake2",
+    "jolt-inlines/blake3",
+    "jolt-inlines/bigint",
+    "examples/btreemap/host",
+    "examples/btreemap/guest",
+    "examples/collatz",
+    "examples/collatz/guest",
+    "examples/fibonacci",
+    "examples/fibonacci/guest",
+    "examples/sha2-ex",
+    "examples/sha2-ex/guest",
+    "examples/blake2-ex",
+    "examples/blake2-ex/guest",
+    "examples/blake3-ex",
+    "examples/blake3-ex/guest",
+    "examples/sha3-ex",
+    "examples/sha3-ex/guest",
+    "examples/sha2-chain",
+    "examples/sha2-chain/guest",
+    "examples/sha3-chain",
+    "examples/sha3-chain/guest",
+    "examples/multi-function",
+    "examples/multi-function/guest",
+    "examples/alloc",
+    "examples/alloc/guest",
+    "examples/stdlib",
+    "examples/stdlib/guest",
+    "examples/muldiv",
+    "examples/muldiv/guest",
+    "examples/overflow",
+    "examples/overflow/guest",
+    "examples/memory-ops",
+    "examples/memory-ops/guest",
+    "examples/recursion",
+    "examples/recursion/guest",
+    "examples/recover-ecdsa",
+    "examples/recover-ecdsa/guest",
+    "examples/random",
+    "examples/random/guest",
+    "examples/malloc",
+    "examples/malloc/guest",
 ]
 
 [features]
@@ -121,29 +121,32 @@ debug = false
 [patch.crates-io]
 ark-ff = { git = "https://github.com/a16z/arkworks-algebra", branch = "dev/twist-shout" }
 ark-ec = { git = "https://github.com/a16z/arkworks-algebra", branch = "dev/twist-shout" }
-jolt-optimizations = { git = "https://github.com/a16z/arkworks-algebra", branch = "dev/twist-shout" }
 ark-serialize = { git = "https://github.com/a16z/arkworks-algebra", branch = "dev/twist-shout" }
 ark-bn254 = { git = "https://github.com/a16z/arkworks-algebra", branch = "dev/twist-shout" }
+ark-grumpkin = { git = "https://github.com/a16z/arkworks-algebra", branch = "dev/twist-shout" }
 allocative = { git = "https://github.com/facebookexperimental/allocative", rev = "85b773d85d526d068ce94724ff7a7b81203fc95e" }
+jolt-optimizations = { git = "https://github.com/a16z/arkworks-algebra", branch = "dev/twist-shout" }
+dory = { git = "https://github.com/markosg04/dory", branch = "dev/twist-shout" }
 
 [workspace.metadata.cargo-machete]
 ignored = ["jolt-sdk"]
 
 [workspace.dependencies]
 # Cryptography and Math
-ark-bn254 = { version = "0.5.0", default-features = false }
+ark-bn254 = { git = "https://github.com/a16z/arkworks-algebra", branch = "dev/twist-shout" }
+ark-grumpkin = { git = "https://github.com/a16z/arkworks-algebra", branch = "dev/twist-shout", default-features = false }
 ark-ec = { version = "0.5.0", default-features = false }
 ark-ff = { version = "0.5.0", default-features = false }
 ark-serialize = { version = "0.5.0", default-features = false, features = [
-  "derive",
+    "derive",
 ] }
 ark-serialize-derive = { version = "0.5.0", default-features = false }
 ark-std = { version = "0.5.0", default-features = false }
-jolt-optimizations = { git = "https://github.com/a16z/arkworks-algebra", branch = "dev/twist-shout" }
-dory = { git = "https://github.com/spaceandtimefdn/sxt-dory", branch = "feat/return_row_commitments" }
 sha3 = "0.10.8"
 blake2 = "0.10"
 blake3 = { version = "1.5.0" }
+jolt-optimizations = { git = "https://github.com/a16z/arkworks-algebra", branch = "dev/twist-shout" }
+dory = { git = "https://github.com/markosg04/dory", branch = "dev/twist-shout" }
 
 # Core Utilities
 clap = { version = "4.5.4", features = ["derive"] }
@@ -208,8 +211,8 @@ proc-macro2 = "1.0.79"
 # Platform-specific
 fnv = { version = "1.0.7", default-features = false }
 object = { version = "0.36.7", features = [
-  "build_core",
-  "elf",
+    "build_core",
+    "elf",
 ], default-features = false }
 addr2line = "0.25.1"
 getrandom = { version = "0.2.16", default-features = false }
--- a/jolt-core/Cargo.toml
+++ b/jolt-core/Cargo.toml
@@ -50,6 +50,7 @@ challenge-254-bit = []
 
 [dependencies]
 ark-bn254.workspace = true
+ark-grumpkin.workspace = true
 jolt-optimizations.workspace = true
 ark-ec.workspace = true
 ark-ff.workspace = true
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -152,9 +152,8 @@ checksum = "a23eb6b1614318a8071c9b2521f36b424b2c83db5eb3a0fead4a6c0809af6e61"
 
 [[package]]
 name = "ark-bn254"
-version = "0.5.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d69eab57e8d2663efa5c63135b2af4f396d66424f88954c21104125ab6b3e6bc"
+version = "0.5.0-alpha.0"
+source = "git+https://github.com/a16z/arkworks-algebra?branch=dev%2Ftwist-shout#2b732305af764d9fd4fcbc570eedf6d289fb2897"
 dependencies = [
  "ark-ec",
  "ark-ff",
@@ -164,7 +163,7 @@ dependencies = [
 [[package]]
 name = "ark-ec"
 version = "0.5.0"
-source = "git+https://github.com/a16z/arkworks-algebra?branch=dev%2Ftwist-shout#60a0d017739620ec2b75a67a1e048681fe54d622"
+source = "git+https://github.com/a16z/arkworks-algebra?branch=dev%2Ftwist-shout#2b732305af764d9fd4fcbc570eedf6d289fb2897"
 dependencies = [
  "ahash",
  "ark-ff",
@@ -185,7 +184,7 @@ dependencies = [
 [[package]]
 name = "ark-ff"
 version = "0.5.0"
-source = "git+https://github.com/a16z/arkworks-algebra?branch=dev%2Ftwist-shout#60a0d017739620ec2b75a67a1e048681fe54d622"
+source = "git+https://github.com/a16z/arkworks-algebra?branch=dev%2Ftwist-shout#2b732305af764d9fd4fcbc570eedf6d289fb2897"
 dependencies = [
  "allocative",
  "ark-ff-asm",
@@ -206,7 +205,7 @@ dependencies = [
 [[package]]
 name = "ark-ff-asm"
 version = "0.5.0"
-source = "git+https://github.com/a16z/arkworks-algebra?branch=dev%2Ftwist-shout#60a0d017739620ec2b75a67a1e048681fe54d622"
+source = "git+https://github.com/a16z/arkworks-algebra?branch=dev%2Ftwist-shout#2b732305af764d9fd4fcbc570eedf6d289fb2897"
 dependencies = [
  "quote",
  "syn 2.0.106",
@@ -215,7 +214,7 @@ dependencies = [
 [[package]]
 name = "ark-ff-macros"
 version = "0.5.0"
-source = "git+https://github.com/a16z/arkworks-algebra?branch=dev%2Ftwist-shout#60a0d017739620ec2b75a67a1e048681fe54d622"
+source = "git+https://github.com/a16z/arkworks-algebra?branch=dev%2Ftwist-shout#2b732305af764d9fd4fcbc570eedf6d289fb2897"
 dependencies = [
  "num-bigint",
  "num-traits",
@@ -225,9 +224,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "ark-grumpkin"
+version = "0.5.0-alpha.0"
+source = "git+https://github.com/a16z/arkworks-algebra?branch=dev%2Ftwist-shout#2b732305af764d9fd4fcbc570eedf6d289fb2897"
+dependencies = [
+ "ark-bn254",
+ "ark-ec",
+ "ark-ff",
+ "ark-std",
+]
+
+[[package]]
 name = "ark-poly"
 version = "0.5.0"
-source = "git+https://github.com/a16z/arkworks-algebra?branch=dev%2Ftwist-shout#60a0d017739620ec2b75a67a1e048681fe54d622"
+source = "git+https://github.com/a16z/arkworks-algebra?branch=dev%2Ftwist-shout#2b732305af764d9fd4fcbc570eedf6d289fb2897"
 dependencies = [
  "ahash",
  "ark-ff",
@@ -241,7 +251,7 @@ dependencies = [
 [[package]]
 name = "ark-serialize"
 version = "0.5.0"
-source = "git+https://github.com/a16z/arkworks-algebra?branch=dev%2Ftwist-shout#60a0d017739620ec2b75a67a1e048681fe54d622"
+source = "git+https://github.com/a16z/arkworks-algebra?branch=dev%2Ftwist-shout#2b732305af764d9fd4fcbc570eedf6d289fb2897"
 dependencies = [
  "ark-serialize-derive 0.5.0 (git+https://github.com/a16z/arkworks-algebra?branch=dev%2Ftwist-shout)",
  "ark-std",
@@ -265,7 +275,7 @@ dependencies = [
 [[package]]
 name = "ark-serialize-derive"
 version = "0.5.0"
-source = "git+https://github.com/a16z/arkworks-algebra?branch=dev%2Ftwist-shout#60a0d017739620ec2b75a67a1e048681fe54d622"
+source = "git+https://github.com/a16z/arkworks-algebra?branch=dev%2Ftwist-shout#2b732305af764d9fd4fcbc570eedf6d289fb2897"
 dependencies = [
  "proc-macro2",
  "quote",
@@ -319,7 +329,7 @@ dependencies = [
  "miniz_oxide",
  "object 0.37.3",
  "rustc-demangle",
- "windows-link 0.2.0",
+ "windows-link 0.2.1",
 ]
 
 [[package]]
@@ -882,7 +892,7 @@ dependencies = [
 [[package]]
 name = "dory"
 version = "1.0.0"
-source = "git+https://github.com/spaceandtimefdn/sxt-dory?branch=feat%2Freturn_row_commitments#814844f68b093e9c240f3c50288d3fb5b370e06c"
+source = "git+https://github.com/markosg04/dory?branch=dev%2Ftwist-shout#c047854b447a8bddb4a7c279849b31d364aeef98"
 dependencies = [
  "ark-bn254",
  "ark-ec",
@@ -998,7 +1008,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.61.1",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -1054,9 +1064,9 @@ checksum = "1d674e81391d1e1ab681a28d99df07927c6d4aa5b027d7da16ba32d1d21ecd99"
 
 [[package]]
 name = "flate2"
-version = "1.1.2"
+version = "1.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4a3d7db9596fecd151c5f638c0ee5d5bd487b6e0ea232e5dc96d5250f6f94b1d"
+checksum = "dc5a4e564e38c699f2880d3fda590bedc2e69f3f84cd48b457bd892ce61d0aa9"
 dependencies = [
  "crc32fast",
  "miniz_oxide",
@@ -1735,6 +1745,7 @@ dependencies = [
  "ark-bn254",
  "ark-ec",
  "ark-ff",
+ "ark-grumpkin",
  "ark-serialize",
  "ark-serialize-derive 0.5.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "ark-std",
@@ -1843,7 +1854,7 @@ dependencies = [
 [[package]]
 name = "jolt-optimizations"
 version = "0.5.0"
-source = "git+https://github.com/a16z/arkworks-algebra?branch=dev%2Ftwist-shout#60a0d017739620ec2b75a67a1e048681fe54d622"
+source = "git+https://github.com/a16z/arkworks-algebra?branch=dev%2Ftwist-shout#2b732305af764d9fd4fcbc570eedf6d289fb2897"
 dependencies = [
  "ark-bn254",
  "ark-ec",
@@ -1936,7 +1947,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d7c4b02199fee7c5d21a5ae7d8cfa79a6ef5bb2fc834d6e9058e89c825efdc55"
 dependencies = [
  "cfg-if",
- "windows-link 0.2.0",
+ "windows-link 0.2.1",
 ]
 
 [[package]]
@@ -2082,6 +2093,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1fa76a2c86f704bdb222d66965fb3d63269ce38518b83cb0575fca855ebb6316"
 dependencies = [
  "adler2",
+ "simd-adler32",
 ]
 
 [[package]]
@@ -2403,7 +2415,7 @@ dependencies = [
  "libc",
  "redox_syscall",
  "smallvec",
- "windows-link 0.2.0",
+ "windows-link 0.2.1",
 ]
 
 [[package]]
@@ -2717,9 +2729,9 @@ dependencies = [
 
 [[package]]
 name = "redox_syscall"
-version = "0.5.17"
+version = "0.5.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5407465600fb0548f1442edf71dd20683c6ed326200ace4b1ef0763521bb3b77"
+checksum = "ed2bf2547551a7053d6fdfafda3f938979645c44812fbfcda098faae3f1a362d"
 dependencies = [
  "bitflags",
 ]
@@ -2886,7 +2898,7 @@ dependencies = [
  "errno",
  "libc",
  "linux-raw-sys 0.11.0",
- "windows-sys 0.61.1",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -2967,7 +2979,7 @@ version = "0.1.28"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "891d81b926048e76efe18581bf793546b4c0eaf8448d72be8de2bbee5fd166e1"
 dependencies = [
- "windows-sys 0.61.1",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -3210,6 +3222,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0fda2ff0d084019ba4d7c6f371c95d8fd75ce3524c3cb8fb653a3023f6323e64"
 
 [[package]]
+name = "simd-adler32"
+version = "0.3.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d66dc143e6b11c1eddc06d5c423cfc97062865baf299914ab64caa38182078fe"
+
+[[package]]
 name = "slab"
 version = "0.4.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3418,7 +3436,7 @@ dependencies = [
  "getrandom 0.3.3",
  "once_cell",
  "rustix 1.1.2",
- "windows-sys 0.61.1",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -3738,9 +3756,9 @@ checksum = "f63a545481291138910575129486daeaf8ac54aee4387fe7906919f7830c7d9d"
 
 [[package]]
 name = "unicode-width"
-version = "0.2.1"
+version = "0.2.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4a1a07cc7db3810833284e8d372ccdc6da29741639ecc70c9ec107df0fa6154c"
+checksum = "b4ac048d71ede7ee76d585517add45da530660ef4390e49b098733c6e897f254"
 
 [[package]]
 name = "untrusted"
@@ -3959,7 +3977,7 @@ version = "0.1.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c2a7b1c03c876122aa43f3020e6c3c3ee5c05081c9a00739faf7503aeba10d22"
 dependencies = [
- "windows-sys 0.61.1",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -3995,9 +4013,9 @@ checksum = "5e6ad25900d524eaabdbbb96d20b4311e1e7ae1699af4fb28c17ae66c80d798a"
 
 [[package]]
 name = "windows-link"
-version = "0.2.0"
+version = "0.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "45e46c0661abb7180e7b9c281db115305d49ca1709ab8242adf09666d2173c65"
+checksum = "f0805222e57f7521d6a62e36fa9163bc891acd422f971defe97d64e70d0a4fe5"
 
 [[package]]
 name = "windows-registry"
@@ -4061,16 +4079,16 @@ version = "0.60.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f2f500e4d28234f72040990ec9d39e3a6b950f9f22d3dba18416c35882612bcb"
 dependencies = [
- "windows-targets 0.53.4",
+ "windows-targets 0.53.5",
 ]
 
 [[package]]
 name = "windows-sys"
-version = "0.61.1"
+version = "0.61.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6f109e41dd4a3c848907eb83d5a42ea98b3769495597450cf6d153507b166f0f"
+checksum = "ae137229bcbd6cdf0f7b80a31df61766145077ddf49416a728b02cb3921ff3fc"
 dependencies = [
- "windows-link 0.2.0",
+ "windows-link 0.2.1",
 ]
 
 [[package]]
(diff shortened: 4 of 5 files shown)


.gitignore
```diff
@@ -51,3 +51,4 @@ jolt-sdk/tests/fixtures
 jolt-sdk/tests/fib_io_device_bytes.rs
 jolt-sdk/tests/fib_proof_bytes.rs
 jolt-sdk/tests/jolt_verifier_preprocessing_bytes.rs
+blog.md
```
markosg04:
```suggestion
```

Cargo.toml
```diff
@@ -65,10 +65,6 @@ members = [
     "examples/memory-ops/guest",
     "examples/recursion",
     "examples/recursion/guest",
-    "examples/recover-ecdsa",
-    "examples/recover-ecdsa/guest",
-    "examples/random",
-    "examples/random/guest",
 ]
```
markosg04, line 76: could we add these back in please?

Cargo.toml
```diff
@@ -115,13 +111,27 @@ inherits = "release"
...
 
 [patch.crates-io]
-ark-ff = { git = "https://github.com/a16z/arkworks-algebra", branch = "feat/fewer-reductions" }
-ark-ec = { git = "https://github.com/a16z/arkworks-algebra", branch = "feat/fewer-reductions" }
-jolt-optimizations = { git = "https://github.com/a16z/arkworks-algebra", branch = "feat/fewer-reductions" }
-ark-serialize = { git = "https://github.com/a16z/arkworks-algebra", branch = "feat/fewer-reductions" }
-ark-bn254 = { git = "https://github.com/a16z/arkworks-algebra", branch = "feat/fewer-reductions" }
+ark-ff = { git = "https://github.com/abiswas3/arkworks-algebra", branch = "dev/twist-shout" }
+ark-ec = { git = "https://github.com/abiswas3/arkworks-algebra", branch = "dev/twist-shout" }
+ark-serialize = { git = "https://github.com/abiswas3/arkworks-algebra", branch = "dev/twist-shout" }
```
markosg04: don't forget to delete the comments here

jolt-core/Cargo.toml
```diff
@@ -135,13 +135,19 @@ harness = false
...
+# name = "poly_bench"
+# harness = false
+#
+# [[bench]]
+# name = "compact_poly_bench"
+# harness = false
+#
 [[bench]]
-name = "poly_bench"
+name = "mul"
```
markosg04: can we use a better name for this? something like `challenges` or related?

jolt-core/benches/binding.rs
markosg04, line 1: nit: it would have been nice to compare the new school bindings to the old school. I think there could be a way to do it using the trivial challenge associated type. If it's not too many loc / wouldn't take too long then i would consider it, otherwise then this is fine for now

jolt-core/benches/mul.rs
markosg04, line 1: In similar spirit to the `binding.rs` comment, we could write this over generic `F` and `F::Challenge` so that we can easily compare between multiple backends and types.

jolt-core/benches/mul.rs
```diff
@@ -0,0 +1,129 @@
...
+                    acc *= b;
+                }
+                acc
+            },
+            BatchSize::SmallInput,
+        )
+    });
+
+    // Fr * u128 via from_u128_mont
+    c.bench_function("Fr * u128 (via from_u128_mont) (100x)", |b| {
```
markosg04: this is not expected to be any faster right?

jolt-core/src/field/ark.rs
```diff
@@ -28,6 +26,8 @@ impl JoltField for ark_bn254::Fr {
         std::mem::transmute(<ark_bn254::FrConfig as MontConfig<4>>::R2)
     };
     type SmallValueLookupTables = [Vec<Self>; 2];
+    type Challenge = MontU128Challenge<ark_bn254::Fr>;
+    //type Challenge = TrivialChallenge<ark_bn254::Fr>;
```
markosg04:
```suggestion
```

jolt-core/src/field/challenge/mont_ark_u128.rs
```diff
@@ -0,0 +1,342 @@
...
+use allocative::Allocative;
+use ark_ff::{BigInt, PrimeField};
+use ark_serialize::{CanonicalDeserialize, CanonicalSerialize};
+use std::fmt::{Debug, Display};
+use std::hash::Hash;
+use std::marker::PhantomData;
+use std::ops::{Add, Mul, Sub};
+/// Bespoke implementation of Challenge type that is a subset of the JoltField
+/// with the property that the 2 least significant digits are 0'd out, and it needs
+/// 125 bits to represent.
```
markosg04: can we explain a bit the reasoning for this type and which operations are defined (subfield vs subset, bottom 2 limbs 0'd out)? Can we also explain why it is 125 bits and not 128. Let's also make it a module-level comment

Related: it would be nice to update the jolt book with a page under optimizations that explains the deal with this. This comment can be short in comparison with just the above and point the reader to the page

jolt-core/src/field/challenge/mod.rs
```diff
@@ -0,0 +1,6 @@
+pub mod mont_ark_u128;
```
markosg04, line 3: do these need to be re-exported?

jolt-core/src/field/challenge/mont_ark_u128.rs
```diff
@@ -0,0 +1,342 @@
...
+    Hash,
+    CanonicalSerialize,
+    CanonicalDeserialize,
+    Allocative,
+)]
+pub struct MontU128Challenge<F: JoltField> {
+    value: [u64; 4],
+    _marker: PhantomData<F>,
+}
+
```
markosg04, line 36: Is there a reason `MontU128Challenge` is defined over generic F? I don't think this is completely generic for any montgomery backend because the constructor is tied to the fact that Fr of bn254 is 254 bits.

jolt-core/src/field/challenge/mont_ark_u128.rs
```diff
@@ -0,0 +1,342 @@
...
+    #[inline(always)]
+    fn into_F(self) -> ark_bn254::Fr {
+        ark_bn254::Fr::from_bigint_unchecked(BigInt::new(self.value())).unwrap()
+    }
+}
+macro_rules! impl_field_ops_inline {
+    ($t:ty, $f:ty) => {
+        /* ----------------------
+         * $t ⊗ $t
+         * ---------------------- */
```
markosg04: we can delete these comments

jolt-core/src/field/challenge/mont_ark_u128.rs
```diff
@@ -0,0 +1,342 @@
...
+        self.value
+    }
+}
+impl IntoField<ark_bn254::Fr> for MontU128Challenge<ark_bn254::Fr> {
+    #[inline(always)]
+    fn into_F(self) -> ark_bn254::Fr {
+        ark_bn254::Fr::from_bigint_unchecked(BigInt::new(self.value())).unwrap()
+    }
+}
+macro_rules! impl_field_ops_inline {
```
markosg04: (1) can we add just a small comment about what this macro is implementing
(2) and we should probably move it to its own file, so that we can factor it out of both `trivial.rs` and `mont_ark_u128.rs`

jolt-core/src/field/challenge/trivial.rs
```diff
@@ -0,0 +1,346 @@
+use crate::field::{IntoField, JoltField};
+use allocative::Allocative;
+use ark_serialize::{CanonicalDeserialize, CanonicalSerialize};
+use std::fmt::{Debug, Display};
+use std::hash::Hash;
+use std::ops::*;
+/// Trivial implementation of Challenge type that just wraps the field element
```
markosg04: (i know I wrote this comment) can we just move it to as a module level comment if we are keeping this implementation around?

jolt-core/src/field/mod.rs
```diff
@@ -15,6 +13,9 @@ pub trait FieldOps<Rhs = Self, Output = Self>:
 {
 }
 
+pub trait IntoField<F: JoltField> {
```
markosg04: Can we instead use rust's `Into`:
```rust
impl Into<ark_bn254::Fr> for MontU128Challenge<ark_bn254::Fr> {
    #[inline(always)]
    fn into(self) -> ark_bn254::Fr {
        ark_bn254::Fr::from_bigint_unchecked(BigInt::new(self.value())).unwrap()
    }
}
```
and then replace the trait bound with `Into<Self>`

jolt-core/src/field/mod.rs
```diff
@@ -41,6 +42,12 @@ pub trait JoltField:
     + CanonicalDeserialize
     + Hash
     + MaybeAllocative
+    + Add<Self::Challenge, Output = Self>              // F + Challenge -> F
+    + for<'a> Add<&'a Self::Challenge, Output = Self>  // F + &Challenge -> F
+    + Sub<Self::Challenge, Output = Self>              // F - Challenge -> F
+    + for<'a> Sub<&'a Self::Challenge, Output = Self>  // F - &Challenge -> F
+    + Mul<Self::Challenge, Output = Self>              // F * Challenge -> F
+    + for<'a> Mul<&'a Self::Challenge, Output = Self>  // F * &Challenge -> F
```
markosg04: i think we can delete these comments, the bounds themselves are pretty clear

jolt-core/src/field/mod.rs
```diff
@@ -54,6 +61,51 @@ pub trait JoltField:
...
+        + Debug
+        + Display
+        + Default
+        + Eq
+        + Hash
+        + CanonicalSerialize
+        + CanonicalDeserialize
+        + Allocative
+        + From<u128>
+        + IntoField<Self> // <-- enforce `into_F` exists
```
markosg04:
```suggestion
        + Into<Self>
```

jolt-core/src/field/mod.rs
```diff
@@ -54,6 +61,51 @@ pub trait JoltField:
...
+        //------------------------------------------------------------------
+        + Add<Self::Challenge, Output = Self> // F + Challenge -> F
+        + for<'a> Add<&'a Self::Challenge, Output = Self> // F + &Challenge -> F
+        + Sub<Self::Challenge, Output = Self> // F - Challenge -> F
+        + for<'a> Sub<&'a Self::Challenge, Output = Self> // F - &Challenge -> F
+        + Mul<Self::Challenge, Output = Self> // F * Challenge -> F
+        + for<'a> Mul<&'a Self::Challenge, Output = Self> // F * &Challenge -> F
+        //------------------------------------------------------------------
+        // Challenge ⊗ F → F
+        //------------------------------------------------------------------
```
markosg04: we can delete all these comments

jolt-core/src/field/mod.rs
```diff
@@ -226,4 +278,7 @@ where
 }
 
 pub mod ark;
-pub mod tracked_ark;
+pub mod challenge;
+//pub mod mont_ark;
```
markosg04: are the tracked values broken after this change or something?

jolt-core/src/poly/commitment/commitment_scheme.rs
```diff
@@ -93,7 +93,7 @@ pub trait CommitmentScheme: Clone + Sync + Send + 'static {
     fn prove<ProofTranscript: Transcript>(
         setup: &Self::ProverSetup,
         poly: &MultilinearPolynomial<Self::Field>,
-        opening_point: &[Self::Field],
+        opening_point: &[<Self::Field as JoltField>::Challenge],
```
markosg04, line 96: I'm just going to make a comment about this so I don't forget: we could probably have nicer looking syntax by removing `Field` as a type on the trait and instead use `F` as a generic directly (`JoltField` already has `Sized` bound). We can leave this for later.

jolt-core/src/poly/commitment/kzg.rs
```diff
@@ -420,7 +423,8 @@ mod test {
             let (ck, vk) = SRS::trim(pp, degree);
             let p = UniPoly::random::<ChaCha20Rng>(degree, rng);
             let comm = UnivariateKZG::<Bn254>::commit(&ck, &p)?;
-            let point = Fr::rand(rng);
+            let point =
+                <<Bn254 as Pairing>::ScalarField as JoltField>::Challenge::from(rng.gen::<u128>());
```
markosg04: if we are grabbing random values for `Challenge` a lot in tests, we could add a `Rand` trait bound or something similar on `Challenge`.

jolt-core/src/poly/dense_mlpoly.rs
```diff
@@ -177,7 +177,7 @@ impl<F: JoltField> DensePolynomial<F> {
 
             if low.is_zero() {
                 if high.is_one() {
-                    new_evals[i] = *r;
+                    new_evals[i] = *r + F::zero();
```
markosg04: is this basically just a cast into `F`? Can we use a proper cast if so? Applies everywhere else in this file.

jolt-core/src/poly/dense_mlpoly.rs
```diff
@@ -245,16 +245,22 @@ impl<F: JoltField> DensePolynomial<F> {
...
     // returns Z(r) in O(n) time
-    pub fn evaluate(&self, r: &[F]) -> F {
-        // r must have a value for each variable
-        assert_eq!(r.len(), self.get_num_vars());
-        let chis = EqPolynomial::evals(r);
-        assert_eq!(chis.len(), self.Z.len());
-        compute_dotproduct(&self.Z, &chis)
+    pub fn evaluate(&self, r: &[F::Challenge]) -> F {
+        //// r must have a value for each variable
+        //assert_eq!(r.len(), self.get_num_vars());
```
markosg04: let's clean this up

jolt-core/src/poly/dense_mlpoly.rs
```diff
@@ -245,16 +245,22 @@ impl<F: JoltField> DensePolynomial<F> {
...
+        //let chis = EqPolynomial::evals(r);
+        //assert_eq!(chis.len(), self.Z.len());
+        //compute_dotproduct(&self.Z, &chis)
+
+        let m = r.len() / 2;
+        let (r2, r1) = r.split_at(m);
+        let (eq_one, eq_two) = rayon::join(|| EqPolynomial::evals(r2), || EqPolynomial::evals(r1));
+        self.split_eq_evaluate(r.len(), &eq_one, &eq_two)
```
markosg04, line 268: this seems like a new change outside the original scope of this PR, which is ok, but I just want to confirm, then:

(1) this is just leveraging Dao-Thaler optimization?
(2) if so, can we comment explaining what we are doing here?

jolt-core/src/poly/dense_mlpoly.rs
```diff
@@ -484,11 +490,22 @@ impl<F: JoltField> Index<usize> for DensePolynomial<F> {
...
+        let (r2, r1) = r.split_at(m);
+        let (eq_one, eq_two) = rayon::join(
+            || EqPolynomial::evals_field(r2),
+            || EqPolynomial::evals_field(r1),
+        );
+
+        self.split_eq_evaluate(r.len(), &eq_one, &eq_two)
+    }
+
+    fn batch_evaluate(polys: &[&Self], r: &[F::Challenge]) -> Vec<F> {
```
markosg04: we are seemingly not using `batch_evaluate` anywhere?

jolt-core/src/poly/dense_mlpoly.rs
```diff
@@ -566,37 +584,37 @@ mod tests {
...
-        // (1, 1) = 1
-        // g(x_0,x_1) => c_0*(1 - x_0)(1 - x_1) + c_1*(1-x_0)(x_1) + c_2*(x_0)(1-x_1) + c_3*(x_0)(x_1)
-        // g(3, 4) = 8*(1 - 3)(1 - 4) + 8*(1-3)(4) + 8*(3)(1-4) + 8*(3)(4) = 48 + -64 + -72 + 96  = 8
-        // g(5, 10) = 8*(1 - 5)(1 - 10) + 8*(1 - 5)(10) + 8*(5)(1-10) + 8*(5)(10) = 96 + -16 + -72 + 96  = 8
-        assert_eq!(
-            dense_poly.evaluate(vec![Fr::from(3), Fr::from(4)].as_slice()),
-            Fr::from(8)
-        );
-    }
+    // #[test]
```
markosg04: can we bring back these tests or their new equivalents?

jolt-core/src/poly/eq_poly.rs
```diff
@@ -9,7 +9,30 @@ const PARALLEL_THRESHOLD: usize = 16;
...
 impl<F: JoltField> EqPolynomial<F> {
     /// Computes the MLE evaluation EQ(x, y)
-    pub fn mle(x: &[F], y: &[F]) -> F {
+    pub fn mle(x: &[F::Challenge], y: &[F::Challenge]) -> F {
+        assert_eq!(x.len(), y.len());
+        x.par_iter()
+            .zip(y.par_iter())
+            .map(|(x_i, y_i)| *x_i * y_i + (F::one() - x_i) * (F::one() - y_i))
+            .product()
+    }
```
markosg04: i would prefer to try and keep only one method if possible to help reduce the diff.

I think something like this would work:

```rust
pub fn mle<X, Y>(x: &[X], y: &[Y]) -> F
    where
        X: Copy + Send + Sync + std::ops::Sub<Output = F> + std::ops::Mul<Y, Output = F>,
        Y: Copy + Send + Sync + std::ops::Sub<Output = F>,
        F: std::ops::Sub<X, Output = F> + std::ops::Sub<Y, Output = F>,
    {
        assert_eq!(x.len(), y.len());
        x.par_iter()
            .zip(y.par_iter())
            .map(|(x_i, y_i)| *x_i * *y_i + (F::one() - *x_i) * (F::one() - *y_i))
            .product()
    }
```
Basically just leverage the beauty of generic approach to make use of trait bounds that are available. We might have to implement `Sub` or handle it in a different way, but such a solution should exist

jolt-core/src/poly/eq_poly.rs
```diff
@@ -136,7 +196,38 @@ impl<F: JoltField> EqPolynomial<F> {
...
+
+    /// Computes the table of coefficients:
+    ///
+    /// scaling_factor * eq(r, x) for all x in {0, 1}^n
+    ///
+    /// computing biggest layers of the dynamic programming tree in parallel.
+    #[tracing::instrument(skip_all, "EqPolynomial::evals_parallel_field")]
+    #[inline]
+    pub fn evals_parallel_field(r: &[F], scaling_factor: Option<F>) -> Vec<F> {
```
markosg04: similar to the MLE case, I think we should be able to avoid having two methods. Either the field method is not necessary or we can use generics to keep it as one method

jolt-core/src/zkvm/lookup_table/and.rs
```diff
@@ -16,7 +16,19 @@ impl<const XLEN: usize> JoltLookupTable for AndTable<XLEN> {
...
+        let mut result = F::zero();
+        for i in 0..XLEN {
+            let x_i = r[2 * i];
+            let y_i = r[2 * i + 1];
+            result += F::from_u64(1u64 << (XLEN - 1 - i)) * x_i * y_i;
+        }
+        result
+    }
+
+    fn evaluate_mle_field<F: JoltField>(&self, r: &[F]) -> F {
```
markosg04: let's try and come up with generic solutions to avoid having `evaluate_mle` and `evaluate_mle_field`, since it blows up the diff. Check out my comments on `EqPolynomial` for how we could try and approach this. Applies everywhere

markosg04 (changes requested):
Firstly, thank you very very much for going through the pain of getting this to compile. I know it is quite complicated because of the many edge cases, and we want to maintain both clarity in the code but also brevity, which is particularly challenging with this change (especially because it is an optimization change).

 I did not finish going through all files, but I think on this first pass I've left enough feedback to warrant a revision before another pass. Generally there are two main points in this review:

(1) general cleanup and feedback (because this change touches core prover machinery I was rather nitty, so I apologize in advance)
(2) I think we should try and reduce the diff. Because the solution is now generic, we can leverage rust's type system to make generic methods to avoid many of the helpers introduced, which I think should reduce the diff by at least half if not more.

On (2): I think it may have been difficult to see such solutions on a first attempt to get this to compile, so now might be a good chance to explore how to reduce the diff via generic methods. Particularly any function that splits between `Challenge` and `F` types could likely be reconciled into one. I would recommend ignoring most of the nits and first try and reduce the diff as much as possible, it will help with the next review cycle.

Once more, thanks a ton for taking the time on this -- it is one of the larger changes to the codebase since T&S integration so it's important we get it right

jolt-core/benches/mul.rs
abiswas3: No it's not. This is more of a test that we do not lose performance going back and forth between types. In a much much earlier version we did due to bit shift operations to extract the big integer from a u128. So this is a sanity check.

jolt-core/src/field/challenge/mont_ark_u128.rs
abiswas3: I have notes/short blog on this. Can add it here, plus the book.

jolt-core/src/field/challenge/mont_ark_u128.rs
abiswas3: The mult part of this macro is different however. So it might be hard to move to a new file, but I'll try

jolt-core/src/field/challenge/mont_ark_u128.rs
abiswas3: You're right. The only reason I could say is that some methods of challenges can be written generically over any F that implements `JoltField`.
However, for the methods that matter -- you are right in that we need the concrete type.

Cargo.toml
abiswas3: Yeah -- i never meant to touch these. Let me see why these got removed. Weird, likely merge conflicts. Thanks for spotting this.

jolt-core/benches/binding.rs
abiswas3: You mean have one bench that compares old school vs new school benching instead of having the user comment lines out as mentioned above? Yeah, that should not be terrible. 
I'll add that to the revision.

jolt-core/src/poly/dense_mlpoly.rs
abiswas3: Yes. This was a remount of code before i went and implemented intoField method.
The is a worse way to cast (although the cost is imperceivable). I'll change this.

jolt-core/src/poly/dense_mlpoly.rs
abiswas3: Yes. This polynomial evaluation optimisation got lost in earlier merges. I'll add in the reference of why this Dao-Thaler split eq-evaluate is the best way to evaluate a polynomial at a random point.

jolt-core/src/poly/dense_mlpoly.rs
abiswas3: Yeah we can, they were written by me to test polynomial evaluation a while back. 
They are not mission critical, but I can bring them back.

jolt-core/src/poly/eq_poly.rs
abiswas3: Just leaving a comment so we don't forget about the discussion earlier. 
This works for the above Mle function, but there are cases where these functions are part of traits that are turned into trait objects, where this fails. 
Currently we are getting rid of the dyn TraitObjects so we can generic this everywhere else.

Cargo.toml
abiswas3: They are all back now

jolt-core/benches/binding.rs
abiswas3: This is now addressed through features. By default we use MontU128 but by just passing in the other flag we can just use the other challenge type.

```zsh
cargo bench -p jolt-core --bench binding
cargo bench =p jolt-core --no-default-features --features "host,trivial-challenge" --bench binding
```

jolt-core/src/field/challenge/mod.rs
abiswas3: It does **need** to be everywhere we access these through one level higher. So changing this would be annoying, and as is this does not really affect performance.

jolt-core/src/field/challenge/mont_ark_u128.rs
abiswas3: I've put a placeholder link -- which will get populated later today with details. 
Once it's been reviewed by Justin, I'll put it in the Jolt Book, along with other optimisations.

jolt-core/src/field/challenge/mont_ark_u128.rs
abiswas3: Actually keeping the generics makes using the macros much easier, otherwise you need to separate the t x t traits separately.

jolt-core/src/field/challenge/mont_ark_u128.rs
abiswas3: I've tidied this up significantly by Grouping traits (see comments). I feel the current state is fine.

jolt-core/src/poly/commitment/kzg.rs
abiswas3: Done -- there's now a Challenge::random(&mut rng) we can use. And is used in some places. It's just a wrapper around the above.

jolt-core/src/poly/dense_mlpoly.rs
abiswas3: See [this](https://randomwalks.xyz/publish/fast_polynomial_evaluation.html) for more details. I'm in the process of cleaning these up; and putting a cleaner version into the Jolt book if required.

Cargo.toml
```diff
@@ -116,14 +116,34 @@ overflow-checks = true
...
+# ark-bn254 = { git = "https://github.com/a16z/arkworks-algebra", branch = "dev/twist-shout" }
+#
 allocative = { git = "https://github.com/facebookexperimental/allocative", rev = "85b773d85d526d068ce94724ff7a7b81203fc95e" }
 
+[patch.'https://github.com/a16z/arkworks-algebra']
+# jolt-optimizations = {path="../arkworks-algebra/jolt-optimizations"}
+# ark-serialize = { path = "../arkworks-algebra/serialize" }
+jolt-optimizations = { git = "https://github.com/abiswas3/arkworks-algebra", branch = "feat/fewer-reductions" }
+ark-serialize = { git = "https://github.com/abiswas3/arkworks-algebra", branch = "feat/fewer-reductions" }
 [workspace.metadata.cargo-machete]
```
markosg04: needs clean up

jolt-core/Cargo.toml
```diff
@@ -21,37 +21,39 @@ license-file = "LICENSE"
...
+  "ark-ff/parallel",
+  "ark-std/parallel",
+  "dory/parallel",
 ]
 # This is for building jolt-core without prover capabilities, e.g. for recursion
 # jolt-core needs std and rayon to compile, so these are the minimal set of features.
 minimal = ["ark-ec/std", "ark-ff/std", "ark-std/std", "ark-ff/asm", "rayon"]
 allocative = ["dep:inferno"]
 test_incremental = []
+mont-challenge = []
```
markosg04: let's just make the 125 bit challenges the default, so we can remove the `mont-challenge` flag, and then let's make a flag called `challenge-254-bit` that is optional. Then, in the main Fr impl we can gate it that way

markosg04: in general i want to rename the trivial challenge now that we are keeping it, will address else where

jolt-core/Cargo.toml
```diff
@@ -2,16 +2,16 @@
...
+  # author of original Spartan paper and code base
+  "Srinath Setty <srinath@microsoft.com>",
+  # authors who contributed to the Arkworks Spartan fork
+  "Zhenfei Zhang <zhenfei.zhang@hotmail.com>",
+  # authors who contributed to Lasso/Jolt
+  "Michael Zhu <mzhu@a16z.com>",
+  "Sam Ragsdale <sragsdale@a16z.com>",
+  "Noah Citron <ncitron@a16z.com>",
+  "Arasu Arun <arasu.arun5@gmail.com>",
+  "Quang Dao <dvquang85@gmail.com>",
```
markosg04: i see a lot of weird indents, maybe need to run `cargo fmt`

jolt-core/benches/poly_bench.rs
```diff
@@ -22,19 +22,22 @@ fn setup_inputs(n: u64) -> (MultilinearPolynomial<Fr>, Vec<Fr>) {
...
         let id_dot = format!("dot-product-{}", exp);
         group.bench_function(&id_dot, |b| {
             b.iter(|| poly.evaluate_dot_product(eval_point.as_slice()))
         });
 
-        let id_opt = format!("inside-out-{}", exp);
-        group.bench_function(&id_opt, |b| b.iter(|| poly.evaluate(eval_point.as_slice())));
+        //let id_opt = format!("inside-out-{}", exp);
+        //group.bench_function(&id_opt, |b| b.iter(|| poly(eval_point.as_slice())));
     }
```
markosg04: let's clean this

jolt-core/src/field/ark.rs
```diff
@@ -30,6 +32,12 @@ impl JoltField for ark_bn254::Fr {
     type Unreduced<const N: usize> = BigInt<N>;
     type SmallValueLookupTables = [Vec<Self>; 2];
 
+    #[cfg(feature = "mont-challenge")]
+    type Challenge = MontU128Challenge<ark_bn254::Fr>;
+
+    #[cfg(feature = "trivial-challenge")]
+    type Challenge = TrivialChallenge<ark_bn254::Fr>;
+
```
markosg04: as noted elsewhere let's just make the small challenge a default and get rid of `mont-challenge`

jolt-core/src/field/challenge/mont_ark_u128.rs
markosg04: it's fine for it to go in a subsequent PR if more time is needed

jolt-core/src/field/mod.rs
```diff
@@ -1,11 +1,12 @@
+use allocative::Allocative;
 use ark_ff::biginteger::S224;
+use ark_ff::UniformRand;
+//use ark_serialize::{CanonicalDeserialize, CanonicalSerialize};
```
markosg04:
```suggestion
```

jolt-core/src/poly/dense_mlpoly.rs
```diff
@@ -107,7 +107,7 @@ impl<F: JoltField> DensePolynomial<F> {
             .for_each(|(a, b)| {
                 let m = *b - *a;
                 if m.is_one() {
-                    *a += *r;
+                    *a += *r + F::zero();
```
markosg04: can we use a cast? Applies everywhere with this pattern

jolt-core/src/poly/eq_poly.rs
```diff
@@ -7,16 +7,21 @@ use std::marker::PhantomData;
...
 impl<F: JoltField> EqPolynomial<F> {
-    /// Computes the MLE evaluation EQ(x, y)
-    pub fn mle(x: &[F], y: &[F]) -> F {
+    pub fn mle<X, Y>(x: &[X], y: &[Y]) -> F
+    where
+        X: Copy + Send + Sync,
+        Y: Copy + Send + Sync,
+        F: JoltField + std::ops::Sub<X, Output = F> + std::ops::Sub<Y, Output = F>,
+        X: std::ops::Mul<Y, Output = F>,
+    {
```
markosg04: since we use these ops frequently in this file let's import them instead of fully qualified path

jolt-core/src/poly/eq_poly.rs
```diff
@@ -226,7 +255,12 @@ impl<F: JoltField> EqPlusOnePolynomial<F> {
             let step = 1 << (ell - i);
             let half_step = step / 2;
 
-            let r_lower_product = (F::one() - r[i]) * r.iter().skip(i + 1).copied().product::<F>();
+            let mut r_lower_product = F::one();
+            for &x in r.iter().skip(i + 1) {
+                r_lower_product = r_lower_product * x; // To get the benefits of multiplication
+            }
+            r_lower_product *= F::one() - r[i];
+            // let r_lower_product = (F::one() - r[i]) * r.iter().skip(i + 1).copied().product::<F>();
```
markosg04:
```suggestion
```

jolt-core/src/poly/eq_poly.rs
```diff
@@ -257,9 +292,11 @@ mod tests {
     fn test_evals() {
         let mut rng = test_rng();
         for len in 5..22 {
-            let r = (0..len).map(|_| Fr::random(&mut rng)).collect::<Vec<_>>();
+            let r = (0..len)
+                .map(|_| <Fr as JoltField>::Challenge::from(rng.gen::<u128>()))
+                .collect::<Vec<_>>();
```
markosg04, line 298: If we use the new `rand` impl then these lines might shrink which would be nice if it does

jolt-core/src/poly/identity_poly.rs
```diff
@@ -46,13 +46,14 @@ impl<F: JoltField> PolynomialBinding<F> for IdentityPolynomial<F> {
             }
             BindingOrder::HighToLow => {
                 self.bound_value += self.bound_value;
-                self.bound_value += r;
+                self.bound_value = self.bound_value + r;
```
markosg04, line 49: does += not work anymore?

jolt-core/src/poly/multilinear_polynomial.rs
```diff
@@ -1,10 +1,12 @@
 use crate::{
+    field::{ChallengeFieldOps, FieldChallengeOps},
     poly::{one_hot_polynomial::OneHotPolynomial, rlc_polynomial::RLCPolynomial},
     utils::{compute_dotproduct, small_scalar::SmallScalar},
 };
 use allocative::Allocative;
 use ark_ff::biginteger::S128;
 use ark_serialize::{CanonicalDeserialize, CanonicalSerialize, Valid};
+// use ark_std::iterable::Iterable;
```
markosg04:
```suggestion
```

jolt-core/src/poly/range_mask_polynomial.rs
```diff
@@ -29,7 +29,7 @@ impl<F: JoltField> RangeMaskPolynomial<F> {
             let range_end_bit = range_end.pop_msb();
             if range_end_bit == 1 {
                 lt_range_end += eq_range_end * (F::one() - r_i);
-                eq_range_end *= *r_i;
+                eq_range_end = eq_range_end * r_i;
```
markosg04, line 32: `*=` doesn't apply anymore?

abiswas3: I intentionally kept assigns out of F * C to make sure i didn't accidentally miss any. It was a way to ensure the transfer was good, and way for me to distinguish. I can put them back in if we want.

jolt-core/src/poly/identity_poly.rs
abiswas3: See below

Cargo.toml
abiswas3: This will go away compeltely once we merge my arkworks-algebra changges. Then we use the current one main has -- and things will work smoothly.

jolt-core/Cargo.toml
abiswas3: I thought I did run cargo fmt. And there's a linter in vim too. I'll double check again

jolt-core/src/poly/dense_mlpoly.rs
abiswas3: Yeah missed the is one. Will fix it

jolt-core/src/field/mod.rs
abiswas3: It was initially as I did not update it with Challenge mults, but now everything is back in order.

jolt-core/Cargo.toml
abiswas3: Cargo fmt on my end is not giving any diffs. Perhaps the original was off? Or I'm on a differnt version of fmt??

jolt-core/Cargo.toml
abiswas3: Done.

jolt-core/src/field/ark.rs
abiswas3: Done

jolt-core/src/poly/eq_poly.rs
abiswas3: Fixed all invocations where it was easy.

jolt-core/Cargo.toml
0xAndoroid: cargo fmt doesn't touch `cargo.toml` file. I believe you have a VS Code plugin installed that formats it, and has 2 spaces as indentation set. 
Let's revert these files to have 4 spaces of indentation.

abiswas3: Thanks. I'll revert these.

jolt-core/Cargo.toml
markosg04, line 1: can we revert the formatting to the original indents? I think you might have an extension or something that auto formatted it in a different way

jolt-core/benches/poly_bench.rs
```diff
@@ -20,21 +29,29 @@ fn setup_inputs(n: u64) -> (MultilinearPolynomial<Fr>, Vec<Fr>) {
...
+/// Compares performance of:
+/// 1. `evaluate()` - Standard split-and-merge algorithm
+/// 2. `evaluate_dot_product()` - Dot product based algorithm
+///
+/// Both algorithms evaluate dense multilinear polynomials (no sparse coefficients)
+/// across different polynomial sizes (2^12 to 2^24 coefficients)
 fn bench_all(c: &mut Criterion) {
     let mut group = c.benchmark_group("evals");
-    group.measurement_time(std::time::Duration::from_secs(60));
+    //group.measurement_time(std::time::Duration::from_secs(60));
```
markosg04:
```suggestion
```

jolt-core/src/field/challenge/trivial.rs
```diff
@@ -0,0 +1,392 @@
...
+/// - Similar for `&F::Challenge` (reference types)
+///
+/// **Challenge with Challenge:**
+/// - `F::Challenge + F::Challenge`
+/// - `F::Challenge - F::Challenge`
+/// - `F::Challenge * F::Challenge`
+/// - Reference variants
+///
+/// All operations are marked `#[inline(always)]` for optimal performance in
+/// hot path polynomial operations.
```
markosg04:
```suggestion
```

jolt-core/src/field/challenge/mont_ark_u128.rs
```diff
@@ -0,0 +1,397 @@
...
+/// - Similar for `&F::Challenge` (reference types)
+///
+/// **Challenge with Challenge:**
+/// - `F::Challenge + F::Challenge`
+/// - `F::Challenge - F::Challenge`
+/// - `F::Challenge * F::Challenge`
+/// - Reference variants
+///
+/// All operations are marked `#[inline(always)]` for optimal performance in
+/// hot path polynomial operations.
```
markosg04:
```suggestion
```

jolt-core/src/field/mod.rs
```diff
@@ -16,6 +17,68 @@ pub trait FieldOps<Rhs = Self, Output = Self>:
 {
 }
 
+// Define the helper traits once
```
markosg04:
```suggestion
```

jolt-core/src/subprotocols/mles_product_sum.rs
```diff
@@ -296,11 +297,15 @@ mod tests {
     #[test]
     fn test_compute_mles_product_sum_with_2_mles() {
         const N_MLE: usize = 2;
-        let rng = &mut test_rng();
-        let r: &[Fr; 1] = &rng.gen();
+        let mut rng = &mut test_rng();
+        let r_whole = [<Fr as JoltField>::Challenge::rand(&mut rng)];
+        let r: &[<Fr as JoltField>::Challenge; 1] = &r_whole;
+        // let r: &[Fr; 1] = &rng.gen();
```
markosg04:
```suggestion
```

jolt-core/src/subprotocols/mles_product_sum.rs
```diff
@@ -328,11 +335,14 @@ mod tests {
     #[test]
     fn test_compute_mles_product_sum_with_8_mles() {
         const N_MLE: usize = 8;
-        let rng = &mut test_rng();
-        let r: &[Fr; 1] = &rng.gen();
+        let mut rng = &mut test_rng();
+        let r_whole = [<Fr as JoltField>::Challenge::random(&mut rng)];
+        let r: &[<Fr as JoltField>::Challenge; 1] = &r_whole;
+        // let r: &[Fr; 1] = &rng.gen();
```
markosg04:
```suggestion
```

jolt-core/src/subprotocols/mles_product_sum.rs
```diff
@@ -344,11 +354,14 @@ mod tests {
     #[test]
     fn test_compute_mles_product_sum_with_16_mles() {
         const N_MLE: usize = 16;
-        let rng = &mut test_rng();
-        let r: &[Fr; 1] = &rng.gen();
+        let mut rng = &mut test_rng();
+        //let r: &[Fr; 1] = &rng.gen();
```
markosg04:
```suggestion
```

jolt-core/src/transcripts/blake2b.rs
```diff
@@ -230,6 +230,31 @@ impl Transcript for Blake2bTranscript {
         }
         q_powers
     }
+
+    // New methods that return F::Challenge
```
markosg04:
```suggestion
```

markosg04: ^ I left this comment in the instructive diff so we an get rid of it

jolt-core/src/field/challenge/trivial.rs
```diff
@@ -0,0 +1,392 @@
...
+    Debug,
+    Default,
+    PartialEq,
+    Eq,
+    Hash,
+    CanonicalSerialize,
+    CanonicalDeserialize,
+    Allocative,
+)]
+pub struct TrivialChallenge<F: JoltField> {
```
markosg04: Let's rename this to `Mont254BitChallenge` to be more in line with the other one

jolt-core/src/transcripts/transcript.rs
```diff
@@ -21,6 +21,10 @@ pub trait Transcript: Default + Clone + Sync + Send + 'static {
     fn challenge_vector<F: JoltField>(&mut self, len: usize) -> Vec<F>;
     // Compute powers of scalar q : (1, q, q^2, ..., q^(len-1))
     fn challenge_scalar_powers<F: JoltField>(&mut self, len: usize) -> Vec<F>;
+    // New methods that return F::Challenge
+    fn challenge_scalar_special<F: JoltField>(&mut self) -> F::Challenge;
+    fn challenge_vector_special<F: JoltField>(&mut self, len: usize) -> Vec<F::Challenge>;
+    fn challenge_scalar_powers_special<F: JoltField>(&mut self, len: usize) -> Vec<F>;
```
markosg04: let's use `special` -> `optimized` everywhere

jolt-core/src/transcripts/blake2b.rs
```diff
@@ -258,4 +283,38 @@ mod tests {
             );
         }
     }
+
+    #[test]
+    fn test_challenge_special_trivial() {
+        use ark_std::UniformRand;
+        //TODO: this test needs to be changed like your old test.
```
markosg04: is addressed? if so delete

markosg04 (changes requested):
Final nits, otherwise looks good. See our chat for some comments about the macro refator.

Let's also not forget to add to the jolt book the notes on this optimization.

Good work. Thanks a lot!

jolt-core/Cargo.toml
abiswas3: i think andrews merge fixed this

jolt-core/src/zkvm/lookup_table/pow2_w.rs
```diff
@@ -17,7 +17,11 @@ impl<const XLEN: usize> JoltLookupTable for Pow2WTable<XLEN> {
         1 << (index % 32) as u64
     }
 
-    fn evaluate_mle<F: JoltField>(&self, r: &[F]) -> F {
+    fn evaluate_mle<F, C>(&self, r: &[C]) -> F
+    where
+        C: ChallengeFieldOps<F>,
+        F: JoltField + FieldChallengeOps<C>,
+    {
```
moodlezoup, line 24: Is there reason we have these trait bounds instead of:
```suggestion
    fn evaluate_mle<F: JoltField>(&self, r: &[F::Challenge]) -> F
    {
```

jolt-core/src/zkvm/lookup_table/prefixes/mod.rs
```diff
@@ -104,26 +107,31 @@ pub trait SparseDensePrefix<F: JoltField>: 'static + Sync {
...
+        r_x: Option<C>,
         c: u32,
         b: LookupBits,
         j: usize,
-    ) -> F;
-
+    ) -> F
+    where
+        C: ChallengeFieldOps<F>,
+        F: FieldChallengeOps<C>;
```
moodlezoup, line 119: same here (and in update_prefix_checkpoint), can we just use `r_x: Option<F::Challenge>` instead of the trait bounds?

moodlezoup (commented):
nice work! Just a couple of questions on the trait bounds for some functions

jolt-core/src/zkvm/lookup_table/pow2_w.rs
abiswas3: Yes. If you look at the tests, and I think the case where we evaluate with a bit vector (though that could be for another MLE), the r is something like `F::one()` or `F::zero()`. In the MontU128 Challenge world, there is no canonical one or zero - as these are not sub-groups. 
So the old tests would break. 
As an intermediate step we used to have an `evaluate_mle` as you describe, and a separate `evaluate_mle_field` -- for the tests (even though as of today all real invocations outside tests use challenges). 
To avoid code duplication, we went ahead and used these generics.

jolt-core/src/zkvm/lookup_table/prefixes/mod.rs
abiswas3: See comment above.

jolt-core/src/zkvm/lookup_table/pow2_w.rs
moodlezoup: got it, makes sense

jolt-core/src/zkvm/lookup_table/prefixes/mod.rs
abiswas3: @markosg04 I'll fix the toml tomorrow. I only have vim right now, and the formatter is being weird for toml. I'll open nano and do this tomorrow morning before NYC time start of day.

PR #970, opened 2025-09-26
Feat/delayed reduction

0xAndoroid:


jolt-core/src/field/ark.rs
```diff
@@ -225,77 +229,73 @@ impl JoltField for ark_bn254::Fr {
...
     #[inline]
-    fn linear_combination_u64(pairs: &[(Self, u64)], add_terms: &[Self]) -> Self {
-        let mut tmp = ark_ff::BigInt::<4>::mul_u64_w_carry::<5>(&pairs[0].0 .0, pairs[0].1);
-        for (a, b) in &pairs[1..] {
-            let carry = tmp.add_with_carry(&ark_ff::BigInt::<4>::mul_u64_w_carry::<5>(&a.0, *b));
-            debug_assert!(!carry, "spurious carry in linear_combination_u64");
-        }
+    fn mul_unreduced<const L: usize>(self, other: Self) -> BigInt<L> {
+        self.0.mul_trunc::<4, L>(&other.0)
+    }
```
markosg04, line 233: all of these functions currently return `BigInt<L>`, shouldn't they return `Self::Unreduced`?

Also, we use `const L` here, but the associated type is defined with `N` for number of limbs if im understanding correctly. Should we pick one or is there a reason why we have two?

jolt-core/src/field/ark.rs
markosg04, line 1: should we also consider just adding one additional test in this file to sanity check the implementation of `Self::Unreduced` compared to the normal way of arithmetic?

jolt-core/src/field/mod.rs
```diff
@@ -49,6 +49,33 @@ pub trait JoltField:
     /// The squared Montgomery factor R^2 = 2^(128*N) mod p
     const MONTGOMERY_R_SQUARE: Self;
 
+    /// Unreduced field element representation with N 64 bit limbs
+    type Unreduced<const N: usize>: Clone
```
markosg04, line 54: just pointing out here we define # of Limbs as `N` but elsewhere it seemed that we were using `L`

jolt-core/src/field/mod.rs
```diff
@@ -49,6 +49,33 @@ pub trait JoltField:
...
+        + MulU64WithCarry<Output<5> = Self::Unreduced<5>>
+        + Add<Output = Self::Unreduced<N>>
+        + for<'a> Add<&'a Self::Unreduced<N>, Output = Self::Unreduced<N>>
+        + AddAssign
+        + for<'a> AddAssign<&'a Self::Unreduced<N>>
+        + AddAssign<Self::Unreduced<4>>
+        + AddAssign<Self::Unreduced<5>>
+        + AddAssign<Self::Unreduced<6>>
+        + AddAssign<Self::Unreduced<7>>
+        + AddAssign<Self::Unreduced<8>>;
```
markosg04, line 79: I think this solution of just specifying the limbs "manually" is ok for now. It's not that many, it's rather clear, and easy enough to change in the future if we swap fields.

I tried playing around with fancier compile time enforcement solutions (like marker traits in some way) but rust type system (at least stable rust) doesn't really support such things with associated constants.

The only other thing is macro but macro is ugly and overkill here, i think.

jolt-core/src/field/mod.rs
```diff
@@ -114,27 +158,57 @@ pub trait JoltField:
...
+        acc: &mut Self::Acc<P>,
+    );
+}
+
+pub trait FromLimbs<const N: usize> {
+    fn from_limbs(limbs: [u64; N]) -> Self;
+}
+
+pub trait FromS224 {
+    fn from_s224(value: ark_ff::biginteger::S224) -> Self;
```
markosg04: nit: import rather than fully qualified path?

jolt-core/src/poly/one_hot_polynomial.rs
```diff
@@ -232,14 +232,20 @@ impl<F: JoltField> OneHotPolynomialProverOpening<F> {
...
 
-                    [B_evals[0] * inner_sum[0], B_evals[1] * inner_sum[1]]
+                    [
+                        B_evals[0].mul_unreduced::<9>(inner_sum[0]),
+                        B_evals[1].mul_unreduced::<9>(inner_sum[1]),
+                    ]
                 })
                 .reduce(
-                    || [F::zero(); 2],
+                    || [F::Unreduced::<9>::default(); 2],
```
markosg04: maybe we should consider defining constants for some of these cases like 9 in some way. It is more verbose but it is somewhat unclear to new reader what `F::Unreduced::<9>::default()` is (at least for max case or something? the name doesn't have to be super long, it can be abbreviated or something but then the reader can just look up what the const is and see what it represents).

separately, maybe we should have a `Zero/Identity` trait bound on `Unreduced` or something so we can still use that instead of default (I think it would be clearer).

jolt-core/src/poly/one_hot_polynomial.rs
```diff
@@ -232,14 +232,20 @@ impl<F: JoltField> OneHotPolynomialProverOpening<F> {
                             |running, new| [running[0] + new[0], running[1] + new[1]],
                         );
 
-                    [B_evals[0] * inner_sum[0], B_evals[1] * inner_sum[1]]
+                    [
+                        B_evals[0].mul_unreduced::<9>(inner_sum[0]),
+                        B_evals[1].mul_unreduced::<9>(inner_sum[1]),
+                    ]
```
markosg04, line 239: any way to bring back `*` instead of `.mul_unreduced::<9>` or are we just stuck with it as a consequence of the generic (would it be super annoying / is it possible to implement `ops`?)

jolt-core/src/poly/one_hot_polynomial.rs
```diff
@@ -232,14 +232,20 @@ impl<F: JoltField> OneHotPolynomialProverOpening<F> {
...
-                    || [F::zero(); 2],
+                    || [F::Unreduced::<9>::default(); 2],
                     |running, new| [running[0] + new[0], running[1] + new[1]],
                 );
 
-            univariate_poly_evals.to_vec()
+            unreduced_univariate_poly_evals
+                .into_iter()
+                .map(|evals| F::from_montgomery_reduce(evals))
+                .collect()
```
markosg04, line 249: i feel like we are going to be doing this pattern in basically all sumchecks. Should we introduce a `batch_reduce()` or `evals_reduce()` method or something to turn this into 1 line and make it clear everywhere?

jolt-core/src/poly/one_hot_polynomial.rs
```diff
@@ -263,21 +269,30 @@ impl<F: JoltField> OneHotPolynomialProverOpening<F> {
...
-                        d_e_in[x_in] * d_e_out[x_out] * ra_eval(j)
+                    .map(|x_in| {
+                        let unreduced_inner_sum = (0..num_x_out)
+                            .into_par_iter()
+                            .map(|x_out| {
+                                let j = (x_in << num_x_out_bits) | x_out;
+                                d_e_out[x_out].mul_unreduced::<9>(ra_eval(j))
+                            })
+                            .reduce(F::Unreduced::<9>::default, |running, new| running + new);
+                        let inner_sum = F::from_montgomery_reduce(unreduced_inner_sum);
```
markosg04, line 295: what do you think about implementing `From` or an extension trait or something? Is it not worth having the macro to handle all the generic cases

jolt-core/src/poly/prefix_suffix.rs
```diff
@@ -285,18 +287,130 @@ impl<F: JoltField, const ORDER: usize> PrefixSuffixDecomposition<F, ORDER> {
...
+                *reduced_q_coeff = F::from_barrett_reduce(*q_coeff);
+            }
+        }
+
+        self.Q = std::array::from_fn(|i| DensePolynomial::new(std::mem::take(&mut reduced_Q[i])));
+    }
+
+    /// Initialize Q for two PrefixSuffixDecomposition instances in a single pass over indices
+    #[tracing::instrument(skip_all)]
+    pub fn init_Q_dual(
```
markosg04, line 322: can we leave a comment pointing to docs or something or 1 liner for what `Q` is

jolt-core/src/poly/spartan_interleaved_poly.rs
```diff
@@ -590,18 +601,18 @@ impl<const NUM_SVO_ROUNDS: usize, F: JoltField> SpartanInterleavedPolynomial<NUM
...
-                            let cz0 = cz_acc[0].reduce_to_field::<F>();
-                            let az1 = az_acc[1].reduce_to_field::<F>();
-                            let bz1 = bz_acc[1].reduce_to_field::<F>();
-                            let cz1 = cz_acc[1].reduce_to_field::<F>();
+                            let az0 = az_acc[0].reduce_to_field();
+                            let bz0 = bz_acc[0].reduce_to_field();
+                            let cz0 = cz_acc[0].reduce_to_field();
+                            let az1 = az_acc[1].reduce_to_field();
+                            let bz1 = bz_acc[1].reduce_to_field();
+                            let cz1 = cz_acc[1].reduce_to_field();
```
markosg04, line 616: this is not really related to this PR but just going through some of the small value changes as a consequence i find the use of these azi, bzi, czi hard to follow and verbose. Not saying I have a better solution off top of my head but maybe we could revisit if there is a nicer abstraction/way to implement this, maybe I could ticket it if you agree

jolt-core/src/utils/thread.rs
```diff
@@ -55,3 +55,62 @@ pub fn unsafe_zero_slice<F: JoltField + Sized>(slice: &mut [F]) {
...
+    unsafe {
+        let ptr = vector.as_mut_ptr();
+        std::ptr::write_bytes(ptr, 0, size);
+        vector.set_len(size);
+    }
+    vector
+}
+
+/// Fast allocation of zero vector of BigInt
+/// TODO: refactor so that both this and unsafe_allocate_zero_vec are one function
```
markosg04: I think `unsafe_allocate_zero_vec_bigint` can be deleted? I don't see it being used anywhere anymore after you introduced `unsafe_allocate_zero_vec_unreduced`.

jolt-core/src/utils/thread.rs
```diff
@@ -55,3 +55,62 @@ pub fn unsafe_zero_slice<F: JoltField + Sized>(slice: &mut [F]) {
...
+        let ptr = std::alloc::alloc_zeroed(layout) as *mut ark_ff::BigInt<N>;
+
+        if ptr.is_null() {
+            panic!("Zero vec allocation failed");
+        }
+
+        result = Vec::from_raw_parts(ptr, size, size);
+    }
+    result
+}
```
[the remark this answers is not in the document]
markosg04: this function should probably be deleted anyways

jolt-core/src/field/ark.rs
0xAndoroid: BigInt is the Self::Unredured for Arc.
I'll change L to N.

jolt-core/src/field/ark.rs
0xAndoroid: Not too sure what you mean

jolt-core/src/poly/one_hot_polynomial.rs
0xAndoroid: No, because we need that const generic

jolt-core/src/poly/one_hot_polynomial.rs
0xAndoroid: We have mont and barret reduction, so we can't

jolt-core/src/poly/one_hot_polynomial.rs
0xAndoroid: I have changed default to zero.
I think adding 5 new constants would be more verbose than valuable though.

PR #975, opened 2025-09-30
feat: snark composition

markosg04:
adds the initial recursion & snark composition proof machinery. Brings the main dory verification from 1.5B RISC-V cycles down to 330 million:
- adds a square and multiply sumcheck
- adds hyrax and related things
- changes the commitment scheme API to allow for offloading ops
- adds some new accumulator methods for recursion specific context
- grumpkin-related support for sumchecks
- stage 6 DAG for recursion mode

also does some miscellaneous things:
- srs -> urs
- aligns all the feature branches back into `dev/twist-shout` across dory and arkworks
- probably some other small fixes along the way that I don't quite remember

In general, I am somewhat OK with the flow of the stage 6 for now. My personal main problem is with the recursion helpers being verbose in the `opening_proof.rs` and some jank casts. I am still thinking for better solutions and so just flagging this during the review. Happy to take suggestions for better approaches. If no one has better ideas in this review then a subsequent PR will address

Next steps:
- CI needs to test in recursion mode
- Implement remaining sumchecks for the 330 mil cycles (should bring us to pretty much 0 cycles hopefully).
- further cleanup as we come up with better solutions for the recursion machinery
- measure and improve prover perf in stage 6 (for this initial version it is a few seconds in debug mode)
- Optimize the hyrax scheme by using a giant matrix instead of homomorphism (we have many small polys so currently homomorphic combining is inefficient)

On the last point: we are producing about 50k polys of 4 vars. We should be able to fit everything including all the remaining witnesses from the 330M cycles into a 2^10 x 2^10 matrix, so we have 2^11 Grumpkin G1 (Fr) ops for verifier (2 MSM of 2^10), which should be really fast (few ms) to verify...

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -152,7 +152,19 @@ checksum = "a23eb6b1614318a8071c9b2521f36b424b2c83db5eb3a0fead4a6c0809af6e61"
 
 [[package]]
 name = "ark-bn254"
+version = "0.5.0-alpha.0"
+source = "git+https://github.com/a16z/arkworks-algebra?branch=dev%2Ftwist-shout#2b732305af764d9fd4fcbc570eedf6d289fb2897"
+dependencies = [
+ "ark-ec",
+ "ark-ff",
+ "ark-std",
+]
+
+[[package]]
+name = "ark-bn254"
 version = "0.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d69eab57e8d2663efa5c63135b2af4f396d66424f88954c21104125ab6b3e6bc"
 dependencies = [
  "ark-ec",
  "ark-ff",
@@ -162,6 +174,7 @@ dependencies = [
 [[package]]
 name = "ark-ec"
 version = "0.5.0"
+source = "git+https://github.com/a16z/arkworks-algebra?branch=dev%2Ftwist-shout#2b732305af764d9fd4fcbc570eedf6d289fb2897"
 dependencies = [
  "ahash",
  "ark-ff",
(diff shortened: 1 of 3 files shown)


jolt-core/src/poly/commitment/dory.rs
```diff
@@ -249,12 +263,55 @@ where
     }
 
     fn scale(&self, k: &Self::Scalar) -> Self {
-        Self(self.0.cyclotomic_exp(k.0.into_bigint()))
+        Self(self.0.pow(k.0.into_bigint()))
```
markosg04, line 266: cyclotomic exp is currently not implemented in our bespoke dory snark (though it could be in future, it would reduce the witness size by about half)

jolt-core/src/zkvm/dag/jolt_dag.rs
```diff
@@ -471,6 +546,62 @@ impl JoltDAG {
...
+                            ProofTranscript,
+                            1,
+                        >,
+                        &crate::subprotocols::snark_composition::RecursionProof<
+                            Fq,
+                            ProofTranscript,
+                            1,
+                        >,
+                    >(sz_proof)
+                };
```
markosg04, line 587: open to suggestions to avoid jank casts (applies prover side too)

jolt-core/src/poly/opening_proof.rs
```diff
@@ -900,6 +955,23 @@ where
...
+                        .iter()
+                        .map(|(polynomial, coeff)| {
+                            (commitments.get(polynomial).unwrap().clone(), *coeff)
+                        })
+                        .collect::<Vec<_>>(),
+                )
+            } else {
+                None
+            };
+
```
markosg04, line 973: also jank, suggestions?

markosg04:
Stale & there is a new PR coming soon for this feature

PR #979, opened 2025-10-02
fix: run two-byte alignment tests

codygunton:
In preparation for RV64IMAC support, https://github.com/a16z/jolt/pull/842 introduced patching of certain RISC-V architecture tests tests so they would not be run by RISCOF. I suppose the reasoning for this was that the tests live in `riscv-arch-test/riscv-test-suite/rv64i_m/privilege` and they all declare `RVTEST_ISA("RV64I_Zicsr")`, which is apparently a constraint that was not being parsed do to an incorrect regex (also fixed by https://github.com/a16z/jolt/pull/842). They certainly appear to be tests of privileged architecture from this lens. However, upon closer inspection, the excluded tests are not testing conditions of Zicsr. Rather, they detect, in our case, whether we can actually branch or jump to two-byte aligned addresses, which is a requirement of the C extension. You can think of those tests as also checking these conditions that exceptions are not thrown in such situations:
https://github.com/riscv/riscv-isa-manual/blob/0e1a870c53e8dd062ecb2b734b265c78732d667f/src/c-st-ext.adoc?plain=1#L26
https://github.com/riscv/riscv-isa-manual/blob/0e1a870c53e8dd062ecb2b734b265c78732d667f/src/rv32.adoc?plain=1#L514

This PR reinstates those tests, which pass without any modifications to the emulator. To do this we unset `rvtest_mtrap_routine`, which causes a privileged trap handler to be used. Such a trap handler being used is not required for the tests to be meaningful. We also remove the Zicsr extension from the `RVTEST_ISA` requirement for the same reason. This causes all of the following tests to run and pass:
```
/privilege/src/misalign-beq-01.S
/privilege/src/misalign-bge-01.S
/privilege/src/misalign-bgeu-01.S
/privilege/src/misalign-blt-01.S
/privilege/src/misalign-bltu-01.S
/privilege/src/misalign-bne-01.S
/privilege/src/misalign-jal-01.S
/privilege/src/misalign2-jalr-01.S
```
Noe that this is in addition to three similar tests that are already being run in Jolt's CI:
```
/privilege/src/misalign1-cjalr-01.S
/privilege/src/misalign1-cjr-01.S
/privilege/src/misalign1-jalr-01.S
```
(cc @zouguangxian, please let me know if I'm missing something here :smile:)


0xAndoroid (approved):

0xAndoroid:
Thank you!

PR #984, opened 2025-10-06
Create new crate for Jolt fields

andrewmilson:


sagar-a16z (approved):

jolt-core/src/poly/multilinear_polynomial.rs
```diff
@@ -345,71 +345,55 @@ impl<F: JoltField> MultilinearPolynomial<F> {
...
-}
 
-impl<F: JoltField> From<Vec<F>> for MultilinearPolynomial<F> {
-    fn from(coeffs: Vec<F>) -> Self {
-        let poly = DensePolynomial::new(coeffs);
-        Self::LargeScalars(poly)
-    }
-}
-
-impl<F: JoltField> From<Vec<u8>> for MultilinearPolynomial<F> {
```
andrewmilson, line 357: All these From<Vec<u8>>, From<Vec<u16>>, etc. had to be removed.
The problem: jolt_field crate would compile if JoltField is implemented on a u8 but that would be a problem here in jolt_core because there'd be two implementations here for the same type (One for From<Vec<JoltField>> and one for From<Vec<u8>>).

PR #986, opened 2025-10-07
Feat/product virtualization

moodlezoup:
Virtualize `Product` witness polynomial, removing one commitment and one R1CS constraint.

### DIFF
--- a/jolt-core/src/zkvm/r1cs/inputs.rs
+++ b/jolt-core/src/zkvm/r1cs/inputs.rs
@@ -614,7 +614,6 @@ where
         .zip(right_input.par_iter_mut())
         .zip(trace.par_iter())
         .for_each(|((left, right), cycle)| {
-            // Instruction inputs and product
             (*left, *right) = LookupQuery::<XLEN>::to_instruction_inputs(cycle);
         });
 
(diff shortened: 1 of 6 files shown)


jolt-core/src/zkvm/r1cs/inputs.rs
```diff
@@ -596,6 +595,32 @@ where
...
+        .zip(right_input.par_iter_mut())
+        .zip(trace.par_iter())
+        .for_each(|((left, right), cycle)| {
+            // Instruction inputs and product
+            (*left, *right) = LookupQuery::<XLEN>::to_instruction_inputs(cycle);
+        });
+
+    (left_input.into(), right_input.into())
+}
+
```
markosg04, line 622: could we avoid iterating again over trace by combining with something else? maybe as a follow up.

markosg04: just making one additional note here as per offline discussion: we are repeating the same witness gen for these polys as in `generate_witness_batch`, but we can't reuse because those values get dropped to reduce mem. Follow up work will unify these approaches (or problem may go away when left and right inputs get virtualized)

0xAndoroid (approved):
lgtm

PR #987, opened 2025-10-07
feat: Untrusted/Trusted Advice

omibo:
This PR adds support for trusted and untrusted advice to the Jolt zkVM.

# Terminology
1. **UntrustedAdvice**: An input to the program that the verifier does not know about.
2. **TrustedAdvice**: An input to the program where the verifier knows **only** its commitment. This commitment is generated by an _external party_, not by the Jolt zkVM prover.

# How to use
- Write a regular guest program and use TrustedAdvice/UntrustedAdvice annotation for advice inputs.
- The provable macro automatically generates the corresponding functions, passing all inputs (public and advice, along with the commitment to trusted advice, if any) to the prover, and only the public inputs and the commitment to trusted advice (if any) to the verifier.
- The prover commits to the untrusted advice at the start of the protocol.
- The rest of the process remains unchanged, except that whenever the verifier needs access to initial_memory_state in the RAM Stage 3 sumcheck, the prover provides the evaluation. The main changes occur in ram/val_evaluation.rs and ram/output_check.rs.
- The prover proves, and the verifier verifies, the correctness of the evaluation proofs for advice inputs at the end of the protocol.

# Other Changes
- `JoltDevice` now includes `trusted_advice` and `untrusted_advice`. These are empty for the verifier and populated for the prover.
- The Tracer module is modified to read values from `trusted_advice` and `untrusted_advice` when needed.
- The guest main function is updated to include a pointer to the memory location where advice inputs reside.
- `remap_address()` is updated so that the starting index begins from `trusted_advice_start`.

# Memory Layout
trusted_advice || untrusted_advice || inputs || outputs || panic || termination || padding || RAM

### DIFF
--- a/jolt-core/src/zkvm/ram/val_evaluation.rs
+++ b/jolt-core/src/zkvm/ram/val_evaluation.rs
@@ -1,5 +1,5 @@
 use num_traits::Zero;
-use std::{cell::RefCell, rc::Rc};
+use std::{cell::RefCell, rc::Rc, sync::Arc};
 
 use crate::{
     field::JoltField,
@@ -13,6 +13,7 @@ use crate::{
             OpeningPoint, ProverOpeningAccumulator, SumcheckId, VerifierOpeningAccumulator,
             BIG_ENDIAN,
         },
+        ra_poly::RaPolynomial,
     },
     subprotocols::sumcheck::SumcheckInstance,
     transcripts::Transcript,
@@ -31,7 +32,7 @@ use rayon::prelude::*;
 #[derive(Allocative)]
 pub struct ValEvaluationProverState<F: JoltField> {
     inc: MultilinearPolynomial<F>,
-    wa: MultilinearPolynomial<F>,
+    wa: RaPolynomial<usize, F>,
     lt: MultilinearPolynomial<F>,
 }
 
@@ -79,14 +80,14 @@ impl<F: JoltField> ValEvaluationSumcheck<F> {
         let _guard = span.enter();
 
         // Compute the wa polynomial using the above table
-        let wa: Vec<F> = trace
+        let wa_indices: Vec<Option<usize>> = trace
             .par_iter()
             .map(|cycle| {
                 remap_address(cycle.ram_access().address() as u64, memory_layout)
-                    .map_or(F::zero(), |k| eq_r_address[k as usize])
+                    .map(|k| k as usize)
             })
             .collect();
-        let wa = MultilinearPolynomial::from(wa);
+        let wa = RaPolynomial::new(Arc::new(wa_indices), eq_r_address);
 
         drop(_guard);
         drop(span);
@@ -206,9 +207,7 @@ impl<F: JoltField, T: Transcript> SumcheckInstance<F, T> for ValEvaluationSumche
                 let inc_evals = ps
                     .inc
                     .sumcheck_evals_array::<DEGREE>(i, BindingOrder::HighToLow);
-                let wa_evals = ps
-                    .wa
-                    .sumcheck_evals_array::<DEGREE>(i, BindingOrder::HighToLow);
+                let wa_evals = ps.wa.sumcheck_evals(i, DEGREE, BindingOrder::HighToLow);
                 let lt_evals = ps
                     .lt
                     .sumcheck_evals_array::<DEGREE>(i, BindingOrder::HighToLow);
@@ -237,13 +236,10 @@ impl<F: JoltField, T: Transcript> SumcheckInstance<F, T> for ValEvaluationSumche
     #[tracing::instrument(skip_all, name = "RamValEvaluationSumcheck::bind")]
     fn bind(&mut self, r_j: F::Challenge, _round: usize) {
         if let Some(prover_state) = &mut self.prover_state {
-            [
-                &mut prover_state.inc,
-                &mut prover_state.wa,
-                &mut prover_state.lt,
-            ]
-            .par_iter_mut()
-            .for_each(|poly| poly.bind_parallel(r_j, BindingOrder::HighToLow));
+            [&mut prover_state.inc, &mut prover_state.lt]
+                .par_iter_mut()
+                .for_each(|poly| poly.bind_parallel(r_j, BindingOrder::HighToLow));
+            prover_state.wa.bind_parallel(r_j, BindingOrder::HighToLow);
         }
     }
 
--- a/jolt-core/src/poly/one_hot_polynomial.rs
+++ b/jolt-core/src/poly/one_hot_polynomial.rs
@@ -43,7 +43,7 @@ pub struct OneHotPolynomial<F: JoltField> {
     /// The array described in Section 6.3 of the Twist/Shout paper.
     G: Vec<F>,
     /// The array described in Section 6.3 of the Twist/Shout paper.
-    H: Arc<RwLock<RaPolynomial<F>>>,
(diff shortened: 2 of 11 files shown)


jolt-core/src/zkvm/dag/jolt_dag.rs
```diff
@@ -275,6 +285,21 @@ impl JoltDAG {
...
 
         tracing::info!("Stage 5 proving");
+
+        // Generate untrusted_advice opening proofs
+        if !state_manager.program_io.untrusted_advice.is_empty() {
+            // Todo(omid): Remove transcript clone
+            let proof = Self::generate_untrusted_advice_proof(
+                &mut state_manager,
+                &preprocessing.generators,
+                transcript.borrow().clone(),
```
omibo: Without using `clone()` on the transcript, verification in the next stage fails (reduce_and_prove and reduce_and_verify).
I believe the reason is that the Dory prover and verifier leave the transcript in inconsistent states, which causes the issue for the next Dory prover/verification.

jolt-core/src/zkvm/dag/jolt_dag.rs
```diff
@@ -515,4 +558,100 @@ impl JoltDAG {
...
+            let word = u64::from_le_bytes(word);
+            initial_memory_state[index] = word;
+            index += 1;
+        }
+
+        let poly = MultilinearPolynomial::from(initial_memory_state);
+        let (commitment, hint) = PCS::commit(&poly, &preprocessing.generators);
+
+        if let Some(ref mut prover_state) = state_manager.prover_state {
+            prover_state.untrusted_advice_polynomial = Some(poly);
```
omibo, line 645: Maybe storing this polynomial uses too much memory. We can drop it here and recalculate it later.

jolt-core/src/zkvm/dag/jolt_dag.rs
```diff
@@ -515,4 +558,100 @@ impl JoltDAG {
...
+    >(
+        state_manager: &mut StateManager<'_, F, ProofTranscript, PCS>,
+        generators: &PCS::ProverSetup,
+        mut transcript: ProofTranscript,
+    ) -> PCS::Proof {
+        let prover_state = state_manager.prover_state.as_ref().unwrap();
+        let untrusted_advice_poly = prover_state.untrusted_advice_polynomial.as_ref().unwrap();
+        let accumulator = state_manager.get_prover_accumulator();
+        let (point, _) = accumulator.borrow().get_untrusted_advice_opening().unwrap();
+        PCS::prove_without_hint(generators, untrusted_advice_poly, &point.r, &mut transcript)
```
omibo: When I use `PCS::prove()`, I get an error due to an MSM length mismatch, which happens because initial_memory_state is small.
I could resize hint and append zeros to it, but since initial_memory_state is small, I think we can use `prove_without_hint()` instead.

jolt-core/src/zkvm/dag/jolt_dag.rs
omibo: This is going to be fixed in the next commit. I only commit to an array of length equal to untrusted_advice.

jolt-core/src/zkvm/ram/val_evaluation.rs
```diff
@@ -133,9 +133,21 @@ impl<F: JoltField> ValEvaluationSumcheck<F> {
         );
         let (r_address, _) = r.split_at(K.log_2());
 
-        let val_init: MultilinearPolynomial<F> =
+        let untrusted_advice_eval = state_manager
+            .verifier_state
+            .as_ref()
+            .unwrap()
+            .accumulator
```
moodlezoup: nit:
```suggestion
            .get_verifier_accumulator()
```

jolt-core/src/zkvm/ram/output_check.rs
```diff
@@ -471,9 +471,33 @@ impl<F: JoltField> ValFinalSumcheck<F> {
...
+            );
+            let (r_address_val_evaluation, _) = r.split_at(state_manager.ram_K.log_2());
+            assert_eq!(r_address_val_evaluation.r, r_address);
+        }
+
+        let untrusted_advice_eval = state_manager
+            .verifier_state
+            .as_ref()
+            .unwrap()
+            .accumulator
```
moodlezoup: nit:
```suggestion
            .get_verifier_accumulator()
```

jolt-core/src/zkvm/dag/jolt_dag.rs
moodlezoup: Hmm, that should not be the case. If you don't do the clone, and run an end-to-end test, do you get a "Fiat-Shamir transcript mismatch" somewhere?

jolt-sdk/macros/src/lib.rs
```diff
@@ -729,7 +814,17 @@ impl MacroBuilder {
...
         }
 
-        args
+        (pub_args, untrusted_advice_args)
+    }
+
+    fn is_private_type(ty: &Type) -> bool {
+        if let Type::Path(type_path) = ty {
+            if let Some(last_segment) = type_path.path.segments.last() {
+                // Check for "Private" or path ending in "Private" (e.g., jolt::Private)
```
moodlezoup: nit:
```suggestion
    fn is_untrusted_advice(ty: &Type) -> bool {
        if let Type::Path(type_path) = ty {
            if let Some(last_segment) = type_path.path.segments.last() {
                // Check for "UntrustedAdvice" or path ending in "UntrustedAdvice" (e.g., jolt::UntrustedAdvice)
```

jolt-core/src/poly/commitment/dory.rs
```diff
@@ -1175,6 +1175,40 @@ impl CommitmentScheme for DoryCommitmentScheme {
...
+        transcript: &mut ProofTranscript,
+    ) -> Self::Proof {
+        // Dory uses the opposite endian-ness as Jolt
+        let point_dory: Vec<JoltFieldWrapper<Self::Field>> = opening_point
+            .iter()
+            .rev()
+            .map(|&p| JoltFieldWrapper(p.into()))
+            .collect();
+
+        let sigma = DoryGlobals::get_num_columns().log_2();
```
moodlezoup, line 1192: for the advice commitments, I think we'll want to use different matrix dimensions than the other polynomials in Jolt. E.g. this sigma is chosen so that we commit using a sqrt(256T) x sqrt(256T) matrix, but for the advice commitments we can use a sqrt(n) x sqrt(n) matrix, where n is the size of the advice

jolt-core/src/zkvm/dag/jolt_dag.rs
```diff
@@ -516,4 +574,245 @@ impl JoltDAG {
...
+        let untrusted_advice_commitment = state_manager.untrusted_advice_commitment.as_ref();
+        let trusted_advice_commitment = state_manager.trusted_advice_commitment.as_ref();
+
+        match (untrusted_advice_commitment, trusted_advice_commitment) {
+            (Some(untrusted_comm), Some(trusted_comm)) => {
+                let accumulator = state_manager.get_verifier_accumulator();
+                let (untrusted_point, untrusted_eval) =
+                    accumulator.borrow().get_untrusted_advice_opening().unwrap();
+                let (_trusted_point, trusted_eval) =
+                    accumulator.borrow().get_trusted_advice_opening().unwrap();
```
moodlezoup: let's add an assertion that the two points are the same

PR #995, opened 2025-10-07
Feat/virtualize remaining products

markosg04:
WIP


jolt-core/src/zkvm/bytecode/mod.rs
```diff
@@ -156,36 +193,25 @@ impl<F: JoltField, PCS: CommitmentScheme<Field = F>, T: Transcript> SumcheckStag
         let F_1 = compute_ra_evals(bytecode_preprocessing, trace, &E_1);
 
         let read_raf = ReadRafSumcheck::new_prover(sm);
-        let booleanity = BooleanitySumcheck::new_prover(sm, r_cycle, F_1.clone());
         let hamming_weight = HammingWeightSumcheck::new_prover(sm, F_1);
```
0xAndoroid, line 196: How about we also move the hamming weight to the stage 2 as well so that we don't have to recalculate `F_1`. Hamming weight is negl cost anyways.

jolt-core/src/zkvm/bytecode/read_raf_checking.rs
```diff
@@ -59,6 +60,8 @@ pub struct ReadRafSumcheck<F: JoltField> {
     gamma: [F; STAGES],
     gamma_cub: F,
     gamma_sqr: F,
+    gamma_four: F,
+    gamma_five: F,
```
0xAndoroid: Let's remove these and just use `gamma` array. So `gamma_cub = gamma[3]` and so on.

0xAndoroid: And we can also make gamma array to be a vector (since it stores 6 elements now).

jolt-core/src/zkvm/bytecode/read_raf_checking.rs
```diff
@@ -385,27 +501,220 @@ impl<F: JoltField> ReadRafSumcheck<F> {
...
+            let (_, claim) = sm.get_virtual_polynomial_opening(
+                VirtualPolynomial::OpFlags(flag),
+                SumcheckId::SpartanOuter,
+            );
+            sum += claim * gamma_powers[gamma_idx];
+            gamma_idx += 1;
+        }
+
+        sum
+    }
```
0xAndoroid, line 516: Use of iterators, as before, seems cleaner to me

jolt-core/src/zkvm/bytecode/read_raf_checking.rs
```diff
@@ -74,10 +77,14 @@ pub struct ReadRafSumcheck<F: JoltField> {
 enum ReadCheckingValType {
     /// Spartan outer sumcheck
     Stage1,
-    /// Registers read-write sumcheck
+    /// Jump flag from ShouldJumpVirtualization
     Stage2,
-    /// Registers val sumcheck wa, PCSumcheck, Instruction Lookups
+    /// PCSumcheck
```
moodlezoup, line 78: idk how much it matters, but I think there's a way to shuffle things around so that we only need 4 stages here, e.g. can move PCSumcheck to stage 4 or 5

jolt-core/src/zkvm/bytecode/read_raf_checking.rs
moodlezoup: nit: it's a bit redundant to have `gamma: [F; STAGES]` and also these `gamma_cub`, `gamma_sqr`, etc.

jolt-core/src/zkvm/bytecode/read_raf_checking.rs
```diff
@@ -385,27 +501,220 @@ impl<F: JoltField> ReadRafSumcheck<F> {
...
+                } else {
+                    F::zero()
+                };
+                let rd_addr_val = F::from_u64(instr.operands.rd as u64);
+                jump_val * gamma_powers[0]
+                    + branch_val * gamma_powers[1]
+                    + rd_addr_val * gamma_powers[2]
+                    + jump_val * gamma_powers[3]
+                    + rd_addr_val * gamma_powers[4]
+                    + write_lookup_output_to_rd_val * gamma_powers[5]
```
0xAndoroid: I think it's just slightly more optimal (and cleaner) to have a mut accumulator variable, and add to it, instead of having F::zero() * gamma_powers(), as it is was done before.

jolt-core/src/zkvm/bytecode/read_raf_checking.rs
```diff
@@ -385,27 +501,220 @@ impl<F: JoltField> ReadRafSumcheck<F> {
...
+    ) -> Vec<F> {
+        sm.get_bytecode()
+            .par_iter()
+            .map(|instruction| {
+                let instr = instruction.normalize();
+                let flags = instruction.circuit_flags();
+                let unexpanded_pc = instr.address;
+                let imm = instr.operands.imm;
+
+                let mut linear_combination: F = F::from_i128(imm);
```
0xAndoroid, line 620: This is what I meant by previous comment.

jolt-core/src/zkvm/bytecode/read_raf_checking.rs
```diff
@@ -385,27 +501,220 @@ impl<F: JoltField> ReadRafSumcheck<F> {
...
+            VirtualPolynomial::OpFlags(CircuitFlags::WriteLookupOutputToRD),
+            SumcheckId::WriteLookupOutputToRDVirtualization,
+        );
+
+        jump_claim * gamma_powers[0]
+            + branch_claim * gamma_powers[1]
+            + rd_wa_claim_write_pc * gamma_powers[2]
+            + jump_claim_write_pc * gamma_powers[3]
+            + rd_wa_claim_write_lookup * gamma_powers[4]
+            + write_lookup_output_to_rd_flag_claim * gamma_powers[5]
```
0xAndoroid: Let's do it same way as it's done in rv3

jolt-core/src/zkvm/instruction_lookups/mod.rs
```diff
@@ -91,32 +91,66 @@ impl<F: JoltField, PCS: CommitmentScheme<Field = F>, T: Transcript> SumcheckStag
         &mut self,
         sm: &mut StateManager<'_, F, T, PCS>,
     ) -> Vec<Box<dyn SumcheckInstance<F, T>>> {
-        let read_raf = ReadRafSumcheck::new_prover(sm, self.eq_r_cycle.take().unwrap());
         let hamming_weight = HammingWeightSumcheck::new_prover(sm, self.G.take().unwrap());
```
0xAndoroid, line 94: Let's move this hamming weight to stage 2 (with booleanity). 
Also, not sure if we'd like to not store, but recompute `eq_r_cycle` here to reduce stage 2 memory by T.

jolt-core/src/zkvm/instruction_lookups/read_raf_checking.rs
```diff
@@ -282,13 +292,16 @@ impl<'a, F: JoltField> ReadRafProverState<F> {
             prefix_checkpoints: vec![None.into(); Prefixes::COUNT],
             suffix_polys,
             v: ExpandingTable::new(M),
-            u_evals,
-            eq_r_cycle: MultilinearPolynomial::from(eq_r_cycle),
+            u_evals_rv: eq_r_cycle_rv.clone(),
+            u_evals_raf: eq_r_cycle.clone(),
```
0xAndoroid, line 296: Can you make `clone` before the return statement, and then add tracing around computation of eq_r_cycle_rv and clones

jolt-core/src/zkvm/r1cs/inputs.rs
```diff
@@ -278,15 +278,14 @@ pub enum JoltR1CSInputs {
...
-    NextIsNoop,       // Virtual (spartan shift sumcheck)
     ShouldJump,
     OpFlags(CircuitFlags),
 }
 
 const NUM_R1CS_INPUTS: usize = ALL_R1CS_INPUTS.len();
 /// This const serves to define a canonical ordering over inputs (and thus indices
 /// for each input). This is needed for sumcheck.
-pub const ALL_R1CS_INPUTS: [JoltR1CSInputs; 41] = [
+pub const ALL_R1CS_INPUTS: [JoltR1CSInputs; 40] = [
```
0xAndoroid: 40 seems a lot for 16 constraints. Are you sure all of these are used in constraints?

jolt-core/src/zkvm/bytecode/read_raf_checking.rs
markosg04: stage 3 is not mem bottleneck but looking at the graph it might be better if it was stage 4. I will try and if it doesn't create problems we can move it there and if it is problematic we can leave it for now. It goes to sqrt space anyways

jolt-core/src/zkvm/spartan/instruction_input.rs
```diff
@@ -0,0 +1,596 @@
...
+    eq_r_cycle_stage_1: GruenSplitEqPolynomial<F>,
+    eq_r_cycle_stage_2: GruenSplitEqPolynomial<F>,
+    prev_claim_stage_1: F,
+    prev_claim_stage_2: F,
+    prev_round_poly_stage_1: Option<UniPoly<F>>,
+    prev_round_poly_stage_2: Option<UniPoly<F>>,
+}
+
+impl<F: JoltField> ProverState<F> {
+    fn gen(
```
0xAndoroid, line 514:
```suggestion
    fn new(
```

jolt-core/src/zkvm/instruction_lookups/read_raf_checking.rs
moodlezoup: I'm probably going to change this anyway, so no need to do this rn

jolt-core/src/zkvm/instruction_lookups/mod.rs
markosg04: Leaving this as a follow up work

jolt-core/src/zkvm/bytecode/mod.rs
markosg04: Leaving this as a follow up work

jolt-core/src/zkvm/spartan/instruction_input.rs
andrewmilson: new is fine if it's free but things that are expensive should indicate it.
Same goes with from() and into().

PR #1000, opened 2025-10-09
Benchmark utils

0xAndoroid:


Cargo.toml
```diff
@@ -228,6 +229,7 @@ spinners = { version = "4.1.1", default-features = false }
 criterion = { version = "0.5.1", features = ["html_reports"] }
 iai-callgrind = "0.10.2"
 serial_test = "3.2.0"
+plotly = "0.13"
```
danielwlz: We can remove this?

jolt-core/Cargo.toml
```diff
@@ -98,6 +99,7 @@ dory.workspace = true
 lazy_static.workspace = true
 allocative.workspace = true
 inferno = { workspace = true, optional = true }
+plotly.workspace = true
```
danielwlz: This too

PR #1002, opened 2025-10-09
Perf/matrix refactor opt

markosg04:
adds a new batch addition routine and some other small perf things

### DIFF
--- a/jolt-core/src/zkvm/dag/jolt_dag.rs
+++ b/jolt-core/src/zkvm/dag/jolt_dag.rs
@@ -3,7 +3,6 @@ use std::collections::HashMap;
 use crate::field::JoltField;
 use crate::poly::commitment::commitment_scheme::CommitmentScheme;
 use crate::poly::commitment::dory::DoryGlobals;
-use crate::poly::multilinear_polynomial::MultilinearPolynomial;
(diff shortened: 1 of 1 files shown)


jolt-core/src/zkvm/dag/jolt_dag.rs
```diff
@@ -499,11 +499,70 @@ impl JoltDAG {
...
+                    all_commitments[idx] = Some(commitment);
+                    all_hints[idx] = Some(hint);
+                }
+            }
+        }
+
+        let commitments: Vec<_> = all_commitments.into_iter().map(|c| c.unwrap()).collect();
+        let hints: Vec<_> = all_hints.into_iter().map(|h| h.unwrap()).collect();
+
+        // Build hint map preserving original order
```
markosg04: this is my attempt at playing around with the sorting of the workload. Seems to be better on small core machines and worse on the threadripper.

moodlezoup: hmm if it's worse on the threadripper let's just scrap it

markosg04: done

PR #1010, opened 2025-10-13
Fix/reduce virtual registers usage in amo instructinos

0xAndoroid:
And also fixes the virtual instruction tests

### DIFF
--- a/tracer/src/instruction/amoaddd.rs
+++ b/tracer/src/instruction/amoaddd.rs
@@ -13,13 +13,13 @@ use crate::{
     emulator::cpu::{Cpu, Xlen},
 };
 
-use super::{format::format_r_amo::FormatRAMO, Cycle, RISCVInstruction, RISCVTrace};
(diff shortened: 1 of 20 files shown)


tracer/src/instruction/format/format_r_amo.rs
```diff
@@ -0,0 +1,114 @@
...
+
+use super::{
+    normalize_register_value, InstructionFormat, InstructionRegisterState, NormalizedOperands,
+};
+
+/// FormatR specialized for AMO (Atomic Memory Operation) instructions
+/// Uses the same format as FormatR but with a custom RegisterState that
+/// constrains rs1 to be a valid memory address for testing
+#[derive(Default, Debug, Clone, Copy, PartialEq, Serialize, Deserialize)]
+pub struct FormatRAMO {
```
moodlezoup: nit: rename this to FormatAMO

PR #1011, opened 2025-10-13
Change HammingBooleanity to use split eq poly

quangvdao:
Discovered this while looking at hamming weight stuff.

### DIFF
--- a/jolt-core/src/zkvm/ram/hamming_booleanity.rs
+++ b/jolt-core/src/zkvm/ram/hamming_booleanity.rs
@@ -4,11 +4,11 @@ use std::rc::Rc;
 use crate::field::JoltField;
 use crate::poly::commitment::commitment_scheme::CommitmentScheme;
 use crate::poly::eq_poly::EqPolynomial;
-use crate::poly::split_eq_poly::GruenSplitEqPolynomial;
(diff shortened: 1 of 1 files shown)


jolt-core/src/zkvm/ram/hamming_booleanity.rs
```diff
@@ -97,52 +97,58 @@ impl<F: JoltField, T: Transcript> SumcheckInstance<F, T> for HammingBooleanitySu
...
+                            let E_in_eval = eq.E_in_current()[x_in];
+                            let h0 = H.get_bound_coeff(2 * j);
+                            let h1 = H.get_bound_coeff(2 * j + 1);
+                            let delta = h1 - h0;
+                            [E_in_eval * (h0.square() - h0), E_in_eval * delta.square()]
+                        })
+                        .reduce(|| [F::zero(); 2], |a, b| [a[0] + b[0], a[1] + b[1]]);
+                    [E_out_eval * inner[0], E_out_eval * inner[1]]
+                })
+                .reduce(|| [F::zero(); 2], |a, b| [a[0] + b[0], a[1] + b[1]])
```
0xAndoroid, line 140: Are we not using delayed reduction here?

quangvdao: Ah woops let me fix

PR #1014, opened 2025-10-13
Compression doc

JuI3s:


book/src/how/optimizations/compression.md
```diff
@@ -0,0 +1,117 @@
+# Torus-based compression
+
+We implement a torus-based compression method to compress an output of the pairing computation for the BN254 curve (viewed as an element in a degree 12 extension $\mathbb{F}_{q^{12}}$ over the base prime field $\mathbb{F}_q$) to two elements in a degree 2 field extension over the same prime field, thus achieving a threefold compression ratio with no information loss. The compression only affects the final exponentiation step of the pairing computation.
```
markosg04: would it be possible to include some motivation for why we want this type of compression? Generally the docs are made for the layman, so we might want to mention general proof compression, what the proof consists of, and how the torus compression helps (GT elements from dory being a significant amount)

markosg04: We may also want to include a reference paper / textbook for further background on the algebra in this page

markosg04: We may also want to contrast this method of proof compression to what currently exists in arkworks: i.e, we have 50% compression for G1, G2, and those methods do not affect the pairing primitive, whereas this one does

book/src/how/optimizations/compression.md
```diff
@@ -0,0 +1,117 @@
...
+\Phi_6(x) = x^2 - x + 1
+$$
+is the $6^{\text{th}}$ cyclotomic polynomial and 
+
+$$
+\Psi_6(x) = \frac{x^6 - 1}{\Phi_6(x)} = (x^3 + 1)(x + 1).
+$$
+
+Let $\xi \in \mathbb{F}_{q^2}$ be a sextic non-residue and identify 
+$$\mathbb{F}_{q^6} = \mathbb{F}_{q^2}(\xi^{\frac{1}{3}}) = \mathbb{F}_{q^2}(\tau)$$
```
markosg04, line 25: mostly for my own understanding: what does this notation here refer to when we treat $\mathbb{F}_{q^2}(\xi^{\frac{1}{3}})$ sort of like function notation?

book/src/how/optimizations/compression.md
```diff
@@ -0,0 +1,117 @@
+# Torus-based compression
```
markosg04, line 1: Perhaps we should make this page just a general "Proof Compression" page and the Torus-based compression can be its own complete section

book/src/how/optimizations/compression.md
```diff
@@ -0,0 +1,117 @@
...
+a + b\sigma = a + (b\sigma^{\frac{2}{3}})\sigma^{\frac{1}{3}} = a + (b\tau)\sigma^{\frac{1}{3}}
+$$
+
+and hence 
+
+$$
+a + b\sigma^{\frac{1}{3}} = a + (b\tau^{-1})\sigma,
+$$
+
+where $a$, $b$, and $\tau$ all are elements of $\mathbb{F}_{q^6} = $\mathbb{F}_{q^2}(\tau)$. To convert between elements in $\mathbb{F}_{q^{12}}$ written in $\mathbb{F}_{q^6}$-bases $(1, \sigma)$ and $(1, \sigma^{\frac{1}{3}})$, it suffices to do a multiplication or a division by $\tau$, and we shall see this can be implemented entirely in $\mathbb{F}_{q^2}$ arithmetic.
```
markosg04: this seems to not render for me on the book

book/src/how/optimizations/compression.md
```diff
@@ -0,0 +1,117 @@
...
+$$
+
+and
+
+$$
+b\tau^{-1} = c_1 + c_2\tau + c_0\xi^{-1}\tau^2,
+$$
+which gives the corresponding conversion formulae.
+
+
```
markosg04: maybe we can also briefly include how this concretely gets integrated into Jolt? or if not now then leave as TODO after it is integrated

book/src/how/optimizations/compression.md
JuI3s: F(a) is the smallest field containing F and a. This is standard notation: https://math.stackexchange.com/questions/1080336/adjoining-a-number-to-a-field

PR #1018, opened 2025-10-14
[JOLT-210] Omid/inline sdk

omibo:
This PR optimizes the SDK for `Sha2`, `Keccak256`, `Blake2`, and `Blake3` inlines to reduce the number of RISC-V instructions required to compute the hash, resulting in improved cycles per byte.

This PR also adds a new example, hash-bench, that reports the number of cycles for all four hash functions, both our inline and reference implementation. For Sha2, Blake2, and Keccak256, the input size is 32 KB; for Blake3, as it only supports 64 bytes for now, the input size is 64 bytes.


moodlezoup:
> Not sure whether we should keep `blake2-ex` and `blake3-ex` in examples/ or remove them, since they are also included in hash-bench.

I think we can just remove them

PR #1019, opened 2025-10-15
[JOLT-157] Optimize instruction `read_raf_checking` memory

moodlezoup:
Reduces memory usage in instruction read-checking/raf-evaluation sumcheck using three optimizations (see commit messages)

### DIFF
--- a/jolt-core/src/zkvm/instruction_lookups/read_raf_checking.rs
+++ b/jolt-core/src/zkvm/instruction_lookups/read_raf_checking.rs
@@ -388,25 +388,25 @@ impl<F: JoltField, T: Transcript> SumcheckInstance<F, T> for ReadRafSumcheck<F> 
                     }
 
                     [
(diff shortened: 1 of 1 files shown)


jolt-core/src/zkvm/instruction_lookups/read_raf_checking.rs
```diff
@@ -326,53 +342,87 @@ impl<F: JoltField, T: Transcript> SumcheckInstance<F, T> for ReadRafSumcheck<F>
...
-                        for j in 0..DEGREE {
-                            running[j] += new[j];
-                        }
-                        running
-                    },
-                )
-                .into_iter()
-                .map(F::from_montgomery_reduce)
+                    || [F::zero(); 4],
+                    |a, b| std::array::from_fn(|i| a[i] + b[i]),
```
0xAndoroid, line 399: We can do delayed reduction here

PR #1021, opened 2025-10-16
Miscellaneous memory optimizations

moodlezoup:
Optimizes the memory usage of the Twist read/write checking sumchecks, the bytecode read/raf sumcheck, and the opening reduction sumcheck (stage 7). 
Also simplifies the opening proof reduction code now that we've eliminated most of the committed dense polynomials in Jolt


jolt-core/src/poly/opening_proof.rs
```diff
@@ -1331,31 +1167,23 @@ where
 
         transcript.append_scalars(&reduced_opening_proof.sumcheck_claims);
 
-        let gamma: F = transcript.challenge_scalar();
-        let mut gamma_powers = vec![F::one()];
-        for i in 1..self.sumchecks.len() {
-            gamma_powers.push(gamma_powers[i - 1] * gamma);
-        }
+        let gamma_powers: Vec<F> = transcript.challenge_scalar_powers(self.sumchecks.len());
```
andrewmilson, line 1170: Nice!

PR #1023, opened 2025-10-17
Omid/fix advice

omibo:
This PR resolves an issue with advice inputs, enabling handling of varying sizes for trusted and untrusted advice.
Previously, verification failed at stage 3 if untrusted advice was larger than trusted advice.
This PR fixes this by placing the larger advice first in memory, followed by the smaller one.


common/src/jolt_device.rs
```diff
@@ -289,19 +300,45 @@ impl MemoryLayout {
...
+                .checked_add(max_trusted_advice_size)
+                .expect("trusted_advice_end overflow");
+            (trusted_start, trusted_end, untrusted_start, untrusted_end)
+        };
+
+        let input_start = if max_trusted_advice_size >= max_untrusted_advice_size {
+            untrusted_advice_end
+        } else {
+            trusted_advice_end
+        };
```
moodlezoup:
```suggestion
        let input_start = std::cmp::max(untrusted_advice_end, trusted_advice_end);
```

PR #1025, opened 2025-10-17
Fix/virtual sequence constraint

moodlezoup:
Adds a constraint to fix a soundness bug where a malicious prover can skip the first instruction(s) in a virtual sequence.

new constraint (pseudocode)
```
if isVirtual(j+1) && !isFirstInSequence(j+1) {
    assert!(DoNotUpdateUnexpandedPC(j))
}
```
"If the next instruction is virtual and not the first in its sequence, then the current instruction must be part of (and not the last instruction in) a virtual sequence")

### DIFF
--- a/jolt-core/src/zkvm/bytecode/read_raf_checking.rs
+++ b/jolt-core/src/zkvm/bytecode/read_raf_checking.rs
@@ -587,7 +587,7 @@ impl<F: JoltField> ReadRafSumcheck<F> {
             .map(|instruction| {
                 let instr = instruction.normalize();
                 let instr_flags = instruction.instruction_flags();
-                let circuit_flgas = instruction.circuit_flags();
+                let circuit_flags = instruction.circuit_flags();
                 let unexpanded_pc = instr.address;
                 let imm = instr.operands.imm;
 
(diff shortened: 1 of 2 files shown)


jolt-core/src/zkvm/instruction/ecall.rs
```diff
@@ -14,6 +14,7 @@ impl<const XLEN: usize> InstructionLookup<XLEN> for ECALL {
 impl Flags for ECALL {
     fn circuit_flags(&self) -> [bool; NUM_CIRCUIT_FLAGS] {
         let mut flags = [false; NUM_CIRCUIT_FLAGS];
+        flags[CircuitFlags::IsFirstInSequence as usize] = self.is_first_in_sequence;
```
omibo, line 17: Can ECALL be a virtual instruction?

jolt-core/src/zkvm/instruction/fence.rs
```diff
@@ -14,6 +14,7 @@ impl<const XLEN: usize> InstructionLookup<XLEN> for FENCE {
 impl Flags for FENCE {
     fn circuit_flags(&self) -> [bool; NUM_CIRCUIT_FLAGS] {
         let mut flags = [false; NUM_CIRCUIT_FLAGS];
+        flags[CircuitFlags::IsFirstInSequence as usize] = self.is_first_in_sequence;
```
omibo, line 17: Can FENCE be a virtual instruction?

tracer/src/utils/inline_helpers.rs
```diff
@@ -124,12 +124,13 @@ impl InstrAssembler {
...
     }
 
-    /// Finalize the instruction buffer: back-fill `inline_sequence_remaining`
+    /// Finalize the instruction buffer: back-fill `virtual_sequence_remaining`
     /// and return ownership of the underlying `Vec`.
     pub(crate) fn finalize(mut self) -> Vec<Instruction> {
         let len = self.sequence.len();
         for (i, instr) in self.sequence.iter_mut().enumerate() {
-            instr.set_inline_sequence_remaining(Some((len - i - 1) as u16));
+            instr.set_is_first_in_sequence(i == 0);
```
omibo, line 132: As this happens only for the first instruction, we can also extract this out.

omibo (approved):
We can probably remove the setting of circuit flags related to virtual instructions when we know the instructions are not virtual (PC-related instructions like beq, bne, etc. ).

PR #1038, opened 2025-10-22
Switch Dory matrix layout

moodlezoup:
Switches Dory matrix layout from "cycle-major" to "address-major" order. Changes `commit_rows` and `vector_matrix_product` implementations to reflect this change. 
Note that this change has implications for commitment time: on one hand we can expect dense polynomials to have more expensive multipairing; one the other, we can expect one-hot `commit_rows` to be faster (leveraging the uniformity of ones in each row)

### DIFF
--- a/jolt-core/src/msm/mod.rs
+++ b/jolt-core/src/msm/mod.rs
@@ -117,7 +117,6 @@ where
         }
     }
 
-    // #[tracing::instrument(skip_all)]
     fn msm_field_elements(
         bases: &[Self::MulBase],
         scalars: &[Self::ScalarField],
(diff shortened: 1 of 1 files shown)


jolt-core/src/msm/mod.rs
```diff
@@ -117,7 +117,7 @@ where
         }
     }
 
-    #[tracing::instrument(skip_all)]
+    // #[tracing::instrument(skip_all)]
```
markosg04: do we want these back or delete them entirely if not? applies to this whole file

jolt-core/src/poly/commitment/dory.rs
```diff
@@ -1544,6 +1386,7 @@ mod tests {
 
     #[test]
     #[serial]
+    #[ignore]
```
markosg04: Should we un-ignore now?

jolt-core/src/poly/commitment/dory.rs
```diff
@@ -1438,6 +1279,7 @@ mod tests {
...
```
markosg04: Should we un-ignore now?

jolt-core/src/msm/mod.rs
moodlezoup: will delete

PR #1040, opened 2025-10-22
refactor: DRY booleanity and hamming

markosg04:
Part of an on-going effort to clean up the sumcheck code. Val and RA Virtualization can be DRY'd up in a similar fashion.

### DIFF
--- a/jolt-core/src/subprotocols/hamming_weight.rs
+++ b/jolt-core/src/subprotocols/hamming_weight.rs
@@ -8,7 +8,6 @@ use std::{cell::RefCell, rc::Rc};
 use crate::{
     field::{JoltField, MulTrunc},
     poly::{
-        commitment::commitment_scheme::CommitmentScheme,
         multilinear_polynomial::{BindingOrder, MultilinearPolynomial, PolynomialBinding},
         opening_proof::{
             OpeningAccumulator, OpeningPoint, ProverOpeningAccumulator, SumcheckId,
@@ -16,11 +15,7 @@ use crate::{
         },
     },
     transcripts::Transcript,
-    utils::math::Math,
-    zkvm::{
-        dag::state_manager::StateManager,
-        witness::{compute_d_parameter, CommittedPolynomial, VirtualPolynomial, DTH_ROOT_OF_K},
-    },
+    zkvm::witness::{CommittedPolynomial, VirtualPolynomial},
 };
 
 use crate::subprotocols::sumcheck::SumcheckInstance;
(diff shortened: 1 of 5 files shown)


jolt-core/src/zkvm/bytecode/booleanity.rs
```diff
@@ -78,7 +41,7 @@ pub struct BooleanitySumcheck<F: JoltField> {
     r_address: Vec<F::Challenge>,
 }
 
-impl<F: JoltField> BooleanitySumcheck<F> {
+impl<F: JoltField> BytecodeBooleanitySumcheck<F> {
     #[tracing::instrument(skip_all, name = "BytecodeBooleanitySumcheck::new_prover")]
     pub fn new_prover(
```
markosg04: i think there could be ways to even avoid this verbose new prover, but i defer that to a future PR. Will be easier to think of the right solution as I go through more sumchecks

jolt-core/src/subprotocols/booleanity.rs
```diff
@@ -0,0 +1,596 @@
...
+
+    fn prover_state_mut(&mut self) -> Option<&mut BooleanityProverState<F>>;
+
+    fn get_r_cycle(&self, accumulator: &dyn OpeningAccumulator<F>) -> Vec<F::Challenge>;
+
+    fn booleanity_degree(&self) -> usize {
+        3
+    }
+
+    fn booleanity_num_rounds(&self) -> usize {
```
0xAndoroid:
```suggestion
    fn num_rounds(&self) -> usize {
```
Given that we are wrapping `B: BooleanitySumcheck` into a `Booleanity` struct, we don't actually need to prefix these methods, since they will not conflict with `SumcheckInstance` at all.
And then we can do `<self as BooleanitySumcheck>::degree()` or so.

jolt-core/src/subprotocols/hamming_weight.rs
```diff
@@ -0,0 +1,284 @@
...
+use crate::subprotocols::sumcheck::SumcheckInstance;
+
+#[derive(Allocative)]
+pub struct HammingWeightProverState<F: JoltField> {
+    /// ra polynomials
+    pub ra: Vec<MultilinearPolynomial<F>>,
+}
+
+/// Configuration trait for hamming weight sumchecks
+pub trait HammingWeightConfig {
```
0xAndoroid: Why do we need separate config trait? Why don't we just put these methods on the `HammingWeightSumcheck` trait?

jolt-core/src/subprotocols/booleanity.rs
```diff
@@ -0,0 +1,596 @@
...
+pub trait BooleanitySumcheck<F: JoltField, T: Transcript>:
+    BooleanityConfig + Send + Sync + MaybeAllocative
+{
+    fn gamma(&self) -> &[F::Challenge];
+
+    fn r_address(&self) -> &[F::Challenge];
+
+    fn prover_state(&self) -> Option<&BooleanityProverState<F>>;
+
+    fn prover_state_mut(&mut self) -> Option<&mut BooleanityProverState<F>>;
```
0xAndoroid: I think we can make these functions return `&BooleanittyProverState` and `&mut BooleanityProverState` without the `Option`, since we are not handling the case of no prover state regardless.

jolt-core/src/zkvm/bytecode/mod.rs
```diff
@@ -178,13 +178,13 @@ impl<F: JoltField, PCS: CommitmentScheme<Field = F>, T: Transcript> SumcheckStag
         sm: &mut StateManager<'_, F, T, PCS>,
     ) -> Vec<Box<dyn SumcheckInstance<F, T>>> {
         let read_checking = ReadRafSumcheck::new_verifier(sm);
-        let hamming_weight = HammingWeightSumcheck::new_verifier(sm);
-        let booleanity = BooleanitySumcheck::new_verifier(sm);
+        let hamming_weight = BytecodeHammingWeightSumcheck::new_verifier(sm);
```
0xAndoroid:
```suggestion
        let hamming_weight = HammingWeightSumcheck::new_verifier(sm);
```
I think it's best to still keep it like this, given that sumchecks for given instance of twist/shout are defined within one module. But I don't feel too strong about this.

jolt-core/src/zkvm/instruction_lookups/booleanity.rs
```diff
@@ -1,66 +1,38 @@
...
-
 #[derive(Allocative)]
-pub struct BooleanitySumcheck<F: JoltField> {
+pub struct InstructionBooleanitySumcheck<F: JoltField> {
     /// Batching challenges γ_i (optimized challenges)
-    /// TODO: special casing for the first challenge to be F::one()
     gamma: [F::Challenge; D],
     prover_state: Option<BooleanityProverState<F>>,
     r_address: Vec<F::Challenge>,
+    r_cycle: Vec<F::Challenge>,
```
0xAndoroid: Why do we need this argument now?

jolt-core/src/zkvm/instruction_lookups/booleanity.rs
```diff
@@ -108,464 +112,65 @@ impl<F: JoltField> BooleanitySumcheck<F> {
...
 
-        let gruen_evals =
-            D_poly.gruen_evals_deg_3(quadratic_coeffs_f[0], quadratic_coeffs_f[1], adjusted_claim);
-        vec![
-            p.eq_r_r * gruen_evals[0],
-            p.eq_r_r * gruen_evals[1],
-            p.eq_r_r * gruen_evals[2],
-        ]
+    fn get_r_cycle(&self, _accumulator: &dyn OpeningAccumulator<F>) -> Vec<F::Challenge> {
+        self.r_cycle.clone()
```
0xAndoroid: This should be pulled from accumulator.

jolt-core/src/zkvm/ram/booleanity.rs
```diff
@@ -250,422 +259,29 @@ impl<F: JoltField, T: Transcript> SumcheckInstance<F, T> for BooleanitySumcheck<
...
+                    .collect();
 
-                // Drop G arrays and F array as they're no longer needed in phase 2
+                // Drop G arrays as they're no longer needed
                 let g = std::mem::take(&mut ps.G);
-                drop_in_background_thread(g);
-
-                let f = std::mem::take(&mut ps.F);
-                drop_in_background_thread(f);
+                crate::utils::thread::drop_in_background_thread(g);
```
0xAndoroid: nit: import `drop_in_background_thread`

jolt-core/src/zkvm/bytecode/mod.rs
```diff
@@ -191,6 +210,28 @@ impl<F: JoltField, PCS: CommitmentScheme<Field = F>, T: Transcript> SumcheckStag
 
 #[inline(always)]
 #[tracing::instrument(skip_all, name = "Bytecode::compute_ra_evals")]
+fn compute_bytecode_h_indices(
```
quangvdao, line 280: The tracing doesn't match the function?

jolt-core/src/subprotocols/booleanity.rs
```diff
@@ -0,0 +1,914 @@
...
+        } else {
+            // For instruction verifier, get from virtual polynomial
+            accumulator
+                .get_virtual_polynomial_opening(
+                    VirtualPolynomial::LookupOutput,
+                    SumcheckId::SpartanOuter,
+                )
+                .0
+                .r
+                .clone()
```
moodlezoup: aren't these the same point?

PR #1043, opened 2025-10-23
Add inline performance metrics to Jolt Book

omibo:
Provides Cycles per Byte and number of cycles for Sha2, Keccak256, and Blake2 hash functions, both reference and inline implementation.


book/src/how/optimizations/inlines.md
```diff
@@ -89,6 +89,30 @@ unsafe {
...
+
+*Note: Blake3 currently supports inputs up to 64 bytes. Full implementation for larger inputs is in development.*
+
+#### Proving Time
+
+Proving time depends on the hardware running the prover. On a MacBook M4, the Jolt prover achieves approximately 400 kHz throughput, processing 400,000 RISC-V cycles per second.
+
+For example, using the inline Blake2B implementation:
+- **32KB hash proving time**: ~0.85 seconds
+- **Effective throughput**: ~38.5 KB/second
```
markosg04: most up to date we can use 500 kHz on MBP and let's also include example of 32-core cpu 1.5 mHz

PR #1044, opened 2025-10-23
perf: optimize ripple-carry propagation in bigint multiplication

Lil-Duckling-22:
Replaced sequential carry propagation loop with optimized while loop that generates instructions more efficiently. 

Reduces the number of generated assembly instructions for cases where carry propagation terminates early, improving performance in zkVM context where instruction count directly impacts proof size.


0xAndoroid:
Thank you for the contribution but we generally do not accept trivial PRs to discourage airdrop farming.

PR #1048, opened 2025-10-23
Omid/test lookup trace

omibo:
This PR provides the following new features/tests:

- For all instructions (except branch instructions and assert instructions), it is checked that the `to_lookup_output` matches the tracer behaviour exactly.
- For random cycle generation, it is considered that if the register is 0, the value should be 0; if the registers are the same, they should equal each other.

Also, it provides the following fixes:

- DIV, REM, DIVW, REMW all failed for left input 0 and right input -1 in Stage5 verification.
- DIV and REM fail for left input `i64::MIN` and right input `-1` in Stage1 verification.
- JALR `to_lookup_output()` did not correctly clear the LSB bit and mismatched the tracer.
- JAL and AUIPC instructions (in tracer, not zkVM) panic in case of overflow.
- JALR did not use `normalize_imm()`.


tracer/src/instruction/format/format_virtual_right_shift_i.rs
```diff
@@ -13,19 +13,25 @@ pub struct FormatVirtualRightShiftI {
...
+
 impl InstructionRegisterState for RegisterStateFormatVirtualI {
     #[cfg(any(feature = "test-utils", test))]
-    fn random(rng: &mut rand::rngs::StdRng) -> Self {
+    fn random(rng: &mut rand::rngs::StdRng, operands: &NormalizedOperands) -> Self {
         use rand::RngCore;
         Self {
             rd: (rng.next_u64(), rng.next_u64()),
-            rs1: rng.next_u64(),
+            rs1: if operands.rs1 == 0 { 0 } else { rng.next_u64() },
```
moodlezoup: shouldn't we have a case where `if operands.rs1 == operands.rd`, then the rs1 value equals the old rd value?
this applies to some of the other formats as well

moodlezoup (commented):
lgtm, just one suggestion

tracer/src/instruction/format/format_virtual_right_shift_i.rs
omibo: Correct. I fix it.

PR #1049, opened 2025-10-23
refactor: nightly imports

0xAndoroid:
Adds all the beauty of imports formatting, but requires cargo fmt nightly.


moodlezoup:
> @moodlezoup feel free to close this PR if you wouldn't like to introduce nightly cargo fmt. I don't feel too strongly about this, but would be nice to have it.

I slightly prefer to keep everything (including fmt) on stable

PR #1053, opened 2025-10-28
Split SumcheckInstance trait to verifier and prover trait

andrewmilson:


moodlezoup (commented):
This looks good to me! Only suggestion is to have the prover and verifier opening accumulators as parameters of `SumcheckInstanceProver::input_claim` and `SumcheckInstanceVerifier::input_claim`, respectively

jolt-core/src/subprotocols/sumcheck_verifier.rs
```diff
@@ -0,0 +1,45 @@
...
+        &self,
+        accumulator: &mut VerifierOpeningAccumulator<F>,
+        transcript: &mut T,
+        sumcheck_challenges: &[F::Challenge],
+    );
+
+    /// Optional univariate-skip verify hook. If implemented, the driver passes the full
+    /// first-round univariate polynomial and the input claim (pre-uniskip). The instance
+    /// should run any strong, const-generic checks internally and return (r0, next_claim).
+    fn uniskip_verify(
```
quangvdao: Given that we still have separate trait for univariate skip first round, and a separate struct for univariate skip proof, perhaps you could remove this function from the sumcheck API.

We can refactor univariate skip into this later on if needed.

jolt-core/src/zkvm/instruction_lookups/ra_virtual.rs
```diff
@@ -36,31 +42,29 @@ use rayon::prelude::*;
...
+    #[allocative(skip)]
+    params: RaSumcheckParams<F>,
 }
 
-impl<F: JoltField> RaSumcheck<F> {
-    #[tracing::instrument(skip_all, name = "InstructionRaSumcheck::new_prover")]
-    pub fn new_prover<ProofTranscript: Transcript, PCS: CommitmentScheme<Field = F>>(
+impl<F: JoltField> RaSumcheckProver<F> {
+    #[tracing::instrument(skip_all, name = "InstructionRaSumcheckProver::gen_from_state_manager")]
+    pub fn gen_from_state_manager<ProofTranscript: Transcript, PCS: CommitmentScheme<Field = F>>(
```
quangvdao: Why change the name of the function here?

jolt-core/src/zkvm/ram/mod.rs
```diff
@@ -559,10 +564,11 @@ where
...
         state_manager: &mut StateManager<'_, F, ProofTranscript, PCS>,
-    ) -> Vec<Box<dyn SumcheckInstance<F, ProofTranscript>>> {
-        let raf_evaluation = RafEvaluationSumcheck::new_verifier(state_manager);
-        let read_write_checking = RamReadWriteChecking::new_verifier(state_manager);
-        let output_check = OutputSumcheck::new_verifier(state_manager);
+    ) -> Vec<Box<dyn SumcheckInstanceVerifier<F, ProofTranscript>>> {
+        let raf_evaluation = RafEvaluationSumcheckVerifier::new_from_state_manager(state_manager);
+        let read_write_checking =
+            RamReadWriteCheckingVerifier::new_from_state_manager(state_manager);
+        let output_check = OutputSumcheckVerifier::new_from_state_manager(state_manager);
```
quangvdao: The naming for prover & verifier side seems to be mismatched, and on prover side there's also inconsistency in naming.

Why did you move away from `new_prover` and `new_verifier`?

andrewmilson: I don't like new, from or into for things that are costly.
Think it's better the name of the function indicates there's some kind of cost to it.
Hence gen_ for prover and new_ for verifier. 
Happy to change though

quangvdao: Makes sense. Then the only comment is to standardize naming "gen_from_state_manager" instead of "gen_prover_from_state_manager".

quangvdao: Could you also make this renaming for the univariate skip stuff as well? They are in only two files right now, outer.rs and product.rs

quangvdao: Oh I suppose you need to separate Prover & Verifier traits for univariate skip as well. Should be quite doable & beneficial in the long run

jolt-core/src/zkvm/instruction_lookups/ra_virtual.rs
andrewmilson: Renamed to just gen

jolt-core/src/zkvm/ram/mod.rs
andrewmilson: Oh thanks didn’t realise I named some of them gen_prover_from_state_manager 🤦‍♂️.
I’ll stick with `gen` for prover and `new` for verifier. I was considering the sumchecks would eventually have two constructors - `gen_from_state_manager` and `gen_from_params` — to decouple sumchecks from the state manager for easier unit testing, but that’d require broader changes, so I’ll leave it for now.

jolt-core/src/subprotocols/sumcheck_prover.rs
```diff
@@ -0,0 +1,72 @@
...
+    fn input_claim(&self) -> F;
+
+    /// Computes the full univariate polynomial to be sent in the uni-skip round.
+    /// Returns a degree-bounded `UniPoly` with exactly `DEGREE_BOUND + 1` coefficients.
+    fn compute_poly(&mut self) -> UniPoly<F>;
+
+    // TODO: add flamegraph support
+    // #[cfg(feature = "allocative")]
+    // fn update_flamegraph(&self, flamegraph: &mut FlameGraphBuilder);
+}
```
markosg04, line 72: nit: i know this was originally in sumcheck.rs but should go in univariate_skip.rs if we are going to have a separate file for uni skip?

jolt-core/src/subprotocols/booleanity.rs
```diff
@@ -489,114 +386,200 @@ where
...
+//         log_t,
+//         gamma,
+//         r_address,
+//         r_cycle,
+//         polynomial_types,
+//         sumcheck_id,
+//         virtual_poly,
+//         prover_state: None,
+//     }
+// }
```
markosg04:
```suggestion
```

jolt-core/src/subprotocols/booleanity.rs
```diff
@@ -489,114 +386,200 @@ where
...
+    pub r_address: Vec<F::Challenge>,
+    /// Cycle binding point
+    pub r_cycle: Vec<F::Challenge>,
+    /// Polynomial types for opening accumulator
+    pub polynomial_types: Vec<CommittedPolynomial>,
+    /// Sumcheck ID for opening accumulator
+    pub sumcheck_id: SumcheckId,
+    /// Optional virtual polynomial for r_cycle
+    pub virtual_poly: Option<VirtualPolynomial>,
+}
```
markosg04, line 528: nit: should this go top of file? same could apply to hamming, up to ur pref

jolt-core/src/zkvm/dag/state_manager.rs
```diff
@@ -276,6 +276,16 @@ where
...
 
+    pub fn get_trace_len(&self) -> usize {
+        if let Some(ref verifier_state) = self.verifier_state {
+            verifier_state.trace_length
+        } else if let Some(ref prover_state) = self.prover_state {
+            prover_state.trace.len()
+        } else {
+            panic!("Neither prover nor verifier state initialized");
+        }
+    }
```
markosg04, line 297:
```suggestion
pub fn get_trace_len(&self) -> usize {
    if let Some(verifier_state) = self.verifier_state.as_ref() {
        verifier_state.trace_length
    } else if let Some(prover_state) = self.prover_state.as_ref() {
        prover_state.trace.len()
    } else {
        panic!("Neither prover nor verifier state initialized");
    }
}
```

nit: does this work? I don't think we use `ref` anywhere else in code and I think in general `as_ref()` these days is much more common and possibly clearer. Same may apply elsewhere

jolt-core/src/zkvm/bytecode/read_raf_checking.rs
```diff
@@ -282,239 +224,643 @@ impl<F: JoltField> ReadRafSumcheck<F> {
...
+                transcript,
+                vec![CommittedPolynomial::BytecodeRa(i)],
+                SumcheckId::BytecodeReadRaf,
+                [r_address, &r_cycle.r].concat(),
+            );
+        });
+    }
+}
+
+struct ReadRafSumcheckParams<F: JoltField> {
```
markosg04, line 629: nit: Usually when i hear `params` i think immutable data. Wdyt of standardizing `state` vs `params` (I think state is what we had everywhere before)

jolt-core/src/zkvm/instruction_lookups/read_raf_checking.rs
```diff
@@ -178,30 +180,24 @@ struct ReadRafProverState<F: JoltField> {
...
-    gamma_sqr: F,
-    /// Prover-only state across rounds; None in verifier.
-    prover_state: Option<ReadRafProverState<F>>,
-
-    /// log2(T): number of cycle variables (last rounds bind cycles).
-    log_T: usize,
+pub struct ReadRafSumcheckProver<F: JoltField> {
+    prover_state: ReadRafProverState<F>,
+    #[allocative(skip)]
+    params: ReadRafSumcheckParams<F>,
```
markosg04: here we have both a state and params struct, but I think elsewhere we unified it into one? Here the state has the mutable data and params are immutable

jolt-core/src/zkvm/spartan/inner.rs
```diff
@@ -207,41 +139,90 @@ impl<F: JoltField, T: Transcript> SumcheckInstance<F, T> for InnerSumcheck<F> {
...
+///
+///   ( A_small(rx_constr, r) + γ·B_small(rx_constr, r) ) · z(r).
+///
+/// Final check (verifier): compute
+///   - eval_a = evaluate_uniform_a_at_point(rx_constr, r),
+///   - eval_b = evaluate_uniform_b_at_point(rx_constr, r),
+///   - eval_z = evaluate_z_mle_with_segment_evals(segment_evals, r, true),
+///     where `segment_evals` are the cached witness openings at `r_cycle` from the outer stage.
+///
+/// Then `expected = (eval_a + γ·eval_b) · eval_z`, and accept iff output_claim == expected.
```
markosg04, line 188: nit: maybe these should just be doc strings so that people reading the code find the explanation when they first open the file

jolt-core/src/zkvm/dag/jolt_dag.rs
```diff
@@ -29,997 +35,968 @@ use crate::zkvm::ProverDebugInfo;
 use allocative::FlameGraphBuilder;
 use anyhow::Context;
 
-pub enum JoltDAG {}
```
markosg04, line 36: what's the reasoning behind moving to just module level functions? Will it be easier to do a clean split this way I am guessing?

jolt-core/src/zkvm/ram/ra_virtual.rs
```diff
@@ -45,111 +44,40 @@ use rayon::prelude::*;
...
 
 #[derive(Allocative)]
-pub struct RaProverState<F: JoltField> {
+pub struct RaSumcheckProver<F: JoltField> {
     /// `ra` polys to be constructed based addresses
     ra_i_polys: Vec<RaPolynomial<u8, F>>,
     /// eq poly
     eq_poly: MultilinearPolynomial<F>,
+    #[allocative(skip)]
+    params: RaSumcheckParams<F>,
```
markosg04, line 53: Pointing this out here just as an example for my other comments: sometimes we have a `SumcheckState` under prover, sometimes `Params` holds all of the state, and then sometimes we have the mutable things not in a separate struct (as is the case here)

jolt-core/src/zkvm/spartan/product.rs
```diff
@@ -916,144 +912,138 @@ impl<F: JoltField, T: Transcript> SumcheckInstance<F, T> for ProductVirtualRemai
...
-impl<F: JoltField, T: Transcript> SumcheckInstance<F, T> for ProductVirtualInner<F> {
+impl<F: JoltField, T: Transcript> SumcheckInstanceProver<F, T> for ProductVirtualInnerProver<F> {
     // Trivial sum-check, no round, no computation of messages or binding, only verification
     fn degree(&self) -> usize {
         0
     }
+
     fn num_rounds(&self) -> usize {
         0
     }
```
markosg04, line 938: i wonder if there is a better way to handle this case in the code cuz having a degree 0 sum-check seems strange to me

markosg04 (approved):
Great work! This was a wicked change. Thank you for going through this pain and for the little cleanups around the codebase as well.

Comments are mostly just some questions / nits related to namings and consistency of the sum-check patterns. Namely, I am unsure if we should use `params` everywhere vs `state` and sometimes we might be mixing things up a bit. Feel free to address as you see fit.

jolt-core/src/zkvm/spartan/inner.rs
markosg04: can apply elsewhere, too

jolt-core/src/subprotocols/booleanity.rs
andrewmilson: tyty

jolt-core/src/zkvm/bytecode/read_raf_checking.rs
andrewmilson: These are immutable aren't they?

jolt-core/src/zkvm/dag/jolt_dag.rs
andrewmilson: It's more idiomatic to use a module to group functions than an enum.

jolt-core/src/zkvm/spartan/product.rs
andrewmilson: Yeah agree. Let's resolve in a seperate PR though.

jolt-core/src/zkvm/instruction_lookups/read_raf_checking.rs
andrewmilson: Whoops thanks removed

jolt-core/src/zkvm/ram/ra_virtual.rs
andrewmilson: The Params struct is for parameters shared between the prover and verifier. Thanks for pointing out the ProverState in the other comment (I now removed it). I removed the ProverState struct because the *SumcheckVerifier and *SumcheckProver structs can store their state directly which avoids unnecessary indirection.

jolt-core/src/zkvm/spartan/inner.rs
andrewmilson: Once the files are split it will be the first thing in the file again

jolt-core/src/zkvm/instruction_lookups/read_raf_checking.rs
andrewmilson: Addressed in other comment.

jolt-core/src/zkvm/dag/state_manager.rs
andrewmilson: It works. I mean they're equivalent and IMO just as readable/clear as each other. 
ref is used in state_manager in a bunch of other places.

jolt-core/src/subprotocols/booleanity.rs
andrewmilson: I think the verifier struct with the doc string of what the sumcheck is for should go at the top

jolt-core/src/subprotocols/sumcheck_prover.rs
andrewmilson: Would rather not introduce univariate_skip.rs now. 
Would rather keep it consistent with how it was before. 
Can be addressed in a seperate PR

PR #1054, opened 2025-10-28
feat: streaming commitments

markosg04:
Adds streaming for the commitment phase of Jolt:
- Introduces a `StreamingCommitmentScheme` trait and its implementation for Dory. This achieves rougly `NUM_THREADS * sqrt(T)` space for the AFGHO (msm then multi pairing) commitments for Dory.
- Introduces and integrates a `LazyTraceIterator` abstraction to lazily produce `&[Cycle]`. For now, this is an extra item on the state manager. As we slowly move towards streaming all sum-checks we can use this object to maintain compatibility with the non-streaming sum-checks. Eventually we will only need to hold this iterator + some check-point abstraction.

We keep the legacy non-streaming version in the code for now in case we need to swap back. Proof serialization and other miscellaneous things have been updated as well.

### DIFF
--- a/jolt-core/src/zkvm/ram/mod.rs
+++ b/jolt-core/src/zkvm/ram/mod.rs
@@ -577,7 +577,7 @@ where
         prover_accumulate_advice(state_manager);
         let K = state_manager.ram_K;
 
-        let (_, trace, program_io, _) = state_manager.get_prover_data();
+        let (_, _, trace, program_io, _) = state_manager.get_prover_data();
         let memory_layout = &program_io.memory_layout;
         let d = compute_d_parameter(K);
         let log_k_chunk = DTH_ROOT_OF_K.log_2();
@@ -730,7 +730,7 @@ where
         &mut self,
         state_manager: &mut StateManager<'_, F, ProofTranscript, PCS>,
     ) -> Vec<Box<dyn SumcheckInstance<F, ProofTranscript>>> {
-        let (_, trace, program_io, _) = state_manager.get_prover_data();
+        let (_, _, trace, program_io, _) = state_manager.get_prover_data();
         let memory_layout = &program_io.memory_layout;
         let d = compute_d_parameter(state_manager.ram_K);
         let num_rounds = DTH_ROOT_OF_K.log_2();


jolt-core/src/zkvm/witness.rs
```diff
@@ -184,40 +181,21 @@ impl CommittedPolynomial {
...
         }
     }
-
-    // TODO(moodlezoup): return Result<Self>
-    pub fn from_index(index: usize) -> Self {
-        unsafe {
-            ALL_COMMITTED_POLYNOMIALS
-                .get()
-                .expect("ALL_COMMITTED_POLYNOMIALS is uninitialized")[index]
-        }
```
moodlezoup, line 194: what's the rationale for deleting `from_index` and `to_index`?

jolt-core/src/zkvm/dag/proof_serialization.rs
```diff
@@ -325,11 +334,36 @@ impl CanonicalSerialize for CommittedPolynomial {
         mut writer: W,
         compress: Compress,
     ) -> Result<(), SerializationError> {
-        self.to_index().serialize_with_mode(&mut writer, compress)
```
moodlezoup, line 328: see question about deleting `to_index` and `from_index`, this (de)serialization stuff was cleaner before

jolt-core/src/zkvm/dag/jolt_dag.rs
```diff
@@ -806,13 +808,89 @@ impl JoltDAG {
...
+
+        let chunks: Vec<Vec<_>> = lazy_trace
+            .as_ref()
+            .expect("Lazy trace not found!")
+            .clone()
+            .pad_using(T, |_| Cycle::NoOp)
+            .chunks(row_len)
+            .into_iter()
+            .map(|chunk| chunk.collect::<Vec<_>>())
+            .collect();
```
moodlezoup: don't these collects defeat the purpose of streaming?

jolt-core/src/poly/commitment/dory.rs
```diff
@@ -1337,6 +1338,128 @@ impl CommitmentScheme for DoryCommitmentScheme {
...
+            let T = DoryGlobals::get_T();
+            let mut row_commitments: Vec<_> = chunks.into_par_iter().map(|r| r[0]).collect();
+            // Pad row commitments since we don't pad streamed trace.
+            let row_pad_count = T / DoryGlobals::get_num_columns() - row_commitments.len();
+            assert!(T / DoryGlobals::get_num_columns() >= row_commitments.len());
+            row_commitments.extend(vec![JoltG1Wrapper::identity(); row_pad_count]);
+            let commitment = JoltBn254::multi_pair(
+                &row_commitments,
+                &state.setup.g2_vec()[..row_commitments.len()],
+            );
```
moodlezoup: I don't think we need to pad here?
```suggestion
            let commitment = JoltBn254::multi_pair(
                &row_commitments,
                &state.setup.g2_vec()[..row_commitments.len()],
            );
```

PR #1056, opened 2025-10-30
optimistaion/better_parallels

abiswas3:
- Nearly **2x** speedups on Multilinear parallel bind for compact and dense poly's -- with 2 lines of code changed (though it took a while to find this).
- There are few more of these optimisations; but I need to double check them as I pulled in all the new code changes.
I'll push a sequence of mini-PR's with speedups 

Screenshot 2025-10-30 at 04 04 40


moodlezoup (approved):
nice find

PR #1058, opened 2025-10-30
Minor comment change

JuI3s:
Very minor comment change to make the description more readable. Happy for discussion.

### DIFF
--- a/book/src/how/optimizations/compression.md
+++ b/book/src/how/optimizations/compression.md
@@ -149,3 +149,7 @@ pub fn compressible_fq12_to_fq12(value: CompressibleFq12) -> Fq12 {
     }
 }
 ```
+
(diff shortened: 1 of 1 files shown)


JuI3s:
> IMO those comments are readable and useful for layman. Thoughts?

I basically just removed the below occurrences (all but one place) which repeats too many times and adjusted a few indentations. The global dory state is just a wrapper around the stuff defined right on top of it so I think simplifying it makes things a lot clearer.

"///Dory works by viewing the coefficients of a polynomial as a matrix.
/// In order to batch Dory opening proofs together for polynomials of
/// different lengths, we fix one dimension of the matrix (the number of
/// columns, i.e. the row length). This is the fixed dimension, the number"

PR #1060, opened 2025-10-30
Update ZkLean extractor to work with 64-bit Twist & Shout

protoben:
This PR updates the ZkLean extractor for the current version of Jolt. It includes the following changes:
- All 64-bit Twist & Shout lookup-table MLEs are extracted into Lean. Both the internal representation of the MLEs and their Lean representations have been modified to allow them to scale to the larger polynomials.
- For each instruction, the corresponding lookup table (if any) and circuit flags are extracted.
- The R1CS extraction is updated to work with the current handling of constraints in the codebase.
- For each extracted lookup table polynomial, we emit a test to ensure that it evaluates the same as the corresponding Rust function.
- Tests for the extractor have been added back into 

We make just two changes to `jolt-core`.
- We derive `IntoStaticStr` for `LookupTables` enum. This helps in automatically emitting relevant names for the lookup-table MLEs in Lean.
- We make the `LC::decompose` function `pub`. This makes it possible for us to extract the R1CS constraints.

### DIFF
--- a/zklean-extractor/src/constants.rs
+++ b/zklean-extractor/src/constants.rs
@@ -5,7 +5,7 @@ use common::{constants, jolt_device::MemoryConfig};
 /// compatibility.
 pub trait JoltParameterSet {
     /// The architecture size.
-    const WORD_SIZE: usize;
+    const XLEN: usize;
     /// The memory config to use
     const MEMORY_CONFIG: MemoryConfig;
 }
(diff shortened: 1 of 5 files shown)


zklean-extractor/src/constants.rs
```diff
@@ -1,35 +1,28 @@
...
-    /// Size of a materialized subtable
-    const M: usize;
-    /// The number of bits in a subtable entry; this should always be `log2(M)`
-    const LOG_M: usize = Self::M.ilog2() as usize;
-    /// The architecture size; although this isn't exposed in the Jolt codebase, it can be inferred
-    /// from `C` and `LOG_M`, since those constants need to be able to decompose a pair of
-    /// `WORD_SIZE` registers
-    const WORD_SIZE: usize = (Self::LOG_M * Self::C) / 2;
+    /// The architecture size.
+    const WORD_SIZE: usize;
```
0xAndoroid: We prefer using XLEN for arch size since word is always 32 bits.

PR #1062, opened 2025-10-31
Remove smart pointers from state manager

andrewmilson:


0xAndoroid (approved):
Great catch. Everything's much cleaner now.

PR #1066, opened 2025-11-03
Remove VerifierState from state manager

andrewmilson:


jolt-core/src/zkvm/dag/jolt_dag.rs
0xAndoroid, line 1: I strongly prefer current way of dag construction. I don't think that this refactor to the dag adds any clarity, while making the code a lot more verbose. I would like to keep the `stageX_instances` initialization.

moodlezoup: My 2 cents here is that I like having the different stage methods for organization
If we end up going with my [proposed](https://gist.github.com/moodlezoup/045545ffb139b0b0d4e238b28fe7394e) `JoltProverBackend` trait, we'll end up having the stages explicit at the whole-DAG level, rather than implementing `SumcheckStagesProver` for each component. So we'd end up somewhere in-between the current state and the changes in this PR.

@andrewmilson are you planning to reintroduce the stage methods with the `JoltProverBackend` trait, or are you proposing that we toss them out entirely?

andrewmilson: Plan is still to introduce stage methods on JoltProverBackend trait. I just want to slightly modify them so we don't pass state_manager and instead store state on the prover struct (I think this will make different backend state management much easier for us). The goal of this stack is to decouple everything in our codebase from state manager (which I think is good in itself for various reasons - testing for instance) which will help make this change and the prover/verifier split possible. Does that sound ok?

moodlezoup: ok, that sounds good to me

andrewmilson: I moved each stage to its own function. 
Let me know what you guys think.

moodlezoup: looks good to me, let's add the symmetric prover functions for each stage in https://github.com/a16z/jolt/pull/1071

0xAndoroid (approved):
looks good to me

PR #1071, opened 2025-11-03
Decouple prover sumchecks from StateManager

andrewmilson:


jolt-core/src/zkvm/verifier.rs
```diff
@@ -0,0 +1,524 @@
...
+
+        #[cfg(test)]
+        {
+            // Need to initialize globals because the verifier computes commitments
+            // in `VerifierOpeningAccumulator::append` inside of a `#[cfg(test)]` block
+            let _guard = DoryGlobals::initialize(
+                crate::zkvm::witness::DTH_ROOT_OF_K,
+                proof.trace_length.next_power_of_two(),
+            );
+        }
```
moodlezoup: I think we no longer need this

jolt-core/src/zkvm/prover.rs
```diff
@@ -0,0 +1,1420 @@
...
+    fn prove_stage1(
+        &mut self,
+    ) -> (
+        UniSkipFirstRoundProof<F, ProofTranscript>,
+        SumcheckInstanceProof<F, ProofTranscript>,
+    ) {
+        #[cfg(not(target_arch = "wasm32"))]
+        print_current_memory_usage("Stage 1 baseline");
+        let span = tracing::span!(tracing::Level::INFO, "Stage 1 sumchecks");
+        let _guard = span.enter();
```
moodlezoup: nit: we can use the function decorator instead of `tracing::span!` 
```suggestion
    #[tracing::instrument(skip_all)]
    fn prove_stage1(
        &mut self,
    ) -> (
        UniSkipFirstRoundProof<F, ProofTranscript>,
        SumcheckInstanceProof<F, ProofTranscript>,
    ) {
        #[cfg(not(target_arch = "wasm32"))]
        print_current_memory_usage("Stage 1 baseline");
```

jolt-core/src/zkvm/prover.rs
```diff
@@ -0,0 +1,1420 @@
...
+        );
+        let mut spartan_product_virtual_claim_check =
+            ProductVirtualInnerProver::new(&self.opening_accumulator, &mut self.transcript);
+        let instruction_ra_evals =
+            instruction_lookups::compute_ra_evals(&self.trace, &self.opening_accumulator);
+        let mut lookups_ra_hamming_weight = instruction_lookups::gen_ra_hamming_weight_prover(
+            &instruction_ra_evals,
+            &mut self.transcript,
+        );
+        self.instruction_ra_evals = Some(instruction_ra_evals);
```
moodlezoup: perhaps we should just move this hamming weight sumcheck to Stage 6 (alongside the instruction booleanity sumcheck) so that we don't have to cache `instruction_ra_evals` like this. The memory footprint of this Hamming weight sumcheck is negligible so we should be able to move it around freely

jolt-core/src/zkvm/prover.rs
```diff
@@ -0,0 +1,1420 @@
...
+        {
+            let mut flamegraph = FlameGraphBuilder::default();
+            for sumcheck in stage2_instances.iter() {
+                sumcheck.update_flamegraph(&mut flamegraph);
+            }
+            write_flamegraph_svg(flamegraph, "stage2_end_flamechart.svg");
+        }
+
+        // Drop product immediately because it has lifetimes.
+        drop(spartan_product_virtual_remainder);
```
moodlezoup: I slightly prefer to keep trace as an `Arc<Vec<Cycle>>` instead introducing the lifetime, so that we can use `drop_in_background` for everything. Same for the shift sumcheck.

jolt-core/src/zkvm/prover.rs
```diff
@@ -0,0 +1,1420 @@
...
+    'a,
+    F: JoltField,
+    PCS: StreamingCommitmentScheme<Field = F>,
+    ProofTranscript: Transcript,
+> {
+    pub preprocessing: &'a JoltProverPreprocessing<F, PCS>,
+    pub program_io: JoltDevice,
+    pub lazy_trace: LazyTraceIterator,
+    pub trace: Vec<Cycle>,
+    pub final_memory_state: Memory,
```
moodlezoup: do we need this on this struct? we already have `final_ram_state`

jolt-core/src/zkvm/prover.rs
```diff
@@ -0,0 +1,1420 @@
...
+    pub untrusted_advice_polynomial: Option<MultilinearPolynomial<F>>,
+    pub trusted_advice_commitment: Option<PCS::Commitment>,
+    pub trusted_advice_polynomial: Option<MultilinearPolynomial<F>>,
```
moodlezoup: nit: throw these three into a separate `JoltAdvice` struct

moodlezoup (commented):
Thanks for your hard work on this! A few comments but this is in pretty good shape I think

jolt-core/src/zkvm/prover.rs
andrewmilson: awesome, done

jolt-core/src/zkvm/prover.rs
andrewmilson: removed, nice catch

PR #1088, opened 2025-11-07
Omid/fix advice

omibo:
This PR fix a bug in advice inputs, adds end to end test for merkle-tree example, and provide advice documentation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1864,6 +1864,7 @@ dependencies = [
  "eyre",
  "jolt-core",
  "jolt-sdk",
+ "mimalloc",
  "rand 0.8.5",
  "rmp-serde",
  "serde",
@@ -2102,6 +2103,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f9fbbcab51052fe104eb5e5d351cf728d30a5be1fe14d9be8a3b097481fb97de"
 
 [[package]]
+name = "libmimalloc-sys"
+version = "0.1.44"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "667f4fec20f29dfc6bc7357c582d91796c169ad7e2fce709468aefeb2c099870"
+dependencies = [
+ "cc",
+ "libc",
+]
+
+[[package]]
 name = "libredox"
 version = "0.1.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2237,6 +2248,15 @@ version = "0.1.0"
 dependencies = [
  "jolt-inlines-sha2",
  "jolt-sdk",
+]
+
+[[package]]
+name = "mimalloc"
+version = "0.1.48"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e1ee66a4b64c74f4ef288bcbb9192ad9c3feaad75193129ac8509af543894fd8"
+dependencies = [
+ "libmimalloc-sys",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -88,6 +88,7 @@ path = "./src/main.rs"
 [dependencies]
 clap.workspace = true
 eyre.workspace = true
+mimalloc.workspace = true
 rand.workspace = true
 sysinfo.workspace = true
 syn.workspace = true
@@ -149,7 +150,9 @@ clap = { version = "4.5.4", features = ["derive"] }
 eyre = "0.6.12"
 anyhow = "1.0.97"
 thiserror = "1.0.58"
-derive_more = { version = "2.0.1", features = ["from"], default-features = false }
+derive_more = { version = "2.0.1", features = [
+    "from",
+], default-features = false }
 strum = { version = "0.26.3", default-features = false }
 strum_macros = "0.26.4"
 itertools = "0.10.0"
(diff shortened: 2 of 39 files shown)


book/src/how/architecture/ram.md
```diff
@@ -20,10 +20,10 @@ Our RISC-V [emulator](./emulation.md) is configured to use `0x80000000` as the D
 For the purposes of the memory checking argument, we remap the memory address to a witness index:
 
 ```rust
-(address - memory_layout.input_start) / 8 + 1
+(address - memory_layout.lowest_address) / 8 + 1
```
moodlezoup: While you're updating this, I believe we removed the +1 here
```suggestion
(address - memory_layout.lowest_address) / 8
```

book/src/imgs/Screenshot 2025-11-06 at 2.58.54 PM.png
markosg04, line 1: can we rename the image?

book/src/usage/guests_hosts/guests.md
```diff
@@ -38,6 +38,21 @@ fn sha3(input: &[u8]) -> [u8; 32] {
...
+## Prover and Verifier Views of Guest Program
+A guest program consists of two main components: program code and inputs. Both the prover and verifier know the program code (and how it compiles to RISC-V instructions).
+
+Jolt supports three types of inputs:
+
+1. **Input**: These are public inputs known to both the prover and verifier. The prover proves that given the bytecode and these inputs, the claimed output is correct.
+
+2. **Untrusted Input**: These inputs are only known to the prover, not the verifier. Here, the prover proves that there exists some input that produces the claimed output, without revealing the actual input to the verifier.
+
+3. **Trusted Advice**: Similar to untrusted input, but the verifier has a commitment to the input generated by an external party (not the prover). The prover proves that there exists an input matching the given commitment that, when executed with the program code, produces the claimed output.
```
markosg04: Let's be careful to not confuse reader with the ZK property and this

rust_out
markosg04, line 1: delete?

jolt-core/src/zkvm/ram/mod.rs
```diff
@@ -113,6 +113,36 @@ pub struct RamDagProver {
...
+/// * `index` - Where to start writing
+/// * `bytes` - Input bytes to be written
+/// * `initial_state` - Memory state at program start
+/// * `final_state` - Memory state at program end
+pub fn populate_memory_states(
+    mut index: usize,
+    bytes: &[u8],
+    mut initial_state: Option<&mut Vec<u64>>,
+    mut final_state: Option<&mut Vec<u64>>,
+) {
```
markosg04, line 128: i feel like some of this memory populating stuff we are doing seems hacky. I don't have any better solution in my head rn though so just flagging

markosg04 (commented):
Left a few nitty comments but would appreciate a pass from @moodlezoup on this one

book/src/usage/guests_hosts/guests.md
omibo: Right. I updated the doc.

jolt-core/src/zkvm/ram/mod.rs
omibo: I agree. I was considering ways to avoid storing these values and instead retrieve them directly from program_io and memory when binding them in the first round.
It needs more thought, and I'll address it in a separate PR.

book/src/usage/guests_hosts/guests.md
```diff
@@ -38,6 +38,23 @@ fn sha3(input: &[u8]) -> [u8; 32] {
...
+## Prover and Verifier Views of Guest Program
+A guest program consists of two main components: program code and inputs. Both the prover and verifier know the program code (and how it compiles to RISC-V instructions).
+
+Jolt supports three types of inputs:
+
+1. **Input**: These are inputs known to both the prover and verifier. The prover proves that given the bytecode and these inputs, the claimed output is correct.
+
+2. **Untrusted Input**: These inputs are known to the prover for generating the proof, but the verifier does not know them for the purpose of verification. Here, the prover proves that there exists some input that produces the claimed output. 
+
+3. **Trusted Advice**: Similar to untrusted input, but the verifier has a commitment to the input generated by an external party (not the prover). The prover proves that there exists an input matching the given commitment that, when executed with the program code, produces the claimed output.
```
moodlezoup:
```suggestion
1. **Public Input**: These are inputs known to both the prover and verifier. The prover proves that given the bytecode and these inputs, the claimed output is correct.

2. **Untrusted Advice**: These inputs are known to the prover for generating the proof, but the verifier does not know them for the purpose of verification. Here, the prover proves that there exists some input that produces the claimed output. 

3. **Trusted Advice**: Similar to untrusted advice, but the verifier has a commitment to the input generated by an external party (not the prover). The prover proves that there exists an input matching the given commitment that, when executed with the program code, produces the claimed output.
```

book/src/how/architecture/ram.md
```diff
@@ -73,6 +73,49 @@ $$
...
+#### Example
+
+Consider a concrete example with:
+- Total memory: 8192 elements ($2^{13}$)
+- Trusted advice: 1024 elements ($2^{10}$)
+- Untrusted advice: 128 elements ($2^7$)
+
+The memory layout would be:
+- Addresses with the 3 MSBs as `000`: trusted advice region (all combinations of the remaining 10 bits)
+- Addresses with the MSBs as `000100`: untrusted advice region (all combinations of the remaining 7 LSBs)
```
moodlezoup: Typo here I think?

```suggestion
The memory layout would be:
- Addresses with the 3 MSBs as `000`: trusted advice region (all combinations of the remaining 10 bits)
- Addresses with the MSBs as `001000`: untrusted advice region (all combinations of the remaining 7 LSBs)
```

moodlezoup (commented):
Nice work, just some nits on the docs

PR #1094, opened 2025-11-08
Improve instruction read raf checking initialization

quangvdao:
Better parallelization for init of InstructionReadRaf, and adding some more detailed tracing spans.

I tested this by benching on fibonacci on threadripper - no more sequential step, reduced time compared to previous run

### DIFF
--- a/jolt-core/src/zkvm/instruction_lookups/read_raf_checking.rs
+++ b/jolt-core/src/zkvm/instruction_lookups/read_raf_checking.rs
@@ -280,6 +280,7 @@ impl<'a, F: JoltField> ReadRafSumcheckProver<F> {
             let _guard = span.enter();
             cycle_data
                 .par_iter()
+                .with_min_len(4096)
                 .fold(
                     || vec![Vec::new(); num_tables],
                     |mut buckets, data| {


quangvdao:
I checked on large instances (~2^27 I think?)

Will double check tomorrow on threadripper

jolt-core/src/zkvm/instruction_lookups/read_raf_checking.rs
```diff
@@ -251,39 +251,55 @@ impl<'a, F: JoltField> ReadRafSumcheckProver<F> {
...
-                    data.table.and_then(|t| {
-                        if LookupTables::<XLEN>::enum_index(&t) == t_idx {
-                            Some(data.idx)
-                        } else {
-                            None
+        let lookup_indices_by_table: Vec<Vec<usize>> = {
+            let span = tracing::span!(tracing::Level::INFO, "build lookup_indices_by_table");
+            let _guard = span.enter();
+            cycle_data
+                .par_iter()
```
moodlezoup, line 282: nit: let's add a `.with_min_len(4096)` here to avoid excessive allocator work

quangvdao:
@markosg04 Just tested on threadripper with fibonacci & sha2-chain on 2^28. Speedup only - gen goes from 4.2s to 2.8s.

PR #1095, opened 2025-11-08
Reduce space usage in `prepare_all_sumchecks`

quangvdao:
We reduce memory by not materializing the full eq table, instead doing the parallel iter over split struct.

TODO: make the iteration cleaner via the split eq poly function in #1097


jolt-core/src/poly/one_hot_polynomial.rs
```diff
@@ -151,35 +140,72 @@ impl<F: JoltField> OneHotPolynomialProverOpening<F> {
...
-        let D_coeffs_for_G = &eq.merged_D.as_ref().unwrap();
-
-        // Compute G as described in Section 6.3
-        let G = nonzero_indices
-            .par_chunks(chunk_size)
+        // Compute G without materializing the dense eq(r_cycle, j) table:
+        // G[k] = sum_{j: nonzero_indices[j] = k} eq(r_cycle, j).
+        // (described in Section 6.3 of the Twist/Shout paper)
+        // Custom parallelization: parallel over E_out, sequential over E_in and the last bit.
+        // This avoids nested Rayon layers while still computing d_j on-the-fly from the split representation.
```
markosg04, line 149: is the nested rayon here strictly worse?

jolt-core/src/poly/one_hot_polynomial.rs
```diff
@@ -151,35 +140,72 @@ impl<F: JoltField> OneHotPolynomialProverOpening<F> {
...
-                    if let Some(k) = k {
-                        result[*k as usize] += D_coeffs_for_G[j];
+            .map(|(x_out, &high)| {
+                let mut local_unreduced: Vec<F::Unreduced<9>> =
+                    unsafe_allocate_zero_vec(polynomial.K);
+                let mut touched_indices: Vec<usize> = Vec::new();
+                let mut touched_flags: Vec<u8> = vec![0u8; polynomial.K];
+                let x_out_base = x_out << (x_in_bits + 1);
+
+                for (x_in, &low) in E_in.iter().enumerate() {
```
markosg04, line 167: might be nice to have a single comment about what this touched flags and touched indices loop is doing

jolt-core/src/poly/one_hot_polynomial.rs
```diff
@@ -151,35 +140,72 @@ impl<F: JoltField> OneHotPolynomialProverOpening<F> {
...
             .reduce(
                 || unsafe_allocate_zero_vec(polynomial.K),
-                |mut running, new| {
-                    running
-                        .par_iter_mut()
-                        .zip(new.into_par_iter())
-                        .for_each(|(x, y)| *x += y);
-                    running
+                |mut a, b| {
+                    for (x, y) in a.iter_mut().zip(b) {
```
markosg04, line 205: can this be par iter? or we are purposefully avoiding it

quangvdao:
@markosg04 Thanks! My plan is to have some time on the threadripper tomorrow to test out any performance diffs. So far none on my local mac.

The code will also be nice with the extra function for merging split eq poly in #1097, so perhaps let's investigate that first.

jolt-core/src/poly/one_hot_polynomial.rs
quangvdao: Unclear, needs testing tmr

jolt-core/src/poly/one_hot_polynomial.rs
```diff
@@ -151,35 +140,72 @@ impl<F: JoltField> OneHotPolynomialProverOpening<F> {
...
                     }
-                    j += 1;
                 }
-                result
+                // Materialize as reduced F and apply the high factor only for touched indices
+                let mut local_g: Vec<F> = unsafe_allocate_zero_vec(polynomial.K);
+                if high.is_zero() {
+                    return local_g;
+                }
+                let apply_high = high != F::one();
```
moodlezoup, line 195: isn't the probability of `high` being 0 or 1 negligible?

PR #1096, opened 2025-11-08
MLE product sum for d=32

quangvdao:
There are some more playing around with precise extrapolation, use of `#[inline]`, and so on. But for now this is probably acceptable.

Also added bench for this compared with naive method for all parameters (d = 4, 8, 16, 32).

### DIFF
--- a/jolt-core/src/subprotocols/mles_product_sum.rs
+++ b/jolt-core/src/subprotocols/mles_product_sum.rs
@@ -115,10 +115,10 @@ fn compute_mles_product_sum_d16<F: JoltField>(
     // combined sum over (x_out, x_in) after Montgomery reduction.
     let sum_evals_arr: [F; 16] = eq_poly.par_fold_out_in_unreduced::<9, 16>(&|g| {
         // Build pairs[(p0, p1); 16] on the stack.
-        let mut pairs: [(F, F); 16] = core::array::from_fn(|i| {
+        let pairs: [(F, F); 16] = core::array::from_fn(|i| {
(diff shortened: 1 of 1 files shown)


quangvdao:
@0xAndoroid I reworked this a bunch so now the files should look a lot nicer, and slightly more efficient.

Can you take another look?

jolt-core/src/subprotocols/mles_product_sum.rs
```diff
@@ -57,17 +90,78 @@ pub fn compute_mles_product_sum<F: JoltField>(
...
+    // `par_fold_out_in_unreduced` handles the split-eq structure and returns the
+    // combined sum over (x_out, x_in) after Montgomery reduction.
+    let sum_evals_arr: [F; 16] = eq_poly.par_fold_out_in_unreduced::<9, 16>(&|g| {
+        // Build pairs[(p0, p1); 16] on the stack.
+        let mut pairs: [(F, F); 16] = core::array::from_fn(|_| (F::zero(), F::zero()));
+        for (i, mle) in mles.iter().enumerate() {
+            let p0 = mle.get_bound_coeff(2 * g);
+            let p1 = mle.get_bound_coeff(2 * g + 1);
+            pairs[i] = (p0, p1);
+        }
```
moodlezoup: nit:
```suggestion
        let mut pairs: [(F, F); 16] = core::array::from_fn(|i| {
            let p0 = mles[i].get_bound_coeff(2 * g);
            let p1 = mles[i].get_bound_coeff(2 * g + 1);
            (p0, p1)            
        });
```

PR #1098, opened 2025-11-09
feat: Decrease one-hot K = 256 => K = 16 for small cycles

quangvdao:
Made possible by #1096 . Need to gate behind feature flag because we are changing compile-time constants

Preliminary benchmark on my M4 mac: sha3 goes from 6.8s to 1.9s!


quangvdao:
This PR should be parked for now until those changes can be made (either by me, preferably by someone else)

PR #1100, opened 2025-11-10
DRY for split-eq double summation / folding pattern

quangvdao:
Seeking advice on whether the current auxiliary function is generic enough

### Theme
DRY split-eq folding, consistent unreduced accumulation, cleaner product/output sumchecks.

### Key Changes
- Split-eq folding
  - Consolidated all bespoke fold-out/in logic into a single reusable helper that accumulates over E_in, does one Montgomery reduction, and scales by E_out. Applied across Spartan (outer, product), booleanity, RAM/register checks, and openings.

- Product virtualization
  - Switched to centralized constraints/evaluation and a fused product evaluator. Dropped local constants/Lagrange plumbing. Extended univariate-skip evals now flow through the shared split-eq fold with R² scaling built into the split-eq state.

- Unreduced inner accumulations
  - RAM and register read/write checks now accumulate with unreduced multiplies in the inner loop, reduce once per outer E_out weight, and keep outer accumulation unreduced until the end. Fewer reductions; better alignment with Montgomery-friendly accumulation.

- Output check
  - Reintroduced split-eq: compute q(X) = io_mask · (val_final − val_io), extract constant/quadratic coefficients, and evaluate via the degree-3 split-eq routine. Removed the EQ table, simplified binding, and fixed r′ endianness to match cached openings.

- API and plumbing
  - Constructors take scalars (ram_K, trace_len, n_cycle_vars, twist_sumcheck_switch_index) instead of a state manager. Normalized naming (twist sumcheck switch index) and made binding/order consistent with where openings are cached.

### Impact
- Less duplication, fewer reductions, clearer parallel fold utilization.
- Modest expected speedups in sumcheck-heavy paths.
- No protocol/transcript changes; internal-only refactors.

### Verification
- All four claims (split-eq consolidation, product cleanup, unreduced inner loops for RAM/registers, split-eq in output check) are correct.


0xAndoroid:
Can we also do this in the `mle_product_sum.rs`?

jolt-core/src/zkvm/instruction_lookups/read_raf_checking.rs
```diff
@@ -620,69 +620,90 @@ impl<F: JoltField, T: Transcript> SumcheckInstanceProver<F, T> for ReadRafSumche
...
-                    ]
-                })
-                .reduce(
-                    || [F::Unreduced::zero(); 4],
-                    |a, b| std::array::from_fn(|i| a[i] + b[i]),
-                );
+            // Fold across (x_out, x_in) using Spartan's split-eq, mirror weights from Branch
+            let [eval_at_0_spartan, eval_at_inf_spartan, eval_at_0_branch, eval_at_inf_branch] =
+                self.eq_r_spartan
+                    .par_fold_out_in(
```
moodlezoup, line 641: I'm tempted to say we should just have a separate `self.eq_r_branch.par_fold_out_in` here... afaict the performance hit would be pretty minimal, wdyt?

moodlezoup (approved):
this was much needed, ty!

jolt-core/src/zkvm/instruction_lookups/read_raf_checking.rs
quangvdao: I think we should avoid that esp. as we move on to streaming. This code is lengthier but has the optimal access pattern into the polys, doing a single pass instead of two passes

quangvdao: Later on if we can register this as two separate consumers that then get executed in lockstep when streaming, then I think that's fine too

moodlezoup: fair enough, let's plan to on the two consumer approach

PR #1101, opened 2025-11-11
refactor: new dory implementation

markosg04:
Some things this refactor does:

- uses the new and beautiful dory implementation (which itself brings new optimizations thanks to @quangvdao )
- pushes most of the backend implementation into dory instead of in Jolt (including serialization)
- Refines some of the jank ways we were optimizing things. Namely, caching. Introduces a good non-hacky system for this available via a dory flag.
- allows for non-square matrices in dory
- Fixes patch hell
- Imports `dory-pcs`
- Full streaming. Commit + Vector matrix product. Dory in Jolt should now be O(sqrt(T)) space

### DIFF
--- a/jolt-core/src/zkvm/prover.rs
+++ b/jolt-core/src/zkvm/prover.rs
@@ -192,7 +192,11 @@ impl<'a, F: JoltField, PCS: StreamingCommitmentScheme<Field = F>, ProofTranscrip
 
         // Setup trace length and padding
         let unpadded_trace_len = trace.len();
-        let padded_trace_len = (trace.len() + 1).next_power_of_two();
+        let padded_trace_len = if unpadded_trace_len < 256 {
+            256 // ensures that T >= k^{1/D}
+        } else {
+            (trace.len() + 1).next_power_of_two()
+        };
         trace.resize(padded_trace_len, Cycle::NoOp);
 
         // Calculate K for DoryGlobals initialization


jolt-core/src/poly/rlc_polynomial.rs
```diff
@@ -1,46 +1,136 @@
...
+        #[cfg_attr(test, allow(unused_variables))] streaming_context: Option<(
+            LazyTraceIterator,
+            Arc<RLCStreamingData>,
+        )>,
     ) -> Self {
         debug_assert_eq!(polynomials.len(), coefficients.len());
+        debug_assert_eq!(poly_ids.len(), coefficients.len());
+
+        // Disable streaming in test mode. In order to check the joint commitment consistency we need fully materialized RLC poly.
+        #[cfg(not(test))]
```
0xAndoroid: Is this supposed to be `cfg(test)`

jolt-core/src/poly/rlc_polynomial.rs
```diff
@@ -184,43 +274,154 @@ impl<F: JoltField> RLCPolynomial<F> {
...
+                let (_, pre_value, post_value) = cycle.rd_write();
+                F::from_i128(post_value as i128 - pre_value as i128)
+            }
+            CommittedPolynomial::RamInc => match cycle.ram_access() {
+                tracer::instruction::RAMAccess::Write(write) => {
+                    F::from_i128(write.post_value as i128 - write.pre_value as i128)
+                }
+                _ => F::zero(),
+            },
+            _ => F::zero(), // One-hot polynomials handled below
```
0xAndoroid: To prevent future footguns, let's just exhaustively specify all the options and panic! / return 0.

jolt-core/src/poly/rlc_polynomial.rs
```diff
@@ -184,43 +274,154 @@ impl<F: JoltField> RLCPolynomial<F> {
...
+            }
+            CommittedPolynomial::RamRa(idx) => remap_address(
+                cycle.ram_access().address() as u64,
+                &preprocessing.memory_layout,
+            )
+            .map(|address| {
+                (address as usize >> (DTH_ROOT_OF_K.log_2() * (preprocessing.ram_d - 1 - idx)))
+                    % DTH_ROOT_OF_K
+            }),
+            _ => None,
```
0xAndoroid: Same as on top

jolt-core/src/zkvm/mod.rs
```diff
@@ -668,6 +668,7 @@ mod tests {
 
     #[test]
     #[serial]
+    #[ignore] // @TODO(markosg04) this test is broken
```
0xAndoroid: was this missed?

jolt-core/src/zkvm/witness.rs
```diff
@@ -188,6 +187,110 @@ impl AllCommittedPolynomials {
...
+        F: JoltField,
+        PCS: StreamingCommitmentScheme<Field = F>,
+    {
+        match self {
+            CommittedPolynomial::RdInc => {
+                let row: Vec<i128> = row_cycles
+                    .iter()
+                    .map(|cycle| {
+                        let (_, pre_value, post_value) = cycle.rd_write();
+                        post_value as i128 - pre_value as i128
```
0xAndoroid, line 230: I believe this logic is used quite frequently in the witness and other places, and it might be beneficial to factor it out. I also think there's a place in this PR that does this (I left a footgun comment there).

jolt-core/src/poly/commitment/dory/wrappers.rs
```diff
@@ -0,0 +1,331 @@
...
+    fn append_to_transcript<S: Transcript>(&self, transcript: &mut S) {
+        transcript.append_serializable(*self);
+    }
+}
+
+#[inline]
+pub fn jolt_to_ark(f: &Fr) -> ArkFr {
+    // SAFETY: ArkFr and Fr have the same memory layout
+    unsafe { std::mem::transmute_copy(f) }
+}
```
0xAndoroid, line 59: Can we just implement `Into<Fr> for ArkFr` in Dory?

jolt-core/src/poly/commitment/dory/wrappers.rs
```diff
@@ -0,0 +1,331 @@
...
+
+#[inline]
+pub fn ark_to_jolt(ark: &ArkFr) -> Fr {
+    // SAFETY: ArkFr and Fr have the same memory layout
+    unsafe { std::mem::transmute_copy(ark) }
+}
```
0xAndoroid, line 65: Same

jolt-core/src/poly/commitment/dory/commitment_scheme.rs
```diff
@@ -0,0 +1,332 @@
...
+    #[tracing::instrument(skip_all, name = "DoryCommitmentScheme::compute_tier1_commitment")]
+    fn compute_tier1_commitment<T: SmallScalar>(
+        setup: &Self::ProverSetup,
+        chunk: &[T],
+    ) -> Self::ChunkState {
+        debug_assert_eq!(chunk.len(), DoryGlobals::get_num_columns());
+
+        let row_len = DoryGlobals::get_num_columns();
+        let g1_slice =
+            unsafe { std::slice::from_raw_parts(setup.g1_vec.as_ptr(), setup.g1_vec.len()) };
```
0xAndoroid, line 245: Is it possible to eliminate all `unsafe` from Jolt repo in favor of having them in the dory repo? This, for example, could be something like `setup.g1_slice()` function. Not feeling too strong about it.

jolt-core/src/poly/commitment/dory/wrappers.rs
0xAndoroid: This would also be nice for other type conversions.

0xAndoroid (approved):
Nice. I left some suggestions, but I don't feel strongly about any of them.

jolt-core/src/poly/rlc_polynomial.rs
markosg04: It is semantically correct. but cfg(not(test)) is kinda unintuitive so I'll swap the order to make it clearer

jolt-core/src/zkvm/witness.rs
markosg04: I agree, but I suspect we will do this more in the sum-check streaming and I'm not yet too sure where the right place is to put such a helper / organize these types of streaming utils. I made note of this for when we have a better idea

jolt-core/src/poly/commitment/dory/commitment_scheme.rs
markosg04: I feel like because this is a jolt-specific implementation detail that it should prolly live in our bridge implementation. It would be nice to abstract away the unsafe things but also if they live here it might prevent us from misusing the function if we forget there is some unsafe rust involved

jolt-core/src/zkvm/mod.rs
markosg04: fixed.

jolt-core/src/poly/commitment/commitment_scheme.rs
```diff
@@ -136,37 +135,26 @@ pub trait CommitmentScheme: Clone + Sync + Send + 'static {
...
-    fn process_chunk_field(
-        cached_data: &Self::CachedData,
-        chunk: &[Self::Field],
-    ) -> Self::ChunkState;
-
-    /// Process a chunk of one-hot values
-    fn process_chunk_onehot(
-        cached_data: &Self::CachedData,
+    /// Compute tier 1 commitment for a chunk of one-hot values
+    fn compute_tier1_commitment_onehot(
```
moodlezoup: I kinda liked `process_chunk` and `finalize` better than `compute_tier1_commitment` and `compute_tier2_commitment`; it seems less Dory-specific

jolt-core/src/poly/commitment/commitment_scheme.rs
```diff
@@ -136,37 +135,26 @@ pub trait CommitmentScheme: Clone + Sync + Send + 'static {
...
-    /// Prepare cached data that will be shared across all polynomial commitments
-    fn prepare_cached_data(setup: &Self::ProverSetup) -> Self::CachedData;
-
-    // First `process_chunk` is a tier 1 commitment (MSM)
-
-    /// Process a chunk of small scalar values
-    fn process_chunk<T: SmallScalar>(
-        cached_data: &Self::CachedData,
+    /// Compute tier 1 commitment for a chunk of small scalar values
+    fn compute_tier1_commitment<T: crate::utils::small_scalar::SmallScalar>(
```
moodlezoup: nit: import instead of fully-qualified path

jolt-core/src/poly/commitment/dory/jolt_dory_routines.rs
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    fn fold_field_vectors(left: &mut [ArkFr], right: &[ArkFr], scalar: &ArkFr) {
+        assert_eq!(left.len(), right.len(), "Lengths must match");
+        left.par_iter_mut()
+            .zip(right.par_iter())
+            .for_each(|(l, r)| {
+                *l = *l * *scalar + *r;
+            });
+    }
```
moodlezoup, line 189: nit: DRY the two `fold_field_vectors`

jolt-core/src/poly/rlc_polynomial.rs
```diff
@@ -1,109 +1,203 @@
...
+                    result.one_hot_rlc.push((coefficients[i], poly));
+                }
+            }
+
+            result
         }
 
-        result
+        // Streaming mode in non-test builds
+        #[cfg(not(test))]
```
moodlezoup: hmm it concerns me that we aren't testing this codepath in our tests... not immediately clear to me what we should do about it

jolt-core/src/poly/commitment/dory/wrappers.rs
```diff
@@ -0,0 +1,331 @@
...
+            .par_chunks(row_len)
+            .map(|row| {
+                let field_row: Vec<Fr> = row
+                    .iter()
+                    .map(|&b| if b { Fr::from(1u64) } else { Fr::from(0u64) })
+                    .collect();
+                ArkG1(
+                    VariableBaseMSM::msm_field_elements(&bases[..field_row.len()], &field_row)
+                        .unwrap(),
+                )
```
moodlezoup: nit: I know we aren't using this, but using `msm_u8` (or computing the boolean MSM inline here) would be better

jolt-core/src/poly/commitment/dory/commitment_scheme.rs
```diff
@@ -0,0 +1,345 @@
...
+    type BatchedProof = Vec<ArkDoryProof>;
+    type OpeningProofHint = Vec<ArkG1>;
+
+    fn setup_prover(max_num_vars: usize) -> Self::ProverSetup {
+        let _span = trace_span!("DoryCommitmentScheme::setup_prover").entered();
+        let setup = ArkworksProverSetup::new_from_urs(&mut OsRng, max_num_vars);
+
+        // Initialize the prepared point cache for faster multi-pairings
+        // Skips cache initialization during tests to avoid shared state issues
+        #[cfg(not(test))]
```
moodlezoup: same thing here, I'd like to test this codepath. What are the shared state issues? Can they be fixed using [#[serial]](https://docs.rs/serial_test/latest/serial_test/)?

PR #1107, opened 2025-11-13
fix(tracer): add overflow checks inside of rem virtual sequences

0xAndoroid:


0xAndoroid:
Closing in favor of #1118

PR #1110, opened 2025-11-17
Remove spartan inner sumcheck

andrewmilson:


jolt-core/src/zkvm/spartan/outer.rs
```diff
@@ -565,19 +464,24 @@ impl<F: JoltField, T: Transcript> SumcheckInstanceProver<F, T> for OuterRemainin
     }
 }
 
-pub struct OuterRemainingSumcheckVerifier<F: JoltField> {
+pub struct OuterRemainingSumcheckVerifier<'a, F: JoltField> {
     params: OuterRemainingSumcheckParams<F>,
+    key: &'a UniformSpartanKey<F>,
```
quangvdao: Can we make this lifetime-less?

jolt-core/src/zkvm/spartan/outer.rs
```diff
@@ -596,10 +500,16 @@ impl<F: JoltField, T: Transcript> SumcheckInstanceVerifier<F, T>
...
-            .get_virtual_polynomial_opening(VirtualPolynomial::SpartanAz, SumcheckId::SpartanOuter);
-        let (_, claim_Bz) = accumulator
-            .get_virtual_polynomial_opening(VirtualPolynomial::SpartanBz, SumcheckId::SpartanOuter);
+        let r1cs_input_evals = ALL_R1CS_INPUTS.map(|input| {
+            accumulator
+                .get_virtual_polynomial_opening((&input).into(), SumcheckId::SpartanOuter)
+                .1
+        });
+
+        let rx_constr = &[sumcheck_challenges[0], self.params.r0_uniskip];
```
quangvdao, line 506: Maybe add a line of comment here on what this is

jolt-core/src/zkvm/spartan/outer.rs
```diff
@@ -596,10 +500,16 @@ impl<F: JoltField, T: Transcript> SumcheckInstanceVerifier<F, T>
...
+        let Az_Bz_prod = self
+            .key
+            .evaluate_Az_Bz_product_at_point(rx_constr, r1cs_input_evals);
```
quangvdao: Also add comment here on what this is

Forpee:
Can someone give a high-level overview of what enabled this?

andrewmilson:
@Forpee yeah, in spartan outer sumcheck the prover sent claims for a_claim = sum_y A_bound_rows(y)z(y), b_claim = sum_y B_bound_rows(y)z(y) to compute a_claim * b_claim. Inner sumcheck did sumcheck over y but it's simpler just not to do this. Now the verifier just computes a_claim and b_claim directly from the R1CS input evaluations. The cost is only 18 extra field mul but we don't append a_claim or b_claim to the transcript (more expensive than 18 field mul) so it's actually simpler and cheaper.

jolt-core/src/zkvm/spartan/outer.rs
andrewmilson: Added comment. Also renamed to evaluate_inner_sum_product_at_point

PR #1114, opened 2025-11-17
Feat/streaming_prover

abiswas3:
A generalised sum-check API. 
*NOTE*: Parallel performance of compute message sub-optimal right now due to increased complexity of building general-sized multi-quadratic for general window_sizes >=1

### DIFF
--- a/jolt-core/src/zkvm/spartan/outer.rs
+++ b/jolt-core/src/zkvm/spartan/outer.rs
@@ -556,6 +556,10 @@ impl<F: JoltField> OuterSharedState<F> {
     }
 
     #[allow(clippy::too_many_arguments)]
+    #[tracing::instrument(
+        skip_all,
+        name = "OuterSharedState::extrapolate_from_binary_grid_to_tertiary_grid"
+    )]
     fn extrapolate_from_binary_grid_to_tertiary_grid(
         &self,
         acc_az: &mut [Acc5U<F>],
@@ -746,6 +750,7 @@ impl<F: JoltField> OuterSharedState<F> {
         self.t_prime_poly = Some(MultiquadraticPolynomial::new(window_size, res));
     }
 
+    #[tracing::instrument(skip_all, name = "OuterSharedState::compute_t_evals")]
     pub fn compute_t_evals(&self, window_size: usize) -> (F, F) {
         let t_prime_poly = self
             .t_prime_poly
@@ -786,6 +791,7 @@ pub struct OuterStreamingWindow<F: JoltField> {
 impl<F: JoltField> StreamingSumcheckWindow<F> for OuterStreamingWindow<F> {
     type Shared = OuterSharedState<F>;
 
+    #[tracing::instrument(skip_all, name = "OuterStreamingWindow::initialize")]
     fn initialize(shared: &mut Self::Shared, window_size: usize) -> Self {
         shared.compute_evaluation_grid_from_trace(window_size);
         Self {
@@ -793,6 +799,7 @@ impl<F: JoltField> StreamingSumcheckWindow<F> for OuterStreamingWindow<F> {
         }
     }
 
+    #[tracing::instrument(skip_all, name = "OuterStreamingWindow::compute_message")]
     fn compute_message(
         &self,
         shared: &Self::Shared,
@@ -805,6 +812,7 @@ impl<F: JoltField> StreamingSumcheckWindow<F> for OuterStreamingWindow<F> {
             .gruen_poly_deg_3(t_prime_0, t_prime_inf, previous_claim)
     }
 
+    #[tracing::instrument(skip_all, name = "OuterStreamingWindow::ingest_challenge")]
     fn ingest_challenge(&mut self, shared: &mut Self::Shared, r_j: F::Challenge, _round: usize) {
         shared.split_eq_poly.bind(r_j);
 
@@ -823,6 +831,10 @@ pub struct OuterLinearStage<F: JoltField> {
 }
 
 impl<F: JoltField> OuterLinearStage<F> {
+    #[tracing::instrument(
+        skip_all,
+        name = "OuterLinearStage::fused_materialise_polynomials_general_with_multiquadratic"
+    )]
     fn fused_materialise_polynomials_general_with_multiquadratic(
         shared: &mut OuterSharedState<F>,
         window_size: usize,
@@ -1007,6 +1019,10 @@ impl<F: JoltField> OuterLinearStage<F> {
         )
     }
 
+    #[tracing::instrument(
+        skip_all,
+        name = "OuterLinearStage::fused_materialise_polynomials_round_zero"
+    )]
     fn fused_materialise_polynomials_round_zero(
         shared: &mut OuterSharedState<F>,
         num_vars: usize,
@@ -1371,6 +1387,7 @@ impl<F: JoltField> LinearSumcheckStage<F> for OuterLinearStage<F> {
     type Shared = OuterSharedState<F>;
     type Streaming = OuterStreamingWindow<F>;
 
+    #[tracing::instrument(skip_all, name = "OuterLinearStage::initialize")]
     fn initialize(
         _streaming: Option<Self::Streaming>,
         shared: &mut Self::Shared,
@@ -1386,10 +1403,12 @@ impl<F: JoltField> LinearSumcheckStage<F> for OuterLinearStage<F> {
         Self { az, bz }
     }
 
+    #[tracing::instrument(skip_all, name = "OuterLinearStage::next_window")]
     fn next_window(&mut self, shared: &mut Self::Shared, window_size: usize) {
         self.compute_evaluation_grid_from_polynomials_parallel(shared, window_size);
     }
 
+    #[tracing::instrument(skip_all, name = "OuterLinearStage::compute_message")]
     fn compute_message(
         &self,
         shared: &Self::Shared,
@@ -1402,6 +1421,7 @@ impl<F: JoltField> LinearSumcheckStage<F> for OuterLinearStage<F> {
             .gruen_poly_deg_3(t_prime_0, t_prime_inf, previous_claim)
     }
 
+    #[tracing::instrument(skip_all, name = "OuterLinearStage::ingest_challenge")]
     fn ingest_challenge(&mut self, shared: &mut Self::Shared, r_j: F::Challenge, _round: usize) {
         shared.split_eq_poly.bind(r_j);
 
@@ -1415,6 +1435,7 @@ impl<F: JoltField> LinearSumcheckStage<F> for OuterLinearStage<F> {
         );
     }
 
+    #[tracing::instrument(skip_all, name = "OuterLinearStage::cache_openings")]
     fn cache_openings<T: Transcript>(
         &self,
         shared: &Self::Shared,


jolt-core/src/poly/multiquadratic_poly.rs
```diff
@@ -0,0 +1,386 @@
...
+        let one = F::one();
+
+        for new_idx in 0..new_size {
+            let old_base_idx = new_idx * 3;
+            let eval_at_0 = self.evals[old_base_idx]; // z_0 = 0
+            let eval_at_1 = self.evals[old_base_idx + 1]; // z_0 = 1
+            let eval_at_inf = self.evals[old_base_idx + 2]; // z_0 = ∞
+
+            self.evals[new_idx] =
+                eval_at_0 * (one - r) + eval_at_1 * r + eval_at_inf * r * (r - one);
```
moodlezoup: nit: can pull `r * (r - one)` out of the for loop

tracer/src/lib.rs
```diff
@@ -86,7 +86,9 @@ pub fn trace(
         memory_config,
     ));
     let lazy_trace_iter_ = lazy_trace_iter.clone();
+    // NOTE: this will materialise the trace in full
     let trace: Vec<Cycle> = lazy_trace_iter.by_ref().collect();
+    //let trace = lazy_trace_iter.by_ref();
```
moodlezoup: nit: delete

quangvdao (changes requested):
Thanks! There are a number of changes I'd like to land before this is merge-able:

1. Can you clean up all the TODOs, fix the performance regression, and streamline the code?
2. When switching to linear time, ideally we should compute the evals of that round as we materialize, rather than doing another pass over the bound evals
3. More generally, I think the following flow might be preferable for computing prover message (I have talked about this a number of times I think):
```
fn compute_message(...) -> {
  // 1. Build the multiquadratic eval for this window, if it does not exist (or number of variables have fallen to 0)
  // 1.a) If we are streaming from trace, and we don't want to materialize for this window (i.e. not the last window), call a `stream_eval` kernel
  // 1.b) If we are streaming from trace, and we want to materialize for this window (i.e. last window of streaming), call a `stream_materialize_eval` kernel
  // 1.c) If we have materialized, call a `materialize_eval` kernel
  
  // 2. From the multiquadratic poly, derive the current round's uni poly.
}
```
This allows us to be flexible on the number of rounds to compute, even when we have the materialized values. it is rare that we will use it on CPU, but we will most certainly use it on GPU, and it would be good to have a reference CPU version. It will also collapse into the linear-time version, if you compute smartly (the multiquadratic evals should omit / not compute the eval at `(1, 1, ..., 1)`).

abiswas3:
> Thanks! There are a number of changes I'd like to land before this is merge-able:
> 
> 1. Can you clean up all the TODOs, fix the performance regression, and streamline the code?
> 2. When switching to linear time, ideally we should compute the evals of that round as we materialize, rather than doing another pass over the bound evals
> 3. More generally, I think the following flow might be preferable for computing prover message (I have talked about this a number of times I think):
> 
> ```
> fn compute_message(...) -> {
>   // 1. Build the multiquadratic eval for this window, if it does not exist (or number of variables have fallen to 0)
>   // 1.a) If we are streaming from trace, and we don't want to materialize for this window (i.e. not the last window), call a `stream_eval` kernel
>   // 1.b) If we are streaming from trace, and we want to materialize for this window (i.e. last window of streaming), call a `stream_materialize_eval` kernel
>   // 1.c) If we have materialized, call a `materialize_eval` kernel
>   
>   // 2. From the multiquadratic poly, derive the current round's uni poly.
> }
> ```
> 
> This allows us to be flexible on the number of rounds to compute, even when we have the materialized values. it is rare that we will use it on CPU, but we will most certainly use it on GPU, and it would be good to have a reference CPU version. It will also collapse into the linear-time version, if you compute smartly (the multiquadratic evals should omit / not compute the eval at `(1, 1, ..., 1)`).

For now we've adapted the API to suit the above. 
We do not skip the (1,....1) for now. 
As we're still trying to get the parellism right.
For small windows like w=1, 2, 3 -- it might be good to have a special manual solutions with loops unrolled, to maximum speed.

jolt-core/src/zkvm/r1cs/inputs.rs
```diff
@@ -264,6 +265,149 @@ pub struct R1CSCycleInputs {
...
+    {
+        // Calculate checkpoint and offset
+        let checkpoint_idx = t / checkpoint_interval;
+        let offset = t % checkpoint_interval;
+
+        // Clone and advance to target cycle
+        let mut iter = checkpoints[checkpoint_idx].clone();
+        for _ in 0..offset {
+            iter.next();
+        }
```
moodlezoup: We should only clone each checkpoint once for the `checkpoint_interval` cycles it's responsible for ($$O(T)$$ per round). What you have here is $$O(T^2)$$

jolt-core/src/poly/split_eq_poly.rs
```diff
@@ -125,6 +142,128 @@ impl<F: JoltField> GruenSplitEqPolynomial<F> {
...
+                    );
+                    &self.E_in_vec[head_in_bits]
+                };
+
+                (e_out, e_in)
+            }
+            BindingOrder::HighToLow => {
+                // Streaming windows are not defined for HighToLow in the current
+                // Spartan code paths; return neutral head tables.
+                (&self.one_table, &self.one_table)
```
moodlezoup: Seems like this should be `unimplemented!` for now?

jolt-core/src/poly/split_eq_poly.rs
```diff
@@ -125,6 +142,128 @@ impl<F: JoltField> GruenSplitEqPolynomial<F> {
...
+                let remaining_w = &self.w[..num_unbound];
+                let window_start = remaining_w.len() - window_size;
+                let (_w_body, w_window) = remaining_w.split_at(window_start);
+                let (w_active, _w_curr_slice) = w_window.split_at(window_size - 1);
+                // We only need the full eq table over the active window bits.
+                EqPolynomial::<F>::evals(w_active)
+            }
+            BindingOrder::HighToLow => {
+                // Not used for the outer Spartan streaming code.
+                vec![F::one()]
```
moodlezoup: same here, prefer `unimplemented!`

jolt-core/src/subprotocols/streaming_schedule.rs
```diff
@@ -0,0 +1,346 @@
...
+//TODO: Robustify half split
+pub trait StreamingSchedule: Send + Sync {
+    /// If this round is the switch-over round then we finally
+    /// materialise the sum-check polynomials in memory.
+    /// Prior, to this round, the streaming data structure
+    /// use to respond to verifier messages is constructed directly
+    /// from the trace.
+    fn is_switch_over_point(&self, round: usize) -> bool;
+    fn before_switch_over_point(&self, round: usize) -> bool;
+    fn after_switch_over_point(&self, round: usize) -> bool;
```
moodlezoup:
```suggestion
    fn switch_over_round(&self) -> usize;
```

moodlezoup: or perhaps `first_nonstreaming_round`

jolt-core/src/zkvm/spartan/outer.rs
```diff
@@ -43,6 +48,7 @@ use allocative::FlameGraphBuilder;
 
 /// Degree bound of the sumcheck round polynomials for [`OuterRemainingSumcheckVerifier`].
 const OUTER_REMAINING_DEGREE_BOUND: usize = 3;
+const INFINITY: usize = 2; // 2 represents ∞ in base-3
```
moodlezoup: This is kind of confusing to me, doesn't 2 represent 2 in base-3? 😅

jolt-core/src/zkvm/spartan/outer.rs
```diff
@@ -204,29 +210,60 @@ impl<F: JoltField, T: Transcript> UniSkipFirstRoundInstanceProver<F, T>
...
-    bz: DensePolynomial<F>,
+    az: Option<DensePolynomial<F>>,
+    bz: Option<DensePolynomial<F>>,
+    t_prime_poly: Option<MultiquadraticPolynomial<F>>, // multiquadratic polynomial used to answer queries in a streaming window
+    r_grid: ExpandingTable<F>, // hadamard product of (1 - r_j, r_j) for bound variables so far to help with streaming
     /// The first round evals (t0, t_inf) computed from a streaming pass over the trace
-    first_round_evals: (F, F),
     #[allocative(skip)]
     params: OuterRemainingSumcheckParams<F>,
+    lagrange_evals_r0: [F; 10],
```
moodlezoup: where does the number 10 come from? would be good to use a named const here

jolt-core/src/zkvm/spartan/outer.rs
```diff
@@ -248,72 +285,745 @@ impl<F: JoltField> OuterRemainingSumcheckProver<F> {
...
+        //    let az_j = acc_az[j].barrett_reduce();
+        //    let bz_first_j = acc_bz_first[j].barrett_reduce();
+        //    let bz_second_j = acc_bz_second[j].barrett_reduce();
+        //    grid_az[j] = az_j;
+        //    grid_bz[j] = bz_first_j + bz_second_j;
+        //}
+
+        let grid_az_ptr = grid_az.as_mut_ptr() as usize;
+        let grid_bz_ptr = grid_bz.as_mut_ptr() as usize;
+        let chunk_size = 4096;
```
moodlezoup: should this be min(4096, jlen)?

jolt-core/src/zkvm/spartan/outer.rs
```diff
@@ -248,72 +285,745 @@ impl<F: JoltField> OuterRemainingSumcheckProver<F> {
...
+
+            for j in start..end {
+                let az_j = acc_az[j].barrett_reduce();
+                let bz_first_j = acc_bz_first[j].barrett_reduce();
+                let bz_second_j = acc_bz_second[j].barrett_reduce();
+
+                unsafe {
+                    *az_ptr.add(j) = az_j;
+                    *bz_ptr.add(j) = bz_first_j + bz_second_j;
+                }
```
moodlezoup: instead of this unsafe pointer stuff, can't we just use `grid_az.par_chunks_mut(chunk_size).zip(grid_bz.par_chunk_mut(chunk_size)).enumerate()`  or something?

jolt-core/src/zkvm/spartan/outer.rs
```diff
@@ -248,72 +285,745 @@ impl<F: JoltField> OuterRemainingSumcheckProver<F> {
...
+        debug_assert_eq!(scaled_w.len(), klen);
+        debug_assert_eq!(grid_az.len(), jlen);
+        debug_assert_eq!(grid_bz.len(), jlen);
+
+        // Unreduced accumulators per j for Az and the two Bz groups.
+        let mut acc_az = vec![Acc5U::<F>::zero(); jlen];
+        let mut acc_bz_first = vec![Acc6S::<F>::zero(); jlen];
+        let mut acc_bz_second = vec![Acc7S::<F>::zero(); jlen];
+
+        if !parallel {
```
moodlezoup: Are we ever going to use the serial version in practice?

moodlezoup (changes requested):
Which of the new `OuterRemainingSumcheckProver` methods are going to be used in practice? I see that several method names are currently prefixed with an underscore, and some appear to be serial algorithms with a parallel counterpart. Let's delete anything that we're not going to use (including some of the stuff that's currently commented out).

jolt-core/src/zkvm/spartan/outer.rs
abiswas3: You're right. This is a remnant from previous naming -- the actual value does not matter. I'll change this, thanks

jolt-core/src/poly/multiquadratic_poly.rs
abiswas3: I'll confirm the ones that'll be used after benching them. They'll likely be some specialised ones for small window sizes for speed. 
I recommend keeping some of the serial implementations -- as the parallel code can often be hard to follow -- and theres's now a fair amount of nesting.

jolt-core/src/zkvm/spartan/outer.rs
abiswas3: It should be `OUTER_SKIP_DOMAIN_SIZE`

jolt-core/src/zkvm/r1cs/inputs.rs
abiswas3: I've removed this as the from checkpoint implementation will change completely anyway.

jolt-core/src/zkvm/spartan/outer.rs
abiswas3: Quang has handled this.

jolt-core/src/subprotocols/streaming_schedule.rs
abiswas3: The new api quang's designed does not make the distinction between streaming vs non-streaming. I'll do a small writeup on this. 
The short version is every sumcheck has a schedule.
As per the schedule: 

+ Each round in the sumcheck -- is either a window start or not. 
+ If a round is a window start -- then we materialise the streaming data structure, which for `SpartanOuter` is `MultiQuadratic` polynomial. But in general is some degree k multi-variable polynomial where $k \ll t$ where $t$ being the number of vars in the original sum-check polynomial. 
+ Then compute prover message always uses the streaming data structure above.

+ The switch over point which we mentally think of as first linear round -- actually materialises the polynomial. This switch over point is a special  window start where we also construct the sum-check data structure again, but this time from the materialised polynomial instead of a pass over the trace. 

+ henceforth we go back to the above mode, but every time we need to regenerate the sum-check data structure, we use the materialised polynomoial, not the trace. 

+ Therefore, the above API -- but you're right -- this can be simplified to returning a usize and doing the check in the sum-check compute prover message directly.

abiswas3: My recommendation is to the keep the above as is, as the before, after and at switch point are 3 different logical parts of the generalised sum-check now. 

The linear sum-check has switch over point at index 0; before switch-over is always false and after is true at all indices after 0. 

Every round is a window start -- and streaming data structure is a univariate polynomial so we only need to store evals 0 and infinity.

moodlezoup: mm I still prefer a single `switch_over_round(&self) -> usize` function

jolt-core/src/zkvm/spartan/outer.rs
```diff
@@ -248,60 +283,623 @@ impl<F: JoltField> OuterRemainingSumcheckProver<F> {
...
+        debug_assert_eq!(n, bz.len());
+
+        let three_pow_dim = 3_usize.pow(num_vars as u32);
+        let grid_size = 1 << num_vars;
+        let (E_out, E_in) = eq_poly.E_out_in_for_window(num_vars);
+        let ans: Vec<F> = if E_in.len() == 1 {
+            // Parallel version with reduction
+            (0..E_out.len())
+                .into_par_iter()
+                .with_min_len(4096)
```
moodlezoup: Performance regression is mostly caused by thread underutilization. Let's remove this `with_min_len`

jolt-core/src/zkvm/spartan/outer.rs
```diff
@@ -248,60 +283,623 @@ impl<F: JoltField> OuterRemainingSumcheckProver<F> {
...
+                            acc[idx] += local_ans[idx];
+                        }
+                        acc
+                    },
+                )
+        } else {
+            let num_xin_bits = E_in.len().log_2();
+            (0..E_out.len())
+                .into_par_iter()
+                .with_min_len(4096)
```
moodlezoup: same here, remove this `with_min_len`

jolt-core/src/zkvm/spartan/outer.rs
```diff
@@ -248,60 +283,623 @@ impl<F: JoltField> OuterRemainingSumcheckProver<F> {
...
+                            num_vars,
+                        );
+
+                        let e_product = E_out[x_out] * E_in[x_in];
+                        for idx in 0..three_pow_dim {
+                            local_ans[idx] += buff_a[idx] * buff_b[idx] * e_product;
+                        }
+                    }
+
+                    local_ans
```
moodlezoup: pull the multiplication by E_out out of the for loop over x_in
```suggestion
                        for idx in 0..three_pow_dim {
                            local_ans[idx] += buff_a[idx] * buff_b[idx] * E_in[x_in];
                        }
                    }

                    for idx in 0..three_pow_dim {
                        local_ans[idx] *= E_out[x_out];
                    }

                    local_ans
```

moodlezoup: I think similar optimizations elsewhere in this file

jolt-core/src/zkvm/spartan/outer.rs
```diff
@@ -248,60 +283,623 @@ impl<F: JoltField> OuterRemainingSumcheckProver<F> {
...
+        let mut buff_b = vec![F::zero(); three_pow_dim];
+        let mut tmp = vec![F::zero(); three_pow_dim];
+        let mut ans = vec![F::zero(); three_pow_dim];
+
+        let (E_out, E_in) = eq_poly.E_out_in_for_window(num_vars);
+        //println!("Size of Az/Bz: {}", n.log_2());
+        //println!("Size of Grid: {grid_size}, Num-vars: {num_vars}");
+        //println!("Num out vars: {}", E_out.len().log_2());
+        //println!("Num in vars: {}", E_in.len().log_2());
+        //
```
moodlezoup: delete

jolt-core/src/zkvm/spartan/outer.rs
```diff
@@ -358,49 +958,457 @@ impl<F: JoltField> OuterRemainingSumcheckProver<F> {
...
+                                bz_grid[x_val] += bz_val * r_eval;
+                            }
+                        }
+
+                        // Store the accumulated grid in chunk-relative position
+                        let buffer_offset = grid_size * (pair_idx - start_pair);
+                        //for j in 0..grid_size {
+                        //    az_chunk[buffer_offset + j] = az_grid[j];
+                        //    bz_chunk[buffer_offset + j] = bz_grid[j];
+                        //}
```
moodlezoup: delete

PR #1117, opened 2025-11-18
Add back e2e tracing span for prove & minor opts for instruction read raf

quangvdao:
Pretty self-explanatory

### DIFF
--- a/jolt-core/src/zkvm/ram/read_write_checking.rs
+++ b/jolt-core/src/zkvm/ram/read_write_checking.rs
@@ -36,10 +36,10 @@ use tracer::instruction::{Cycle, RAMAccess};
 // RAM read-write checking sumcheck
 //
 // Proves the relation:
-//   Σ_{k,j} eq(r', (j, k)) ⋅ ra(k, j) ⋅ (Val(k, j) + γ ⋅ (inc(j) + Val(k, j)))
+//   Σ_{k,j} eq(r_cycle, j) ⋅ ra(k, j) ⋅ (Val(k, j) + γ ⋅ (inc(j) + Val(k, j)))
 //   = rv_claim + γ ⋅ wv_claim
 // where:
-// - r' are the fresh challenges for this sumcheck
+// - r_cycle are the challenges for the cycle variables in this sumcheck (from Spartan outer)
 // - ra(k, j) = 1 if memory address k is accessed at cycle j, and 0 otherwise
 // - Val(k, j) is the value at memory address k right before cycle j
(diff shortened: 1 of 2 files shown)


jolt-core/src/zkvm/instruction_lookups/read_raf_checking.rs
```diff
@@ -554,13 +554,24 @@ impl<F: JoltField> ReadRafSumcheckProver<F> {
...
-                            self.v[phase][k_bound]
-                        })
-                        .product::<F>()
+                    // Initialize accumulator with the first phase value to avoid an extra multiply by F::one().
+                    // PHASES is assumed to be > 0 here.
+                    let (first_prefix, _) = k.split((PHASES - 1) * LOG_M);
+                    // Masking is equivalent to modulo here since M is a power of two (M = 1 << LOG_M).
+                    // Apply the mask in u128 before casting to avoid usize overflow for wide prefixes.
+                    let first_k_bound: usize =
+                        (u128::from(&first_prefix) & ((M - 1) as u128)) as usize;
```
moodlezoup, line 563: why not just `first_prefix % M` here?

jolt-core/src/zkvm/instruction_lookups/read_raf_checking.rs
```diff
@@ -554,13 +554,24 @@ impl<F: JoltField> ReadRafSumcheckProver<F> {
...
+                    let mut acc = self.v[0][first_k_bound];
+
+                    for phase in 1..PHASES {
+                        let (prefix, _) = k.split((PHASES - 1 - phase) * LOG_M);
+                        // Masking is equivalent to modulo here since M is a power of two (M = 1 << LOG_M).
+                        // Apply the mask in u128 before casting to avoid usize overflow for wide prefixes.
+                        let k_bound: usize = (u128::from(&prefix) & ((M - 1) as u128)) as usize;
```
moodlezoup, line 570: same

jolt-core/src/zkvm/instruction_lookups/read_raf_checking.rs
quangvdao: I guess modulus can be more expensive than bit-wise and

moodlezoup: i would hope that the rust compiler would be smart enough to make them equivalent when the modulus is a const power of 2 but idk

0xAndoroid: cc @andrewmilson I saw you working on similar thing today

PR #1121, opened 2025-11-20
Specialized mles product sum kernels for more d

quangvdao:
Tests pass. However haven't benched them

### DIFF
--- a/jolt-core/src/subprotocols/mles_product_sum.rs
+++ b/jolt-core/src/subprotocols/mles_product_sum.rs
@@ -450,6 +450,86 @@ impl_eval_linear_prod_internal!(eval_linear_prod_7_internal, 7);
 impl_eval_linear_prod_internal!(eval_linear_prod_9_internal, 9);
 impl_eval_linear_prod_internal!(eval_linear_prod_10_internal, 10);
 
+/// Evaluate the product of 6 linear polynomials on `U_6 = [1, 2, 3, 4, 5, ∞]`.
+///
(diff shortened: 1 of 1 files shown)


jolt-core/src/subprotocols/mles_product_sum.rs
```diff
@@ -14,16 +14,37 @@ pub fn compute_mles_product_sum<F: JoltField>(
...
+    // stack-allocated `d = 16` implementation) while keeping the final
+    // interpolation logic centralized.
+    let sum_evals: Vec<F> = match d {
+        // Fully stack-allocated paths based on optimized interpolation kernels.
+        13 => compute_mles_product_sum_evals_d13(mles, eq_poly),
+        16 => compute_mles_product_sum_evals_d16(mles, eq_poly),
+        19 => compute_mles_product_sum_evals_d19(mles, eq_poly),
+        22 => compute_mles_product_sum_evals_d22(mles, eq_poly),
+        26 => compute_mles_product_sum_evals_d26(mles, eq_poly),
+        32 => compute_mles_product_sum_evals_d32(mles, eq_poly),
```
0xAndoroid, line 31: Is `15` not needed here?

0xAndoroid (approved):
Thanks

jolt-core/src/subprotocols/mles_product_sum.rs
quangvdao: hmm you're right, I missed that

PR #1127, opened 2025-11-21
Continuing feat/col-major-sparse-matrix

quangvdao:
## NOTE

Does NOT support arbitrary phase1 and phase2 num rounds. However, we do support the cases we care about:

- Bind all cycle variables first (or even all but _one_ cycle variables)
- Bind zero cycle variable, start with some address variables, then materialize

Added TODO for the future to make things work for all parameters

---

### Overview

This PR introduces several major changes to the RAM subsystem:

1. **New RA Reduction Sumcheck** - Consolidates four RAM RA claims into a single reduced claim
2. **Sparse Matrix Restructuring** - New `AddressMajor` and `CycleMajor` matrix representations
3. **Phase-based Read/Write Checking** - Three-phase structure for flexible variable binding
4. **Advice Opening Optimization** - ~~Opens advice at TWO points~~ **Conditionally opens at ONE or TWO points based on phase configuration**

---

#### 1. `opening_proof.rs` (+81 lines)

- **`OpeningId` enum refactored**: Advice variants now take `SumcheckId` parameter
- **New `SumcheckId::RamRaReduction`** variant added
- **Getter/setter methods updated** to take `SumcheckId` parameter

#### 2. `proof_serialization.rs` (+106 lines)

- **Four advice proof fields** (but second pair may be `None` when single opening is used):
  ```rust
  pub trusted_advice_val_evaluation_proof: Option<PCS::Proof>,
  pub trusted_advice_val_final_proof: Option<PCS::Proof>,      // None when single_opening
  pub untrusted_advice_val_evaluation_proof: Option<PCS::Proof>,
  pub untrusted_advice_val_final_proof: Option<PCS::Proof>,    // None when single_opening
  ```

#### 3. `ram/mod.rs` (+224 lines)

- **`prover_accumulate_advice`**: Conditionally opens at ONE or TWO points:
  ```rust
  pub fn prover_accumulate_advice<F: JoltField>(
      ...,
      single_opening: bool,  // NEW: when true, skip second opening
  )
  ```
- **`verifier_accumulate_advice`**: Same conditional pattern

#### 4. `prover.rs` (+127 lines)

- **`prove_trusted_advice` / `prove_untrusted_advice`**: 
  - Returns `(Some(proof), None)` when `needs_single_advice_opening(T)` is true
  - Returns `(Some(proof), Some(proof))` otherwise

#### 5. `verifier.rs` (+98 lines)

- **`verify_trusted_advice_opening_proofs` / `verify_untrusted_advice_opening_proofs`**: 
  - Skips second proof verification when `needs_single_advice_opening(T)` is true

#### 6. `ram/read_write_checking.rs` (+497 lines)

- **NEW: `needs_single_advice_opening(T: usize) -> bool`** function:
  ```rust
  /// Returns true if all cycle variables are bound in phase 1.
  /// When true, advice opening points are identical → only need one opening.
  pub fn needs_single_advice_opening(T: usize) -> bool {
      phase1_num_rounds(0, T) == T.log_2()
  }
  ```
- **Documented supported configurations**:
  - `(T.log_2(), any)` - All cycle vars bound in phase 1 → **single advice opening**
  - `(0, any)` - Skip phase 1 → two advice openings needed

#### 7. `ram/output_check.rs` (+22 lines)

- **Updated to use correct `SumcheckId`** based on `needs_single_advice_opening`:
  ```rust
  let advice_sumcheck_id = if needs_single_advice_opening(trace_len) {
      SumcheckId::RamValEvaluation      // Reuse the single opening
  } else {
      SumcheckId::RamValFinalEvaluation // Use dedicated opening
  };
  ```

#### 8. `ram/ra_reduction.rs` (NEW +868 lines)
- **`RamRaReductionSumcheckProver`**: Consolidates four RA claims into one
- **`RamRaReductionSumcheckVerifier`**: Verifies the reduced claim

#### 9. `ram/val_evaluation.rs` (+16 lines)
- Updated docstring: LT is `LT(j, r_cycle)` (bitstring j < r_cycle)

---

### ~~Why Two Advice Opening Points?~~ → **Advice Opening Optimization**

**Previous behavior**: Always open advice at TWO points.

**New behavior**: Conditionally open at ONE or TWO points:

| Phase Config | `needs_single_advice_opening` | Advice Openings |
|--------------|------------------------------|-----------------|
| `phase1 = T.log_2()` (all cycle vars) | `true` | ONE (points identical) |
| `phase1 = 0` (all address vars first) | `false` | TWO (points differ) |

**Savings when `single_opening = true`**:
- ✅ Skip 2 PCS proof generations
- ✅ Skip 2 polynomial evaluations  
- ✅ Skip 2 PCS verifications
- ✅ Smaller proof size

---

### Required Invariants (from `ram/mod.rs` documentation)

For the RA reduction to work correctly:

1. **OutputCheck and RafEvaluation** MUST be in the same batched sumcheck (Stage 2)
2. **ValEvaluation and ValFinal** MUST be in the same batched sumcheck (Stage 4)
3. **ValEvaluation** MUST read `r_address` from `RamReadWriteChecking`'s opening
4. **ValFinal** MUST read `r_address` from `OutputCheck`'s opening (or `RamReadWriteChecking` if single_opening)

### DIFF
--- a/jolt-core/src/subprotocols/read_write_matrix/address_major.rs
+++ b/jolt-core/src/subprotocols/read_write_matrix/address_major.rs
@@ -4,603 +4,287 @@
 
 use std::cmp::Ordering;
 use std::mem::MaybeUninit;
+use std::sync::Arc;
+use std::sync::Mutex;
 
 use allocative::Allocative;
 use ark_std::Zero;
-use common::jolt_device::MemoryLayout;
+use num::Integer;
 use rayon::prelude::*;
-use tracer::instruction::{Cycle, RAMAccess};
 
 use crate::field::JoltField;
 use crate::field::OptimizedMul;
 use crate::poly::multilinear_polynomial::{BindingOrder, MultilinearPolynomial, PolynomialBinding};
 use crate::poly::unipoly::UniPoly;
-use crate::utils::thread::unsafe_allocate_zero_vec;
-use crate::zkvm::ram::remap_address;
 
 use super::cycle_major::ReadWriteMatrixCycleMajor;
-use super::ColIndex;
 
-/// Represents the ra(k, j) and Val(k, j) polynomials for the RAM
-/// read/write-checking sumcheck in address-major (column-major) order.
-///
-/// # Memory Layout: Struct-of-Arrays (SoA)
-///
-/// Unlike `ReadWriteMatrixCycleMajor` which uses Array-of-Structs (`Vec<ReadWriteEntry>`),
-/// this uses SoA for better cache performance during address binding:
-///
-/// ```text
-/// Sparse entry arrays (variable size, doesn't halve predictably):
-///   rows[i] = cycle index for entry i
-///   cols[i] = address index for entry i
-///   vals[i] = Val(k, j) coefficient for entry i
-///   ras[i]  = ra(k, j) coefficient for entry i
-///
-/// Dense auxiliary arrays (size K, halves each round):
-///   val_init[k]  = initial value at address k (before any access)
-///   val_final[k] = final value at address k (after last access)
-/// ```
-///
-/// # Deriving `next_val` Without Storing It
-///
-/// Key memory optimization: Instead of storing `prev_val`/`next_val` per entry
-/// (which would add 2 field elements = 64 bytes per entry), we derive `next_val`
-/// on-the-fly using `get_next_val(i)`:
-///
-/// ```text
-/// If entry i+1 is in the same column:
-///   next_val(i) = vals[i+1]   // Next access in same column
-/// Else:
-///   next_val(i) = val_final[cols[i]]  // Value persists until end
-/// ```
-///
-/// This works because entries are sorted by `(col, row)`, so within each column,
-/// entries appear in chronological order.
-///
-/// # Invariants
-///
-/// - Entries are sorted by `(col, row)` (address-major order)
-/// - `rows`, `cols`, `vals`, `ras` all have the same length (`nnz()`)
-/// - `val_init` and `val_final` have the same length (`K` = address space size)
-/// - `val_final[k]` equals `next_val` of the last entry in column `k`
+/// Represents a non-zero entry in the ra(k, j) and Val(k, j) polynomials.
+/// Conceptually, both ra and Val can be seen as K x T matrices.
 ///
 /// # Type Parameters
 ///
 /// - `F`: The field type for coefficients.
-/// - `I`: The column index type (e.g., `usize` for RAM, `u8` for registers).
+#[derive(Allocative, Debug, PartialEq, Clone, Copy)]
+pub struct ReadWriteEntry<F: JoltField> {
+    /// The row index. Before binding, row \in [0, T)
+    pub row: usize,
+    /// The column index. Before binding, col \in [0, K)
+    pub col: usize,
+    /// In round i, each ReadWriteEntry represents a coefficient
+    ///   Val(k, j', r)
+    /// which is some combination of Val(k, j', 00...0), ...
+    /// Val(k, j', 11...1).
+    /// `prev_val` contains the unbound coefficient before
+    /// Val(k, j', 00...0) –– abusing notation, `prev_val` is
+    /// Val(k, j'-1, 11...1)
+    pub(crate) prev_val: F,
+    /// In round i, each ReadWriteEntry represents a coefficient
+    ///   Val(k, j', r)
+    /// which is some combination of Val(k, j', 00...0), ...
+    /// Val(k, j', 11...1).
+    /// `next_val` contains the unbound coefficient after
+    /// Val(k, j', 00...0) –– abusing notation, `next_val` is
+    /// Val(k, j'+1, 00...0)
+    pub(crate) next_val: F,
+    /// The Val coefficient for this matrix entry.
+    pub val_coeff: F,
+    /// The ra coefficient for this matrix entry. Note that for RAM,
+    /// ra and wa are the same polynomial.
+    pub ra_coeff: F,
+}
+
+/// Represents the ra(k, j) and Val(k, j) polynomials for the RAM
+/// read/write-checking sumcheck in address-major (column-major) order.
+///
+/// # Type Parameters
+///
+/// - `F`: The field type for coefficients.
 #[derive(Allocative, Debug, Default, Clone)]
-pub struct ReadWriteMatrixAddressMajor<F: JoltField, I: ColIndex = usize> {
-    /// Row indices (cycle indices) for each sparse entry.
-    pub rows: Vec<usize>,
-    /// Column indices (address indices) for each sparse entry.
-    pub cols: Vec<I>,
-    /// Val(k, j) coefficients: the value read/written at each access.
-    pub vals: Vec<F>,
-    /// ra(k, j) coefficients: always 1 for explicit accesses, used for sumcheck.
-    pub ras: Vec<F>,
-    /// Initial Val polynomial over addresses: `val_init[k] = Val(k, j=0)`.
-    /// This is the memory state before the first cycle.
-    val_init: MultilinearPolynomial<F>,
-    /// Final Val polynomial over addresses: `val_final[k]` = value at address `k`
-    /// after the last access to it. Used to derive `next_val` for the last entry
-    /// in each column. For columns with no accesses, `val_final[k] = val_init[k]`.
-    val_final: MultilinearPolynomial<F>,
+pub struct ReadWriteMatrixAddressMajor<F: JoltField> {
+    pub entries: Vec<ReadWriteEntry<F>>,
+    pub(crate) val_init: MultilinearPolynomial<F>,
 }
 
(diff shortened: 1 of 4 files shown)


jolt-core/src/zkvm/verifier.rs
```diff
@@ -412,11 +420,18 @@ impl<'a, F: JoltField, PCS: CommitmentScheme<Field = F>, ProofTranscript: Transc
     }
 
     fn verify_trusted_advice_opening_proofs(&mut self) -> Result<(), anyhow::Error> {
+        use crate::poly::opening_proof::SumcheckId;
```
moodlezoup: nit: hoist this import

jolt-core/src/zkvm/ram/ra_virtual.rs
```diff
@@ -271,99 +266,61 @@ impl<F: JoltField, T: Transcript> SumcheckInstanceVerifier<F, T> for RaSumcheckV
...
+    /// r_cycle_reduced as OpeningPoint (for verifier's mle_endian)
+    r_cycle_opening: OpeningPoint<BIG_ENDIAN, F>,
+    /// r_address_reduced split into chunks according to one-hot decomposition
     r_address_chunks: Vec<Vec<F::Challenge>>,
-    /// Number of decomposition parts
+    /// Number of decomposition chunks
     d: usize,
-    /// Length of the trace
-    T: usize,
+    /// Length of the trace (T)
```
moodlezoup: nit: inaccurate comment

jolt-core/src/zkvm/ram/ra_virtual.rs
```diff
@@ -271,99 +266,61 @@ impl<F: JoltField, T: Transcript> SumcheckInstanceVerifier<F, T> for RaSumcheckV
...
-struct RaSumcheckParams<F: JoltField> {
-    gamma_powers: [F; 3],
-    /// Random challenge r_cycle
-    r_cycle: [Vec<F::Challenge>; 3],
+/// Shared parameters between prover and verifier.
+struct RaVirtualParams<F: JoltField> {
+    /// r_cycle_reduced from RA reduction sumcheck (for prover)
+    r_cycle: Vec<F::Challenge>,
+    /// r_cycle_reduced as OpeningPoint (for verifier's mle_endian)
+    r_cycle_opening: OpeningPoint<BIG_ENDIAN, F>,
```
moodlezoup: this should just be one field

jolt-core/src/zkvm/ram/ra_reduction.rs
```diff
@@ -0,0 +1,868 @@
...
+    subprotocols::{
+        sumcheck_prover::SumcheckInstanceProver, sumcheck_verifier::SumcheckInstanceVerifier,
+    },
+    transcripts::Transcript,
+    utils::{expanding_table::ExpandingTable, math::Math, thread::unsafe_allocate_zero_vec},
+    zkvm::{config::OneHotParams, ram::remap_address, witness::VirtualPolynomial},
+};
+
+/// Degree bound of the sumcheck round polynomials.
+/// Degree 2: one from eq polynomial, one from ra (which is 0 or 1).
```
moodlezoup, line 60:
```suggestion
/// Degree 2: one from eq polynomial, one from ra
```

jolt-core/src/zkvm/ram/ra_reduction.rs
```diff
@@ -0,0 +1,868 @@
...
+        r_cycle_val: &[F::Challenge],
+        gamma: F,
+    ) -> (Vec<F>, Vec<F>) {
+        // Materialize full eq tables for each cycle point
+        let eq_raf = EqPolynomial::<F>::evals(r_cycle_raf);
+        let eq_rw = EqPolynomial::<F>::evals(r_cycle_rw);
+        let eq_val = EqPolynomial::<F>::evals(r_cycle_val);
+
+        // Compute G arrays in parallel using fold-reduce pattern
+        let chunk_size = 1 << 14; // Process 16K cycles at a time
```
moodlezoup: let's calculate the chunk_size dynamically based on current # threads

jolt-core/src/zkvm/ram/ra_reduction.rs
```diff
@@ -0,0 +1,868 @@
...
+                    (acc_A, acc_B)
+                },
+            );
+
+        (G_A, G_B)
+    }
+
+    /// Minimum number of k iterations to parallelize the inner loop.
+    /// Below this threshold, sequential iteration is faster due to parallel overhead.
+    const MIN_INNER_PARALLEL_LEN: usize = 1 << 12; // 4096
```
moodlezoup: does this give a a concrete speedup? If not, I'd prefer to get rid of the serial code to reduce LoC

jolt-core/src/subprotocols/read_write_matrix/cycle_major.rs
```diff
@@ -489,78 +561,92 @@ impl<F: JoltField> SparseMatrixPolynomial<F> {
...
+        // Extract the coefficient slice from self.val_init (which is a LargeScalars polynomial)
+        let val_init_coeffs = match &self.val_init {
+            MultilinearPolynomial::LargeScalars(poly) => &poly.Z,
+            _ => panic!("val_init must be LargeScalars"),
+        };
+        for (k, &v) in val_init_coeffs.iter().enumerate() {
+            for t in 0..T_prime {
+                val[k * T_prime + t] = v;
+            }
+        }
```
moodlezoup: I worry that this will be slow for moderately large parameter sizes. Can we do `val.par_chunks_mut(T_prime).zip(val_init_coeffs.par_iter())`?

jolt-core/src/subprotocols/read_write_matrix/address_major.rs
```diff
@@ -0,0 +1,1178 @@
...
+///
+/// ```text
+/// If entry i+1 is in the same column:
+///   next_val(i) = vals[i+1]   // Next access in same column
+/// Else:
+///   next_val(i) = val_final[cols[i]]  // Value persists until end
+/// ```
+///
+/// This works because entries are sorted by `(col, row)`, so within each column,
+/// entries appear in chronological order.
```
moodlezoup: I'm not convinced that this is correct, and this could be the reason verification is failing for some parameter settings. 
Let's say we have Val(k, j', r_cycle) and are looking for its neighbor Val(k, j' + 1, r_cycle).
Let's further assume Val(k, j' + 1, r_cycle) is not explicitly represented in the `ReadWriteMatrixAddressMajor`, but that Val(k, j' + 2, r_cycle) is.
The absence of Val(k, j' + 1, r_cycle) does _not_ imply that Val(k, j' + 1, r_cycle) = Val(k, j' + 2, r_cycle)

This was why I had explicitly represented `next_val` and `prev_val`

jolt-core/src/subprotocols/read_write_matrix/address_major.rs
```diff
@@ -0,0 +1,1178 @@
...
+        let rows: Vec<usize> = entries.par_iter().map(|(row, _, _, _)| *row).collect();
+        let cols: Vec<usize> = entries.par_iter().map(|(_, col, _, _)| *col).collect();
+        let ras: Vec<F> = vec![F::one(); n]; // ra = 1 for all explicit accesses
+
+        // For vals, we need the val_coeff (pre_value for writes, read_value for reads)
+        let vals: Vec<F> = entries.par_iter().map(|(_, _, val, _)| *val).collect();
+
+        // Step 4: Compute val_final
+        // Initialize from val_init, then update with last entry's next_val per column
+        let mut val_final_vec: Vec<F> = (0..k_size).into_par_iter().map(|k| val_init[k]).collect();
```
moodlezoup:
```suggestion
        let mut val_final_vec: Vec<F> = val_init.clone();
```

jolt-core/src/subprotocols/read_write_matrix/address_major.rs
```diff
@@ -0,0 +1,1178 @@
...
+    ///
+    /// If `dry_run` is true, ignores output buffers and just computes the output length.
+    /// This is used to allocate exact memory before the real bind operation.
+    #[allow(clippy::too_many_arguments)]
+    fn bind_cols(
+        &self,
+        e0: usize,
+        e1: usize, // Even column entries: indices [e0, e1)
+        o0: usize,
+        o1: usize, // Odd column entries: indices [o0, o1)
```
moodlezoup:
```suggestion
        even_indices: Range<usize>,
        odd_indices: Range<usize>,
```

jolt-core/src/subprotocols/read_write_matrix/address_major.rs
```diff
@@ -0,0 +1,1178 @@
...
+        for col in 0..K_prime {
+            if !col_seen[col] {
+                let init_val = self.val_init.get_bound_coeff(col);
+                for row in 0..T_prime {
+                    let idx_flat = col * T_prime + row;
+                    // ra is already zero
+                    val[idx_flat] = init_val;
+                }
+            }
+        }
```
moodlezoup: instead of doing this `col_seen` thing I think it's cleaner to just initialize these indices of `val` at the beginning

PR #1132, opened 2025-11-26
Front-end API for Sumchecks

### LINKED ISSUE #1131: Formal Verification: Front-end API for Sumchecks
For the formal verification effort, it would be good to have a high-level API for constructing sumcheck instances that also allows them to be extracted via `zklean-extractor`. @jprider63 and I have come up with a candidate API that looks like this:
```rust
/// An arithmetic expression over opening values to use as either the LHS or RHS of an individual claim
pub enum ClaimExpr<F> {
    /// A fixed value, e.g., 0 or 1
    Val(F),
    /// Arithmetic operations used in claims
    Add(Box<ClaimExpr<F>>, Box<ClaimExpr<F>>),
    Mul(Box<ClaimExpr<F>>, Box<ClaimExpr<F>>),
    Sub(Box<ClaimExpr<F>>, Box<ClaimExpr<F>>),
    /// Corresponds to a [`CommittedOpening`]
    CommittedVar(CommittedPolynomial),
   /// Corresponds to a [`VirtualOpening`]
    VirtualVar(VirtualPolynomial),
}

/// Interpreted as `input_claim = eq_eval * output_claim`.
pub struct Claim<F: JoltField> {
    /// Sumcheck to take the input opening values from
    pub input_sumcheck_id: SumcheckId,
    pub input_claim_expr: ClaimExpr<F>,
    pub expected_output_claim_expr: ClaimExpr<F>,
    /// Determines whether to use [`EqPolynomial`] or [`EqPlusOnePolynomial`]
    pub is_offset: bool,
}

/// Combines the component claims and performs the linear combination with powers of gamma, representing the combined claim
///      \sum_i gamma^i * input_claim_i = \sum_i gamma^i * eq_eval_i * expected_output_claim_i
pub struct InputOutputClaims<F: JoltField> {
    claims: Vec<Claim<F>>,
    output_sumcheck_id: SumcheckId,
    gamma_pows: Vec<F>,
}

impl<F: JoltField> InputOutputClaims<F> {
    pub fn new(claims: Vec<Claim<F>>, output_sumcheck_id: SumcheckId, gamma: F) -> Self { ... }
    /// Computes \sum_i gamma^i * input_claim_i
    pub fn input_claim(&self, acc: &impl OpeningAccumulator<F>) -> F { ... }
    /// Computes \sum_i gamma^i * eq_eval_i * expected_output_claim_i
    pub fn expected_output_claim(&self, r: OpeningPoint<BIG_ENDIAN, F>, acc: &impl OpeningAccumulator<F>) -> F { ... }
}

pub trait FrontendSumcheck<F: JoltField> {
    fn input_output_claims(gamma: F) -> InputOutputClaims<F>;
}
```

This doesn't seem applicable to all of the sumchecks, but it seems to apply to the ones that we're most interested in extracting for the frontend verification, e.g., `RamReadWriteChecking`, `RegistersReadWriteChecking`, `InstructionVirtualization`, `SpartanShift`, etc. The idea is that the sumcheck verifier instance could implement `FrontendSumcheck` if this interface is applicable. It could then call `InputOutputClaims::input_claim` and `InputOutputClaims::expected_output_claim` from its `SumcheckInstanceVerifier::input_claim` and `SumcheckInstanceVerifier::expected_output_claim` methods. Having the sumchecks in this form would allow us to extract them for use in the frontend verification.

For example, the impl for `RamReadWriteSumcheck` might look like this:
```rust
impl<F: JoltField> FrontendSumcheck<F> for RamReadWriteCheckingVerifier<F> {
    fn input_output_claims(gamma: F) -> InputOutputClaims<F> {
        let ram_read_value: ClaimExpr<F> = VirtualPolynomial::RamReadValue.into();
        let ram_write_value: ClaimExpr<F> = VirtualPolynomial::RamWriteValue.into();
        let ram_ra: ClaimExpr<F> = VirtualPolynomial::RamRa.into();
        let ram_val: ClaimExpr<F> = VirtualPolynomial::RamVal.into();
        let ram_inc: ClaimExpr<F> = CommittedPolynomial::RamInc.into();

        InputOutputClaims::new(
            vec![
                Claim {
                    input_sumcheck_id: SumcheckId::SpartanOuter,
                    input_claim_expr: ram_read_value,
                    expected_output_claim_expr: ram_ra.clone() * ram_val.clone(),
                    is_offset: false,
                },
                Claim {
                    input_sumcheck_id: SumcheckId::SpartanOuter,
                    input_claim_expr: ram_write_value,
                    expected_output_claim_expr: ram_ra * (ram_val + ram_inc),
                    is_offset: false,
                },
            ],
            SumcheckId::RamReadWriteChecking,
            gamma,
        )
    }
}
```

I've experimented with putting several of the sumchecks in this form, and I can make a draft PR with them shortly.

There might be a small amount of overhead involved with interpreting the `ClaimExpr` values when the verifier wants to evaluate, but if that becomes an issue, we believe we could get rid of most or all of that overhead by making the above into a macro.

protoben:
Closes #1131

### DIFF
--- a/jolt-core/src/zkvm/ram/hamming_booleanity.rs
+++ b/jolt-core/src/zkvm/ram/hamming_booleanity.rs
@@ -1,4 +1,5 @@
 use crate::field::JoltField;
+use crate::poly::eq_poly::EqPolynomial;
 use crate::poly::multilinear_polynomial::{BindingOrder, MultilinearPolynomial, PolynomialBinding};
 use crate::poly::opening_proof::{
     OpeningAccumulator, OpeningPoint, PolynomialId, ProverOpeningAccumulator, SumcheckId,
@@ -162,11 +163,12 @@ impl<F: JoltField, T: Transcript> SumcheckInstanceVerifier<F, T>
     for HammingBooleanitySumcheckVerifier<F>
 {
     fn input_claim(&self, accumulator: &VerifierOpeningAccumulator<F>) -> F {
-        let result = Self::input_output_claims().input_claim(&[F::one()], accumulator);
+        let result = self.params.input_claim(accumulator);
 
         #[cfg(test)]
         {
-            let reference_result = self.params.input_claim(accumulator);
+            let reference_result =
+                Self::input_output_claims().input_claim(&[F::one()], accumulator);
             assert_eq!(result, reference_result);
         }
 
@@ -182,37 +184,35 @@ impl<F: JoltField, T: Transcript> SumcheckInstanceVerifier<F, T>
         accumulator: &VerifierOpeningAccumulator<F>,
         sumcheck_challenges: &[F::Challenge],
     ) -> F {
-        let r = self.params.normalize_opening_point(sumcheck_challenges);
-        let result =
-            Self::input_output_claims().expected_output_claim(&r, &[F::one()], accumulator);
+        let H_claim = accumulator
+            .get_virtual_polynomial_opening(
+                VirtualPolynomial::RamHammingWeight,
+                SumcheckId::RamHammingBooleanity,
+            )
+            .1;
+
+        let (r_cycle, _) = accumulator.get_virtual_polynomial_opening(
+            VirtualPolynomial::LookupOutput,
+            SumcheckId::SpartanOuter,
+        );
+
+        let eq = EqPolynomial::<F>::mle(
+            sumcheck_challenges,
+            &r_cycle
+                .r
+                .iter()
+                .cloned()
+                .rev()
+                .collect::<Vec<F::Challenge>>(),
+        );
+
+        let result = (H_claim.square() - H_claim) * eq;
 
         #[cfg(test)]
         {
-            use crate::poly::eq_poly::EqPolynomial;
-
-            let H_claim = accumulator
-                .get_virtual_polynomial_opening(
-                    VirtualPolynomial::RamHammingWeight,
-                    SumcheckId::RamHammingBooleanity,
-                )
-                .1;
-
-            let (r_cycle, _) = accumulator.get_virtual_polynomial_opening(
-                VirtualPolynomial::LookupOutput,
-                SumcheckId::SpartanOuter,
-            );
-
-            let eq = EqPolynomial::<F>::mle(
-                sumcheck_challenges,
-                &r_cycle
-                    .r
-                    .iter()
-                    .cloned()
-                    .rev()
-                    .collect::<Vec<F::Challenge>>(),
-            );
-
-            let reference_result = (H_claim.square() - H_claim) * eq;
+            let r = self.params.normalize_opening_point(sumcheck_challenges);
+            let reference_result =
+                Self::input_output_claims().expected_output_claim(&r, &[F::one()], accumulator);
             assert_eq!(result, reference_result);
         }
 
--- a/jolt-core/src/zkvm/registers/read_write_checking.rs
+++ b/jolt-core/src/zkvm/registers/read_write_checking.rs
@@ -807,39 +807,18 @@ impl<F: JoltField, T: Transcript> SumcheckInstanceVerifier<F, T>
     for RegistersReadWriteCheckingVerifier<F>
 {
     fn input_claim(&self, accumulator: &VerifierOpeningAccumulator<F>) -> F {
-        let claims = Self::input_output_claims();
-        let gamma_pows: Vec<F> =
-            std::iter::successors(Some(F::one()), |prev| Some(*prev * self.params.gamma))
-                .take(claims.claims.len())
-                .collect();
-        let result = claims.input_claim(&gamma_pows, accumulator);
+        let result = self.params.input_claim(accumulator);
 
         #[cfg(test)]
         {
-            let reference_result = self.params.input_claim(accumulator);
+            let claims = Self::input_output_claims();
+            let gamma_pows: Vec<F> =
+                std::iter::successors(Some(F::one()), |prev| Some(*prev * self.params.gamma))
+                    .take(claims.claims.len())
+                    .collect();
+            let reference_result = claims.input_claim(&gamma_pows, accumulator);
             assert_eq!(result, reference_result);
         }
-
-        let (_, rs1_rv_claim) = accumulator.get_virtual_polynomial_opening(
-            VirtualPolynomial::Rs1Value,
-            SumcheckId::RegistersClaimReduction,
-        );
-        let (_, rs1_rv_claim_instruction_input) = accumulator.get_virtual_polynomial_opening(
-            VirtualPolynomial::Rs1Value,
-            SumcheckId::InstructionInputVirtualization,
-        );
-        // TODO: Make error and move to more appropriate place.
-        assert_eq!(rs1_rv_claim, rs1_rv_claim_instruction_input);
-        let (_, rs2_rv_claim) = accumulator.get_virtual_polynomial_opening(
-            VirtualPolynomial::Rs2Value,
-            SumcheckId::RegistersClaimReduction,
-        );
-        let (_, rs2_rv_claim_instruction_input) = accumulator.get_virtual_polynomial_opening(
-            VirtualPolynomial::Rs2Value,
-            SumcheckId::InstructionInputVirtualization,
-        );
-        // TODO: Make error and move to more appropriate place.
-        assert_eq!(rs2_rv_claim, rs2_rv_claim_instruction_input);
 
         result
     }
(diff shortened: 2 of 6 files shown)


jolt-core/src/subprotocols/sumcheck_claim.rs
```diff
@@ -0,0 +1,299 @@
...
+pub enum BatchingPolynomial {
+    Eq(CachedPointRef),
+    EqPlusOne(CachedPointRef),
+    Lt(CachedPointRef),
+    Identity,
+    // TODO: We could handle this using Identity, but we would need to add binary ops and constants
+    // within this struct.
+    UnmapRamAddress,
+    NoBatching,
+}
```
moodlezoup, line 108: I don't really understand how these will be handled on the zklean side, but conceptually this feels a bit awkward. Can we instead make these variants of `ClaimExpr`? I also think `EfficientlyEvaluablePolynomial` or `VerifierEvaluablePolynomial` might be a better name here

jolt-core/src/subprotocols/sumcheck_claim.rs
```diff
@@ -0,0 +1,299 @@
...
+    Cycle,
+    Full,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct CachedPointRef {
+    pub opening: OpeningRef,
+    pub sumcheck: SumcheckId,
+    pub part: ChallengePart,
+    pub reverse: bool,
```
moodlezoup: afaict we should never need to reverse the point; we normalize the point stored in OpeningAccumulator to always be big-endian

jolt-core/src/zkvm/registers/read_write_checking.rs
```diff
@@ -884,3 +939,48 @@ impl<F: JoltField, T: Transcript> SumcheckInstanceVerifier<F, T>
...
+    fn input_output_claims() -> InputOutputClaims<F> {
+        let rs1_value: ClaimExpr<F> = VirtualPolynomial::Rs1Value.into();
+        let rs2_value: ClaimExpr<F> = VirtualPolynomial::Rs2Value.into();
+        let rd_write_value: ClaimExpr<F> = VirtualPolynomial::RdWriteValue.into();
+
+        let registers_val: ClaimExpr<F> = VirtualPolynomial::RegistersVal.into();
+        let rs1_ra: ClaimExpr<F> = VirtualPolynomial::Rs1Ra.into();
+        let rs2_ra: ClaimExpr<F> = VirtualPolynomial::Rs2Ra.into();
+        let rd_wa: ClaimExpr<F> = VirtualPolynomial::RdWa.into();
+        let rd_inc: ClaimExpr<F> = CommittedPolynomial::RdInc.into();
```
moodlezoup, line 934: maybe we should impl `Add`, `Sub`, and `Mul` on `VirtualPolynomial` and `CommittedPolynomial` with `type Output = ClaimExpr<F>`, wdyt?

jolt-core/src/subprotocols/sumcheck_claim.rs
```diff
@@ -0,0 +1,299 @@
...
+        eq_poly::EqPolynomial,
+        identity_poly::UnmapRamAddressPolynomial,
+        multilinear_polynomial::PolynomialEvaluation as _,
+        opening_proof::{OpeningAccumulator, OpeningPoint, SumcheckId, BIG_ENDIAN},
+    },
+    zkvm::witness::{CommittedPolynomial, VirtualPolynomial},
+};
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum OpeningRef {
```
moodlezoup: this seems somewhat redundant with the existing `OpeningId` enum; is there a reason this needs to be a separate type?

jolt-core/src/subprotocols/sumcheck_claim.rs
```diff
@@ -0,0 +1,299 @@
...
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum BatchingPolynomial {
+    Eq(CachedPointRef),
+    EqPlusOne(CachedPointRef),
+    Lt(CachedPointRef),
+    Identity,
+    // TODO: We could handle this using Identity, but we would need to add binary ops and constants
+    // within this struct.
+    UnmapRamAddress,
```
moodlezoup, line 106: Can we put `ram_k` and `ram_start_address` inside this variant and get rid of `BatchingEvaluationParams`?

jolt-core/src/subprotocols/sumcheck_claim.rs
```diff
@@ -0,0 +1,299 @@
...
+            )
+            .evaluate(&r.r),
+            BatchingPolynomial::NoBatching => F::one(),
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub enum ClaimExpr<F> {
+    Val(F),
```
moodlezoup:
```suggestion
    Constant(F),
```

jolt-core/src/subprotocols/sumcheck_claim.rs
protoben: @jprider63 may be able to clarify this better than I can, but the issue is how that would be handled in zklean.

Our current plan is that each of these corresponds to quantifying the claim in some way. So, e.g., `Eq` on the cycle part of the point would be "for all cycles", `Lt` would be "for all previous cycles", etc. If they're incorporated into `ClaimExpr`, I think we would need to fall back on proving a claim about the actual arithmetic expression including these polynomials, which would complicate our proofs considerably.

I'll go ahead and change the name but leave this conversation open for now, in case there's more to say here.

jolt-core/src/subprotocols/sumcheck_claim.rs
protoben: Good point. I think this design maybe predates the normalization? Anyway, I'll get rid of that field.

jolt-core/src/zkvm/registers/read_write_checking.rs
protoben: Oh. Yeah, I think that would be a lot cleaner. I'll give it a shot.

jolt-core/src/subprotocols/sumcheck_claim.rs
protoben: 👍 I'll use that instead.

jolt-core/src/zkvm/registers/read_write_checking.rs
protoben: Sadly, I don't think this will work, due to type-system limitations. The problem is that Rust needs to constrain the type of `F`. So `impl<F> Add for VirtualPolynomial { type Output = ClaimExpr<F>; ... }` results in an error "`F` is not constrained by the impl trait".

We could do something like `impl<F> Add<VirtualPolynomial> for ClaimExpr<F> { ... }`, but then you still need to write things like `ClaimExpr::<F>::from(VirtualPolynomial::RamRa) * VirtualPolynomial::RamVal`. I can push a change like this if desired, but having tried it, I don't think it reduces boilerplate significantly.

moodlezoup: bummer, let's leave it as is then. thanks for looking into it

jolt-core/src/subprotocols/sumcheck_claim.rs
protoben: I renamed `BatchingPolynomial` to `VerifierEvaluablePolynomial`, as well as modifying several references to "batching" in this context in other type names and comments.

jolt-core/src/subprotocols/sumcheck_claim.rs
protoben: Looking at this a bit more closely, I don't think that `OpeningId` is a drop-in replacement for `OpeningRef`. What we need inside `ClaimExpr::Var` and `CachedPointRef` is a simple union of `VirtualPolynomial` and `CommittedPolynomial`, whereas `OpeningId` adds in separate sumcheck ids for each value. Since for us, the sumcheck id is the same for the whole claim, that's not something we want.

I took a stab at removing the redundancy you identified by making a small modification to `OpeningId` by adding a `PolynomialId` union in [this commit](https://github.com/a16z/jolt/pull/1132/commits/ee17101a4f5f761269a18d9460cab152b8498c79). Does this seem like an acceptable solution?

jolt-core/src/subprotocols/sumcheck_claim.rs
protoben: The problem is that we want to be able to use `UnmapRamAddress` statically, before the program itself is known. If you take a look [here](https://github.com/GaloisInc/jolt/blob/feature/sumcheck-front-end-api/jolt-core/src/zkvm/ram/raf_evaluation.rs#L349), where we use `UnmapRamAddress` inside a `SumcheckFrontend` impl, you can see that `ram_k` and `ram_start_address` aren't known yet. They're known when we call `expected_output_claim`, [here](https://github.com/GaloisInc/jolt/blob/feature/sumcheck-front-end-api/jolt-core/src/zkvm/ram/raf_evaluation.rs#L293) which is why we have the evaluation parameters as an argument to the evaluation function for the output claim.

moodlezoup: ah right

jolt-core/src/zkvm/ram/hamming_booleanity.rs
```diff
@@ -158,6 +161,18 @@ impl<F: JoltField> HammingBooleanitySumcheckVerifier<F> {
 impl<F: JoltField, T: Transcript> SumcheckInstanceVerifier<F, T>
     for HammingBooleanitySumcheckVerifier<F>
 {
+    fn input_claim(&self, accumulator: &VerifierOpeningAccumulator<F>) -> F {
+        let result = Self::input_output_claims().input_claim(&[F::one()], accumulator);
+
+        #[cfg(test)]
```
moodlezoup, line 168: Just had another thought––for all these `input_claim` and `expected_output_claim` functions, let's put the `Self::input_output_claims()` version in the `#[cfg(test)]` block and keep the existing implementation as the production code path. The latter should be slightly more efficient.

protoben: Ah, that makes sense. Will do.

protoben: Done.

PR #1136, opened 2025-12-01
Fix: Propagate user attributes through `#[jolt::provable]` macro

### LINKED ISSUE #588: `jolt::provable` attribute omits subsequent attributes
The provable [`proc_macro_attribute`](https://github.com/a16z/jolt/blob/783da5d32010e707f85085d59ae0451f6d8a6b25/jolt-sdk/macros/src/lib.rs#L17-L36) does not take into account subsequent attributes when generating the token stream of a function. For instance, the [`make_execute_function`'s `quote!`](https://github.com/a16z/jolt/blob/783da5d32010e707f85085d59ae0451f6d8a6b25/jolt-sdk/macros/src/lib.rs#L139-L144) does not output any used attributes alongside the function itself. This can lead to arbitrary behavior. If a user adds a custom safety-attribute or logical-attribute, this attribute will be omitted, which could introduce an unknown vulnerability or prevent the program from compiling. For example, the following code will fail to compile:

```rust
// guest code
#[jolt::provable]
#[allow(arithmetic_overflow)]
pub fn overflow_add() -> u8 {
    let x = u8::MAX + u8::MAX;
    x // should return 254
}
```

with the following error:

```sh
error: this arithmetic operation will overflow
  --> guest/src/lib.rs:21:13
   |
21 |     let x = u8::MAX + u8::MAX;
   |             ^^^^^^^^^^^^^^^^^ attempt to compute `u8::MAX + u8::MAX`, which would overflow
   |
   = note: `#[deny(arithmetic_overflow)]` on by default

error: could not compile `guest` (lib) due to 1 previous error
```

In this case, this issue is preventing the program from compiling, but other attributes if omitted could introduce undefined behavior.

Consider either allowing the usage of further attributes, or parsing all function attributes and whitelisting a defined subset of those attributes to prevent undefined behavior.

shreyas-londhe:
Fixes #588

### Problem
The `#[jolt::provable]` proc macro was dropping user-defined attributes (like `#[allow(arithmetic_overflow)]`) when generating functions, causing compilation errors or unexpected behavior.

### Changes
- `make_execute_function`: Forward `self.func.attrs` to the generated host function
- `make_main_func`: Wrap guest code in an inner function that carries the user's attributes (avoids unstable "attributes on expressions" feature)

### Testing
Added `examples/attribute/` with tests for:
- `#[allow(arithmetic_overflow)]`
- `#[allow(unused_variables)]`
- `#[inline]`
- Multiple attributes combined

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -331,21 +331,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1505bd5d3d116872e7271a6d4e16d81d0c8570876c8de68093a09ac269d8aac0"
 
 [[package]]
-name = "attribute"
-version = "0.1.0"
-dependencies = [
- "attribute-guest",
- "jolt-sdk",
-]
-
-[[package]]
-name = "attribute-guest"
-version = "0.1.0"
-dependencies = [
- "jolt-sdk",
-]
-
-[[package]]
 name = "autocfg"
 version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,8 +35,6 @@ members = [
     "jolt-inlines/blake2",
     "jolt-inlines/blake3",
     "jolt-inlines/bigint",
-    "examples/attribute",
-    "examples/attribute/guest",
     "examples/btreemap/host",
     "examples/btreemap/guest",
     "examples/collatz",


jolt-sdk/macros/src/lib.rs
```diff
@@ -758,8 +760,16 @@ impl MacroBuilder {
...
 
+        let attrs = &self.func.attrs;
+        let output = &self.func.sig.output;
         let block = &self.func.block;
-        let block = quote! {let to_return = (|| -> _ { #block })();};
+        let fn_name = self.get_func_name();
+        let inner_fn_name = syn::Ident::new(&format!("__jolt_guest_{}", fn_name), fn_name.span());
+        let block = quote! {
+            #(#attrs)*
+            fn #inner_fn_name() #output #block
```
0xAndoroid: Because we don't make this a closure anymore, it can't capture input variables from the outside. We still need to keep this a closure.
I believe in rust you can still apply most of the attributes to closures. But it'd be nice to document this limitation in the jolt book.

examples/attribute/guest/src/lib.rs
```diff
@@ -0,0 +1,31 @@
...
+#[jolt::provable]
+#[allow(unused_variables)]
+fn test_allow_unused() -> u32 {
+    let unused = 42; // Would warn without #[allow]
+    100
+}
+
+// Test #[inline]
+#[jolt::provable]
+#[inline]
```
0xAndoroid: Quite unclear what would inline attribute achieve on this function. I'd suggest removing this.

jolt-sdk/macros/src/lib.rs
shreyas-londhe: That makes sense, thanks for pointing this out! Unfortunately applying attributes to expressions (closures) is experimental on stable Rust (https://github.com/rust-lang/rust/issues/15701). 

We would need to use nightly version to add `#![feature(stmt_expr_attributes)]` to the generated guest code. Or we can apply the attributes to `main` which would then propagate down.

moodlezoup: Can we do something like:

```rust
        let inputs_vec: Vec<_> = self.func.sig.inputs.iter().collect();
        let inputs = quote! { #(#inputs_vec),* };
        
        let ordered_func_args = self.get_all_func_args_in_order();
        let all_names: Vec<_> = ordered_func_args.iter().map(|(name, _)| name).collect();

        let block = quote! {
            #(#attrs)*
            fn #inner_fn_name(#inputs) #output #block
            let to_return = #inner_fn_name(#all_names);
        };
```

PR #1205, opened 2026-01-15
feat: BlindFold zero-knowledge protocol

0xAndoroid:
## Summary

Adds native zero-knowledge to Jolt via the **BlindFold** protocol — a folding-based scheme that makes all sumcheck proofs ZK without SNARK composition (no Groth16/Plonk wrapper needed).

### Approach

Instead of revealing sumcheck round polynomial coefficients in the clear, the prover sends **Pedersen commitments** to them. The sumcheck verifier's algebraic checks (round consistency, degree bounds, final claim binding) are encoded into a small **verifier R1CS** circuit. A single Nova fold + Spartan proof over this R1CS proves all sumcheck rounds were executed correctly without revealing the witness.

### Protocol overview

1. **ZK sumcheck rounds**: Each sumcheck round commits polynomial coefficients via Pedersen (`C = Σ cᵢGᵢ + rH`) instead of sending them in the clear. Fiat-Shamir challenges are derived from commitments.

2. **Verifier R1CS construction**: Sumcheck verification logic — round consistency (`g(0)+g(1)=claim`), evaluation chaining (`g(rⱼ)=next_claim`), uni-skip power-sum constraints, final output bindings, and input claim derivations — is encoded as sparse R1CS constraints. Public inputs (challenges, initial claims) are **baked into matrix coefficients** so they don't appear in the witness.

3. **Nova folding**: The real R1CS instance (from actual sumcheck data) is folded with a randomly-sampled satisfying instance. The cross-term `T = (AZ₁)∘(BZ₂) + (AZ₂)∘(BZ₁) − u₁(CZ₂) − u₂(CZ₁)` is committed row-wise. Folding produces a relaxed R1CS instance `(A·Z)∘(B·Z) = u·(C·Z) + E` that hides the real witness.

4. **Spartan over folded R1CS**: An outer sumcheck proves `Σ_x eq(τ,x)·[Az(x)·Bz(x) − u·Cz(x) − E(x)] = 0`, followed by an inner sumcheck reducing the linear claim to a single witness evaluation.

5. **Hyrax-style openings**: The witness `W` and error `E` are arranged as `R'×C` grids. Pedersen row commitments enable evaluation proofs at arbitrary points without revealing the full vectors. Combined rows (`Σᵢ eq(r,i)·row_i`) are opened to verify `W(ry)` and `E(rx)`.

6. **Dory ZK evaluation commitments**: The Dory PCS opening proof produces an evaluation commitment `y_com` (Pedersen commitment to the claimed evaluation) instead of a cleartext value. BlindFold's extra constraints verify this binding.

### Key design decisions

- **No composition overhead**: Unlike Groth16/Plonk wrapping, BlindFold operates at the same algebraic level as Jolt's existing sumcheck machinery.
- **Baked public inputs**: Fiat-Shamir-derived values are embedded in R1CS matrix coefficients rather than witness variables, keeping the witness vector minimal.
- **Commitment reuse**: Sumcheck round commitments double as W row commitments for the Hyrax grid, avoiding redundant group operations.
- **Relaxed R1CS**: Nova-style relaxation (`u·Cz + E`) enables folding without cross-term explosion.

### New modules

- `subprotocols/blindfold/` — R1CS builder, Nova folding, Spartan prover/verifier, Hyrax openings, witness assignment, output/input constraint gadgets
- `poly/commitment/pedersen.rs` — Pedersen commitment scheme for small vectors
- `curve.rs` — `JoltCurve`/`JoltGroupElement` traits abstracting elliptic curve operations

### Changes across existing modules

- `sumcheck.rs` / `univariate_skip.rs` — `prove_zk` / `verify_zk` variants that commit coefficients and collect BlindFold stage data
- `zkvm/prover.rs` / `verifier.rs` — BlindFold integration at Stage 8 (after all sumcheck stages complete)
- `dory/commitment_scheme.rs` — ZK evaluation commitments (`y_com`) and blinding propagation
- `poly/opening_proof.rs` — `ProverOpeningAccumulator` extended with ZK stage data collection
- Various `zkvm/` submodules — `SumcheckInstanceParams` trait implementations for BlindFold constraint generation

## Test plan

- [x] `cargo nextest run -p jolt-core muldiv --cargo-quiet` (e2e correctness)
- [x] `cargo nextest run -p jolt-core blindfold --cargo-quiet` (protocol unit tests)
- [x] `cargo clippy --all --features allocative,host -- -D warnings`


jolt-core/src/guest/verifier.rs
```diff
@@ -28,13 +28,18 @@ pub fn preprocess(
...
+    FS: Transcript,
+>(
     inputs_bytes: &[u8],
     trusted_advice_commitment: Option<<PCS as CommitmentScheme>::Commitment>,
     outputs_bytes: &[u8],
-    proof: JoltProof<F, PCS, FS>,
+    proof: JoltProof<F, C, PCS, FS>,
     preprocessing: &JoltVerifierPreprocessing<F, PCS>,
-) -> Result<(), ProofVerifyError> {
+) -> Result<(), anyhow::Error> {
```
0xAndoroid: I think it's better to return ProofVerifyError. This also means that verifier.verify() needs to return ProofVerifyError.

jolt-core/src/poly/commitment/dory/commitment_scheme.rs
```diff
@@ -30,14 +34,77 @@ use tracing::trace_span;
...
+    for point in opening_point {
+        let scalar: F = (*point).into();
+        point_scalars.push(scalar);
+    }
+    transcript.append_scalars(b"dory_opening_point", &point_scalars);
+
+    let mut bytes = Vec::new();
+    ark_serialize::CanonicalSerialize::serialize_compressed(y_com, &mut bytes)
+        .expect("serialization");
+    transcript.append_bytes(b"dory_eval_commitment", &bytes);
```
0xAndoroid: Let's change the `append_point` on the transcript to accept JoltCurve generic, and then use it everywhere we append JoltCurve to the transcript.

jolt-core/src/poly/commitment/dory/dory_globals.rs
```diff
@@ -470,8 +469,6 @@ impl DoryGlobals {
     /// in repeated pairing operations.
     ///
     /// init_cache handles smart re-initialization internally:
-    /// - If cache doesn't exist, creates it
-    /// - If cache is too small, replaces with larger one
```
0xAndoroid, line 474: add back these comments

jolt-core/src/poly/opening_proof.rs
```diff
@@ -352,12 +425,17 @@ impl<F> ProverOpeningAccumulator<F>
...
         Self {
             openings: BTreeMap::new(),
             #[cfg(test)]
             appended_virtual_openings: std::cell::RefCell::new(vec![]),
             log_T,
+            #[cfg(feature = "zk")]
+            zk_stage_data: Vec::new(),
+            #[cfg(feature = "zk")]
+            uniskip_stage_data: Vec::new(),
+            zk_mode,
```
0xAndoroid: We can make zk_mode into a compile time constant.

jolt-core/src/subprotocols/streaming_sumcheck.rs
```diff
@@ -45,28 +47,18 @@ pub trait LinearSumcheckStage<F: JoltField>: Sized + MaybeAllocative + Send + Sy
...
     );
 }
 
-pub trait SharedStreamingSumcheckState<F: JoltField>:
-    Sized + MaybeAllocative + Send + Sync
-{
-    fn degree(&self) -> usize;
-    fn num_rounds(&self) -> usize;
-    fn input_claim(&self, accumulator: &ProverOpeningAccumulator<F>) -> F;
-}
```
0xAndoroid, line 63: Seems like this was left from bad merge. So probably need to be brought back.

jolt-core/src/subprotocols/sumcheck.rs
```diff
@@ -161,14 +176,233 @@ impl BatchedSumcheck {
...
+            initial_batched_claim,
+        )
     }
 
-    pub fn verify<F: JoltField, ProofTranscript: Transcript>(
-        proof: &SumcheckInstanceProof<F, ProofTranscript>,
+    /// Prove a batched sumcheck with Pedersen commitments (ZK mode).
+    ///
+    /// Instead of appending raw polynomial coefficients to the transcript,
+    #[cfg(feature = "zk")]
```
0xAndoroid: weird placement of cfg

jolt-core/src/subprotocols/sumcheck.rs
```diff
@@ -297,11 +616,265 @@ impl<F: JoltField, ProofTranscript: Transcript> SumcheckInstanceProof<F, ProofTr
...
+        transcript: &mut ProofTranscript,
+    ) -> Result<(F, Vec<F::Challenge>), ProofVerifyError> {
+        match self {
+            Self::Standard(proof) => proof.verify(claim, num_rounds, degree_bound, transcript),
+            Self::Zk(proof) => {
+                let challenges =
+                    proof.verify_transcript_only(num_rounds, degree_bound, transcript)?;
+                // For ZK mode, we don't compute the final claim here
+                // BlindFold verification ensures the R1CS constraints are satisfied
+                Ok((F::zero(), challenges))
```
0xAndoroid, line 827: If `zk` feature is not enabled we should return error

Cargo.toml
```diff
@@ -162,7 +162,7 @@ sha3 = "0.10.8"
 blake2 = "0.10"
 blake3 = { version = "1.5.0" }
 jolt-optimizations = { git = "https://github.com/a16z/arkworks-algebra", branch = "dev/twist-shout" }
-dory = { package = "dory-pcs", version = "0.2.0", features = ["backends", "cache", "disk-persistence"] }
+dory = { package = "dory-pcs", git = "https://github.com/a16z/dory", branch = "feat/zk", features = ["backends", "cache", "disk-persistence"] }
```
0xAndoroid: TODO: revert this after dory change is merged.

book/src/how/blindfold.md
```diff
@@ -0,0 +1,173 @@
+# Zero Knowledge: BlindFold
+
+Jolt achieves zero-knowledge natively via the **BlindFold** protocol — a folding-based scheme that makes all sumcheck proofs ZK without SNARK composition. Unlike most zkVMs that wrap their IOP in Groth16 or Plonk to get ZK, BlindFold operates at the same algebraic level as Jolt's existing sumcheck machinery, avoiding composition overhead entirely.
```
moodlezoup: let's add a citation for the Vega paper here

book/src/how/blindfold.md
```diff
@@ -0,0 +1,173 @@
...
+The verifier never sees polynomial coefficients, intermediate claims, or evaluation values from any sumcheck stage.
+
+## Security
+
+- **Hiding**: Pedersen commitments are computationally hiding under the discrete logarithm assumption. The prover's sumcheck messages reveal no information about the witness polynomials.
+- **Binding**: Pedersen commitments are computationally binding. A malicious prover cannot open a commitment to different coefficients.
+- **Folding soundness**: The folded instance satisfies relaxed R1CS if and only if both constituent instances do (with overwhelming probability over the folding challenge).
+- **Extraction**: An extractor can recover the real witness from two accepting transcripts with different folding challenges (via rewinding).
+
+## References
```
moodlezoup, line 168: let's add the Vega paper and https://blindfoldzkp.github.io/ here

book/src/roadmap/zk.md
moodlezoup, line 1: we can just delete this page now

jolt-core/src/poly/commitment/dory/commitment_scheme.rs
```diff
@@ -191,6 +265,11 @@ impl CommitmentScheme for DoryCommitmentScheme {
     /// can homomorphically combine the row commitments for multiple polynomials into the
     /// row commitments for the RLC of those polynomials. This is more efficient than computing
     /// the row commitments for the RLC from scratch.
+    ///
+    /// For shorter polynomials (e.g., T-length dense polys in a K*T RLC), we replicate their
+    /// row commitments K times to match the K*T matrix layout. This is because a dense poly
+    /// p(cycle) embedded in K*T is constant in the address dimension, so its row commitments
+    /// repeat with period T/num_cols.
```
moodlezoup: i don't follow –– why do we need to do this now, when we weren't before?

jolt-core/src/poly/opening_proof.rs
```diff
@@ -179,10 +179,79 @@ pub enum OpeningId {
     TrustedAdvice(SumcheckId),
 }
 
+impl OpeningId {
+    pub fn virt(poly: VirtualPolynomial, sc: SumcheckId) -> Self {
```
moodlezoup, line 178: nit:
```suggestion
    pub fn virtual(poly: VirtualPolynomial, sc: SumcheckId) -> Self {
```

jolt-core/src/subprotocols/blindfold/layout.rs
```diff
@@ -0,0 +1,149 @@
+use std::collections::HashSet;
+
+use crate::poly::opening_proof::OpeningId;
+
+use super::{OutputClaimConstraint, StageConfig};
+
+#[derive(Debug)]
+pub enum LayoutStep<'a> {
```
moodlezoup, line 21: Some more doc comments on the enum variants would be helpful

jolt-core/src/subprotocols/blindfold/protocol.rs
```diff
@@ -0,0 +1,919 @@
...
+/// The real instance is NOT included — verifier reconstructs from round_commitments,
+/// eval_commitments, public_inputs. The random instance IS included — verifier reads
+/// it from the proof and absorbs into transcript, never learning the random witness.
+#[derive(Clone, Debug, CanonicalSerialize, CanonicalDeserialize)]
+pub struct BlindFoldProof<F: JoltField, C: JoltCurve> {
+    pub random_u: F,
+    pub random_round_commitments: Vec<C::G1>,
+    pub random_noncoeff_row_commitments: Vec<C::G1>,
+    pub random_e_row_commitments: Vec<C::G1>,
+    pub random_eval_commitments: Vec<C::G1>,
```
moodlezoup: i feel like we should just throw these all into a `pub random_instance: RelaxedR1CSInstance<F, C>`?

jolt-core/src/subprotocols/blindfold/protocol.rs
```diff
@@ -0,0 +1,919 @@
...
+    pub az_r: F,
+    pub bz_r: F,
+    pub cz_r: F,
+    pub inner_sumcheck_proof: Vec<CompressedUniPoly<F>>,
+    /// Hyrax W opening: combined row (C elements)
+    pub w_combined_row: Vec<F>,
+    pub w_combined_blinding: F,
+    /// Hyrax E opening: combined row (C_E elements)
+    pub e_combined_row: Vec<F>,
+    pub e_combined_blinding: F,
```
moodlezoup: maybe add a type for Hyrax opening proof as well

moodlezoup: or better yet, add Hyrax to `poly/commitment/` and implement the `CommitmentScheme` trait?

jolt-core/src/subprotocols/blindfold/protocol.rs
```diff
@@ -0,0 +1,919 @@
...
+                    let num_aux = constraint.estimate_aux_var_count();
+                    num_openings + num_aux
+                } else {
+                    let n = fout.num_evaluations;
+                    if n > 1 {
+                        n + n - 1
+                    } else {
+                        n
+                    }
+                };
```
moodlezoup: what's going on here?

jolt-core/src/subprotocols/blindfold/spartan.rs
```diff
@@ -0,0 +1,794 @@
...
+pub fn hyrax_combined_row<F: JoltField>(flat: &[F], hyrax_C: usize, ry_row: &[F]) -> Vec<F> {
+    HyraxParams::combined_row_static(flat, hyrax_C, ry_row)
+}
+
+pub fn hyrax_evaluate<F: JoltField>(combined_row: &[F], ry_col: &[F]) -> F {
+    HyraxParams::evaluate(combined_row, ry_col)
+}
+
+pub fn hyrax_combined_blinding<F: JoltField>(row_blindings: &[F], ry_row: &[F]) -> F {
+    HyraxParams::combined_blinding(row_blindings, ry_row)
```
moodlezoup: these wrappers seem unnecessary

jolt-core/src/subprotocols/blindfold/protocol.rs
```diff
@@ -0,0 +1,919 @@
...
+
+            let r_j = transcript.challenge_scalar_optimized::<F>();
+            challenges.push(r_j);
+            claim = poly.evaluate(&r_j);
+        }
+
+        let az_r = proof.az_r;
+        let bz_r = proof.bz_r;
+        let cz_r = proof.cz_r;
+
```
moodlezoup, line 394: aren't we missing a verifier check here? i.e. shouldn't we be invoking `spartan_verifier.expected_claim` at some point?

jolt-core/src/subprotocols/blindfold/output_constraint.rs
```diff
@@ -0,0 +1,450 @@
...
+                    vec![ValueSource::Opening(*opening)],
+                )
+            })
+            .collect();
+        Self::sum_of_products(terms)
+    }
+
+    pub fn batch(
+        constraints: &[Option<OutputClaimConstraint>],
+        _num_batching_coefficients: usize,
```
moodlezoup: nit: remove

moodlezoup (commented):
Some initial comments/questions, still reading 🥵

jolt-core/src/subprotocols/blindfold/r1cs.rs
```diff
@@ -0,0 +1,1014 @@
...
+        let az = self.a.mul_vector(z);
+        let bz = self.b.mul_vector(z);
+        let cz = self.c.mul_vector(z);
+
+        for row in 0..self.num_constraints {
+            if az[row] * bz[row] != cz[row] {
+                return false;
+            }
+        }
+        true
```
moodlezoup:
```suggestion
        self.check_satisfaction(z).is_ok()
```

jolt-core/src/subprotocols/sumcheck.rs
```diff
@@ -161,14 +176,253 @@ impl BatchedSumcheck {
...
+
+        let max_num_rounds = sumcheck_instances
+            .iter()
+            .map(|sumcheck| sumcheck.num_rounds())
+            .max()
+            .unwrap();
+
+        // In ZK mode, don't absorb cleartext claims — polynomial commitments provide binding.
+        // Batching coefficients are still unpredictable (from transcript state after commitments).
+        let batching_coeffs: Vec<F> = transcript.challenge_vector(sumcheck_instances.len());
```
moodlezoup, line 219: this feels a little sketchy to me; we should make sure that this is secure

jolt-core/src/curve.rs
```diff
@@ -0,0 +1,391 @@
...
+    fn hash_to_g1(domain: &[u8]) -> Self::G1 {
+        bn254_hash_to_g1(domain)
+    }
+}
+
+/// BN254 hash-to-curve using try-and-increment.
+///
+/// For BN254 G1, curve equation is y² = x³ + 3.
+/// Repeatedly hashes domain with counter until finding valid curve point.
+fn bn254_hash_to_g1(domain: &[u8]) -> Bn254G1 {
```
moodlezoup: Instead of rolling our own hash-to-curve here, I think we should be able to use `G1Projective::rand(&mut rng)`, where `rng` can be ChaCha seeded using `domain`

jolt-core/src/poly/opening_proof.rs
```diff
@@ -179,10 +179,79 @@ pub enum OpeningId {
...
 /// (point, claim)
 pub type Opening<F> = (OpeningPoint<BIG_ENDIAN, F>, F);
 pub type Openings<F> = BTreeMap<OpeningId, Opening<F>>;
 
+/// ZK data collected during prove_zk for later use by BlindFold.
+///
+/// When ZK sumcheck is used, this stores the polynomial coefficients and
+/// blinding factors that are needed to construct the BlindFold witness.
+/// The commitments are stored as serialized bytes to keep the accumulator
+/// curve-agnostic.
```
moodlezoup: I kind of feel like there should be a separate `BlindfoldAccumulator` data structure

jolt-core/src/zkvm/verifier.rs
```diff
@@ -628,12 +1513,33 @@ impl<'a, F: JoltField, PCS: CommitmentScheme<Field = F>, ProofTranscript: Transc
                 CommittedPolynomial::UntrustedAdvice,
                 advice_claim * lagrange_factor,
             ));
+            scaling_factors.push(lagrange_factor);
+            include_untrusted_advice = true;
         }
 
         // 2. Sample gamma and compute powers for RLC
+        // Claims NOT absorbed — binding comes from polynomial commitments already in transcript.
```
moodlezoup: i think we should still absorb the claims in the non-zk case

jolt-core/src/zkvm/verifier.rs
```diff
@@ -203,52 +352,216 @@ impl<'a, F: JoltField, PCS: CommitmentScheme<Field = F>, ProofTranscript: Transc
...
+            &uni_skip_params,
             &self.opening_accumulator,
         );
 
-        let _r_stage1 = BatchedSumcheck::verify(
+        let instances: Vec<&dyn SumcheckInstanceVerifier<F, ProofTranscript>> =
+            vec![&spartan_outer_remaining];
+
+        let batching_coefficients: Vec<F> = {
+            let mut transcript_clone = self.transcript.clone();
```
moodlezoup: wait why is it okay to clone the transcript here?

0xAndoroid: Oh, that's pretty bad. Essentially it would derive the batching coeffs from cloned transcript, and then proceed to derive them again in the BatchedSumcheck::verify. It's secure, but terrible. Fixed.

book/src/how/blindfold.md
```diff
@@ -0,0 +1,175 @@
+# Zero Knowledge: BlindFold
+
+Jolt achieves zero-knowledge natively via the **[BlindFold](https://eprint.iacr.org/2025/2094)** protocol — a folding-based scheme that makes all sumcheck proofs ZK without SNARK composition. Unlike most zkVMs that wrap their IOP in Groth16 or Plonk to get ZK, BlindFold operates at the same algebraic level as Jolt's existing sumcheck machinery, avoiding composition overhead entirely.
```
markosg04: is this the right comparison? I think the reason non-blindfold would be expensive is because we would have to do a bunch of sigma protocols and do a bunch of homomorphic (cryptographic work). I think this is kinda what you are alluding to by "algebraic level" -- maybe we should just instead say something along the lines of we do mostly field work and a small amount of cryptographic work thanks to the folding (aka a small r1cs instance)

book/src/how/blindfold.md
```diff
@@ -0,0 +1,175 @@
...
+### Phase 1: ZK sumcheck rounds
+
+During each of the 7 sumcheck stages, `prove_zk` replaces the standard sumcheck prover:
+
+**Per round**, the prover:
+1. Computes the batched univariate polynomial as usual
+2. Commits to all coefficients via Pedersen: $C_j = \sum_i c_{j,i} \cdot G_i + \rho_j \cdot H$
+3. Appends $C_j$ (not the coefficients) to the Fiat-Shamir transcript
+4. Derives the challenge $r_j$ from the transcript
+5. Caches coefficients $c_{j,i}$, blinding $\rho_j$, and commitment $C_j$
```
markosg04: By cache do you mean save for later construction of subsequent r1cs?

jolt-core/src/subprotocols/blindfold/protocol.rs
0xAndoroid: That else just counts vars in a linear case. Regardless this function is unused, so I just removed it.

jolt-core/src/poly/opening_proof.rs
0xAndoroid: `virtual` is a keyword in Rust

jolt-core/src/curve.rs
```diff
@@ -0,0 +1,391 @@
...
+    fn pairing(g1: &Self::G1, g2: &Self::G2) -> Self::GT;
+
+    /// Multi-pairing: ∏ᵢ e(g1s[i], g2s[i])
+    fn multi_pairing(g1s: &[Self::G1], g2s: &[Self::G2]) -> Self::GT;
+
+    /// Multi-scalar multiplication in G1: Σᵢ scalars[i] * bases[i]
+    fn g1_msm<F: JoltField>(bases: &[Self::G1], scalars: &[F]) -> Self::G1;
+
+    /// Multi-scalar multiplication in G2: Σᵢ scalars[i] * bases[i]
+    fn g2_msm<F: JoltField>(bases: &[Self::G2], scalars: &[F]) -> Self::G2;
```
markosg04, line 85: flagging that we probably now have some redundancy / inconsistency with msm abstractions (SmallScalar, msm module, arkworks itself)

jolt-core/src/curve.rs
```diff
@@ -0,0 +1,391 @@
...
+
+    /// Generate a random G1 element
+    fn random_g1<R: rand_core::RngCore>(rng: &mut R) -> Self::G1;
+
+    /// Hash to a G1 curve point with unknown discrete log.
+    ///
+    /// SECURITY: The returned point must have an unknown discrete log relationship
+    /// to the standard generator and any other points derived from scalar multiplication.
+    /// This is critical for Pedersen commitment security.
+    fn hash_to_g1(domain: &[u8]) -> Self::G1;
```
markosg04: do we need this? I think it is part of a dead code path

jolt-core/src/zkvm/prover.rs
```diff
@@ -391,6 +436,10 @@ impl<'a, F: JoltField, PCS: StreamingCommitmentScheme<Field = F>, ProofTranscrip
         let one_hot_params =
             OneHotParams::new(log_T, preprocessing.shared.bytecode.code_size, ram_K);
 
+        // Use deterministic Pedersen generators for BlindFold protocol
+        // This ensures prover and verifier use the same generators
+        let pedersen_generators = PedersenGenerators::<C>::deterministic(4096);
```
markosg04: do we need this? should be using dory urs I think, and may be dead code

jolt-core/src/curve.rs
```diff
@@ -0,0 +1,391 @@
...
+/// For ark_bn254::Fr, this is a direct conversion.
+#[inline]
+fn jolt_field_to_fr<F: JoltField>(f: &F) -> Fr {
+    // Serialize the field element and deserialize as Fr
+    // This is safe because JoltField elements are assumed to be in the same field
+    let mut bytes = [0u8; 32];
+    f.serialize_uncompressed(&mut bytes[..])
+        .expect("serialization should succeed");
+    Fr::from_le_bytes_mod_order(&bytes)
+}
```
markosg04, line 279: in an ideal world / a future refactor: Jolt is fully generic over F, and some Commit (maybe Curve) traits, rn arkworks leaks through pretty heavily

jolt-core/src/curve.rs
markosg04, line 1: A meta comment about this abstraction: I think the right generic to have in the Jolt PIOP code should be a `JoltCommitment` rather than a `JoltCurve`. I think we should have a `JoltCurve` in order to blanket impl a whole bunch of useful operations and be able to eventually support multiple curve backends, but in order to support lattices and hashes eventually with blindfold, everything should be generic over a `JoltCommitmnet` (in some sense, I don't think blindfold needs to assume any implementation detail about the commitment beyond the properties like homomorphism, etc)

And there is a definitely a lot of leakage of the implementation details of these structures in the codebase (as was the case before this PR) which shall be addressed in the refactors

jolt-core/src/poly/commitment/pedersen.rs
markosg04, line 1: Do we need this code? At a glance seems to be used a in code path in prover and verifier that we may be able to get rid of

jolt-core/src/subprotocols/sumcheck.rs
```diff
@@ -297,11 +615,259 @@ impl<F: JoltField, ProofTranscript: Transcript> SumcheckInstanceProof<F, ProofTr
...
+            transcript.append_point(b"sumcheck_commitment", commitment);
+            let r_i: F::Challenge = transcript.challenge_scalar_optimized::<F>();
+            r.push(r_i);
+        }
+
+        Ok(r)
+    }
+}
+
+/// Sumcheck proof enum - replaces old SumcheckInstanceProof throughout codebase.
```
markosg04: nit: there are probably a few of these ai-implementation-meta-comments around the codebase, u could prolly ask claude to just look for all of them and delete them, not big deal tho

jolt-core/src/subprotocols/sumcheck.rs
```diff
@@ -297,11 +615,259 @@ impl<F: JoltField, ProofTranscript: Transcript> SumcheckInstanceProof<F, ProofTr
...
+#[derive(Debug, Clone)]
+pub enum SumcheckInstanceProof<F: JoltField, C: JoltCurve, ProofTranscript: Transcript> {
+    /// Non-ZK: coefficients visible to verifier
+    Standard(StandardSumcheckProof<F, ProofTranscript>),
+    /// ZK: only commitments visible, coefficients hidden in BlindFold
+    Zk(ZkSumcheckProof<F, C, ProofTranscript>),
+}
+
+impl<F: JoltField, C: JoltCurve, ProofTranscript: Transcript> CanonicalSerialize
+    for SumcheckInstanceProof<F, C, ProofTranscript>
```
markosg04, line 727: (again not necessarily initially introduced by ur PR) having these enum manual impls of serialization everywhere is kinda ugly, which we could either get rid of via macro or by serde later and macro

jolt-core/src/subprotocols/sumcheck.rs
```diff
@@ -297,11 +615,259 @@ impl<F: JoltField, ProofTranscript: Transcript> SumcheckInstanceProof<F, ProofTr
...
+        }
+    }
+}
+
+impl<F: JoltField, C: JoltCurve, ProofTranscript: Transcript>
+    SumcheckInstanceProof<F, C, ProofTranscript>
+{
+    /// Create a standard (non-ZK) sumcheck proof.
+    pub fn new_standard(compressed_polys: Vec<CompressedUniPoly<F>>) -> Self {
+        Self::Standard(StandardSumcheckProof::new(compressed_polys))
```
markosg04: wdyt of terminology `Standard` vs `Clear` generally speaking?

jolt-core/src/subprotocols/sumcheck_verifier.rs
```diff
@@ -46,20 +42,27 @@ pub trait SumcheckInstanceVerifier<F: JoltField, T: Transcript> {
...
+    #[cfg(feature = "zk")]
+    fn input_claim_constraint(&self) -> InputClaimConstraint;
+
+    #[cfg(feature = "zk")]
+    fn input_constraint_challenge_values(&self, accumulator: &dyn OpeningAccumulator<F>) -> Vec<F>;
+
+    #[cfg(feature = "zk")]
+    fn output_claim_constraint(&self) -> Option<OutputClaimConstraint>;
+    #[cfg(feature = "zk")]
+    fn output_constraint_challenge_values(&self, _sumcheck_challenges: &[F::Challenge]) -> Vec<F>;
```
markosg04, line 68: it seems that sometimes we have some things feature gated and others don't. Is this because of some weird clippy complaints about dead code sometimes when things are not re-exported?

jolt-core/src/subprotocols/sumcheck.rs
```diff
@@ -297,11 +615,259 @@ impl<F: JoltField, ProofTranscript: Transcript> SumcheckInstanceProof<F, ProofTr
...
+
+        for &degree in &self.poly_degrees {
+            if degree > degree_bound {
+                return Err(ProofVerifyError::InvalidInputLength(degree_bound, degree));
+            }
+        }
+
+        let mut r: Vec<F::Challenge> = Vec::new();
+        for commitment in &self.round_commitments {
+            transcript.append_point(b"sumcheck_commitment", commitment);
```
markosg04: append_point should probably be append_commitment (or just append, with types implementing some type of `appendable` trait / should inherit via serialization) eventually / we are gonna rework transcript to be more agnostic to the data being appended

jolt-core/src/poly/opening_proof.rs
0xAndoroid: I agree. Done

jolt-core/src/subprotocols/blindfold/output_constraint.rs
markosg04, line 1: I think all of this is all infra for a self-rolled intermediate representation of sorts for the PIOP verifier. Will def want to unify with the galois extractor work

jolt-core/src/subprotocols/blindfold/mod.rs
```diff
@@ -0,0 +1,544 @@
+//! BlindFold Zero-Knowledge Protocol
+//!
+//! This module implements the BlindFold protocol for making Jolt's sumcheck proofs
+//! zero-knowledge. The core idea is to encode sumcheck verification into a small R1CS
+//! circuit and apply a folding scheme.
```
markosg04, line 5: We are going to want to reuse this r1cs infra for the bespoke spartan + hyperkzg wrapper too

markosg04 (commented):
Initial peek. I am gonna look again later, gonna take a look at dory first then revisit this one.

jolt-core/src/subprotocols/blindfold/protocol.rs
0xAndoroid: Implementing `CommitmentScheme` trait is quite useless right now since most of the implementation isn't needed.

jolt-core/src/subprotocols/blindfold/protocol.rs
0xAndoroid: The check is a couple of lines down.

jolt-core/src/subprotocols/sumcheck_verifier.rs
0xAndoroid: I'm not sure what you mean. I feature gated everything ZK related.

jolt-core/src/curve.rs
0xAndoroid: I delegate this to the future refactor when lattice picture is clearer

jolt-core/src/poly/commitment/pedersen.rs
0xAndoroid: It's actively used, so not sure what you mean.

jolt-core/src/subprotocols/sumcheck.rs
0xAndoroid: changed to append_commitment.

0xAndoroid: the appendable trait can go in future refactor.

jolt-core/src/subprotocols/sumcheck.rs
0xAndoroid: serde is the way to go imo, but that's for later refactor.

jolt-core/src/poly/commitment/dory/commitment_scheme.rs
0xAndoroid: We don't. Byproduct of claude's hallucinations.

0xAndoroid: Removed this in latest commit.

PR #1223, opened 2026-01-23
Jolt runtime advice

moodlezoup:
## Runtime Advice

- Created an AdviceTape to the tracer CPU to record and read advice bytes
- Added JOLT_ADVICE_WRITE_ECALL_NUM for writing guest memory into advice tape
- Introduced virtual instructions to read from/write to advice tape. 
- Implemented runtime advice primitives in the SDK
- Added `#[jolt::advice]` proc-macro for dual-mode advice functions
- Introduced `AdviceWriter` and `AdviceReader` using `embedded-io`
- Added `check_advice!` macro with `VirtualAssertEQ` instruction
- Added `modinv` example to showcase runtime advice feature


0xAndoroid (commented):
## Missing Panic Check After First Pass

**Location:** `jolt-sdk/macros/src/lib.rs:711`

The `JoltDevice` (4th tuple element) containing the panic flag is discarded:

```rust
let (_lazy_trace, _, _, _, advice_tape) = guest_trace(...)
```

If the `compute_advice` pass panics, the code proceeds to the second pass with an incomplete advice tape. Users will see a cryptic "Failed to read from advice tape" error instead of the actual panic message from the first pass.

**Suggested fix:** Capture `JoltDevice`, check its `.panic` flag, and propagate the error before proceeding to the second pass.

```suggestion
                    // First pass: run compute_advice version to populate advice tape
                    let (_lazy_trace, _, _, jolt_device, advice_tape) = guest_trace(
                        &compute_advice_elf_contents,
                        None,
                        &input_bytes,
                        &untrusted_advice_bytes,
                        &trusted_advice_bytes,
                        &memory_config,
                        None, // Start with empty advice tape
                    );

                    // Check if first pass panicked
                    if jolt_device.panic {
                        panic!("compute_advice pass panicked");
                    }
```

examples/modinv/guest/src/lib.rs
```diff
@@ -0,0 +1,90 @@
...
+
+    inv_advice
+}
+
+/// Naive modular inverse implementation that computes directly without runtime advice.
+///
+/// This version performs the Extended Euclidean Algorithm entirely within the guest,
+/// without leveraging the advice system. This allows us to compare the cycle counts
+/// to demonstrate the efficiency gains from using runtime advice.
+#[jolt::provable]
```
0xAndoroid: Do we need this function to be provable?

jolt-sdk/src/lib.rs
```diff
@@ -73,3 +92,198 @@ impl<T> core::ops::Deref for UntrustedAdvice<T> {
...
+/// This macro generates a VirtualAssertEQ instruction that ensures the given boolean expression
+/// evaluates to true. If the condition is false, the proof will fail.
+///
+/// # Example
+/// ```ignore
+/// check_advice!(idx.len() == x.len());
+/// check_advice!(x[idx[i]] < x[idx[i + 1]]);
+/// ```
+#[macro_export]
+macro_rules! check_advice {
```
0xAndoroid, line 114: Just flagging that functionality of this is the same as `.unwrap_or_spoil_proof()` in the secp inline, which is implemented for any `Result<T, E>`. I think both are necessary for good devex.
We can also move that UnwrapOrSpoilProof trait into sdk and allow using it to check the advice.

jolt-sdk/macros/src/lib.rs
```diff
@@ -420,7 +420,12 @@ impl MacroBuilder {
                 program.set_func(#fn_name_str);
                 #set_std
                 #set_mem_size
-                program.build_with_channel(target_dir, #channel);
+
+                // Build the compute_advice version first
+                program.build_with_features(target_dir, #channel, &["compute_advice"]);
+
+                // Build the normal version (without compute_advice)
+                program.build_with_features(target_dir, #channel, &[]);
```
0xAndoroid, line 423: Do we have to build twice even if advice feature not used?

0xAndoroid (approved):
Looks good to me, just a couple of notes

jolt-sdk/macros/src/lib.rs
moodlezoup: I haven't found a good way to detect whether the program uses `jolt::advice` from within a different proc-macro

Separately, the guest dev needs to add `compute_advice` as a feature in their Cargo.toml. This is kind of annoying, but the good news is that if the `compute_advice` feature is _not_ in their Cargo.toml, the first build will quickly fail (and the error gets swallowed in `build_with_features`)

examples/modinv/guest/src/lib.rs
moodlezoup: nope, good catch

PR #1253, opened 2026-02-05
Optimize registers read/write-checking memory

moodlezoup:
Reduces memory usage of the registers read/write checking sumcheck by ~half, and speeds it up slightly

### DIFF
--- a/jolt-core/src/subprotocols/read_write_matrix/registers.rs
+++ b/jolt-core/src/subprotocols/read_write_matrix/registers.rs
@@ -282,6 +282,8 @@ impl<F: JoltField> ReadWriteMatrixCycleMajor<F, RegistersCycleMajorEntry<F, Look
 }
 
 impl<F: JoltField, C: OneHotCoeff<F>> CycleMajorMatrixEntry<F> for RegistersCycleMajorEntry<F, C> {
+    type AddressMajor = RegistersAddressMajorEntry<F>;
+
     fn row(&self) -> usize {
         (self.row_col & ROW_MASK) as usize
     }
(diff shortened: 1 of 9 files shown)


jolt-core/src/subprotocols/read_write_matrix/registers.rs
```diff
@@ -382,28 +468,114 @@ impl<F: JoltField> ReadWriteMatrixCycleMajor<F, RegistersCycleMajorEntry<F>> {
...
 /// Conceptually, both ra and Val can be seen as K x T matrices.
 ///
+/// # Memory Optimization: Packed `row_col` field
+///
+/// The `row` and `col` fields are packed into a single u64:
+/// - Lower 56 bits: row index (cycle count, row \in [0, T))
+/// - Upper 8 bits: column index (register index, col \in [0, K), K=128)
+///
+/// This saves 8 bytes per entry on 64-bit systems.
+///
```
markosg04: will usize and careful packing here break in wasm?

jolt-core/src/subprotocols/read_write_matrix/registers.rs
```diff
@@ -214,41 +234,90 @@ impl<F: JoltField> ReadWriteMatrixCycleMajor<F, RegistersCycleMajorEntry<F>> {
...
+    type AddressMajor = RegistersAddressMajorEntry<F>;
+
     fn row(&self) -> usize {
-        self.row
+        (self.row_col & ROW_MASK) as usize
     }
 
     fn column(&self) -> usize {
-        self.col as usize
+        (self.row_col >> COL_SHIFT) as usize
```
markosg04: the column value I think gets passed into u8 but here we cast to usize? is this intended? there may be few places we are doing this

markosg04 (approved):
nice work!

jolt-core/src/subprotocols/read_write_matrix/registers.rs
moodlezoup: this trait is also used for the RAM Twist instance, where the column index needs to be usize

jolt-core/src/subprotocols/read_write_matrix/registers.rs
moodlezoup: afaict it should be fine, even if usize is 32 bits. do you have a specific concern?

moodlezoup: reverted the packing optimization since we no longer need it (reordering struct fields gets us to the same place)

PR #1267, opened 2026-02-16
feat(zkvm): self-describing framed proof serialization

quangvdao:
## Summary

Replaces derive-macro-based `CanonicalSerialize`/`CanonicalDeserialize` for `JoltProof` with a manual, length-delimited section format intended for strictness and DoS resistance.

### Wire format

\`[magic: 4B "JOLT"][version: 1B][flags: 1B][section₀][section₁]…\`

Each section is \`[varint payload_len][payload bytes]\`. Sections are sequential and untagged — the deserializer reads them in the fixed order defined by the proof schema.

**Header (6 bytes):**
- Magic: `b"JOLT"` (4 bytes)
- Version: `1` (1 byte)
- Flags: bit 0 = `is_zk` (ZK mode vs standard), bits 1–7 reserved (must be 0)

### Changes

- **`zkvm::transport`** — new module with varint u64 read/write, magic+version header helpers, and capped section-length reading
- **`JoltProof` ser/de** — manual, length-delimited section format with:
  - Per-section caps (1 KiB for params, 128 KiB for other sections)
  - Intra-section trailing-byte checks
  - `debug_assert!` on write to catch cap mismatches during development
  - Final EOF check to reject trailing garbage
  - ZK/non-ZK mode flag with clear cross-mode error messages
- **`OpeningId` encoding** — packed header byte (2-bit kind + 6-bit sumcheck_id) with varint escape for sumcheck_id ≥ 63
- **`Claims` unification** — standalone `Claims` ser/de now uses varint count (matching JoltProof's inline format); JoltProof delegates to it via `write_section!`/`read_single_section!`
- **Varint hardening** — overflow check on 10th byte (shift=63)

### Security / DoS properties

- Section lengths capped before any allocation
- Commitment and claims counts capped at 10,000
- Varint decoding rejects overflows
- Reserved flag bits rejected (forward compatibility)
- Cross-mode deserialization gives clear error messages
- EOF check rejects trailing bytes after final section

### Test plan

- [x] `cargo clippy` clean in both `--features host` and `--features host,zk`
- [x] All proof_serialization + transport unit tests pass in both modes
- [x] OpeningId roundtrip tests cover all parameterized variants (CommittedPolynomial, VirtualPolynomial) × multiple SumcheckId values
- [ ] `muldiv` e2e test passes in both modes


tracer/src/emulator/mmu.rs
```diff
@@ -585,6 +586,14 @@ impl Mmu {
             u64::from_le_bytes(pre_value_bytes)
         };
 
+        if effective_address <= RAM_START_ADDRESS - 8 {
```
sagar-a16z: let's remove these changes since these examples should be working now without crashing. A write outside our defined memory regions is almost always just a write to a null ptr, which means something is actually wrong.

jolt-core/src/zkvm/proof_serialization.rs
```diff
@@ -57,17 +65,354 @@ pub struct JoltProof<F: JoltField, C: JoltCurve, PCS: CommitmentScheme<Field = F
...
 
+#[inline]
+fn io_err(e: std::io::Error) -> SerializationError {
+    SerializationError::IoError(e)
+}
+
+macro_rules! write_section {
+    ($w:expr, $c:expr, $($item:expr),+ $(,)?) => {{
+        let len: u64 = 0 $(+ $item.serialized_size($c) as u64)+;
+        debug_assert!(len <= MAX_SECTION_LEN, "section size {len} exceeds MAX_SECTION_LEN");
```
sagar-a16z: A release-mode prover can produce proofs it can't read back. Should be `assert!` or return `Err`.

jolt-core/src/zkvm/proof_serialization.rs
```diff
@@ -57,17 +65,354 @@ pub struct JoltProof<F: JoltField, C: JoltCurve, PCS: CommitmentScheme<Field = F
...
+        let one_hot_config = OneHotConfig::deserialize_with_mode(&mut limited, compress, validate)?;
+        let dory_layout = DoryLayout::deserialize_with_mode(&mut limited, compress, validate)?;
+        check_trailing_bytes(&limited)?;
+
+        // Commitments
+        let mut limited = transport::read_section(&mut reader, MAX_SECTION_LEN).map_err(io_err)?;
+        let n = transport::read_varint_u64(&mut limited).map_err(io_err)?;
+        let n_usize = usize::try_from(n).map_err(|_| SerializationError::InvalidData)?;
+        if n_usize > 10_000 {
+            return Err(SerializationError::InvalidData);
```
sagar-a16z: Should be a named constant like `MAX_CLAIMS_COUNT`. Also 10,000 × 384 bytes/GT = ~3.8 MB which exceeds the 128 KiB section cap enforced by `read_section` above. The section cap is the real constraint, no? (~341 commitments max).

jolt-core/src/zkvm/proof_serialization.rs
```diff
@@ -209,38 +591,35 @@ impl CanonicalDeserialize for OpeningId {
...
-                ))
-            }
-            _ if fused < OPENING_ID_VIRTUAL_BASE => {
-                let sumcheck_id = fused - OPENING_ID_COMMITTED_BASE;
+        let header = u8::deserialize_with_mode(&mut reader, compress, validate)?;
+        let kind = header >> 6;
+        let small = header & 0x3F;
+
+        let sumcheck_u64 = if small == OPENING_ID_SUMCHECK_ESCAPE {
+            transport::read_varint_u64(&mut reader).map_err(io_err)?
```
sagar-a16z: Accepts non-canonical encodings. i.e sumcheck_id=0 can be encoded inline (1 byte) or via escape+varint (2+ bytes). Add after the varint read:
                                                                                                                                                                                                                       
```rust                                      
  if sumcheck_u64 < OPENING_ID_SUMCHECK_ESCAPE as u64 {                                                                                                                                                                
      return Err(SerializationError::InvalidData);
  }     
```

jolt-core/src/zkvm/proof_serialization.rs
```diff
@@ -57,17 +65,354 @@ pub struct JoltProof<F: JoltField, C: JoltCurve, PCS: CommitmentScheme<Field = F
...
+            dory_layout,
+        })
+    }
+}
+
 #[cfg(not(feature = "zk"))]
 pub struct Claims<F: JoltField>(pub Openings<F>);
 
+#[cfg(not(feature = "zk"))]
+const MAX_CLAIMS_COUNT: u64 = 10_000;
```
sagar-a16z: 10,000 × ~33 bytes/claim = ~330 KB which exceeds MAX_SECTION_LEN (128 KiB). The section cap is the binding constraint (~3,700 claims max). These should be harmonized.

jolt-core/src/zkvm/mod.rs
```diff
@@ -37,6 +37,7 @@ pub mod r1cs;
 pub mod ram;
 pub mod registers;
 pub mod spartan;
+pub mod transport;
```
sagar-a16z: should be `pub(crate)`? it's internal?

jolt-core/src/zkvm/transport.rs
```diff
@@ -0,0 +1,185 @@
...
+
+#[inline]
+pub fn write_magic_version<W: Write>(w: &mut W, magic: &[u8], version: u8) -> io::Result<()> {
+    w.write_all(magic)?;
+    w.write_all(&[version])
+}
+
+#[inline]
+pub fn read_magic_version<R: Read>(r: &mut R, magic: &[u8]) -> io::Result<u8> {
+    let mut buf = vec![0u8; magic.len()];
```
sagar-a16z: Heap-allocates for a 4-byte read. Use let `mut buf = [0u8; 4]` instead.

jolt-core/src/zkvm/transport.rs
```diff
@@ -0,0 +1,185 @@
...
+        return Err(io::Error::new(
+            io::ErrorKind::InvalidData,
+            "section too large",
+        ));
+    }
+    Ok(r.take(len))
+}
+
+#[cfg(test)]
+mod tests {
```
sagar-a16z, line 110: We should have proper e2e tests - maybe also modify examples to use this new formatting.

                                                                                                                                                                                                           
Side note: the Dory crate does Vec::with_capacity(num_rounds) where num_rounds is an untrusted u32 from the stream, before any data hits the Take reader. An attacker can trigger a ~7.8 TB allocation. The section cap limits I/O but not the upfront allocation. Worth adding a sanity check on the num_rounds in the proof.

0xAndoroid:
Closing this as we're moving away from jolt-core. Not just that we're now using `serde` instead of `CanonicalSerialize`/`CanonicalDeserialize`.

PR #1270, opened 2026-02-18
Better Sum-check Docstrings For spartan (Stages 1-3)

abiswas3:
As part of ongoing formal verification of Jolt in Lean, we have had to extract all the sum-checks and constraints in Jolt in their simplest explicit form. 
More precisely:

+ Each sum-check is described only in terms of Virtual Or Real Polynomials. 
+ The Schwartz-Zippel randomness and Sum-check randomness is explicitly specified.
+ The verifier opening claim is explicitly stated.
+ The caching of prover openings is tracked in the documentations.
+ We will auto-derive a reduction DAG from this, and this will help iron out bugs.

Other stages will follow shortly, this week. 
**In the near future, these docs will be programmatically autogenerated from the implementation**. 
For now, they have been hand-checked for accuracy, and then `Claudius Caesar` did a diff with the code and my notes. 

Running
 
```bash
cargo clean --doc &&cargo doc --no-deps --document-private-items -p jolt-core && python3 -m http.server 8080 --directory target/doc
```

Should auto generate docs that look like the attached screenshots.

Screenshot 2026-02-18 at 19 30 41


0xAndoroid:
Closing as the contents are out of date, and not worth updating for now since the we're moving away from jolt-core.

PR #1301, opened 2026-03-02
Move multiply product check into Spartan outer sumcheck

quangvdao:
## Summary

- **Remove `Product` virtual polynomial** from the product-virtualization stage and the `RightLookupEqProductIfMul` R1CS constraint. Instead, prove the multiply-instruction identity `is_mul * (rl - li*ri) = 0` directly in the Spartan outer sumcheck as an additive degree-3 (inside eq) term.
- **Optimized extended-integer arithmetic** for the product term: the first product round uses fused materialization + eval computing `t(2)` and `t(∞)` from trace using `S64`/`S128`/`S160`/`S192` deferred-reduction accumulation (`FMAdd<F, S192>`, `FMAdd<F, S160>` via `mul_trunc`), eliminating all field conversions (`F::from_i128`, `F::from_u128`) from the inner loop. Subsequent rounds use compact `MultilinearPolynomial` storage (`bool`/`u64`/`u128`/`i128`) that binds to field elements on challenge ingestion.
- **Rebalance R1CS constraint groups**: move `RamAddrEqZeroIfNotLoadStore` from the first group to the second group (its `Bz` is `u64` and `Az` is dense, making it a poor fit for the first group's boolean-optimized path).
- **Merge old Stage 3 into Stage 2**: reorder sumcheck instances (RAM-related first, then Spartan-related) and eliminate one proving stage.

## Test plan

- [x] `cargo nextest run -p jolt-core muldiv` passes
- [x] `cargo clippy -p jolt-core --all-targets --features allocative,host -- -D warnings` clean
- [ ] Full test suite (`cargo nextest run -p jolt-core`)

Made with [Cursor](https://cursor.com)


quangvdao:
Close since stale; promising optimization imo but not worth the complexity right now

PR #1325, opened 2026-03-05
fix: harden Claude CI workflows against prompt injection and unauthorized access

persimmon16:
## Summary

- Replace insecure single-stage `pull_request_target` with a two-stage workflow that never places fork code on disk alongside secrets
- **Stage 1** (`prepare-claude-review.yml`): Runs on `pull_request` (no secrets), gates on author association or `claude-review-approved` label, uploads PR number as artifact
- **Stage 2** (`claude-code-review.yml`): Two independent jobs:
  - **Path A (auto)**: `workflow_run` trigger consumes stage 1 artifact, validates PR number + SHA cross-reference, checks out base branch only, runs Claude review
  - **Path B (manual)**: `issue_comment` trigger fires when a CODEOWNER comments `/claude-review` on a fork PR — no artifact needed, inherently safe
- Add `author_association` gating to `claude.yml` so only MEMBER/COLLABORATOR/OWNER can invoke `@claude`
- Add `--allowed-tools` restriction to `claude.yml` limiting Bash to `gh` and `cargo` commands
- Pin all GitHub Actions to immutable commit SHAs (defense against tag rewriting)
- Remove `id-token: write` from both workflows (eliminates OIDC token minting risk)
- Add explicit `github_token` to both review workflows, bypassing OIDC token exchange
- Consolidate `CODEOWNERS` rules in root file, protecting `.github/workflows/` from unauthorized modification

## Context

PR #1323 switched from OAuth to API key auth but the `setupGitHubToken()` path still relied on OIDC token exchange, which GitHub Actions never provides for fork PR runs. This caused failures on fork PRs like #1310.

The initial fix (commit `a253a3db`) used `pull_request_target` to run in base repo context. While functional, this places untrusted fork code on disk in a workflow that holds secrets and a write-capable `GITHUB_TOKEN`. A fork-modified `CLAUDE.md` could attempt prompt injection against the LLM reviewer.

Additionally, `claude.yml` allowed any GitHub user to comment `@claude` on issues/PRs to invoke the agent with unrestricted tool access — effectively giving any external user command execution in the CI runner.

The two-stage architecture resolves the code-review tension, while author-association gating and tool restrictions close the `claude.yml` attack surface.

## Security

**Two-stage isolation (code-review):**
- Fork code is **never** on disk in the secret-bearing stage
- `CLAUDE.md` always comes from the trusted base branch — prompt injection via modified project instructions is eliminated
- `workflow_run` trigger always uses base branch workflow YAML — forks cannot modify stage 2

**Author-association gating (claude.yml):**
- Only MEMBER, COLLABORATOR, or OWNER can invoke `@claude` — external users are rejected at the workflow `if:` condition level
- Gating uses GitHub's `author_association` field which is set by GitHub infrastructure and cannot be forged by the commenter

**Tool restrictions (claude.yml):**
- `--allowed-tools` limits Bash to `gh pr:*`, `gh issue:*`, and `cargo:*` patterns
- Agent retains ability to interact with GitHub and build/test code, but cannot run arbitrary shell commands
- Defense-in-depth: even if a trusted user's comment is somehow weaponized, the agent cannot exfiltrate secrets via arbitrary commands

**Artifact tampering defense (code-review Path A):**
- Artifact contains only a PR number (integer) — minimal attack surface
- Stage 2 validates the PR number is a positive integer
- Cross-references PR's `headRefOid` against `github.event.workflow_run.head_sha` (set by GitHub, unforgeable)
- SHA mismatch causes hard failure

**Supply chain hardening:**
- All GitHub Actions pinned to immutable commit SHAs — tag rewriting attacks are eliminated
- `CODEOWNERS` requires maintainer review for workflow file changes

**Expression injection prevention:**
- All `${{ }}` expressions touching user-controlled data use `env:` indirection, never direct interpolation in `run:` blocks

## Test plan

- [ ] Verify stage 1 (`Prepare Claude Review`) triggers on PRs from MEMBER authors
- [ ] Verify stage 2 (`Claude Code Review` / `claude-review-auto`) triggers via `workflow_run` and posts review
- [ ] Verify SHA validation passes (artifact PR number matches workflow_run head SHA)
- [ ] Test Path B: CODEOWNER comments `/claude-review` on a fork PR — verify `claude-review-manual` triggers
- [ ] Test negative: fork author comments `/claude-review` on own PR — verify no trigger (author_association rejected)
- [ ] Test label gate: open PR from fork without label — verify stage 1 skips; apply `claude-review-approved` label — verify stage 1 triggers
- [ ] Test `@claude` from non-member: comment `@claude hello` from external account — verify `claude.yml` does NOT trigger
- [ ] Test `@claude` from member: comment `@claude hello` from org member — verify workflow runs normally
- [ ] Test tool restrictions: verify Claude cannot run arbitrary Bash commands in `claude.yml`
- [ ] Verify all action SHA pins resolve to expected versions
- [ ] Re-test on a fork PR (e.g. #1310) to validate end-to-end fork PR review

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,2 +1,0 @@
-# Protect CI/CD workflow definitions — changes require maintainer review
-.github/workflows/ @a16z/jolt-maintainers
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,2 @@
+# Protect CI/CD workflow definitions — changes require maintainer review
+.github/workflows/ @a16z/jolt-maintainers
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -22,7 +22,7 @@ jobs:
 
     steps:
       - name: Download PR metadata
-        uses: actions/download-artifact@v4
+        uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093 # v4
         with:
           name: pr-metadata
           run-id: ${{ github.event.workflow_run.id }}
@@ -56,12 +56,12 @@ jobs:
           echo "Validated PR #$PR_NUMBER (SHA: $WORKFLOW_HEAD_SHA)"
 
       - name: Checkout base branch
-        uses: actions/checkout@v6
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
         with:
           fetch-depth: 1
 
       - name: Run Claude Code Review
-        uses: anthropics/claude-code-action@v1
+        uses: anthropics/claude-code-action@26ec041249acb0a944c0a47b6c0c13f05dbc5b44 # v1
         with:
           anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
           github_token: ${{ secrets.GITHUB_TOKEN }}
@@ -88,12 +88,12 @@ jobs:
 
     steps:
       - name: Checkout base branch
-        uses: actions/checkout@v6
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
         with:
           fetch-depth: 1
 
       - name: Run Claude Code Review
-        uses: anthropics/claude-code-action@v1
+        uses: anthropics/claude-code-action@26ec041249acb0a944c0a47b6c0c13f05dbc5b44 # v1
         with:
           anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
           github_token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -1,11 +1,85 @@
 name: Claude Code Review
 
 on:
-  pull_request_target:
-    types: [opened, synchronize, ready_for_review, reopened]
+  workflow_run:
+    workflows: ["Prepare Claude Review"]
+    types: [completed]
+  issue_comment:
+    types: [created]
 
 jobs:
-  claude-review:
+  # PATH A: Auto-trigger via workflow_run (trusted authors, or label-approved fork PRs)
+  claude-review-auto:
+    if: |
+      github.event_name == 'workflow_run' &&
+      github.event.workflow_run.conclusion == 'success'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
+      issues: write
+      actions: read
+
+    steps:
+      - name: Download PR metadata
+        uses: actions/download-artifact@v4
+        with:
+          name: pr-metadata
+          run-id: ${{ github.event.workflow_run.id }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Validate PR metadata
+        id: pr
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          WORKFLOW_HEAD_SHA: ${{ github.event.workflow_run.head_sha }}
+          REPO: ${{ github.repository }}
+        run: |
+          PR_NUMBER=$(cat pr_number.txt)
+
+          # Validate PR number is a positive integer (defense against artifact tampering)
+          if ! [[ "$PR_NUMBER" =~ ^[0-9]+$ ]]; then
+            echo "::error::Invalid PR number in artifact: not an integer"
+            exit 1
+          fi
+
+          # Cross-reference: verify the PR's head SHA matches the workflow_run's head SHA
+          # github.event.workflow_run.head_sha is set by GitHub and cannot be modified by forks
+          ACTUAL_SHA=$(gh pr view "$PR_NUMBER" --repo "$REPO" --json headRefOid -q '.headRefOid')
+          if [ "$ACTUAL_SHA" != "$WORKFLOW_HEAD_SHA" ]; then
+            echo "::error::SHA mismatch -- artifact PR ($ACTUAL_SHA) != workflow trigger ($WORKFLOW_HEAD_SHA)"
+            echo "::error::This may indicate artifact tampering. Aborting."
+            exit 1
+          fi
+
+          echo "number=$PR_NUMBER" >> "$GITHUB_OUTPUT"
+          echo "Validated PR #$PR_NUMBER (SHA: $WORKFLOW_HEAD_SHA)"
+
+      - name: Checkout base branch
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 1
+
+      - name: Run Claude Code Review
+        uses: anthropics/claude-code-action@v1
+        with:
+          anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          plugin_marketplaces: "https://github.com/anthropics/claude-code.git"
+          plugins: "code-review@claude-code-plugins"
+          prompt: "/code-review:code-review ${{ github.repository }}/pull/${{ steps.pr.outputs.number }}"
+
+  # PATH B: Manual trigger via CODEOWNER thumbs-up comment on a PR
+  claude-review-manual:
+    if: |
+      github.event_name == 'issue_comment' &&
+      github.event.issue.pull_request &&
+      (
+        github.event.comment.author_association == 'MEMBER' ||
+        github.event.comment.author_association == 'COLLABORATOR' ||
+        github.event.comment.author_association == 'OWNER'
+      ) &&
+      contains(github.event.comment.body, '👍')
     runs-on: ubuntu-latest
     permissions:
       contents: read
@@ -13,18 +87,16 @@ jobs:
       issues: write
 
     steps:
-      - name: Checkout repository
+      - name: Checkout base branch
         uses: actions/checkout@v6
         with:
           fetch-depth: 1
-          ref: ${{ github.event.pull_request.head.sha }}
 
       - name: Run Claude Code Review
-        id: claude-review
         uses: anthropics/claude-code-action@v1
         with:
           anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
           github_token: ${{ secrets.GITHUB_TOKEN }}
           plugin_marketplaces: "https://github.com/anthropics/claude-code.git"
           plugins: "code-review@claude-code-plugins"
-          prompt: "/code-review:code-review ${{ github.repository }}/pull/${{ github.event.pull_request.number }}"
+          prompt: "/code-review:code-review ${{ github.repository }}/pull/${{ github.event.issue.number }}"
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -69,7 +69,7 @@ jobs:
           plugins: "code-review@claude-code-plugins"
           prompt: "/code-review:code-review ${{ github.repository }}/pull/${{ steps.pr.outputs.number }}"
 
-  # PATH B: Manual trigger via CODEOWNER thumbs-up comment on a PR
+  # PATH B: Manual trigger via /claude-review comment on a PR
   claude-review-manual:
     if: |
       github.event_name == 'issue_comment' &&
@@ -79,7 +79,7 @@ jobs:
         github.event.comment.author_association == 'COLLABORATOR' ||
         github.event.comment.author_association == 'OWNER'
       ) &&
-      contains(github.event.comment.body, '👍')
+      contains(github.event.comment.body, '/claude-review')
     runs-on: ubuntu-latest
     permissions:
       contents: read
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -1,43 +1,30 @@
 name: Claude Code Review
 
 on:
-  pull_request:
+  pull_request_target:
     types: [opened, synchronize, ready_for_review, reopened]
-    # Optional: Only run on specific file changes
-    # paths:
-    #   - "src/**/*.ts"
-    #   - "src/**/*.tsx"
-    #   - "src/**/*.js"
-    #   - "src/**/*.jsx"
 
 jobs:
   claude-review:
-    # Optional: Filter by PR author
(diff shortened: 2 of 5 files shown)


persimmon16 (commented):
## Security Assessment — Two-Stage CI Workflow for Fork PR Reviews

**Scope:** GitHub Actions CI/CD workflow security — fork PR secret isolation, expression injection, artifact integrity

---

### Context

The existing `claude-code-review.yml` uses a `pull_request` trigger with `id-token: write` and OIDC token exchange. This fails on fork PRs because GitHub Actions doesn't provide OIDC tokens for fork-originated workflow runs. PR #1325 replaces this with a two-stage architecture:
- **Stage 1** (`prepare-claude-review.yml`): `pull_request` trigger, no secrets, uploads PR number artifact
- **Stage 2** (`claude-code-review.yml`): `workflow_run` + `issue_comment` triggers, has secrets, checks out base branch only

---

### Findings

#### 1. SOUND: Two-Stage Isolation Architecture

The fundamental security property holds: fork-authored code is **never** on disk in any job that has access to `ANTHROPIC_API_KEY` or a write-scoped `GITHUB_TOKEN`. Stage 2's `actions/checkout@v6` without an explicit `ref:` defaults to `github.ref`, which for `workflow_run` events is the default branch (main). The `workflow_run` trigger always executes the **base branch's** version of the workflow YAML, so a fork cannot modify the stage 2 pipeline itself.

#### 2. SOUND: Artifact Tampering Defense

The artifact contains only a PR number (integer). Defenses are layered:
1. **Integer validation** (`^[0-9]+$` regex) — prevents injection of non-numeric payloads
2. **SHA cross-reference** — `gh pr view` fetches the PR's `headRefOid` and compares against `github.event.workflow_run.head_sha` (set by GitHub infrastructure, unforgeable by forks)
3. **Run-ID scoping** — `actions/download-artifact@v4` with `run-id` ensures the artifact comes from the specific triggering stage 1 run

A TOCTOU concern exists (force-push between stage 1 and stage 2), but this fails safe — SHA mismatch causes a hard abort.

#### 3. SOUND: Expression Injection Prevention

All user-controlled GitHub expression contexts are handled safely:

| Expression | Location | Safety |
|---|---|---|
| `workflow_run.head_sha` | `env:` block → `$WORKFLOW_HEAD_SHA` | env indirection, not shell-interpolated |
| `comment.body` | `if:` condition only | expression engine, never shell |
| `issue.number` | `with:` input to action | action input, not shell |
| `pull_request.number` | `env:` block → `$PR_NUMBER` | env indirection |
| `steps.pr.outputs.number` | `with:` input to action | validated as integer; action input |

No `${{ }}` expressions appear directly inside `run:` blocks with user-controlled data.

#### 4. SOUND: Permission Scoping

Permissions are appropriate and minimal. Removal of `id-token: write` is correct. `secrets.GITHUB_TOKEN` is ephemeral and scoped. Write permissions cannot merge PRs (requires `contents: write`).

#### 5. LOW: Loose Trigger Match on Path B

```yaml
contains(github.event.comment.body, '👍')
```

This matches **any** comment by an authorized user containing the thumbs-up emoji, even unintentionally. Could cause unnecessary review runs and wasted API credits.

**Suggestion:** Consider a more explicit trigger like `/claude-review` or use `startsWith` instead of `contains`.

#### 6. LOW: `labeled` Event Type in Stage 1

The `labeled` event fires for **any** label addition, not just `claude-review-approved`. The `if:` condition correctly filters, but the workflow still starts briefly on public repos. Minor billing concern only.

#### 7. INFORMATIONAL: `claude.yml` Permission Escalation

`claude.yml` permissions changed from `read` to `write` for `pull-requests` and `issues`. Necessary for Claude to post responses, but the `@claude` trigger has no `author_association` gate. On a public repo, any GitHub user can trigger the workflow. The `claude-code-action` likely has its own guardrails, but worth documenting.

#### 8. INFORMATIONAL: No Concurrency Controls

Stage 2 has no `concurrency` group. Rapid pushes could spawn duplicate Claude reviews on the same PR. Not a security issue, but wastes API credits.

---

### Severity Summary

| # | Finding | Severity | Exploitable |
|---|---|---|---|
| 1 | Two-stage isolation | Sound | N/A |
| 2 | Artifact tampering defense | Sound | N/A |
| 3 | Expression injection prevention | Sound | N/A |
| 4 | Permission scoping | Sound | N/A |
| 5 | Loose thumbs-up trigger | Low | No — requires authorized user |
| 6 | `labeled` event billing | Low | No — job skips early |
| 7 | Write perms + no author gate | Informational | No — action has guardrails |
| 8 | No concurrency controls | Informational | No — operational only |

### Overall

**This PR is a significant security improvement.** The two-stage architecture correctly eliminates the `pull_request_target` footgun where fork code and secrets coexist. Artifact integrity checks are well-designed with proper defense-in-depth. Expression injection prevention is thorough. No high or critical findings.

**Recommended:** Approve with the minor suggestions above (tighter trigger match, concurrency group).

.github/workflows/prepare-claude-review.yml
```diff
@@ -0,0 +1,32 @@
...
+
+jobs:
+  prepare:
+    # Trusted authors: auto-trigger on any PR event
+    # External/first-time contributors: require 'claude-review-approved' label from a CODEOWNER
+    if: |
+      github.event.pull_request.author_association == 'MEMBER' ||
+      github.event.pull_request.author_association == 'COLLABORATOR' ||
+      github.event.pull_request.author_association == 'OWNER' ||
+      contains(github.event.pull_request.labels.*.name, 'claude-review-approved')
```
mattg-a16z: We're going to want to dedupe this pretty hard. We should have separate path logic for the label than the other PR triggers, that way we don't run into strange conditions where a label being added by another workflow triggers an expensive scan.

```
if: |
    (
      github.event.action != 'labeled' && 
      (
         github.event.pull_request.author_association == 'MEMBER' ||
         github.event.pull_request.author_association == 'COLLABORATOR' ||
         github.event.pull_request.author_association == 'OWNER' ||
         contains(github.event.pull_request.labels.*.name, 'claude-review-approved')
      ) || (
         github.event.action == 'labeled' &&
         github.event.label.name == 'claude-reivew-approved'
      )
    )
```

.github/workflows/claude-code-review.yml
```diff
@@ -1,43 +1,102 @@
...
+  claude-review-manual:
+    if: |
+      github.event_name == 'issue_comment' &&
+      github.event.issue.pull_request &&
+      (
+        github.event.comment.author_association == 'MEMBER' ||
+        github.event.comment.author_association == 'COLLABORATOR' ||
+        github.event.comment.author_association == 'OWNER'
+      ) &&
+      contains(github.event.comment.body, '👍')
```
mattg-a16z: This feels a bit unintuitive, but it can be changed later.

.github/workflows/claude-code-review.yml
```diff
@@ -1,43 +1,102 @@
...
+          if [ "$ACTUAL_SHA" != "$WORKFLOW_HEAD_SHA" ]; then
+            echo "::error::SHA mismatch -- artifact PR ($ACTUAL_SHA) != workflow trigger ($WORKFLOW_HEAD_SHA)"
+            echo "::error::This may indicate artifact tampering. Aborting."
+            exit 1
+          fi
+
+          echo "number=$PR_NUMBER" >> "$GITHUB_OUTPUT"
+          echo "Validated PR #$PR_NUMBER (SHA: $WORKFLOW_HEAD_SHA)"
+
+      - name: Checkout base branch
```
mattg-a16z, line 58: Do we know if this is needed? Isn't claude code going to do this part on its own?

mattg-a16z (commented):
Mostly good, see comments.

.github/workflows/prepare-claude-review.yml
persimmon16: Good call — fixed in 077164b3. The `if:` now separates the two paths:

- **Non-`labeled` events** (opened, synchronize, etc.): use the author-association + existing-label check as before
- **`labeled` events**: only fire when the specific `claude-review-approved` label is the one being added (via `github.event.label.name`)

This prevents unrelated label additions (from bots, other workflows, or manual triage) from triggering expensive Claude scans.

Note: your suggested snippet had a small typo (`claude-reivew-approved` -> `claude-review-approved`) which I corrected in the commit.

.github/workflows/claude-code-review.yml
persimmon16: Agreed it is not the most discoverable UX. The thumbs-up trigger was chosen because it is quick for CODEOWNERs reviewing fork PRs — a single emoji is faster than typing a command — but it does risk accidental triggers.

If you would prefer a more explicit trigger like `/claude-review`, I can update this in a follow-up commit. The change would be swapping `contains(github.event.comment.body, '👍')` for `contains(github.event.comment.body, '/claude-review')`. Let me know your preference and I will push it.

.github/workflows/claude-code-review.yml
persimmon16: Yes, the checkout step is required. `claude-code-action` does not check out the repo itself — it expects the working directory to already contain the repository files. Every example in the [official docs](https://github.com/anthropics/claude-code-action/blob/main/docs/solutions.md) uses `actions/checkout` before `claude-code-action`.

In this two-stage architecture the checkout is also security-critical: by checking out without an explicit `ref:`, it defaults to the base branch (main) on `workflow_run` events. This ensures fork code is never on disk in the secret-bearing stage — the agent only sees trusted base-branch files (including `CLAUDE.md`).

.github/CODEOWNERS
moodlezoup, line 1: let's just move this into the main [CODEOWNERS](https://github.com/a16z/jolt/blob/main/CODEOWNERS)

.github/workflows/claude-code-review.yml
moodlezoup: yeah I prefer `/claude-review`

.github/workflows/claude-code-review.yml
```diff
@@ -1,43 +1,102 @@
...
-    #   github.event.pull_request.author_association == 'FIRST_TIME_CONTRIBUTOR'
+  # PATH A: Auto-trigger via workflow_run (trusted authors, or label-approved fork PRs)
+  claude-review-auto:
+    if: |
+      github.event_name == 'workflow_run' &&
+      github.event.workflow_run.conclusion == 'success'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
```
moodlezoup, line 19: I don't think Claude needs write perms here
```suggestion
      pull-requests: read
```

.github/workflows/claude-code-review.yml
```diff
@@ -1,43 +1,102 @@
...
+        github.event.comment.author_association == 'OWNER'
+      ) &&
+      contains(github.event.comment.body, '👍')
     runs-on: ubuntu-latest
     permissions:
       contents: read
-      pull-requests: read
-      issues: read
-      id-token: write
+      pull-requests: write
```
moodlezoup, line 86:
```suggestion
      pull-requests: read
```

.github/workflows/claude.yml
```diff
@@ -12,38 +12,61 @@ on:
...
+        )
+      )
     runs-on: ubuntu-latest
     permissions:
       contents: read
-      pull-requests: read
-      issues: read
-      id-token: write
-      actions: read # Required for Claude to read CI results on PRs
+      pull-requests: write
```
moodlezoup, line 57:
```suggestion
      pull-requests: read
```

moodlezoup (commented):
looks good, just a couple nits

.github/CODEOWNERS
persimmon16: Done — merged into root `CODEOWNERS` and deleted `.github/CODEOWNERS` in d2c18b74.

.github/workflows/claude-code-review.yml
persimmon16: Updated in d2c18b74 — trigger is now `/claude-review`.

.github/workflows/claude-code-review.yml
persimmon16: `pull-requests: write` is required here — `claude-code-action` uses the `github_token` to post review comments back on the PR. With only `read`, the action runs but silently fails to post its results.

.github/workflows/claude-code-review.yml
persimmon16: Same as above — `write` is needed for `claude-code-action` to post review comments on the PR.

.github/workflows/claude.yml
persimmon16: This one especially needs `write` — the whole purpose of `claude.yml` is for Claude to respond to `@claude` mentions by posting comments on PRs and issues.

.github/workflows/claude-code-review.yml
```diff
@@ -1,43 +1,102 @@
...
+        uses: anthropics/claude-code-action@26ec041249acb0a944c0a47b6c0c13f05dbc5b44 # v1
         with:
           anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
+          github_token: ${{ secrets.GITHUB_TOKEN }}
           plugin_marketplaces: "https://github.com/anthropics/claude-code.git"
           plugins: "code-review@claude-code-plugins"
-          prompt: "/code-review:code-review ${{ github.repository }}/pull/${{ github.event.pull_request.number }}"
-          # See https://github.com/anthropics/claude-code-action/blob/main/docs/usage.md
-          # or https://code.claude.com/docs/en/cli-reference for available options
+          prompt: "/code-review:code-review ${{ github.repository }}/pull/${{ github.event.issue.number }}"
```
mattg-a16z, line 102: This looks strange, but should work, at least when the issue has a PR associated with it. Do we know what it will do if there is no PR attached to the issue it is asked to review?

mattg-a16z (commented):
LGTM.

One comment, but it should be fine.

.github/workflows/claude-code-review.yml
persimmon16: It does look weird, yeah — but line 76 checks `github.event.issue.pull_request`, which is only set when the comment is on a PR. Plain issue comments won't get past the `if:`.

As for why `issue.number` instead of `pull_request.number` — `issue_comment` events just don't populate `github.event.pull_request`, even on PRs. GitHub treats PRs as issues internally, so the numbers are the same. It's a quirk of the Actions event model.

PR #1344, opened 2026-03-17
Efficient bytecode commitment with dory precommitted geometry

RadNi:
## Summary

This PR adds committed-program support for bytecode and program-image openings throughout the zkVM flow. This allows prover to efficiently prove bytecode / program-image polynomials openings instead of verifier directly evaluating them which significantly reduces the recursion cycles.

The main changes are:
- add committed preprocessing and commitment plumbing for bytecode chunks and the program image
- add bytecode and program-image claim reductions so these openings can be reduced into the shared Stage 8 Dory opening flow
- introduce precommitted Dory geometry and total-var sizing across prover and verifier
- split stage 6 into 6a and 6b to separate address and cycle phases
- update SDK preprocessing helpers, examples, and docs for committed-bytecode flows
- add bytecode committed variant of preprocessing to sdk and updated all the examples to support it

### DIFF
--- a/jolt-core/src/zkvm/bytecode/read_raf_checking.rs
+++ b/jolt-core/src/zkvm/bytecode/read_raf_checking.rs
@@ -1645,6 +1645,16 @@ pub struct BytecodeReadRafSumcheckParams<F: JoltField> {
 }
 
 impl<F: JoltField> BytecodeReadRafSumcheckParams<F> {
+    pub fn stage_gammas(&self) -> [&[F]; N_STAGES] {
+        [
+            &self.stage1_gammas,
+            &self.stage2_gammas,
+            &self.stage3_gammas,
+            &self.stage4_gammas,
+            &self.stage5_gammas,
+        ]
+    }
+
     #[tracing::instrument(skip_all, name = "BytecodeReadRafSumcheckParams::gen")]
     pub fn gen<PCS: CommitmentScheme>(
         program: Option<&ProgramPreprocessing<PCS>>,
--- a/jolt-core/src/zkvm/bytecode/read_raf_checking.rs
+++ b/jolt-core/src/zkvm/bytecode/read_raf_checking.rs
@@ -1569,7 +1569,7 @@ impl<F: JoltField> BytecodeReadRafSumcheckParams<F> {
             &stage5_gammas,
         );
 
-        let int_poly = IdentityPolynomial::new(one_hot_params.bytecode_k.log_2());
+        let int_poly = IdentityPolynomial::new(one_hot_params.bytecode_len.log_2());
 
         let (_, raf_claim) = opening_accumulator
             .get_virtual_polynomial_opening(VirtualPolynomial::PC, SumcheckId::SpartanOuter);
@@ -1634,8 +1634,8 @@ impl<F: JoltField> BytecodeReadRafSumcheckParams<F> {
             stage5_gammas,
             input_claim,
             one_hot_params: one_hot_params.clone(),
-            K: one_hot_params.bytecode_k,
-            log_K: one_hot_params.bytecode_k.log_2(),
+            K: one_hot_params.bytecode_len,
+            log_K: one_hot_params.bytecode_len.log_2(),
             d: one_hot_params.bytecode_d,
             log_T: n_cycle_vars,
             val_polys,
--- a/jolt-core/src/zkvm/bytecode/read_raf_checking.rs
+++ b/jolt-core/src/zkvm/bytecode/read_raf_checking.rs
@@ -15,6 +15,7 @@ use crate::subprotocols::blindfold::{
 };
 use crate::{
     field::JoltField,
+    poly::commitment::commitment_scheme::CommitmentScheme,
     poly::{
         eq_poly::EqPolynomial,
         identity_poly::IdentityPolynomial,
@@ -35,15 +36,19 @@ use crate::{
         sumcheck_verifier::{SumcheckInstanceParams, SumcheckInstanceVerifier},
     },
     transcripts::Transcript,
-    utils::{math::Math, small_scalar::SmallScalar, thread::unsafe_allocate_zero_vec},
+    utils::{
+        errors::ProofVerifyError, math::Math, small_scalar::SmallScalar,
+        thread::unsafe_allocate_zero_vec,
+    },
     zkvm::{
         bytecode::BytecodePreprocessing,
-        config::OneHotParams,
+        config::{OneHotParams, ProgramMode},
         instruction::{
             CircuitFlags, Flags, InstructionFlags, InstructionLookup, InterleavedBitsMarker,
             NUM_CIRCUIT_FLAGS,
         },
         lookup_table::{LookupTables, NUM_LOOKUP_TABLES},
+        program::ProgramPreprocessing,
         witness::{CommittedPolynomial, VirtualPolynomial},
     },
 };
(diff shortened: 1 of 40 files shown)


jolt-core/src/zkvm/claim_reductions/bytecode.rs
```diff
@@ -0,0 +1,769 @@
...
+                tracing::info!(
+                    "BytecodeClaimReduction cache cycle len={} bound_value={} bound_eq={} scale={} cycle_claim={} synced_cycle_claim={}",
+                    self.value_poly.len(),
+                    self.value_poly.get_bound_coeff(0),
+                    self.eq_poly.get_bound_coeff(0),
+                    self.scale,
+                    c_mid,
+                    synced_cycle_claim,
+                );
+            }
```
0xAndoroid: Seems like there's still a lot of debugging leftovers in the PR

examples/backtrace/src/main.rs
```diff
@@ -15,13 +15,18 @@ fn main() {
     #[cfg(any(feature = "nostd", feature = "std"))]
     let should_panic = env_flag("JOLT_BT_TRIGGER").unwrap_or(true);
     #[cfg(any(feature = "nostd", feature = "std"))]
+    let bytecode_chunk = std::env::args()
+        .skip_while(|arg| arg != "--committed-bytecode")
+        .nth(1)
+        .map(|arg| arg.parse().unwrap());
+    #[cfg(any(feature = "nostd", feature = "std"))]
```
0xAndoroid: I think we shouldn't include this in every example. I'd prefer to have one new example with committed bytecode in fib.

jolt-core/src/zkvm/claim_reductions/bytecode.rs
```diff
@@ -0,0 +1,697 @@
...
+        } else {
+            self.params.address_alignment_rounds()
+        };
+        let num_trailing_variables =
+            trailing_cap.saturating_sub(self.params.num_rounds_for_current_phase());
+        let scaling_factor = self.scale * F::one().mul_pow_2(num_trailing_variables);
+        let prev_unscaled = previous_claim * scaling_factor.inverse().unwrap();
+        let poly_unscaled = self.compute_message_unscaled(prev_unscaled);
+        poly_unscaled * scaling_factor
+    }
```
moodlezoup: `compute_message_unscaled` (341–366), `cycle_intermediate_claim` (368–376), `final_claim_if_ready` (378–384), `compute_message` (396–417), and `ingest_challenge` below all duplicate the same-named methods on `PrecommittedProver` in `precommitted.rs:572–657`. The only real differences are the `bind_aux_polys(r_j)` call in `ingest_challenge` and using `num_rounds_for_current_phase()` instead of `num_rounds()` in `compute_message`.

`program_image.rs` and `advice.rs` both compose `PrecommittedProver` and avoid this — bytecode should too. Adding an auxiliary-bind hook (e.g. `fn bind_aux(&mut self, r_j)` defaulting to no-op) on `PrecommittedParams` would let `BytecodeClaimReductionProver` become `core: PrecommittedProver<F, _>` + `chunk_value_polys`, eliminating ~75 lines that otherwise have to be kept in sync by hand.

---

jolt-core/src/zkvm/claim_reductions/precommitted.rs
```diff
@@ -0,0 +1,782 @@
...
+    }
+
+    pub fn is_address_phase_active_round(&self, round: usize) -> bool {
+        self.address_phase_rounds.contains(&round)
+    }
+
+    #[inline]
+    pub fn is_address_phase_round(&self, round: usize) -> bool {
+        self.address_phase_rounds.contains(&round)
+    }
```
moodlezoup: `is_address_phase_active_round` and `is_address_phase_round` have identical bodies. Pick one and delete the other — having both is a footgun for future readers who'll wonder what subtle distinction the naming implies.

```suggestion
    #[inline]
    pub fn is_address_phase_round(&self, round: usize) -> bool {
        self.address_phase_rounds.contains(&round)
    }
```

---

jolt-core/src/zkvm/claim_reductions/precommitted.rs
```diff
@@ -0,0 +1,782 @@
...
+            self.cycle_phase_total_rounds
+        } else {
+            self.address_phase_total_rounds
+        }
+    }
+
+    pub fn round_offset(&self, is_cycle_phase: bool, max_num_rounds: usize) -> usize {
+        let _ = (is_cycle_phase, max_num_rounds);
+        0
+    }
```
moodlezoup: `round_offset` ignores both arguments and unconditionally returns `0`. It then flows through `BytecodeClaimReductionParams::round_offset` → `BytecodeClaimReductionProver::round_offset` (in the `SumcheckInstanceProver` impl), all returning `0`. Either drop the method and have callers return `0` directly, or finish the abstraction. The current shape reads like an incomplete refactor.

---

jolt-core/src/zkvm/verifier.rs
```diff
@@ -1754,78 +2159,202 @@ impl JoltSharedPreprocessing {
...
     ) -> Result<Self, ark_serialize::SerializationError> {
-        let bytecode =
-            BytecodePreprocessing::deserialize_with_mode(&mut reader, compress, validate)?;
-        let ram = RAMPreprocessing::deserialize_with_mode(&mut reader, compress, validate)?;
+        let program = ProgramPreprocessing::deserialize_with_mode(&mut reader, compress, validate)?;
+        let program_meta = ProgramMetadata::deserialize_with_mode(&mut reader, compress, validate)?;
         let memory_layout = MemoryLayout::deserialize_with_mode(&mut reader, compress, validate)?;
         let max_padded_trace_length =
             usize::deserialize_with_mode(&mut reader, compress, validate)?;
+        let bytecode_chunk_count = usize::deserialize_with_mode(&mut reader, compress, validate)?;
```
moodlezoup, line 2120: `bytecode_chunk_count` is read straight off the wire here with no validation, even though `new_committed` (line 2252) enforces nonzero / power-of-two / ≤256 / divides `bytecode_len`. The digest binds the value, so a tampered preprocessing fails Fiat-Shamir cleanly — but in the meantime a malformed serialized file panics deep inside chunk arithmetic instead of erroring at the read site. Calling `validate_committed_bytecode_chunk_count(bytecode_chunk_count)` (and the divisibility check) here, or in `Valid::check`, is cheap defense-in-depth.

---

jolt-core/src/zkvm/claim_reductions/program_image.rs
```diff
@@ -0,0 +1,567 @@
...
+
+    #[allow(clippy::too_many_arguments)]
+    pub fn new(
+        program_io: &JoltDevice,
+        ram_min_bytecode_address: u64,
+        padded_len_words: usize,
+        ram_K: usize,
+        scheduling_reference: PrecommittedSchedulingReference,
+        accumulator: &dyn OpeningAccumulator<F>,
+        _transcript: &mut impl Transcript,
```
moodlezoup: `_transcript: &mut impl Transcript` is unused — callers still have to thread a transcript in only to have it discarded. Either use it (e.g. to absorb the start address or padded length) or drop it from the signature; an underscore-prefixed `&mut` parameter in a public API is misleading.

---

moodlezoup (commented):
A few items from the review of the new precommitted machinery — mostly tech-debt cleanups, plus one defense-in-depth gap on preprocessing deserialization. No soundness issues found.

One follow-up that didn't fit inline: per the spec, the module header in `transpilable_verifier.rs` (lines 25–40, unchanged in this PR) still describes "Stages 1-6" / "Stage 6: CycleVariables phase", but `verify_stage6a()` and `verify_stage6b()` are the actual entry points now. The spec explicitly flags this: "Module comments in `transpilable_verifier.rs` should also be updated to describe Stage 6a and Stage 6b instead of the old monolithic Stage 6."

---

jolt-core/src/zkvm/bytecode/read_raf_checking.rs
moodlezoup, line 1: Now that we have split this sumcheck into address-phase and cycle-phase instances, we should get rid of the old monolithic code entirely

jolt-core/src/subprotocols/booleanity.rs
moodlezoup, line 1: Same or the Booleanity sumcheck -- we should fully cutover to the split instances

PR #1368, opened 2026-03-25
feat: add jolt-crypto crate

markosg04:
## Changes

-

## Testing

- [ ] Ran tests for modified crates
- [ ] `cargo clippy` and `cargo fmt` pass

## Breaking Changes

None

PR #1367, opened 2026-03-25
feat: add jolt-poly crate

markosg04:
## Changes

-

## Testing

- [ ] Ran tests for modified crates
- [ ] `cargo clippy` and `cargo fmt` pass

## Breaking Changes

None

PR #1366, opened 2026-03-25
feat: add jolt-instructions crate

markosg04:
## Changes

-

## Testing

- [ ] Ran tests for modified crates
- [ ] `cargo clippy` and `cargo fmt` pass

## Breaking Changes

None

PR #1369, opened 2026-03-25
feat(jolt-trace): add jolt-trace crate

markosg04:
## Changes

-

## Testing

- [ ] Ran tests for modified crates
- [ ] `cargo clippy` and `cargo fmt` pass

## Breaking Changes

None


markosg04:
* **#1369** <a href="https://app.graphite.com/github/pr/a16z/jolt/1369?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/a16z/jolt/1369?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1368** <a href="https://app.graphite.com/github/pr/a16z/jolt/1368?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1367** <a href="https://app.graphite.com/github/pr/a16z/jolt/1367?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1366** <a href="https://app.graphite.com/github/pr/a16z/jolt/1366?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1365** <a href="https://app.graphite.com/github/pr/a16z/jolt/1365?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1364** <a href="https://app.graphite.com/github/pr/a16z/jolt/1364?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1363** <a href="https://app.graphite.com/github/pr/a16z/jolt/1363?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1362** <a href="https://app.graphite.com/github/pr/a16z/jolt/1362?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

0xAndoroid:
🔍 Starting automated code review process. I'll analyze the changes for semantic consistency, bugs, tech debt, and security concerns. Results will be posted shortly.

---

crates/jolt-trace/src/cycle_row_impl.rs
```diff
@@ -0,0 +1,524 @@
...
+        let cflags = self.circuit_flags();
+        if cflags[CircuitFlags::WriteLookupOutputToRD] {
+            self.rd_write().map_or(0, |(_, _, post)| post)
+        } else if cflags[CircuitFlags::Jump] {
+            self.rd_write().map_or(0, |(_, _, post)| post)
+        } else {
+            0
+        }
+    }
+}
```
0xAndoroid: Bug: `lookup_output()` implementation doesn't match its doc comment contract.

The doc says "For branches: the comparison result (0 or 1)" but the implementation returns `0` for all branch instructions (BEQ, BGE, BGEU, BLT, BLTU, BNE). Branch instructions don't set `WriteLookupOutputToRD` or `Jump` flags, so they fall through to the `else { 0 }` arm.

In jolt-core, `to_lookup_output()` for branches actually computes the comparison. For example, BGE computes `((x as i64) >= (y as i64)) as u64` (see `jolt-core/src/zkvm/instruction/bge.rs:53-62`). This value is used in the R1CS witness as `should_branch_lookup_output` (see `jolt-core/src/zkvm/r1cs/inputs.rs:503`).

Since `CycleRow` is not consumed by jolt-core yet, this is latent — but it will cause incorrect witness generation when this trait replaces the existing `LookupQuery::to_lookup_output()` path. You'd need to either:
1. Evaluate the lookup table here (requires access to the table materializer), or
2. Re-derive the comparison from operands using the instruction flags (similar to how `lookup_index` derives its value from flags).

---

crates/jolt-trace/src/cycle_row_impl.rs
```diff
@@ -0,0 +1,524 @@
...
+        Instruction::VirtualXORROT32(_) => Flags::circuit_flags(&xor_rotate::VirtualXorRot32),
+        Instruction::VirtualXORROT24(_) => Flags::circuit_flags(&xor_rotate::VirtualXorRot24),
+        Instruction::VirtualXORROT16(_) => Flags::circuit_flags(&xor_rotate::VirtualXorRot16),
+        Instruction::VirtualXORROT63(_) => Flags::circuit_flags(&xor_rotate::VirtualXorRot63),
+        Instruction::VirtualXORROTW16(_) => Flags::circuit_flags(&xor_rotate::VirtualXorRotW16),
+        Instruction::VirtualXORROTW12(_) => Flags::circuit_flags(&xor_rotate::VirtualXorRotW12),
+        Instruction::VirtualXORROTW8(_) => Flags::circuit_flags(&xor_rotate::VirtualXorRotW8),
+        Instruction::VirtualXORROTW7(_) => Flags::circuit_flags(&xor_rotate::VirtualXorRotW7),
+        Instruction::NoOp => CircuitFlagSet::default().set(CircuitFlags::DoNotUpdateUnexpandedPC),
+        _ => unreachable!(),
```
0xAndoroid: Nit: The instruction coverage here matches jolt-core's `define_rv32im_trait_impls!` set exactly, which is correct — the tracer expands unsupported base instructions (SLL, LB, LW, atomics, etc.) into virtual sequences during emulation.

However, `unreachable!()` is fragile: if a new un-expanded instruction is added to the tracer, this panics at runtime with no compile-time warning. Consider either:
- Adding a comment explaining why this is safe (the tracer expands all other instructions), or
- Matching `Instruction::INLINE(_)` explicitly (like `lookup_table_kind_for_instruction` does at line 472) instead of relying on `_`.

---

crates/jolt-trace/src/program.rs
```diff
@@ -0,0 +1,697 @@
...
+    let program_size = program_end - RAM_START_ADDRESS;
+    let allocator = VirtualRegisterAllocator::default();
+
+    instructions = instructions
+        .into_iter()
+        .flat_map(|instr: Instruction| instr.inline_sequence(&allocator, xlen))
+        .collect();
+
+    (instructions, raw_bytes, program_size)
+}
```
0xAndoroid: Missing return value: jolt-core's `guest::program::decode()` returns 4 values `(Vec<Instruction>, Vec<(u64, u8)>, u64, u64)` — the 4th being `e_entry` (ELF entry point address). This function only returns 3.

`e_entry` is required by `JoltSharedPreprocessing::new()` and is used extensively in the prover pipeline (see every call site in `jolt-core/src/zkvm/prover.rs`). When jolt-trace eventually replaces the host decode path, callers won't have access to `e_entry`.

The underlying `tracer::decode()` does return `e_entry` — it's just being discarded on line 385. Consider returning it as the 4th tuple element to match jolt-core's interface.

---

crates/jolt-trace/src/program.rs
```diff
@@ -0,0 +1,697 @@
...
+    /// Compile the guest program via the `jolt` CLI with the given extra Cargo features.
+    ///
+    /// No-op if the ELF has already been built (unless `extra_features` contains
+    /// `"compute_advice"`, which produces a separate ELF).
+    #[expect(clippy::expect_used, clippy::unwrap_used, clippy::panic)]
+    #[tracing::instrument(skip_all, name = "Program::build_with_features")]
+    pub fn build_with_features(&mut self, target_dir: &str, extra_features: &[&str]) {
+        if self.elf.is_some() {
+            return;
+        }
```
0xAndoroid, line 145: Bug (also flagged by Graphite): The early-return check on line 137 only checks `self.elf.is_some()` but doesn't account for the `compute_advice` variant. If `build()` was already called (setting `self.elf`), a subsequent `build_with_features(dir, &["compute_advice"])` silently returns without building the compute-advice ELF.

The `is_compute_advice` check should be moved before the early return:
```rust
let is_compute_advice = extra_features.contains(&"compute_advice");
if is_compute_advice {
    if self.elf_compute_advice.is_some() { return; }
} else {
    if self.elf.is_some() { return; }
}
```

---

crates/jolt-trace/src/cycle_row_impl.rs
```diff
@@ -0,0 +1,524 @@
...
+    }
+
+    fn is_virtual(&self) -> bool {
+        self.virtual_sequence_remaining().is_some()
+    }
+
+    fn rs1_read(&self) -> Option<(u8, u64)> {
+        self.rs1_read()
+    }
+
```
0xAndoroid: Nit: This calls `self.rs1_read()` which for a `Cycle` ends up calling `Cycle::rs1_read()` (the inherent method), not `CycleRow::rs1_read()`. This works because the inherent method has the same signature, but it's a subtle dispatch — the trait method delegates to the inherent method of the same name. Might be worth a comment to clarify this is intentional (or use `Cycle::rs1_read(self)` explicitly to make the delegation visible).

---

0xAndoroid (changes requested):
## Code Review: `jolt-trace` crate extraction

Good architectural direction — extracting the host-side tracing into a standalone crate with the `CycleRow` trait abstraction cleanly decouples the tracer from the proving system. The `Program` builder, flag dispatch tables, and test coverage are solid.

### Issues found

**Bugs (2):**

1. **`lookup_output()` returns incorrect values for branch instructions** (`cycle_row_impl.rs:145-155`). The doc comment promises comparison results (0 or 1) for branches, but the implementation returns 0. In jolt-core, branches like BEQ/BGE compute actual comparisons via `LookupQuery::to_lookup_output()`. This will produce incorrect witnesses when `CycleRow` replaces the existing path.

2. **`build_with_features()` early-return skips compute_advice builds** (`program.rs:137`). Also flagged by Graphite — the `self.elf.is_some()` check doesn't distinguish between regular and compute_advice variants.

**Missing data (1):**

3. **`decode()` drops `e_entry`** (`program.rs:384-395`). jolt-core returns 4 values including the ELF entry point, which is required by `JoltSharedPreprocessing::new()`. The new function returns only 3, silently discarding `e_entry` from `tracer::decode()`.

### Minor observations

- `_ => unreachable!()` in flag dispatch is correct for the current instruction set (tracer expands others) but fragile. A comment or explicit `INLINE(_)` match would help.
- Dropping `_advice_tape` in `trace()` matches jolt-core's existing pattern — consistent.
- `trace_to_file` taking `&Path` instead of `&PathBuf` is a nice improvement over jolt-core.
- Removing the unused `F: JoltField` generic from `trace_analyze` is correct.

---

0xAndoroid (approved):

PR #1382, opened 2026-03-28
Harden verifier boundaries and transcript binding

quangvdao:
Posted by Cursor assistant (model: GPT-5.4) on behalf of the user (Quang Dao) with approval.

## Summary
- validate `trace_length` and `ram_K` before verifier size math and allocations
- bind `rw_config`, `one_hot_config`, and `dory_layout` into Fiat-Shamir
- return typed verifier errors for missing openings and malformed opening state
- make opening exhaustiveness always-on, add a Dory runtime guard, and tighten related regressions

## Validation
- `cargo fmt -q`
- `cargo clippy -p jolt-core --features host --message-format=short -q --all-targets -- -D warnings`
- `cargo clippy -p jolt-core --features host,zk --message-format=short -q --all-targets -- -D warnings`
- `cargo nextest run -p jolt-core verifier_ --cargo-quiet --features host`
- `cargo nextest run -p jolt-core verifier_ --cargo-quiet --features host,zk`
- `cargo nextest run -p jolt-core advice_e2e_dory --cargo-quiet --features host`

### DIFF
--- a/jolt-core/src/poly/commitment/dory/dory_globals.rs
+++ b/jolt-core/src/poly/commitment/dory/dory_globals.rs
@@ -5,13 +5,10 @@ use allocative::Allocative;
 use dory::backends::arkworks::{init_cache, ArkG1, ArkG2};
 use std::sync::{
     atomic::{AtomicU8, Ordering},
-    RwLock,
+    Mutex, MutexGuard, OnceLock, RwLock,
 };
 #[cfg(test)]
-use std::{
-    sync::OnceLock,
-    time::{SystemTime, UNIX_EPOCH},
-};
+use std::time::{SystemTime, UNIX_EPOCH};
 
 /// Dory matrix layout for OneHot polynomials.
 ///
@@ -185,6 +182,12 @@ pub struct DoryContextGuard {
     previous_context: DoryContext,
 }
 
+pub struct DoryRuntimeGuard {
+    _guard: MutexGuard<'static, ()>,
+}
+
+static DORY_RUNTIME_GUARD: OnceLock<Mutex<()>> = OnceLock::new();
+
 impl Drop for DoryContextGuard {
     fn drop(&mut self) {
         CURRENT_CONTEXT.store(self.previous_context as u8, Ordering::SeqCst);
@@ -195,6 +198,13 @@ impl Drop for DoryContextGuard {
 pub struct DoryGlobals;
 
 impl DoryGlobals {
+    pub fn acquire_runtime_guard() -> DoryRuntimeGuard {
+        let mutex = DORY_RUNTIME_GUARD.get_or_init(|| Mutex::new(()));
+        DoryRuntimeGuard {
+            _guard: mutex.lock().unwrap(),
+        }
+    }
+
     #[cfg(test)]
     pub(crate) fn configure_test_cache_root() {
         static TEST_CACHE_ROOT: OnceLock<()> = OnceLock::new();
--- a/jolt-core/src/poly/opening_proof.rs
+++ b/jolt-core/src/poly/opening_proof.rs
@@ -11,7 +11,6 @@ use crate::{
 };
 use allocative::Allocative;
 use num_derive::FromPrimitive;
-#[cfg(test)]
 use std::cell::RefCell;
 use std::collections::{BTreeMap, HashMap};
 use std::sync::Arc;
@@ -22,6 +21,7 @@ use super::{
 use crate::{
     field::JoltField,
     transcripts::Transcript,
+    utils::errors::ProofVerifyError,
     zkvm::witness::{CommittedPolynomial, VirtualPolynomial},
 };
 
@@ -212,9 +212,7 @@ where
     opening_ids_by_poly: BTreeMap<PolynomialId, Vec<OpeningId>>,
     /// Maps an `OpeningId` that was deduplicated to its canonical representative.
     pub aliases: BTreeMap<OpeningId, OpeningId>,
-    #[cfg(test)]
     pub appended_virtual_openings: RefCell<Vec<OpeningId>>,
-    #[cfg(test)]
     pub appended_committed_openings: RefCell<Vec<OpeningId>>,
     pub log_T: usize,
     #[allocative(skip)]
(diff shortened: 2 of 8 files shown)


jolt-core/src/zkvm/claim_reductions/ram_ra.rs
```diff
@@ -629,8 +629,8 @@ impl<F: JoltField> RaReductionParams<F> {
         let (r_address_val, r_cycle_val) = r_val.split_at_r(log_K);
 
         // Verify unified address (these should hold by construction after Stage 2 alignment).
-        debug_assert_eq!(r_address_raf, r_address_rw);
-        debug_assert_eq!(r_address_raf, r_address_val);
+        assert_eq!(r_address_raf, r_address_rw);
```
0xAndoroid, line 632: **[Score 50/100 — Medium]** Interaction with `record_missing_opening` dummy values.

Upgrading `debug_assert_eq!` → `assert_eq!` is the right call for release-mode safety. However, `record_missing_opening` (opening_proof.rs:654) now returns dummy all-zero points when an opening is missing. If a malformed proof is missing *one* of the three RamRa openings (but not all three), the dummy's all-zero `r_address` won't match the real openings' `r_address`, causing this `assert_eq!` to panic **before** `take_missing_opening_error()` gets a chance to surface the typed `MissingOpening` error.

The verifier still safely rejects (panic = rejection), but this violates the intent of the deferred error pattern introduced in this PR. In practice, the three RamRa openings are populated by stages 2-4, and `take_missing_opening_error()` after each stage would catch them first — so this is only reachable if sumcheck verification in stages 2-4 somehow succeeds with dummy values.

Consider either:
- Guarding with a fallible check (`if r_address_raf != r_address_rw { ... }`) that returns a dummy `Self` so `take_missing_opening_error()` can surface the real error, or
- Adding a comment that this path is unreachable due to inter-stage error checks

jolt-core/src/poly/opening_proof.rs
```diff
@@ -702,10 +725,15 @@ where
...
-                        "Inconsistent duplicate opening claims: {key:?} vs {existing_id:?}"
-                    );
+                    if *claim != existing_claim {
+                        let mut malformed = self.malformed_proof.borrow_mut();
+                        if malformed.is_none() {
+                            *malformed = Some(format!(
+                                "inconsistent duplicate opening claims: {key:?} vs {existing_id:?}"
+                            ));
+                        }
+                        return;
```
0xAndoroid, line 735: **[Score 50/100 — Medium]** Defense-in-depth: inconsistent opening leaves stale value accessible.

When inconsistent duplicate claims are detected, this records the `MalformedProof` error and returns without inserting or aliasing the new key. The original opening (at `existing_id`) remains in the accumulator with its first-seen claim value.

This is sound — `take_missing_opening_error()` catches the `MalformedProof` after the stage, and sumcheck/Fiat-Shamir binding independently rejects. However, as defense-in-depth, consider also poisoning the existing opening so that any subsequent `get_*` for either key triggers `record_missing_opening`:

```rust
// After recording malformed error:
self.openings.remove(&existing_id);
```

This ensures the verifier never uses a claim value from a proof with internally inconsistent openings, even within a single stage.

jolt-core/src/poly/commitment/dory/dory_globals.rs
```diff
@@ -195,6 +198,13 @@ impl Drop for DoryContextGuard {
 pub struct DoryGlobals;
 
 impl DoryGlobals {
+    pub fn acquire_runtime_guard() -> DoryRuntimeGuard {
+        let mutex = DORY_RUNTIME_GUARD.get_or_init(|| Mutex::new(()));
+        DoryRuntimeGuard {
+            _guard: mutex.lock().unwrap(),
```
0xAndoroid, line 204: **[Score 25/100 — Low]** `unwrap()` on mutex lock — poisoning propagation.

If a thread panics while holding `DoryRuntimeGuard`, the mutex is poisoned and all future `acquire_runtime_guard()` calls will also panic. This is arguably correct (a panic during proving indicates a logic error), but it makes the process permanently broken rather than allowing recovery.

Consider `lock().unwrap_or_else(|e| e.into_inner())` to recover from poisoned state, or use `parking_lot::Mutex` which doesn't support poisoning.

jolt-core/src/poly/opening_proof.rs
```diff
@@ -647,7 +646,31 @@ where
...
+            malformed_proof: RefCell::new(None),
+        }
+    }
+
+    fn record_missing_opening(&self, key: OpeningId) -> (OpeningPoint<BIG_ENDIAN, F>, F) {
+        let mut missing = self.missing_opening.borrow_mut();
+        if missing.is_none() {
+            *missing = Some(key);
+        }
+        let dummy_len = self.log_T.max(1) + 128;
```
0xAndoroid, line 659: **[Score 25/100 — Low]** Magic number `128` in dummy point length.

The `+ 128` ensures the dummy point is long enough for any downstream `split_at(log_K)` call (since `LOG_K` for RV64 is at most 128). Consider documenting this derivation or using a named constant to make the intent clear and prevent it from silently becoming wrong if max address-space sizes change.

jolt-core/src/poly/opening_proof.rs
```diff
@@ -847,3 +875,24 @@ pub fn compute_advice_lagrange_factor<F: JoltField>(
...
         .product()
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use ark_bn254::Fr;
+
+    #[test]
+    fn verifier_accumulator_reports_missing_opening() {
```
0xAndoroid, line 885: **[Score 25/100 — Low]** Missing test coverage for `MalformedProof` path.

This test exercises the `MissingOpening` error path via `record_missing_opening`. Consider adding a companion test for the `MalformedProof` path — triggered by inserting two openings at the same point with different claims via `populate_or_alias_opening` (line 728). This would exercise the priority ordering in `take_missing_opening_error()` where `MalformedProof` takes precedence over `MissingOpening`.

0xAndoroid (commented):
Code review: 5 findings (2 medium, 3 low). Net positive PR for security — adds boundary validation, Fiat-Shamir binding, typed errors, and always-on exhaustiveness. Findings are defense-in-depth improvements, not exploitable vulnerabilities.

quangvdao:
Close since stale

PR #1406, opened 2026-04-04
feat(jolt-eval): invariant and objective framework for AI-assisted development

moodlezoup:
## Summary

- **jolt-eval crate**: mechanically checkable invariants and objectives for the Jolt zkVM, designed for AI agent productivity and minimizing human verification surface
- **Invariant framework**: `#[invariant(Test, Fuzz, RedTeam)]` macro synthesizes tests, fuzz targets, and AI red-team harnesses from a single invariant definition
- **Objective system**: unified `Objective` trait for both static-analysis (LLOC, cognitive complexity, Halstead bugs) and Criterion-benchmarked performance measurements
- **ObjectiveFunction struct**: `const`-constructible struct combining objectives into a scalar for the optimizer to minimize; registered via `ObjectiveFunction::all()` / `by_name()`
- **Type-safe measurement keys**: `OptimizationObjective` enum replaces string-keyed HashMaps; three nested enums (`StaticAnalysisObjective`, `PerformanceObjective`, `OptimizationObjective`) with discriminant-based Hash/Eq
- **AI red-team binary** (`redteam`): runs Claude in an isolated worktree to find invariant violations via structured JSON counterexamples
- **AI optimize binary** (`optimize`): iterative optimization loop — agent proposes changes, diff is applied, objectives re-measured, invariants checked, accepted/reverted with git branch + per-iteration commits
- **CLI features**: `--list`, `--test` (built-in e2e sort targets), `--verbose` (shows prompts/responses), mutual exclusivity via clap
- **E2e sort tests**: `naive_sort` optimization target, `candidate_sort` red-team target with subtle bug (correct for ≤16 elements, skips last element for larger arrays), both MockAgent and `#[ignore]` real-agent tests

## Test plan

- [x] `cargo nextest run -p jolt-eval` — 87 tests pass, 2 skipped (real-agent `#[ignore]` tests)
- [x] `cargo clippy -p jolt-eval --all-targets` — zero warnings
- [x] `cargo run --release -p jolt-eval --bin optimize -- --list` — lists 5 objective functions
- [x] `cargo run --release -p jolt-eval --bin redteam -- --list` — lists red-teamable invariants + e2e targets
- [x] `cargo run --release -p jolt-eval --bin optimize -- --test --verbose` — e2e sort optimization with real Claude agent
- [x] `cargo run --release -p jolt-eval --bin redteam -- --test --verbose` — e2e sort red-team with real Claude agent
- [x] `cargo bench -p jolt-eval` — Criterion benchmarks run


jolt-eval/src/objective/mod.rs
```diff
@@ -0,0 +1,343 @@
...
+    }
+
+    /// Override for performance objectives benchmarked by Criterion.
+    fn run(&self, _setup: Self::Setup) {}
+}
+
+// Data-containing enums — Hash/Eq based on discriminant only
+
+/// Static-analysis objectives.
+#[derive(Clone, Copy, PartialEq, Hash)]
```
0xAndoroid: Comment says "Hash/Eq based on discriminant only" but `StaticAnalysisObjective` and `PerformanceObjective` both derive `Hash`/`PartialEq`, which compare all fields including `target_dir`. The custom `OptimizationObjective::hash()` at line 269 hashes the discriminant AND delegates to the inner enum's derived hash. Test `optimization_objective_hashmap_key` at line 333 panics with "no entry found for key" — confirmed locally.

Either implement custom `Hash`/`PartialEq` on the inner enums that ignores payload fields (matching stated intent), or fix the test and comment to reflect that field data participates in equality.

jolt-eval/src/invariant/soundness.rs
```diff
@@ -0,0 +1,554 @@
...
+            include_hunk = !line.contains("..");
+        }
+        if include_hunk {
+            result.push_str(line);
+            result.push('\n');
+        }
+    }
+
+    result
+}
```
0xAndoroid, line 344: `filter_patch` only blocks hunks with `..` in the path, but a patch can add `build.rs`, modify `Cargo.toml` (adding path dependencies or a `build = "..."` key), or create `.cargo/config.toml` — all within the sandbox, no `..` needed. Cargo auto-executes build scripts during `compile_guest` with full host privileges. `catch_unwind` only catches panics in the current process, not in cargo child processes.

Recommend an allowlist approach:

```suggestion
pub fn filter_patch(patch: &str) -> String {
    const ALLOWED_PATHS: &[&str] = &["src/lib.rs", "src/main.rs"];

    let mut result = String::new();
    let mut include_hunk = true;

    for line in patch.lines() {
        if line.starts_with("diff --git") {
            // Extract path from "diff --git a/path b/path"
            let has_traversal = line.contains("..");
            let path = line.split_whitespace()
                .nth(3)
                .unwrap_or("")
                .trim_start_matches("b/");
            include_hunk = !has_traversal
                && ALLOWED_PATHS.iter().any(|allowed| path == *allowed);
        } else if line.starts_with("--- ") || line.starts_with("+++ ") {
            // Don't let --- /+++ lines override the diff --git decision
        }
        if include_hunk {
            result.push_str(line);
            result.push('\n');
        }
    }

    result
}
```

jolt-eval/src/invariant/soundness.rs
```diff
@@ -0,0 +1,554 @@
...
+impl Drop for PatchGuard {
+    fn drop(&mut self) {
+        if self.applied {
+            let _ = Command::new("git")
+                .current_dir(&self.dir)
+                .args(["checkout", "."])
+                .status();
+        }
+    }
+}
```
0xAndoroid, line 278: `git checkout .` only reverts modified tracked files — new files added by a patch persist after drop. This contaminates the sandbox between invariant checks (e.g., a `build.rs` added by one check survives into the next). The optimize loop's reject path correctly uses both `git checkout .` and `git clean -fd`; `PatchGuard` should do the same.

```suggestion
struct PatchGuard {
    dir: PathBuf,
    applied: bool,
}

impl Drop for PatchGuard {
    fn drop(&mut self) {
        if self.applied {
            let _ = Command::new("git")
                .current_dir(&self.dir)
                .args(["checkout", "."])
                .status();
            let _ = Command::new("git")
                .current_dir(&self.dir)
                .args(["clean", "-fd"])
                .status();
        }
    }
}
```

jolt-eval/src/objective/optimize.rs
```diff
@@ -0,0 +1,419 @@
...
+            best_measurements = new_measurements;
+            env.accept(iter);
+            let msg = format!(
+                "perf(auto-optimize): {} iteration {iter} (score {new_score:.10})",
+                objective.name,
+            );
+            let _ = Command::new("git")
+                .current_dir(repo_dir)
+                .args(["add", "-A"])
+                .status();
```
0xAndoroid: `git add -A` stages everything in the repo, not just files within the objective's `diff_paths()`. The agent runs with `--dangerously-skip-permissions` and can modify files outside scope — those get silently committed. Stage only the objective's allowed paths instead.

```suggestion
            let scope_paths = objective.diff_scope();
            let mut add_cmd = Command::new("git");
            add_cmd.current_dir(repo_dir).arg("add");
            if let crate::agent::DiffScope::Include(paths) = &scope_paths {
                add_cmd.arg("--");
                for p in paths {
                    add_cmd.arg(p);
                }
            } else {
                add_cmd.arg("-A");
            }
            let _ = add_cmd.status();
```

jolt-eval/src/invariant/mod.rs
```diff
@@ -0,0 +1,335 @@
...
+                                return Some(candidate.to_string());
+                            }
+                            break;
+                        }
+                    }
+                    _ => {}
+                }
+            }
+        }
+    }
```
0xAndoroid, line 269: The backward brace-scanning counts raw `{`/`}` bytes without handling braces inside JSON string values. For `{"key": "value with } brace"}`, the inner `}` throws off the depth counter. The serde validation guard prevents returning invalid JSON, but produces false negatives — `extract_json` returns `None` when valid JSON is present. This wastes red-team iterations when the agent produces JSON with brace characters in string values and doesn't use a ` ```json ` code fence.

The ` ```json ` path (lines 230-238) is unaffected, so this only matters for the fallback.

0xAndoroid (commented):
Review from 4 parallel analysis agents (semantic, bugs, tech debt, security), validated via PAL MCP (GPT-5.4). 6 issues found, all confirmed real.

jolt-eval/src/invariant/soundness.rs
```diff
@@ -0,0 +1,554 @@
...
+
+/// Remove diff hunks that reference paths containing `..` to prevent
+/// escaping the sandbox.
+pub fn filter_patch(patch: &str) -> String {
+    let mut result = String::new();
+    let mut include_hunk = true;
+
+    for line in patch.lines() {
+        if line.starts_with("diff --git") || line.starts_with("--- ") || line.starts_with("+++ ") {
```
0xAndoroid, line 334: Each `---`/`+++` line independently reassigns `include_hunk`. A crafted diff where `diff --git` has `..` but `+++ b/src/lib.rs` doesn't would flip `include_hunk` back to `true`. Mitigated by `git apply` rejecting structurally invalid diffs, but the filter intent is per-file-section, not per-header-line. The allowlist approach in the other comment resolves this too (only `diff --git` lines control state).

0xAndoroid (approved):
Maybe a claude skill in the .claude/skills of essentially adapted readme would help a bit with setting up these invariants, but otherwise this looks great!

PR #1414, opened 2026-04-07
feat/global-cse-opt

0xParti:
## Summary

Extends the transpiler's existing per-constraint CSE to global scope, detecting shared `TranscriptHash` chains across constraints and hoisting them into global variables. Previously each constraint had its own isolated CSE context, duplicating identical Poseidon hash computations. Also adds a Rust/Go cross-validation system for verifying codegen correctness. Builds on #1322 .

## Changes

- **Global CSE:** Detect shared `TranscriptHash` chains across constraints and hoist them into global variables, computed once and reused. Previously each constraint had isolated CSE, duplicating identical Poseidon hash computations.
- **Cross-validation system:** Compare intermediate assertion values (LHS/RHS) between a concrete AST evaluator in Rust and the generated gnark circuit in Go. Pass `--crossval` to auto-generate a cross-validation circuit and witness.
- **New files:** `ast_evaluator.rs` (concrete AST evaluation), `cross_validation.rs` (Rust test), `crossval_test.go` (Go test)
- **CseContext enum** instead of `usize::MAX` sentinel for CSE scope tracking
- **Dynamic assertion count** in `crossval_test.go`
- `.gitignore` for generated Groth16 artifacts
- Clippy and fmt fixes

## Testing

- [ ] cargo clippy and cargo fmt pass
- [ ] cargo test -p transpiler — cross-validation test
- [ ] cd transpiler/go && go test -v -run TestStagesCircuitProveVerify — Groth16 prove/verify
- [ ] cd transpiler/go && go test -v -run TestEndToEndPipeline -timeout 30m — full E2E pipeline

On the fibonacci test example, global CSE reduces constraints from ~2.7M to ~621K–677K. Proof size remains 164 bytes. Cross-validation Rust/Go passes on all intermediate values.

## Security Considerations

CSE is a pure optimization: it deduplicates constraint computation but does not change the constraint system's semantics. Each hoisted variable is assigned exactly once and referenced by the same constraints that previously computed it inline, so proof soundness is unaffected. The cross-validation system is test-only infrastructure and does not affect the proof pipeline.

## Breaking Changes

None

### DIFF
--- a/transpiler/src/ast_evaluator.rs
+++ b/transpiler/src/ast_evaluator.rs
@@ -29,7 +29,7 @@ fn eval_atom(atom: &Atom, witness: &HashMap<u16, Fr>) -> Fr {
         Atom::Scalar(limbs) => scalar_to_fr(limbs),
         Atom::Var(idx) => *witness
             .get(idx)
-            .unwrap_or_else(|| panic!("missing witness for Var({})", idx)),
+            .unwrap_or_else(|| panic!("missing witness for Var({idx})")),
         Atom::NamedVar(_) => panic!("NamedVar should not appear in serialized bundle"),
     }
 }
@@ -85,7 +85,9 @@ fn eval_node(
                     let data = eval_edge(data_edge, nodes, cache, witness);
                     let mut hasher =
                         Poseidon::<Fr>::new_circom(3).expect("failed to create Poseidon hasher");
-                    hasher.hash(&[state, rounds, data]).expect("Poseidon hash failed")
+                    hasher
+                        .hash(&[state, rounds, data])
+                        .expect("Poseidon hash failed")
                 }
                 _ => panic!("only Poseidon transcript is supported for evaluation"),
             }
--- a/transpiler/src/ast_evaluator.rs
+++ b/transpiler/src/ast_evaluator.rs
@@ -1,0 +1,232 @@
+//! Concrete evaluation of AST nodes using real field arithmetic.
+//!
+//! Walks the `AstBundle` node arena and evaluates each node to a concrete `Fr` value.
+//! Used for cross-validation: compare Rust AST evaluation against gnark circuit execution.
+
+use ark_bn254::Fr;
+use ark_ff::{Field, PrimeField};
+use light_poseidon::{Poseidon, PoseidonHasher};
+use std::collections::HashMap;
+use zklean_extractor::ast_bundle::Constraint;
+use zklean_extractor::mle_ast::{Atom, Edge, Node, NodeId, Scalar, TranscriptHashData};
+
+/// Evaluate an Edge to a concrete Fr value.
+fn eval_edge(
+    edge: &Edge,
+    nodes: &[Node],
+    cache: &mut HashMap<NodeId, Fr>,
+    witness: &HashMap<u16, Fr>,
+) -> Fr {
+    match edge {
+        Edge::Atom(atom) => eval_atom(atom, witness),
+        Edge::NodeRef(id) => eval_node(*id, nodes, cache, witness),
+    }
+}
+
+/// Evaluate an Atom to Fr.
+fn eval_atom(atom: &Atom, witness: &HashMap<u16, Fr>) -> Fr {
+    match atom {
+        Atom::Scalar(limbs) => scalar_to_fr(limbs),
+        Atom::Var(idx) => *witness
+            .get(idx)
+            .unwrap_or_else(|| panic!("missing witness for Var({})", idx)),
+        Atom::NamedVar(_) => panic!("NamedVar should not appear in serialized bundle"),
+    }
+}
+
+/// Convert a [u64; 4] scalar (raw, NOT Montgomery) to Fr.
+fn scalar_to_fr(limbs: &Scalar) -> Fr {
+    // Scalar is value-form (not Montgomery). Use from_le_bytes_mod_order.
+    let mut bytes = [0u8; 32];
+    for (i, limb) in limbs.iter().enumerate() {
+        bytes[i * 8..(i + 1) * 8].copy_from_slice(&limb.to_le_bytes());
+    }
+    Fr::from_le_bytes_mod_order(&bytes)
+}
+
+/// Evaluate a node, caching results.
+fn eval_node(
+    node_id: NodeId,
+    nodes: &[Node],
+    cache: &mut HashMap<NodeId, Fr>,
+    witness: &HashMap<u16, Fr>,
+) -> Fr {
+    if let Some(&val) = cache.get(&node_id) {
+        return val;
+    }
+
+    let result = match &nodes[node_id] {
+        Node::Atom(atom) => eval_atom(atom, witness),
+
+        Node::Add(a, b) => {
+            eval_edge(a, nodes, cache, witness) + eval_edge(b, nodes, cache, witness)
+        }
+        Node::Sub(a, b) => {
+            eval_edge(a, nodes, cache, witness) - eval_edge(b, nodes, cache, witness)
+        }
+        Node::Mul(a, b) => {
+            eval_edge(a, nodes, cache, witness) * eval_edge(b, nodes, cache, witness)
+        }
+        Node::Div(a, b) => {
+            let denom = eval_edge(b, nodes, cache, witness);
+            eval_edge(a, nodes, cache, witness) * denom.inverse().expect("div by zero")
+        }
+        Node::Neg(a) => -eval_edge(a, nodes, cache, witness),
+        Node::Inv(a) => eval_edge(a, nodes, cache, witness)
+            .inverse()
+            .expect("inv of zero"),
+
+        Node::TranscriptHash(hash_data, state_edge, rounds_edge) => {
+            let state = eval_edge(state_edge, nodes, cache, witness);
+            let rounds = eval_edge(rounds_edge, nodes, cache, witness);
+
+            match hash_data {
+                TranscriptHashData::Poseidon(data_edge) => {
+                    let data = eval_edge(data_edge, nodes, cache, witness);
+                    let mut hasher =
+                        Poseidon::<Fr>::new_circom(3).expect("failed to create Poseidon hasher");
+                    hasher.hash(&[state, rounds, data]).expect("Poseidon hash failed")
+                }
+                _ => panic!("only Poseidon transcript is supported for evaluation"),
+            }
+        }
+
+        Node::ByteReverse(e) => {
+            let val = eval_edge(e, nodes, cache, witness);
+            let bigint = val.into_bigint();
+            let mut bytes = [0u8; 32];
+            for (i, limb) in bigint.0.iter().enumerate() {
+                bytes[i * 8..(i + 1) * 8].copy_from_slice(&limb.to_le_bytes());
+            }
+            bytes.reverse();
+            Fr::from_le_bytes_mod_order(&bytes)
+        }
+
+        Node::Truncate128Reverse(e) => {
+            let val = eval_edge(e, nodes, cache, witness);
+            let bigint = val.into_bigint();
+            let mut le_bytes = [0u8; 32];
+            for (i, limb) in bigint.0.iter().enumerate() {
+                le_bytes[i * 8..(i + 1) * 8].copy_from_slice(&limb.to_le_bytes());
+            }
+            // Take low 16 bytes, reverse, interpret as field element, multiply by 2^128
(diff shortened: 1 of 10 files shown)


0xAndoroid:
Starting code review process via `/ci-code-review`. Will post findings shortly.

---

moodlezoup (commented):
I'm going to let Claude finish its review but this looks good to me. Thanks for adding the cross-validation tests

transpiler/src/gnark_codegen.rs
```diff
@@ -683,13 +792,110 @@ pub fn generate_circuit_from_bundle_with_stats(
...
+        }
+
+        assert_eq!(
+            global_codegen.cse_counter, num_global,
+            "global CSE counter ({}) != global bindings count ({}): index mapping would be inconsistent",
+            global_codegen.cse_counter, num_global
+        );
+
+        if global_codegen.uses_poseidon() {
+            stats.uses_poseidon = true;
```
0xAndoroid, line 814: This update happens after the import block has already been serialized into `output` (line ~752). If every Poseidon chain is shared across ≥2 constraints and ends up in `global_cse.bindings`, each per-constraint `generate_expr` resolves hash subtrees to `gcse[i]` pre-populated entries and never sets `uses_poseidon`, so the import check runs false. This block then flips `stats.uses_poseidon = true`, but the already-written `output` string is missing `"jolt_verifier/poseidon"` while `computeGlobalCse` emits `poseidon.Hash(...)` — the generated Go won't compile.

Fix: pre-scan `bundle.global_cse.bindings` for `TranscriptHash`/`ByteReverse`/`Truncate128*`/`AppendU64Transform` before emitting imports, or build the global CSE body into a separate buffer and inject after imports.

---

zklean-extractor/src/ast_bundle.rs
```diff
@@ -279,42 +294,162 @@ impl AstBundle {
...
+    }
+
+    /// Topological sort a subset of nodes in post-order (children before parents).
+    fn topological_sort_subset(&self, subset: &HashSet<NodeId>) -> Vec<NodeId> {
+        let mut result = Vec::new();
+        let mut visited: HashSet<NodeId> = HashSet::new();
+        let mut stack: Vec<(NodeId, bool)> = Vec::new();
+
+        // Start from all nodes in the subset
+        for &node_id in subset {
```
0xAndoroid, line 379: `subset` is a `HashSet<NodeId>` iterated here as DFS seeds. Rust's default `RandomState` randomizes iteration order across runs, so when the subset has disjoint sub-DAGs (multiple effective roots), the resulting post-order is a concatenation whose sub-sequence ordering is nondeterministic. `global_cse.bindings` positions — and therefore the `gcse[i]` indices in the generated Go — vary between invocations on identical inputs. Single-run consistency holds (same Vec feeds both `new_global` and `global_node_map`), but reproducible builds / golden diffs break.

```suggestion
        // Sort seeds for deterministic output across runs
        let mut seeds: Vec<NodeId> = subset.iter().copied().collect();
        seeds.sort_unstable();
        for node_id in seeds {
```

---

transpiler/src/ast_evaluator.rs
```diff
@@ -0,0 +1,234 @@
...
+            let bigint = val.into_bigint();
+            let x = bigint.0[0]; // u64 value
+            let swapped = x.swap_bytes();
+            Fr::from(swapped) * Fr::from(2u64).pow([192])
+        }
+    };
+
+    cache.insert(node_id, result);
+    result
+}
```
0xAndoroid, line 148: `eval_node`'s match adds a fifth site that must stay in sync with the exhaustive `Node` match arms in `gnark_codegen.rs::node_children` / `is_node_constant_in` / `evaluate_constant_node_in` and `ast_bundle.rs::node_children`. Adding a new `Node` variant requires updating all five. An inherent `Node::children()` method (plus an `Edge::as_node_ref()` helper) on the upstream enum would centralize coverage and let the compiler enforce it.

---

transpiler/src/ast_evaluator.rs
```diff
@@ -0,0 +1,234 @@
...
+
+        Node::TranscriptHash(hash_data, state_edge, rounds_edge) => {
+            let state = eval_edge(state_edge, nodes, cache, witness);
+            let rounds = eval_edge(rounds_edge, nodes, cache, witness);
+
+            match hash_data {
+                TranscriptHashData::Poseidon(data_edge) => {
+                    let data = eval_edge(data_edge, nodes, cache, witness);
+                    let mut hasher =
+                        Poseidon::<Fr>::new_circom(3).expect("failed to create Poseidon hasher");
```
0xAndoroid, line 87: The Rust side uses `light_poseidon::new_circom(3)`, the Go side uses `jolt_verifier/poseidon`. The only test pinning the Rust evaluator (`test_poseidon_hash_zeros` at line 225) just asserts the output is non-zero — it doesn't check against a Go-produced reference vector. If constants/MDS/round counts drift between the two packages, the cross-validation harness would silently pass on divergent witnesses until somebody noticed a proving failure. Consider hardcoding a Go-computed hash vector into the Rust tests so the two implementations are independently pinned.

---

transpiler/src/ast_evaluator.rs
```diff
@@ -0,0 +1,234 @@
...
+                let l = eval_edge(lhs_edge, nodes, &mut cache, witness);
+                let r = eval_edge(rhs_edge, nodes, &mut cache, witness);
+                (l, r)
+            }
+            // sum_zero or other: the whole expression should be 0
+            _ => {
+                let val = eval_node(constraint.root, nodes, &mut cache, witness);
+                (val, Fr::from(0u64))
+            }
+        };
```
0xAndoroid, line 185: `evaluate_assertions` branches only on the root node shape (`Sub` vs other) and ignores `constraint.assertion`. For `Assertion::EqualPublicInput` / `EqualNode` constraints, `RHS` would be set to 0 instead of the public-input value / other-node value. Currently safe because `main.rs:334` only emits `add_constraint_eq_zero`, but the doc comment here describes a broader contract that the implementation doesn't honor. Either match on `constraint.assertion` to produce the right RHS, or narrow the doc to "only EqualZero is supported".

---

transpiler/src/gnark_codegen.rs
```diff
@@ -211,14 +223,45 @@ impl<'a> GnarkCodeGen<'a> {
             ref_counts.insert(node_id, 2);
         }
 
+        // Pre-populate `generated` for global CSE nodes so they resolve to gcse[i]
+        let mut generated = HashMap::new();
+        for (&node_id, &gcse_idx) in global_node_map {
+            generated.insert(node_id, format!("gcse[{gcse_idx}]"));
+        }
```
0xAndoroid, line 230: Pre-populating `generated` with `gcse[i]` entries works, but `generate_expr` phase 1 (lines 313-318) still recurses into children of these nodes unconditionally. Phase 2 correctly skips the pre-populated entries themselves, but all non-global descendants end up with expression strings in `self.generated` — some large enough to trigger `MAX_INLINE_EXPR_LEN` and become dead `cse_K_N` bindings nothing references. Short-circuiting the phase-1 descent with `if self.generated.contains_key(&node_id) { /* no children */ }` would prune the dead work.

---

transpiler/src/gnark_codegen.rs
```diff
@@ -701,7 +907,9 @@ pub fn generate_circuit_from_bundle_with_stats(
                     pc.name
                 ));
                 stats.constant_skipped += 1;
-                continue;
+                if !crossval {
+                    continue;
```
0xAndoroid, line 911: The `val != 0` branch a few lines below falls through to emit the full `verifyConstraint{idx}` function, but `generate_circuit_from_bundle` panics at `stats.constant_failed > 0` before returning the code. So the emitted function is dead in the panicking path — only `_with_stats` callers ever see it, and the transpiler CLI doesn't use that. Either skip emission mirroring the `val == 0` path or document that the function is intentionally preserved for stats-only consumers.

---

transpiler/src/gnark_codegen.rs
```diff
@@ -846,36 +1089,11 @@ fn rewrite_cse_to_slice(line: &str, constraint_idx: usize) -> String {
         }
     }
```
0xAndoroid, line 1091: String-based `rewrite_cse_names_to_slice` is safe today (witness names are PascalCased as `Cse_*`, literals are pure digits inside `bigInt("...")`, the rewriter scans for lowercase `cse_{K}_`), but it's a fragile contract to maintain. A future change to the sanitization rule, CSE naming scheme, or the addition of a comment containing `cse_3_5` would silently corrupt generated Go. Keeping bindings as structured records (`{ idx, expr, hoisted_as }`) and formatting at the end would eliminate the string-rewriting step entirely.

---

0xAndoroid (commented):
Automated review of the global CSE change. The key correctness finding: if every `TranscriptHash` gets globally hoisted, the per-constraint codegen never sets `stats.uses_poseidon`, but the `poseidon` import is decided before the global CSE block runs — the generated Go will reference `poseidon.Hash` without importing it. Also flagging non-deterministic `gcse[i]` ordering from `HashSet`-seeded topological sort (reproducibility), plus a few tech-debt / test-hygiene notes.

---

transpiler/src/gnark_codegen.rs
akoidefi: Good catch. Took the buffer option: global CSE is rendered first to a `String`, which sets `stats.uses_poseidon`, and only then we write imports and inject the buffer. Regression test `test_poseidon_import_emitted_when_all_hoisted` pins the behaviour.

zklean-extractor/src/ast_bundle.rs
akoidefi: Fixed already on the `feat/universal-circuit` branch we maintain in our fork, where deterministic GCSE is required for byte-identical cross-program circuits. The backport to this PR didn't happen at the time. Applied your suggestion verbatim plus a regression test that rebuilds the bundle 30 times and asserts the bindings vector is stable.

transpiler/src/ast_evaluator.rs
akoidefi: Added `Node::child_node_ids` on the enum as single source of truth. `ast_bundle.rs::node_children` now delegates to it. The two matches in `gnark_codegen.rs` (`is_node_constant_in`, `evaluate_constant_node_in`) and the one in `ast_evaluator.rs::eval_node` stay because they do semantic work, not traversal. The pure traversal duplication is gone.

transpiler/src/ast_evaluator.rs
akoidefi: Done. `test_poseidon_hash_zeros` replaced with `test_light_poseidon_output_pinned`, which pin the decimal outputs for `[0,0,0]` and `[1,2,3]`. Same decimals are pinned on the Go side in `transpiler/go/poseidon/poseidon_test.go`

transpiler/src/ast_evaluator.rs
akoidefi: Added an explicit match. `EqualZero` keeps the old path; `EqualPublicInput` and `EqualNode` `panic!` with a message that name the file to update. Two `#[should_panic]` tests pin this

transpiler/src/gnark_codegen.rs
akoidefi: Done exactly as suggested. Phase 1 now checks `self.generated.contains_key` before pushing children on the stack.

transpiler/src/gnark_codegen.rs
akoidefi: Skip it. `val != 0` now `continue`s when `!crossval`, same as `val == 0`. Under `crossval = true` we keep the body so the cross-validation harness can inspect the failing expression. `test_constant_failure_skips_function_emission` pins both halves.

transpiler/src/gnark_codegen.rs
akoidefi: Went with the structural refactor. Expressions are now `Vec<ExprFragment>` where fragments are `Lit(String)`, `CseRef(usize)` or `GcseRef(usize)`. Rendering happens once at the end in `render_expr`, which picks `cse_K_N`, `cse[N]` or `gcse[N]` by context. `rewrite_cse_names_to_slice` is deleted. Touches ~600 lines of `gnark_codegen.rs` because `generated`, `bindings` and `ProcessedConstraint` all move to `Expr`. Three renderer unit tests plus the E2E suite cover the change.

PR #1428, opened 2026-04-10
test(jolt-transcript): add DigestTranscript regression coverage

umeeSthein:
## Summary

Add focused regression tests for the generic `DigestTranscript` implementation introduced in the recent transcript refactor.

These tests cover a few generic behaviors that were not exercised directly before:
- `challenge_vector` advances the internal round counter correctly
- `compare_to` accepts matching transcript histories
- `compare_to` detects divergence and panics as expected

The coverage is kept intentionally narrow to avoid duplicating the broader alias-level tests that already exist for `Blake2bTranscript` and `KeccakTranscript`.


0xAndoroid:
The tests proposed are redundant.

PR #1430, opened 2026-04-12
perf: unreduced accumulation in D=4 sum-of-products

mw2000:
## Summary

- Replaces the macro-generated `compute_mles_product_sum_evals_sum_of_products_d4` with a dedicated implementation that uses `eval_prod_4_accumulate` to defer Montgomery reductions across product terms
- The 4 final pointwise multiplies stay as `UnreducedProductAccum` and are reduced once after all `n_products` terms are summed, saving one Barrett reduction per output lane per product term
- Also adds `eval_prod_4_accumulate` to the `eval_linear_prod_accumulate` dispatch (used by `InstructionReadRafSumcheckProver`)

## Context

The D=4 sum-of-products path is the hot inner loop of `InstructionRaSumcheckProver::compute_message` in stage 6 (the RA virtual sumcheck). For the fibonacci benchmark, this is called 22 times with `n_products=8`, totaling ~1.3s of the ~3.5s stage 6.

The existing code used `eval_prod_4_assign` which writes fully reduced `F` values, then accumulates them with `+=`. The new path uses `eval_prod_4_accumulate` which writes unreduced u128 limbs via `mul_to_product_accum`, deferring the Barrett reduction until after all product terms are summed.

## Benchmark

`sum_of_products_d4_n8` microbenchmark (matching the fibonacci hot path config):
- Baseline: 2.30ms
- Optimized: 2.12ms
- **~6% improvement**

### DIFF
--- a/jolt-core/benches/mles_product_sum.rs
+++ b/jolt-core/benches/mles_product_sum.rs
@@ -10,7 +10,8 @@ use jolt_core::{
         unipoly::UniPoly,
     },
     subprotocols::mles_product_sum::{
-        compute_mles_product_sum, eval_linear_prod_naive_assign, finish_mles_product_sum_from_evals,
+        compute_mles_product_sum, compute_mles_product_sum_evals_sum_of_products_d4,
+        eval_linear_prod_naive_assign, finish_mles_product_sum_from_evals,
     },
 };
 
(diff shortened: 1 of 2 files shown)


0xAndoroid (approved):
nice, thanks

PR #1439, opened 2026-04-15
fix: address ZeroOS security review findings

sagar-a16z:
## Summary

Soundness fixes:
- P-256 ECDSA: reject trivial zero GLV decomposition (a=0, b=0) that collapses the Shamir MSM and allows signature forgery for any message and public key. secp256k1 is not affected.
- rd=x0 allowlist: exclude SCW/SCD/LRW/LRD/INLINE from the generic rd=x0 trace rewriting path so their custom trace() methods run, preserving VirtualAdvice patching and reservation state management.

Completeness fix:
- EBREAK: expand into a JAL-to-self virtual sequence so the Jump flag disables the NextUnexpPCUpdateOtherwise constraint at the NoOp padding boundary.

Emulator fix:
- MULHSU exec(): sign-extend rs1 via `as i128 as u128` instead of zero-extending via `as u128`. Only affects non-tracing emulator mode; proof generation uses the correct inline_sequence path.

Documentation:
- Document the M-mode-only privilege and trap model in the book and in ecall.rs/mret.rs code comments. ECALL's constant mstatus write and MRET's omitted mstatus restoration are correct by design for the single-privilege, no-interrupt execution model.

## Testing

## Breaking Changes

None

### DIFF
--- a/jolt-inlines/p256/src/sdk.rs
+++ b/jolt-inlines/p256/src/sdk.rs
@@ -827,7 +827,7 @@ pub fn ecdsa_verify(z: P256Fr, r: P256Fr, s: P256Fr, q: P256Point) -> Result<(),
     if q.is_infinity() {
         return Err(P256Error::QAtInfinity);
     }
-    if r.is_zero() || s.is_zero() {
+    if r.is_zero() || s.is_zero() || z.is_zero() {
         return Err(P256Error::ROrSZero);
     }
 
--- a/jolt-inlines/p256/src/sdk.rs
+++ b/jolt-inlines/p256/src/sdk.rs
@@ -842,6 +842,34 @@ pub fn ecdsa_verify(z: P256Fr, r: P256Fr, s: P256Fr, q: P256Point) -> Result<(),
     // Step 3: Get R2 = u2*Q and decomposition via Fake GLV advice
     let (r2, a2_val, a2_sign, b2_val, b2_sign) = fake_glv_scalar_mul(&u2, &q);
 
+    verify_ecdsa_inner(
+        &u1, &u2, &r, &q, r1, a1_val, a1_sign, b1_val, b1_sign, r2, a2_val, a2_sign, b2_val,
+        b2_sign,
+    )
+}
+
+/// Inner verification logic, separated from `ecdsa_verify` so it can be tested
+/// with injected decomposition values (e.g. zero GLV attack vectors).
+#[expect(clippy::too_many_arguments)]
+#[inline(always)]
+pub(crate) fn verify_ecdsa_inner(
+    u1: &P256Fr,
+    u2: &P256Fr,
+    r: &P256Fr,
+    q: &P256Point,
+    r1: P256Point,
+    a1_val: u128,
+    a1_sign: bool,
+    b1_val: u128,
+    b1_sign: bool,
+    r2: P256Point,
+    a2_val: u128,
+    a2_sign: bool,
+    b2_val: u128,
+    b2_sign: bool,
+) -> Result<(), P256Error> {
+    let g = P256Point::generator();
+
     // Step 4: Verify R1, R2 are on curve
     if !r1.is_on_curve() {
         spoil_proof();
@@ -868,12 +896,12 @@ pub fn ecdsa_verify(z: P256Fr, r: P256Fr, s: P256Fr, q: P256Point) -> Result<(),
     let b2_fr = make_fr(b2_val, b2_sign);
 
     // Check b1*u1 = a1 (mod n)
-    let check1 = b1_fr.mul(&u1);
+    let check1 = b1_fr.mul(u1);
     if check1.e() != a1_fr.e() {
         spoil_proof();
     }
     // Check b2*u2 = a2 (mod n)
-    let check2 = b2_fr.mul(&u2);
+    let check2 = b2_fr.mul(u2);
     if check2.e() != a2_fr.e() {
         spoil_proof();
     }
@@ -896,7 +924,7 @@ pub fn ecdsa_verify(z: P256Fr, r: P256Fr, s: P256Fr, q: P256Point) -> Result<(),
     let scalars = [a1_val, a2_val, b1_val, b2_val];
     let points_arr = [
         if a1_sign { g.neg() } else { g },
-        if a2_sign { q.neg() } else { q },
+        if a2_sign { q.neg() } else { q.clone() },
         r1_adj,
         r2_adj,
     ];
--- a/jolt-inlines/p256/src/sdk.rs
+++ b/jolt-inlines/p256/src/sdk.rs
@@ -878,6 +878,12 @@ pub fn ecdsa_verify(z: P256Fr, r: P256Fr, s: P256Fr, q: P256Point) -> Result<(),
         spoil_proof();
     }
 
+    // Reject trivial zero decomposition: a=0, b=0 satisfies b*u=a for any u,
+    // collapsing the Shamir MSM and leaving R1/R2 unconstrained.
+    if b1_val == 0 || b2_val == 0 {
+        spoil_proof();
+    }
+
     // Step 6: Prepare points for 4-scalar 128-bit Shamir
     // We verify: a1*G + a2*Q - b1*R1 - b2*R2 = O
     // Which is: a1*G + a2*Q + (-b1)*R1 + (-b2)*R2 = O
--- a/tracer/src/instruction/ebreak.rs
+++ b/tracer/src/instruction/ebreak.rs
@@ -55,6 +55,7 @@ impl RISCVTrace for EBREAK {
         let mut asm = InstrAssembler::new(self.address, self.is_compressed, xlen, allocator);
         let vr = allocator.allocate();
         asm.emit_j::<JAL>(*vr, 0);
+        drop(vr);
         asm.finalize()
     }
 }
--- a/tracer/src/instruction/ebreak.rs
+++ b/tracer/src/instruction/ebreak.rs
@@ -7,9 +7,16 @@
 
 use serde::{Deserialize, Serialize};
 
-use crate::{declare_riscv_instr, emulator::cpu::Cpu};
+use crate::{
+    declare_riscv_instr,
+    emulator::cpu::{Cpu, Xlen},
+    utils::inline_helpers::InstrAssembler,
+    utils::virtual_registers::VirtualRegisterAllocator,
+};
 
-use super::{format::format_i::FormatI, RISCVInstruction, RISCVTrace};
+use super::{
+    format::format_i::FormatI, jal::JAL, Cycle, Instruction, RISCVInstruction, RISCVTrace,
+};
 
 declare_riscv_instr!(
     name   = EBREAK,
@@ -27,4 +34,27 @@ impl EBREAK {
     }
 }
 
-impl RISCVTrace for EBREAK {}
+impl RISCVTrace for EBREAK {
+    fn trace(&self, cpu: &mut Cpu, trace: Option<&mut Vec<Cycle>>) {
+        let inline_sequence = self.inline_sequence(&cpu.vr_allocator, cpu.xlen);
+        let mut trace = trace;
+        for instr in inline_sequence {
+            instr.trace(cpu, trace.as_deref_mut());
+        }
+    }
+
+    /// Expand EBREAK into a JAL-to-self (`j .`), which stalls the PC and
+    /// terminates execution. Using JAL gives the cycle a Jump flag, which
+    /// disables the NextUnexpPCUpdateOtherwise constraint at the NoOp
+    /// padding boundary.
+    fn inline_sequence(
+        &self,
+        allocator: &VirtualRegisterAllocator,
+        xlen: Xlen,
+    ) -> Vec<Instruction> {
+        let mut asm = InstrAssembler::new(self.address, self.is_compressed, xlen, allocator);
+        let vr = allocator.allocate();
+        asm.emit_j::<JAL>(*vr, 0);
+        asm.finalize()
+    }
+}
--- a/tracer/src/instruction/mod.rs
+++ b/tracer/src/instruction/mod.rs
@@ -552,7 +552,16 @@ macro_rules! define_rv32im_enums {
                 let normalized = self.normalize();
                 // Rewrite instructions with rd=x0 via inline_sequence so the
                 // constraint system never sees rd=x0.
-                if normalized.operands.rd == Some(0) {
+                if normalized.operands.rd == Some(0)
+                    && !matches!(
+                        self,
+                        Instruction::SCW(_)
+                            | Instruction::SCD(_)
+                            | Instruction::LRW(_)
+                            | Instruction::LRD(_)
+                            | Instruction::INLINE(_)
+                    )
+                {
                     let inline_sequence = self.inline_sequence(&cpu.vr_allocator, cpu.xlen);
                     let mut trace = trace;
                     for instr in inline_sequence {
--- a/tracer/src/instruction/mulhsu.rs
+++ b/tracer/src/instruction/mulhsu.rs
@@ -31,7 +31,7 @@ impl MULHSU {
                         >> 32,
                 ),
                 Xlen::Bit64 => {
-                    ((cpu.x[self.operands.rs1 as usize] as u128)
+                    ((cpu.x[self.operands.rs1 as usize] as i128 as u128)
                         .wrapping_mul(cpu.x[self.operands.rs2 as usize] as u64 as u128)
                         >> 64) as i64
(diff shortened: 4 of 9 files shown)


0xAndoroid:
🔍 Starting automated code review process. I'll analyze the changes and post my findings shortly.

---

jolt-inlines/p256/src/sdk.rs
```diff
@@ -878,6 +878,12 @@ pub fn ecdsa_verify(z: P256Fr, r: P256Fr, s: P256Fr, q: P256Point) -> Result<(),
         spoil_proof();
     }
 
+    // Reject trivial zero decomposition: a=0, b=0 satisfies b*u=a for any u,
+    // collapsing the Shamir MSM and leaving R1/R2 unconstrained.
+    if b1_val == 0 || b2_val == 0 {
+        spoil_proof();
+    }
```
0xAndoroid, line 913: **Soundness question:** The check rejects `b1_val == 0 || b2_val == 0`, but should `a1_val` and `a2_val` also be checked?

If `a1_val == 0`, then the generator `G` term drops out of the Shamir MSM (`a1*G + a2*Q - b1*R1 - b2*R2 = O`), which could weaken the verification. Similarly for `a2_val == 0` dropping the public key `Q` term.

That said, `a=0, b=0` is already caught by the `b` check since the decomposition relation `b*u = a` with `b=0` forces `a=0`. But is it possible to have `a=0, b≠0`? That would require `b*u ≡ 0 (mod n)` with `b≠0`, which is impossible for prime `n`. So the `b` check alone appears sufficient. Just want to confirm this reasoning is correct for P-256's group order.

---

tracer/src/instruction/ebreak.rs
```diff
@@ -27,4 +34,27 @@ impl EBREAK {
...
+    /// disables the NextUnexpPCUpdateOtherwise constraint at the NoOp
+    /// padding boundary.
+    fn inline_sequence(
+        &self,
+        allocator: &VirtualRegisterAllocator,
+        xlen: Xlen,
+    ) -> Vec<Instruction> {
+        let mut asm = InstrAssembler::new(self.address, self.is_compressed, xlen, allocator);
+        let vr = allocator.allocate();
+        asm.emit_j::<JAL>(*vr, 0);
```
0xAndoroid, line 57: **Minor concern — leaked virtual register:** `allocator.allocate()` returns a virtual register that is stored in `vr`, but it's never `drop()`-ed before `asm.finalize()`. Other inline sequences in the codebase (e.g., ECALL, MRET) explicitly `drop()` temporaries to return them to the allocator pool.

Is this intentional because EBREAK terminates execution so the allocator pool doesn't matter afterward? If so, a brief comment explaining that would be helpful. If not, consider adding `drop(vr)` before `finalize()`.

---

tracer/src/instruction/mod.rs
```diff
@@ -552,7 +552,16 @@ macro_rules! define_rv32im_enums {
...
-                if normalized.operands.rd == Some(0) {
+                if normalized.operands.rd == Some(0)
+                    && !matches!(
+                        self,
+                        Instruction::SCW(_)
+                            | Instruction::SCD(_)
+                            | Instruction::LRW(_)
+                            | Instruction::LRD(_)
+                            | Instruction::INLINE(_)
+                    )
```
0xAndoroid, line 563: **Correctness: Should EBREAK also be in this exclusion list?**

EBREAK now has its own custom `trace()` implementation (added in this same PR) that calls `self.inline_sequence()` directly. If EBREAK ever has `rd == Some(0)` in its normalized operands, the generic rd=x0 path here would intercept it before the custom `trace()` runs, bypassing the JAL-to-self expansion.

Looking at `declare_riscv_instr` for EBREAK with `format = FormatI`, the decoded operands may include `rd`. If the EBREAK encoding (0x00100073) decodes with rd=0 (bits 11:7 are all zero in the EBREAK encoding), then this generic path *would* intercept it.

Can you confirm whether EBREAK's normalized rd is `Some(0)`? If so, it needs to be added to this exclusion list for the EBREAK fix to actually take effect.

---

tracer/src/instruction/mulhsu.rs
```diff
@@ -31,7 +31,7 @@ impl MULHSU {
                         >> 32,
                 ),
                 Xlen::Bit64 => {
-                    ((cpu.x[self.operands.rs1 as usize] as u128)
+                    ((cpu.x[self.operands.rs1 as usize] as i128 as u128)
```
0xAndoroid, line 34: Fix looks correct. `rs1` is the signed operand in MULHSU, so it must be sign-extended via `as i128` (which extends the sign bit of the `i64`) before converting to `u128` for the unsigned multiply. The previous `as u128` zero-extended, which treated negative `rs1` values as large positive numbers.

Confirmed this matches the pattern in `mulh.rs` (both operands signed → both use `as i128`) and `mulhu.rs` (both unsigned → both use `as u64 as u128`).

---

jolt-inlines/p256/src/sdk.rs
0xAndoroid, line 913: **Testing gap:** There doesn't appear to be a test that exercises the `b1_val == 0 || b2_val == 0` rejection path. Consider adding a targeted test that crafts a zero GLV decomposition to verify `spoil_proof()` is reached. This would guard against future regressions in this critical soundness check.

---

0xAndoroid (commented):
## Code Review Summary

This PR addresses findings from the ZeroOS security review. The changes cover 4 functional fixes (2 soundness, 1 completeness, 1 emulator correctness) and documentation improvements. Overall, the fixes are well-targeted and the code quality is high.

### P-256 ECDSA: Zero GLV Decomposition Rejection (`jolt-inlines/p256/src/sdk.rs`)

**Verdict: Correct, critical fix.**

The zero decomposition (`b1=0, b2=0`) trivially satisfies `b*u = a (mod n)` for any `u`, collapsing the 4-scalar Shamir MSM and leaving R1/R2 unconstrained — enabling signature forgery. Checking only `b` values (not `a`) is sufficient: for prime group order `n`, `b*u ≡ 0 (mod n)` with `b≠0` is impossible, so `b=0 ⟹ a=0`. secp256k1 is correctly noted as unaffected — its recomposition check (`k1 + k2*lambda == k`) implicitly rejects the zero case.

**Suggestion:** Add a regression test that crafts a zero decomposition to verify `spoil_proof()` fires.

### rd=x0 Allowlist (`tracer/src/instruction/mod.rs`)

**Verdict: Correct, necessary fix.**

SCW/SCD custom `trace()` methods inject VirtualAdvice success flags and manage reservation state (clearing both word and doubleword widths). LRW/LRD custom `trace()` methods set reservations before executing their inline sequences. INLINE has its own rd=x0 remapping. The generic rd=x0 rewriting path bypasses all of this, breaking correctness.

Note: EBREAK is *not* in the exclusion list, but this appears intentional and correct — the generic rd=x0 path also calls `inline_sequence()`, which now returns the JAL-to-self, so both paths produce the same result.

### EBREAK Completeness Fix (`tracer/src/instruction/ebreak.rs`)

**Verdict: Correct.**

Expanding EBREAK into `JAL vr, 0` (jump-to-self) sets the Jump flag on the cycle, which disables the `NextUnexpPCUpdateOtherwise` R1CS constraint at the EBREAK→NoOp padding boundary. Without this, the constraint would enforce `NextPC == PC + 4` at the termination point, conflicting with EBREAK's PC-stalling behavior.

Minor nit: the allocated virtual register `vr` is not explicitly dropped before `finalize()` — this is likely fine since EBREAK terminates execution, but could be noted with a comment.

### MULHSU Sign Extension Fix (`tracer/src/instruction/mulhsu.rs`)

**Verdict: Correct, straightforward fix.**

`rs1` is the signed operand in MULHSU. The old cast `as u128` zero-extended, treating negative `rs1` as large positive values. The fix `as i128 as u128` correctly sign-extends first. This matches the pattern used in `mulh.rs` (both signed) and `mulhu.rs` (both unsigned). Only affects the non-tracing emulator path; the inline_sequence proof path was already correct.

### Documentation (`book/`, `ecall.rs`, `mret.rs`)

**Verdict: Good.**

The privilege model documentation is thorough and well-placed. The M-mode-only / no-interrupt justifications for the simplified ECALL mstatus write and MRET omission are clearly explained. Consolidating the rationale into module-level docs (rather than scattered inline comments) improves maintainability.

---

jolt-inlines/p256/src/sdk.rs
sagar-a16z: Your reasoning is correct. P-256's group order n is prime (n = 0xFFFFFFFF00000000FFFFFFFFFFFFFFFFBCE6FAADA7179E84F3B9CAC2FC632551), so the ring Z/nZ is a field and every
  nonzero element is invertible.

  The decomposition check enforces `b * u ≡ a (mod n)`. If `b ≠ 0 (mod n)` and `u ≠ 0 (mod n)`, then `b * u` is necessarily nonzero (no zero divisors in a field), so `a` must
  also be nonzero. The only way to get `a = 0` is `b = 0` or `u = 0`.

  Since `u1 = z/s` and `u2 = r/s` with `r, s ≠ 0` (checked at line 830-831), both `u` values are nonzero. And `b_val` is a `u128` magnitude (always < n), so `b_val == 0` is
  equivalent to `b ≡ 0 (mod n)`.

  Therefore checking `b` alone is sufficient — `a=0, b≠0` is algebraically impossible.

tracer/src/instruction/ebreak.rs
sagar-a16z: Good catch — fixed in d2126bf. Added `drop(vr)` before `finalize()` to return it to the allocator pool, matching the pattern in ECALL (ecall.rs:83) and MRET (mret.rs:85).

  Strictly speaking it doesn't matter here since EBREAK terminates execution and no subsequent instruction will need a virtual register, but we should be clean and consistent

tracer/src/instruction/mod.rs
sagar-a16z: EBREAK does have `rd == Some(0)` — encoding 0x00100073 has bits [11:7] = 0. So yes, the generic rd=x0 path at line 555 intercepts it. But this is fine because the path works
  correctly for EBREAK:

  1. `Instruction::trace()` (line 565): calls `self.inline_sequence()`
  2. `Instruction::inline_sequence()` (line 657): EBREAK **is in the allowlist** at line 660, so it calls `dispatch_inline_sequence()`
  3. `dispatch_inline_sequence()`: calls `EBREAK::inline_sequence()` → returns the JAL-to-self
  4. Back in `Instruction::trace()` (line 567): `trace_raw()` on the JAL

  This produces the same result as calling `EBREAK::trace()` directly — both paths call `EBREAK::inline_sequence()` and trace the resulting JAL. The exclusion list at line
  556-563 is specifically for instructions whose custom `trace()` methods do CPU state manipulation beyond what `inline_sequence()` captures (reservation set/clear for LR/SC,
  advice patching for INLINE). EBREAK's custom `trace()` doesn't do any state manipulation — it just iterates the inline sequence, same as the generic path.

  No change needed.

jolt-inlines/p256/src/sdk.rs
sagar-a16z: Added

0xAndoroid:
Starting automated code review process. Will review the changes and post findings shortly.

---

0xAndoroid:
$REVIEW_BODY


---

0xAndoroid:
Lol, the workflow broke because Opus 4.7 got released

0xAndoroid:
Starting code review process...

---

0xAndoroid:
## Code Review: ZeroOS security review findings

### Overview

This PR addresses findings from the ZeroOS security review with four categories of changes:
1. **Soundness (P-256 ECDSA)** — reject trivial zero GLV decomposition (`a=0, b=0`) that would otherwise permit signature forgery.
2. **Completeness (EBREAK)** — expand into a `JAL`-to-self virtual sequence so the `Jump` flag disables `NextUnexpPCUpdateOtherwise` at the NoOp padding boundary.
3. **Completeness (rd=x0 allowlist)** — exclude `SCW/SCD/LRW/LRD/INLINE` from the generic rd=x0 trace-rewriting path so their custom `trace()` methods run.
4. **Emulator-only (MULHSU)** — sign-extend `rs1` via `as i128 as u128` rather than zero-extending via `as u128`.
5. **Documentation** — privilege / trap model for the M-mode-only ZeroOS runtime (book + module-level docs on `ecall.rs`/`mret.rs`).

### Code correctness

- **`jolt-inlines/p256/src/sdk.rs`** — the refactor into `verify_ecdsa_inner` is clean and makes the fix unit-testable. The `b1_val == 0 || b2_val == 0` guard correctly closes the attack: with `b=0`, the decomposition check `b·u = a` is trivially satisfied and the Shamir MSM coefficients collapse, leaving `R1`/`R2` unconstrained. The signature change from `q: P256Point` → `q: &P256Point` requires the added `q.clone()` in the neg/no-neg branch (line 927); that is correct.
- **`tracer/src/instruction/ebreak.rs`** — allocating a VR for `JAL`'s `rd` bypasses the rd=x0 rewrite rule (EBREAK is not on the new allowlist). The VR value is captured as an index in the emitted `Instruction`; `drop(vr)` after `emit_j` is therefore sound. Good.
- **`tracer/src/instruction/mod.rs`** — the allowlist correctly preserves LR/SC reservation-state management and INLINE's expansion pipeline. Worth confirming no other instruction with a custom `trace()` relies on being visible at rd=x0.
- **`tracer/src/instruction/mulhsu.rs`** — `cpu.x` is `[i64; 32]` (`emulator/cpu.rs:168`), so the original `as u128` was zero-extending a signed value. `as i128 as u128` correctly sign-extends. Note this is RV64 only; the RV32 branch already uses `as i64 as i128` and is fine.

### Potential issues / suggestions

1. **Defense-in-depth for ECDSA zero checks (P-256)** — the fix checks `b_val == 0`. The symmetric case `a_val == 0, b_val ≠ 0` requires `u = 0` (since `n` is prime), which requires either `z = 0` or `r = 0`. `r.is_zero()` and `s.is_zero()` are checked at `sdk.rs:830`, but **`z.is_zero()` is not**. While a hash output of zero is cryptographically implausible, an explicit `z.is_zero()` guard (or equivalently `a1_val == 0 || a2_val == 0`) would be cheap and would close the last theoretical edge case. Consider adding it.
2. **EBREAK inline sequence** — confirm that allocating a VR inside `inline_sequence()` and dropping it before `finalize()` returns is the intended lifetime. Other inline sequences in the tracer tend to keep VRs alive across `finalize()`; the pattern here is uncommon but appears correct because `emit_j` copies the register index into the emitted `Instruction`.
3. **Consistency of the allowlist** — the exclusion uses `matches!` on `Instruction` variants inside a macro. If new variants with custom `trace()` are added (e.g. future atomics), they'll silently fall through the rewrite and may break. A longer-term fix would be a trait-level `skips_rd_x0_rewrite()` default method overridden by the affected types.

### Test coverage

- **Good:** `test_zero_glv_decomposition_rejected` exercises the exact attack vector via `verify_ecdsa_inner`, with `R1 = G`, `R2 = ∞`. The `#[should_panic(expected = "proof spoiled")]` hook relies on `spoil_proof()` panicking — that matches the current implementation.
- **Missing:** no integration test for EBREAK in the proving pipeline (i.e. a guest that actually executes `ebreak` and is proven end-to-end). Unit-level inline_sequence assertions would be cheap.
- **Missing:** no test exercising `SCW/SCD/LRW/LRD` with `rd=x0` in a trace to confirm the allowlist preserves reservation semantics.
- **Missing:** no unit test for `MULHSU` with a negative `rs1` in non-tracing mode (the exact bug the fix addresses).

Per `CLAUDE.md`, please also confirm `cargo nextest run -p jolt-core muldiv --features host` and `--features host,zk` both pass.

### Performance

No meaningful impact. The P-256 guard is two `u128` comparisons, and the rd=x0 `matches!` adds a handful of discriminant checks per rewrite — both negligible.

### Security

The core soundness fix is correct and well-scoped. secp256k1 is unaffected, consistent with the PR description (different scalar-mul path). Recommend adding the `z.is_zero()` guard noted above for defense-in-depth.

### Documentation

The M-mode-only privilege/trap model write-up in `emulation.md` and the module-level docs on `ecall.rs`/`mret.rs` are clear and make the ECALL-constant-`mstatus` and MRET-no-`mstatus`-restore choices auditable. Nit: `emulation.md` mentions `vr34–vr39` for CSR virtual registers — worth double-checking that range matches the allocator definition to avoid doc drift.

### Summary

LGTM pending (a) a quick confirmation on the `z.is_zero()` defense-in-depth question and (b) the `muldiv` e2e test passing in both standard and ZK modes.


---

PR #1442, opened 2026-04-16
Fix audit findings from 2026-04-15 report

0xAndoroid:
## Summary

Addresses Justin Thaler's 2026-04-15 audit notes (with Claude assistance, delivered `jolt-audit-report.html`). Report header: **"No soundness vulnerabilities found."** All five findings were completeness / hardening / crypto-hygiene items.

Three fixed, one documented, two deferred with rationale. See [`AUDIT_ANALYSIS.md`](./AUDIT_ANALYSIS.md) for per-finding reasoning.

### Fixed

- **Finding 2 — `has_side_effects()` default (tracer).** Removed the trait default body so every direct impl must declare its side-effect status. `VirtualAdvice` gets an explicit `false`. Prevents a future instruction from silently being replaced with a NOP under the `rd = x0` lowering.
- **Finding 3 — RAM ValCheck alignment `debug_assert` (jolt-core/zkvm/ram).** Promoted the prover + verifier alignment check between `RamVal` and `RamValFinal` address points to runtime `assert_eq!`. One check per proof — cost is negligible, invariant is load-bearing.

### Documented (not changed)

- **Finding 1 — LD/SD address wrap (jolt-core/zkvm/r1cs).** The R1CS `RamAddress = Rs1Value + Imm` constraint is evaluated in the BN254 field (~254 bits), not mod 2^64. LD/SD are the only load/store ops not decomposed into virtual sequences, so they are directly exposed. **This is a completeness limitation only** — a malicious prover cannot exploit it. Real programs use addresses near `0x80000000`, far from the 2^64 wrap boundary. Added a block comment at the constraint site explaining the limitation. Closing the gap requires bit-decomposition + conditional 2^64 subtraction, a soundness-sensitive R1CS change that we elected not to make as part of this audit pass.

### Deferred (noted for follow-up)

- **Finding 4 — `DoryGlobals` process-wide mutable state.** Real, but fixing it requires threading a `DoryContext` handle through the entire commitment / opening / streaming call graph. Would only trigger if concurrent in-process proof generation becomes a supported workflow. Single-proof-per-process usage is safe.
- **Finding 5 — Dory transcript label flattening.** The audit calls this "low-effort" but it isn't: Jolt's `Transcript::append_*` require `label: &'static [u8]`, while Dory passes `&[u8]`. Fixing properly means adding a dynamic-label variant across all Transcript impls (Blake2b / Keccak / Poseidon) and invalidating existing serialized proofs. No soundness concern per audit. Deferred as a transcript-API design question.

## Uncertainty flags

- Finding 1 carries **completeness-only** risk. We elected to document rather than modify the R1CS because the constraint sits in soundness-sensitive code and the practical impact is zero. If the maintainers prefer a real fix (bit-decomposition + 2^64 subtraction on overflow), we can take that as follow-up work.
- Findings 4 and 5 are **explicitly deferred**, not rejected. Both are real, both are out of scope for this PR on effort grounds.

Draft because Finding 1's document-vs-fix call is a maintainer decision.

## Test plan

- [x] `cargo clippy --all --features host -q --all-targets -- -D warnings` — clean
- [x] `cargo clippy --all --features host,zk -q --all-targets -- -D warnings` — clean
- [x] `cargo nextest run -p jolt-core muldiv --cargo-quiet --features host` — pass
- [x] `cargo nextest run -p jolt-core muldiv --cargo-quiet --features host,zk` — pass
- [ ] CI full suite

### DIFF
--- a/AUDIT_ANALYSIS.md
+++ b/AUDIT_ANALYSIS.md
@@ -1,0 +1,239 @@
+# Audit Report Analysis — 2026-04-15
+
+**Source:** `jolt-audit-report.html` by Justin Thaler (with Claude assistance)
+**Scope:** jolt-core, tracer — soundness focus
+**Report verdict:** "No soundness vulnerabilities found."
+
+## Summary
+
+| # | Finding | Severity | Verdict | Action |
+|---|---------|----------|---------|--------|
+| 1 | LD/SD address arithmetic doesn't wrap at 64 bits | LOW | REAL (completeness) | Document as unsupported |
+| 2 | `has_side_effects()` defaults to false | INFO | REAL | Remove trait default; explicit per impl |
+| 3 | RAM ValCheck alignment `debug_assert` only | INFO | REAL | Promote to runtime `assert_eq!` |
+| 4 | DoryGlobals uses process-wide mutable state | INFO | REAL | Out of scope (large refactor) |
+| 5 | Dory transcript label flattening | INFO | REAL (non-trivial) | Deferred (API change required) |
+
+Two findings fixed in this PR (2, 3). One documented (1). Two deferred with rationale (4, 5).
+
+---
+
+## Finding 1 — LD/SD address arithmetic doesn't wrap at 64 bits
+
+**File:** `jolt-core/src/zkvm/r1cs/constraints.rs:237–241`
+
+### What the audit claims
+
+The R1CS constraint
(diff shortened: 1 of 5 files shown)


0xAndoroid:
None of the issues being fixed are meaningful, so closing.

PR #1402, opened 2026-04-02
spec: ecdsa inputs sanitation

0xAndoroid:
## Summary
- Add spec for moving ECDSA input validation into `ecdsa_verify()` for both secp256k1 and P-256
- Restrict `_unchecked` constructors to `pub(crate)` visibility
- Addresses the security footgun documented in #1391

## Test plan
- [ ] Review spec for completeness
- [ ] Run `@claude analyze` on the PR for second-opinion analysis

PR #1444, opened 2026-04-16
fix: add on-curve validation for public key Q in ECDSA verify

sagar-a16z:
Check q.is_on_curve() inside both P-256 and secp256k1 ecdsa_verify before using Q in the Shamir MSM. Without this, a caller using an unchecked point constructor could pass an off-curve Q, enabling invalid-curve attacks against the ECDSA verification.

Measured cost: +1,870 cycles on P-256 ECDSA verify (282,630 → 284,500), identical to the per-check cost of the existing R1/R2 on-curve checks. This matches the standard set by non-zkVM ECDSA libraries (RustCrypto, OpenSSL) which validate the public key internally.

### DIFF
--- a/jolt-inlines/p256/src/sdk.rs
+++ b/jolt-inlines/p256/src/sdk.rs
@@ -817,11 +817,9 @@ fn shamir_4x128(scalars: [u128; 4], points: [P256Point; 4]) -> P256Point {
 ///
 /// This halves the doublings: 128 instead of 256, achieving ~1.7x speedup.
 ///
-/// Security: it is the responsibility of the caller to ensure
-/// 1. z, r, and s are well formed
-/// 2. q is a valid point on the curve
-///
-/// Note that these checks are automatically performed by the from_u64_arr constructors.
+/// Security: this function validates all inputs internally — z, r, s must be
+/// well-formed field elements (enforced by the `P256Fr` type), and q must be a
+/// non-infinity point on the P-256 curve (checked via `is_on_curve()`).
 #[inline(always)]
 pub fn ecdsa_verify(z: P256Fr, r: P256Fr, s: P256Fr, q: P256Point) -> Result<(), P256Error> {
     if q.is_infinity() {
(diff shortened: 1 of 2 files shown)


0xAndoroid:
Starting code review process...

---

jolt-inlines/p256/src/sdk.rs
```diff
@@ -827,6 +827,9 @@ pub fn ecdsa_verify(z: P256Fr, r: P256Fr, s: P256Fr, q: P256Point) -> Result<(),
     if q.is_infinity() {
         return Err(P256Error::QAtInfinity);
     }
+    if !q.is_on_curve() {
+        return Err(P256Error::NotOnCurve);
```
0xAndoroid, line 829: The docstring above (lines 820–822) still states that on-curve validity of `q` is the caller's responsibility. With this check in place that's no longer true — worth updating so downstream users don't either duplicate the check or (worse) skip it under the impression the contract still requires them to do it.

---

jolt-inlines/secp256k1/src/sdk.rs
```diff
@@ -873,6 +873,9 @@ pub fn ecdsa_verify(
     if q.is_infinity() {
         return Result::Err(Secp256k1Error::QAtInfinity);
     }
+    if !q.is_on_curve() {
+        return Result::Err(Secp256k1Error::NotOnCurve);
```
0xAndoroid, line 875: Same note as the p256 sibling: the docstring above (lines 860–862) still says on-curve validation is the caller's responsibility, but `ecdsa_verify` now enforces it. Worth updating the comment to match.

---

0xAndoroid (commented):
Clean, targeted security fix. One minor follow-up: the `ecdsa_verify` docstrings now contradict the new behavior — worth updating.

---

0xAndoroid:
Oh, I think this was already worked on in #1402 I just forgot about it. Seems like there are more checks there.

PR #1450, opened 2026-04-20
feat(jolt-openings): add PCS traits and opening reduction crate

markosg04:
## Changes

-

## Testing

- [ ] Ran tests for modified crates
- [ ] `cargo clippy` and `cargo fmt` pass

## Breaking Changes

None

PR #1448, opened 2026-04-20
feat(jolt-r1cs): add uniform R1CS data crate

markosg04:
## Changes

-

## Testing

- [ ] Ran tests for modified crates
- [ ] `cargo clippy` and `cargo fmt` pass

## Breaking Changes

None

PR #1451, opened 2026-04-20
feat(jolt-dory): add Dory polynomial commitment scheme

markosg04:
## Changes

-

## Testing

- [ ] Ran tests for modified crates
- [ ] `cargo clippy` and `cargo fmt` pass

## Breaking Changes

None


markosg04:
* **#1452** <a href="https://app.graphite.com/github/pr/a16z/jolt/1452?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1451** <a href="https://app.graphite.com/github/pr/a16z/jolt/1451?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/a16z/jolt/1451?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1450** <a href="https://app.graphite.com/github/pr/a16z/jolt/1450?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1449** <a href="https://app.graphite.com/github/pr/a16z/jolt/1449?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1448** <a href="https://app.graphite.com/github/pr/a16z/jolt/1448?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

crates/jolt-dory/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+
+[lints]
+workspace = true
+
+[dependencies]
+jolt-crypto = { path = "../jolt-crypto" }
+jolt-field = { path = "../jolt-field" }
+jolt-poly = { path = "../jolt-poly" }
+jolt-transcript = { path = "../jolt-transcript" }
+jolt-openings = { path = "../jolt-openings" }
```
0xAndoroid: make these workspace deps

crates/jolt-dory/README.md
0xAndoroid, line 1: merge readme.md into the doc comments of the crate

crates/jolt-dory/src/scheme.rs
```diff
@@ -0,0 +1,576 @@
...
+    // SAFETY: Bn254GT and ArkGT are both repr(transparent) over Fq12.
+    unsafe { &*(std::ptr::from_ref(gt).cast::<ArkGT>()) }
+}
+
+#[inline]
+pub(crate) fn jolt_g1_vec_to_ark(v: Vec<Bn254G1>) -> Vec<ArkG1> {
+    // SAFETY: Bn254G1 and ArkG1 have identical size/align (repr(transparent)
+    // over G1Projective), so Vec layout is identical.
+    unsafe { std::mem::transmute(v) }
+}
```
0xAndoroid, line 66: why are all these conversion methods not using Into and From?

0xAndoroid (commented):
Deep review by 4 parallel analysis agents (semantic consistency, deep bugs, tech debt, security) with PAL second-opinion validation. Most consequential: `verify_zk` silently drops `_eval_commitment` (soundness gap confirmed by external expert review) and the cross-crate `transmute` SAFETY claim is false on the dory side (`ArkFr`/`ArkG1`/`ArkGT` aren't `#[repr(transparent)]`). Full list inline.

crates/jolt-dory/src/scheme.rs
```diff
@@ -0,0 +1,576 @@
...
+        commitment: &Self::Output,
+        point: &[Fr],
+        _eval_commitment: &Self::HidingCommitment,
+        proof: &Self::Proof,
+        setup: &Self::VerifierSetup,
+        transcript: &mut impl Transcript<Challenge = Self::Field>,
+    ) -> Result<(), OpeningsError> {
+        let ark_point: Vec<ArkFr> = point.iter().rev().map(jolt_fr_to_ark).collect();
+        // In ZK mode, dory::verify uses proof's y_com/e2 instead of evaluation.
+        let dummy_eval = <ArkFr as DoryField>::zero();
```
0xAndoroid, line 338: `_eval_commitment` is silently dropped — `dory::verify` only validates `proof.y_com` against itself via Σ₁/Σ₂ and the verifier setup. A malicious prover can substitute any `y_com'` in the proof bytes (recomputing Σ₁/Σ₂ for the new value) and the verifier accepts. The hiding-commitment binding promised by the trait is never enforced. Either compare `proof.0.y_com.expect("...") == jolt_g1_to_ark(_eval_commitment)` and reject on mismatch, or absorb `eval_commitment` into the transcript before `dory::verify` so Σ₁/Σ₂ challenges depend on it. The `Mock` impl in `jolt-openings` does enforce the binding.

crates/jolt-dory/src/scheme.rs
```diff
@@ -0,0 +1,576 @@
...
+    poly: &P,
+    num_rows: usize,
+    num_cols: usize,
+    setup: &ArkworksProverSetup,
+) -> Vec<ArkG1> {
+    let g1_bases = &setup.g1_vec[..num_cols];
+    let identity = <InnerBN254 as PairingCurve>::G1::identity();
+
+    let mut row_commitments = vec![identity; num_rows];
+    poly.for_each_nonzero(&mut |flat_idx, _val| {
```
0xAndoroid, line 384: `for_each_nonzero(|flat, _val| ...)` discards the value and treats every nonzero as `1`. Today only `OneHotPolynomial` triggers `is_sparse() == true`, but the trait contract on `for_each_nonzero` doesn't pin values to 1, so any future sparse impl with non-binary values silently produces a wrong commitment — same root opens to two different polynomials (binding break). Either `debug_assert!(val == Fr::one())` here, or accumulate `g1_bases[col].scalar_mul(&val)` so non-one values are handled.

crates/jolt-dory/src/scheme.rs
```diff
@@ -0,0 +1,576 @@
...
+        let row = flat_idx / num_cols;
+        let col = flat_idx % num_cols;
+        row_commitments[row] =
```
0xAndoroid: No bounds check on `row`: if a sparse `for_each_nonzero` impl yields a `flat_idx >= num_rows * num_cols` (the trait doesn't pin `flat_idx < 2^num_vars`), this panics inside the closure with no diagnostic. Cheap defense: `debug_assert!(row < num_rows && col < num_cols, "for_each_nonzero out-of-bounds flat_idx");`.

crates/jolt-dory/src/scheme.rs
```diff
@@ -0,0 +1,576 @@
...
+use jolt_field::Fr;
+use jolt_openings::{AdditivelyHomomorphic, CommitmentScheme, OpeningsError, ZkOpeningScheme};
+use jolt_poly::MultilinearPoly;
+use jolt_transcript::{AppendToTranscript, Label, LabelWithCount, Transcript};
+
+use crate::transcript::JoltToDoryTranscript;
+use crate::types::{DoryCommitment, DoryHint, DoryProof, DoryProverSetup, DoryVerifierSetup};
+
+// All jolt types below are #[repr(transparent)] over the same arkworks
+// inner type as their dory-pcs counterpart, guaranteeing identical layout.
```
0xAndoroid, line 27: The 'both jolt and dory-pcs counterpart are #[repr(transparent)]' premise is false on the dory side. `dory-pcs-0.3.0/src/backends/arkworks/{ark_field.rs:11, ark_group.rs:15-22}` defines `ArkFr`/`ArkG1`/`ArkG2`/`ArkGT` as plain tuple structs (`#[derive(Clone, Copy, ...)] pub struct ArkFr(pub Fr)`) — no `#[repr(transparent)]`. Single-field tuple struct layout is unspecified per Rust reference; layout coincides today, but a `dory-pcs` 0.3.x patch or any rustc layout-randomization breaks every transmute silently. Replace with explicit field-projection (`ArkFr(jolt.into_inner())`, `Bn254G1::from(ark.0)` etc.) — no unsafe needed since the wrappers expose `pub` inner fields.

crates/jolt-dory/src/scheme.rs
```diff
@@ -0,0 +1,576 @@
...
+            .fold(ArkGT::identity(), |acc, x| acc + x);
+
+        DoryCommitment(ark_to_jolt_gt(&combined))
+    }
+
+    fn combine_hints(hints: Vec<Self::OpeningHint>, scalars: &[Self::Field]) -> Self::OpeningHint {
+        assert_eq!(hints.len(), scalars.len());
+        if hints.is_empty() {
+            return DoryHint::default();
+        }
```
0xAndoroid: Returning `DoryHint::default()` (an empty `Vec`) when `hints.is_empty()` is a footgun: a caller that flows the empty hint into `open` lands in `Some(empty_hint)`, takes the `Some` arm at line 173, and yields `row_commitments = vec![]`. Dory then resizes with identities and produces a proof against the all-zero-rows commitment — divergent from any real commitment. Either drop the early return and assert non-empty, or distinguish 'no hint' from 'empty hint' so `open` falls through to `commit_rows_dense`.

crates/jolt-dory/src/scheme.rs
```diff
@@ -0,0 +1,576 @@
...
+
+        let num_rows = hints[0].0.len();
```
0xAndoroid, line 258: `hints[0].0.len()` then `combined.iter_mut().zip(hint.0.iter())` silently truncates if hints have different row counts. Add an assertion on the row-length invariant.

```suggestion
        let num_rows = hints[0].0.len();
        assert!(hints.iter().all(|h| h.0.len() == num_rows), "combine_hints: ragged hint lengths");
        let mut combined = vec![Bn254G1::default(); num_rows];
```

crates/jolt-dory/src/scheme.rs
```diff
@@ -0,0 +1,576 @@
...
+pub(crate) fn ark_to_jolt_g1(ark: ArkG1) -> Bn254G1 {
+    // SAFETY: Bn254G1 and ArkG1 are both repr(transparent) over G1Projective.
+    unsafe { std::mem::transmute(ark) }
+}
+
+#[derive(Clone)]
+pub struct DoryScheme;
+
+impl DoryScheme {
+    /// Generates prover SRS from a deterministic SHA3-seeded RNG.
```
0xAndoroid: 'Generates prover SRS from a deterministic SHA3-seeded RNG' isn't accurate. `ArkworksProverSetup::new_from_urs` calls `ProverSetup::new` which uses `OsRng` (`G1Projective::rand(&mut rand_core::OsRng)`). With `disk-persistence` enabled, the same machine returns cached generators, but two machines will silently produce different SRS for the same `max_num_vars`. Drop the SHA3 wording and document the disk-cache requirement (and the cross-machine implication for production deployments).

crates/jolt-dory/src/scheme.rs
```diff
@@ -0,0 +1,576 @@
...
+        transcript: &mut impl Transcript<Challenge = Self::Field>,
+    ) -> Self::Proof {
+        let num_vars = point.len();
+        let adapter = DorySourceAdapter::new(poly);
+        let sigma = num_vars.div_ceil(2);
+        let nu = num_vars - sigma;
+
+        let row_commitments = match hint {
+            Some(h) => jolt_g1_vec_to_ark(h.0),
+            None => commit_rows_dense(poly, sigma, &setup.0),
```
0xAndoroid, line 178: When `hint = None` and `poly.is_sparse()`, this falls through to `commit_rows_dense`, materializing the full row table for a one-hot polynomial. The numeric result coincides with `commit_rows_sparse` only because `MultilinearPoly::for_each_row` happens to materialize sparse rows correctly; the trait doesn't actually require this for sparse impls. Mirror the `commit` dispatch:

```
let row_commitments = match hint {
    Some(h) => jolt_g1_vec_to_ark(h.0),
    None if poly.is_sparse() => commit_rows_sparse(poly, 1 << nu, 1 << sigma, &setup.0),
    None => commit_rows_dense(poly, sigma, &setup.0),
};
```

Same pattern applies in `open_zk` at line 287.

crates/jolt-dory/src/streaming.rs
```diff
@@ -0,0 +1,77 @@
...
+    #[tracing::instrument(skip_all, name = "DoryScheme::stream_feed")]
+    fn feed(partial: &mut Self::PartialCommitment, chunk: &[Fr], setup: &Self::ProverSetup) {
+        let g1_bases = &setup.0.g1_vec[..chunk.len()];
+        let scalars: Vec<ArkFr> = chunk.iter().map(jolt_fr_to_ark).collect();
+        let row_commitment = G1Routines::msm(g1_bases, &scalars);
+        partial.row_commitments.push(ark_to_jolt_g1(row_commitment));
+    }
+
+    #[tracing::instrument(skip_all, name = "DoryScheme::stream_finish")]
+    fn finish(partial: Self::PartialCommitment, setup: &Self::ProverSetup) -> Self::Output {
```
0xAndoroid, line 50: `finish` doesn't validate that the streamed row count matches `2^(num_vars - sigma)` or pad to a power-of-two. Feeding a non-power-of-two number of rows still produces a commitment via `multi_pair_g2_setup`, but it won't match `DoryScheme::commit` on the same poly (which always sizes to `1 << nu`). Either pad here with `Bn254G1::default()` (identity) up to the next power-of-two, or document the precondition and assert. The `streaming_matches_direct` test never exercises a mismatched chunking.

crates/jolt-dory/src/transcript.rs
```diff
@@ -0,0 +1,92 @@
+//! Transcript adapter bridging jolt-transcript to dory-pcs.
+//!
+//! The bridging must produce the exact same transcript state as jolt-core's
+//! `JoltToDoryTranscript` (in `jolt-core/src/poly/commitment/dory/wrappers.rs`).
+//! jolt-core's adapter uses labeled methods (e.g. `append_bytes(b"dory_bytes", data)`),
```
0xAndoroid: 'Exact same transcript state as jolt-core's `JoltToDoryTranscript`' overstates the contract. jolt-core uses a different transcript framework (no per-absorb domain-tag byte, full 64-byte Blake2b-mod-r challenges); jolt-transcript prepends a `0x00` absorb domain tag and squeezes 16-byte challenges. The byte streams *don't* match across frameworks. The actual invariant — prover/verifier parity within `jolt-dory` — holds by construction. Reword to: 'Bridges jolt-transcript into dory-pcs's `DoryTranscript` trait. Prover/verifier parity is by construction; cross-framework parity with jolt-core's adapter is not guaranteed.'

crates/jolt-dory/src/transcript.rs
```diff
@@ -0,0 +1,92 @@
...
+
+impl<'a, T: Transcript<Challenge = Fr>> JoltToDoryTranscript<'a, T> {
+    pub fn new(transcript: &'a mut T) -> Self {
+        Self { transcript }
+    }
+}
+
+impl<T: Transcript<Challenge = Fr>> Default for JoltToDoryTranscript<'_, T> {
+    fn default() -> Self {
+        panic!("JoltToDoryTranscript must be constructed via JoltToDoryTranscript::new")
```
0xAndoroid: `Default::default()` panicking is dead code: the field is `&'a mut T` (non-Option) and `dory::primitives::transcript::Transcript` doesn't require `Default`. Vestigial copy from jolt-core's `Option<&mut T>` design. Delete the `Default` impl (and the `clippy::panic` suppression it justifies).

crates/jolt-dory/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Overview
+
+Wraps the [Dory](https://eprint.iacr.org/2020/1274) polynomial commitment scheme for BN254 with transparent setup, logarithmic proof size, and logarithmic verification. Supports streaming commitment and additive homomorphism for batch opening reduction.
+
+Implements `CommitmentScheme`, `AdditivelyHomomorphic`, `StreamingCommitment`, and `ZkOpeningScheme` from `jolt-openings`.
+
+## Public API
+
+- **`DoryScheme`** — Implements all four PCS traits. Static methods: `setup_prover`, `setup_verifier`, `extract_pedersen_setup`.
```
0xAndoroid: There's no static method `extract_pedersen_setup` on `DoryScheme` — the Pedersen-setup derivation is exposed via `impl DeriveSetup<DoryProverSetup> for PedersenSetup<Bn254G1>`. Replace with: 'Static methods: `setup_prover`, `setup_verifier`. Also implements `DeriveSetup<DoryProverSetup>` for `PedersenSetup<Bn254G1>` (use `PedersenSetup::derive(&prover_setup, capacity)`).'

crates/jolt-dory/src/lib.rs
```diff
@@ -0,0 +1,18 @@
...
+//!
+//! Wraps `dory-pcs` behind the `jolt-openings` trait hierarchy:
+//! [`CommitmentScheme`](jolt_openings::CommitmentScheme),
+//! [`AdditivelyHomomorphic`](jolt_openings::AdditivelyHomomorphic),
+//! [`StreamingCommitment`](jolt_openings::StreamingCommitment),
+//! [`ZkOpeningScheme`](jolt_openings::ZkOpeningScheme).
+
+pub mod scheme;
+pub mod streaming;
+pub mod types;
```
0xAndoroid: `pub mod` doubles the API surface: callers can reach `jolt_dory::scheme::DoryScheme` *and* `jolt_dory::DoryScheme` (via the `pub use` below). Make these private so only the curated re-exports are reachable.

```suggestion
mod scheme;
mod streaming;
mod types;
```

crates/jolt-dory/src/scheme.rs
```diff
@@ -0,0 +1,576 @@
...
+            &proof.0,
+            setup.0.clone().into_inner(),
+            &mut dory_transcript,
+        )
+        .map_err(|_| OpeningsError::VerificationFailed)
+    }
+}
+
+/// Dense commit: full MSM per row.
+fn commit_rows_dense<P: MultilinearPoly<Fr> + ?Sized>(
```
0xAndoroid, line 355: `commit_rows_dense` (and `commit_rows_sparse` below, plus `combine`/`combine_hints`) are sequential. jolt-core uses `par_chunks` per row, scalar-typed MSM variants for small scalars, and `jolt_optimizations::{batch_g1_additions_multi, vector_scalar_mul_add_gamma_g1_online}`. For a 12-var polynomial (64 rows × 64-col MSM) the loss is ≈ NCPU; for prove-time-critical batched openings (`combine_hints` over hundreds of polys × tens of rows) the gap is much larger. Restoring parallelism on these paths is the single biggest perf delta vs jolt-core.

crates/jolt-dory/src/types.rs
```diff
@@ -0,0 +1,236 @@
...
+        canonical_deserialize(deserializer).map(Self)
+    }
+}
+
+#[derive(Clone)]
+pub struct DoryProverSetup(pub ArkworksProverSetup);
+
+// SAFETY: ArkworksProverSetup contains only group element vectors (plain data).
+// Missing auto-traits come from arkworks type-level plumbing, not interior mutability.
+unsafe impl Send for DoryProverSetup {}
```
0xAndoroid: 'Missing auto-traits come from arkworks type-level plumbing' is hand-wavy. Upstream `ProverSetup<E>` (`dory-pcs/src/setup.rs:25`) is plain `g1_vec, g2_vec, h1, h2, ht` — no `PhantomData`, no marker, no interior mutability. With `E = BN254`, `Send + Sync` should auto-derive. Try removing the manual `unsafe impl`s; if compilation still fails, the comment should pinpoint the actual obstructing field, otherwise delete the unsafe.

crates/jolt-dory/src/scheme.rs
```diff
@@ -0,0 +1,576 @@
...
+        let adapter = DorySourceAdapter::new(poly);
+        let sigma = num_vars.div_ceil(2);
+        let nu = num_vars - sigma;
+
+        let row_commitments = match hint {
+            Some(h) => jolt_g1_vec_to_ark(h.0),
+            None => commit_rows_dense(poly, sigma, &setup.0),
+        };
+
+        let ark_point: Vec<ArkFr> = point.iter().rev().map(jolt_fr_to_ark).collect();
```
0xAndoroid, line 181: Both `open`/`open_zk`/`verify`/`verify_zk` reverse the point endianness but don't apply jolt-core's `reorder_opening_point_for_layout` (used in the v1 wrapper for AddressMajor / CycleMajor). Migrators from jolt-core that expected the reordering will silently get wrong evaluations. Add a doc comment on the public methods stating the convention: 'point is assumed to be in natural variable order; address-major / cycle-major reordering must be done by the caller.'

crates/jolt-dory/src/scheme.rs
```diff
@@ -0,0 +1,576 @@
...
+    fn num_vars(&self) -> usize {
+        self.source.num_vars()
+    }
+
+    fn evaluate(&self, point: &[ArkFr]) -> ArkFr {
+        let native_point: Vec<Fr> = point.iter().map(ark_to_jolt_fr).collect();
+        jolt_fr_to_ark(&self.source.evaluate(&native_point))
+    }
+
+    fn commit<E, Mo, M1>(
```
0xAndoroid, line 427: `DorySourceAdapter::commit` is dead code today: `dory::prove` only invokes `polynomial.commit(...)` when `row_commitments` is None (`evaluation_proof.rs:118`), and `DoryScheme::open*` always pre-computes them. If a future dory-pcs version starts calling this path internally, sparse polys would silently lose their fast-path. Either delete with `unimplemented!("DoryScheme::commit pre-computes row commitments before invoking dory::prove")` or route through the same dispatch as `DoryScheme::commit`.

crates/jolt-dory/benches/dory.rs
```diff
@@ -0,0 +1,185 @@
...
+
+criterion_group!(
+    benches,
+    bench_setup_prover,
+    bench_commit,
+    bench_open,
+    bench_verify,
+    bench_streaming_commit,
+    bench_combine,
+);
```
0xAndoroid, line 319: Coverage gaps: no bench for `combine_hints` (the prover-time-critical batch-opening path), no `open_zk`/`verify_zk` benches, `num_vars` capped at 12 (production uses ≥ 24), and no sparse / OneHot commit. Without these, perf regressions on the actual hot paths are invisible to CI.

crates/jolt-dory/src/scheme.rs
```diff
@@ -0,0 +1,576 @@
...
+                &adapter,
+                &ark_point,
+                row_commitments,
+                <ArkFr as DoryField>::zero(),
+                nu,
+                sigma,
+                &setup.0,
+                &mut dory_transcript,
+            )
+            .expect("Dory proof generation should not fail");
```
0xAndoroid: `expect("Dory proof generation should not fail")` masks `DoryError::InvalidPointDimension` / `InvalidSize`, both reachable on caller misuse (e.g. `point.len() != polynomial num_vars`). Same `expect` at line 304 in `open_zk`. Either pre-validate inputs and panic with a precondition-named message, or surface the error rather than collapsing it into a generic 'should not fail'.

crates/jolt-dory/src/scheme.rs
```diff
@@ -0,0 +1,576 @@
...
+    #[tracing::instrument(skip_all, name = "DoryScheme::setup_verifier", fields(max_num_vars))]
+    pub fn setup_verifier(max_num_vars: usize) -> DoryVerifierSetup {
+        let prover_setup = Self::setup_prover(max_num_vars);
+        DoryVerifierSetup(prover_setup.0.to_verifier_setup())
+    }
+}
+
+impl DeriveSetup<DoryProverSetup> for PedersenSetup<Bn254G1> {
+    fn derive(source: &DoryProverSetup, capacity: usize) -> Self {
+        let len = capacity.min(source.0.g1_vec.len());
```
0xAndoroid: `capacity.min(source.0.g1_vec.len())` silently truncates if the caller asks for more capacity than the dory SRS supports. A Pedersen setup with too-few generators panics only later, when `commit` is called with too many values. Assert/return error here:

```suggestion
        assert!(capacity <= source.0.g1_vec.len(), "capacity ({}) exceeds SRS size ({})", capacity, source.0.g1_vec.len());
        let generators = ark_to_jolt_g1_vec(source.0.g1_vec[..capacity].to_vec());
```

crates/jolt-dory/tests/commit_open_verify.rs
```diff
@@ -0,0 +1,260 @@
...
+    assert!(result.is_err(), "wrong transcript domain must be rejected");
+}
+
+#[test]
+fn property_based_round_trip() {
+    for seed in 0..10u64 {
+        let num_vars = 2 + (seed as usize % 4); // 2..5
+        round_trip::<Blake2bTranscript>(num_vars, 800 + seed, b"prop-rt");
+    }
+}
```
0xAndoroid, line 263: No negative tests for `verify_zk`. Adding `wrong_eval_commitment_rejected_zk` (prove honestly, tamper `eval_com`, verify) would have surfaced the `_eval_commitment` not being bound. With the verify_zk fix in place that test should fail-then-pass; today it'd pass — the test itself is the regression detector.

crates/jolt-dory/fuzz/fuzz_targets/deser_commitment.rs
```diff
@@ -0,0 +1,12 @@
...
+use libfuzzer_sys::fuzz_target;
+
+fuzz_target!(|data: &[u8]| {
+    let config = bincode::config::standard();
+
+    let _ = serde_json::from_slice::<DoryCommitment>(data);
+    let _ = bincode::serde::decode_from_slice::<DoryCommitment, _>(data, config);
+    let _ = serde_json::from_slice::<DoryProof>(data);
+    let _ = bincode::serde::decode_from_slice::<DoryProof, _>(data, config);
+});
```
0xAndoroid, line 12: Only checks deserialization doesn't crash. Suggest a second target that takes a valid proof, mutates one byte at a fuzz-chosen offset, and asserts `DoryScheme::verify` rejects — soundness regressions on the verifier are far more likely to bite than deserialization panics.

crates/jolt-dory/benches/dory.rs
0xAndoroid: We should add other things to be benched. And its okay to use only 12 vars.

crates/jolt-dory/fuzz/fuzz_targets/deser_commitment.rs
0xAndoroid: yes

crates/jolt-dory/src/lib.rs
0xAndoroid: yes

crates/jolt-dory/src/scheme.rs
0xAndoroid: Dory will be made repr(transparent) in the next update. Can be safely ignored for now

crates/jolt-dory/src/scheme.rs
0xAndoroid: drop the comment

crates/jolt-dory/src/scheme.rs
0xAndoroid: add assert or smth

crates/jolt-dory/src/scheme.rs
0xAndoroid: add

crates/jolt-dory/src/scheme.rs
0xAndoroid: well, let's just print the DoryError then

crates/jolt-dory/src/scheme.rs
0xAndoroid: does it ever happen that hints is empty? We should panic if it never happens

crates/jolt-dory/src/scheme.rs
0xAndoroid: add this assert

crates/jolt-dory/src/scheme.rs
0xAndoroid: You should just drop the eval_commitment argument on this function in favor of using y_comm on the proof

crates/jolt-dory/src/scheme.rs
0xAndoroid: Yes, we should restore parallelism

crates/jolt-dory/src/scheme.rs
0xAndoroid: I think sparse and one hot are equivallent statement in Jolt.

crates/jolt-dory/src/scheme.rs
0xAndoroid: sure

crates/jolt-dory/src/scheme.rs
0xAndoroid: yeah, just change it to the unimpl

crates/jolt-dory/src/streaming.rs
0xAndoroid: match the behavior with DoryScheme::commit

crates/jolt-dory/src/transcript.rs
0xAndoroid: sure

crates/jolt-dory/src/transcript.rs
0xAndoroid: delete the default impl

crates/jolt-dory/src/types.rs
0xAndoroid: try removing

crates/jolt-dory/tests/commit_open_verify.rs
0xAndoroid: add tests

0xAndoroid (approved):

0xAndoroid (commented):
Found two issues in the Dory serialization boundary.

crates/jolt-dory/src/types.rs
```diff
@@ -0,0 +1,225 @@
...
+
+impl Serialize for DoryCommitment {
+    fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        dory_serialize(jolt_gt_ref_to_ark(&self.0), serializer)
+    }
+}
+
+impl<'de> Deserialize<'de> for DoryCommitment {
+    fn deserialize<D: Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        let ark_gt: ArkGT = dory_deserialize(deserializer)?;
```
0xAndoroid: This deserializes through `ArkGT` and then transmutes into `Bn254GT`, which skips the validation on `Bn254GT` that rejects zero and non-r-torsion elements. Since commitments are verifier inputs, this should apply the same subgroup checks before constructing `DoryCommitment`.

crates/jolt-dory/src/types.rs
```diff
@@ -0,0 +1,225 @@
...
+
+impl Serialize for DoryProof {
+    fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        canonical_serialize(&self.0, serializer)
+    }
+}
+
+impl<'de> Deserialize<'de> for DoryProof {
+    fn deserialize<D: Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        canonical_deserialize(deserializer).map(Self)
```
0xAndoroid: Proof deserialization accepts dory-pcs attacker-controlled `num_rounds` before the verifier can compare it to the setup. The upstream deserializer allocates vectors with that count, so this wrapper should cap or pre-validate the round count during serde decoding.

0xAndoroid:
Proposal for the ZK opening API boundary: make the proof the single owner of the evaluation commitment, or make the external commitment explicitly checked by the verifier.

Right now `open_zk` returns `(proof, y_com, blind)`, while Dory verification reads `proof.y_com`. That creates two representations of the same commitment. Long term, I think the cleaner design is one of these:

1. Fold `y_com` and the blind into a typed ZK proof object, and expose accessors for the outer protocol to bind/extract them.
2. Keep `y_com` external, but require `verify_zk(..., y_com, proof, ...)` and make Dory reject if it differs from the commitment encoded in the proof.

Either way, generic callers should not be able to carry one hiding commitment in the outer transcript while Dory verifies a different one embedded in the proof. This probably belongs partly in dory-pcs, since the source of truth for `y_com` is currently the Dory proof format.

crates/jolt-dory/src/scheme.rs
```diff
@@ -0,0 +1,604 @@
...
+        proof: &Self::Proof,
+        setup: &Self::VerifierSetup,
+        transcript: &mut impl Transcript<Challenge = Self::Field>,
+    ) -> Result<(), OpeningsError> {
+        let ark_point: Vec<ArkFr> = point.iter().rev().map(jolt_fr_to_ark).collect();
+        let ark_eval = jolt_fr_to_ark(&eval);
+        let ark_commitment = jolt_gt_to_ark(&commitment.0);
+        let mut dory_transcript = JoltToDoryTranscript::new(transcript);
+
+        dory::verify::<ArkFr, InnerBN254, G1Routines, G2Routines, _>(
```
0xAndoroid, line 214: This wrapper should probably enforce the expected proof mode before delegating to Dory. dory-pcs currently infers transparent vs ZK mode from optional proof fields, so the API is less deterministic than the trait names suggest: standard `verify` should reject proofs containing ZK fields, and `verify_zk` should reject proofs missing them.

I don't think this is necessarily a standalone soundness issue if the underlying Dory checks reject inconsistent proofs, but making the mode check explicit here would prevent accidental cross-mode acceptance and make the contract easier to audit.

Issue #17, opened 2023-07-11
Use Quarks Grand Product Argument


sragss:

PatStiles:
Is this a reference to Section 5 of https://eprint.iacr.org/2020/1275.pdf?

PatStiles:
This refers to this optimization correct? 
![image](https://github.com/a16z/Lasso/assets/33334338/eb26e6ea-02d2-4c6d-b192-d27ddbe7d57e)

0xfuturistic:
could you elaborate more @sragss?

PR #1455, opened 2026-04-21
spec: port jolt-transcript to spongefish

shreyas-londhe:
## Summary

Spec-only PR. Proposes porting `crates/jolt-transcript` from its hand-rolled Fiat-Shamir construction to [spongefish](https://github.com/arkworks-rs/spongefish), the audited duplex-sponge library.

Key decisions captured in the spec:

- **Spongefish everywhere off-chain.** Three existing transcript variants (Blake2b, Keccak, Poseidon) become feature-selected sponges plugged into one spongefish construction via `DuplexSpongeInterface`. `spongefish::Blake2b512` and `spongefish::Keccak` come ready-made; Poseidon gets a new `DuplexSpongeInterface` impl over `light-poseidon`.
- **Positional API, no per-call labels.** Matches spongefish-native shape and how production consumers (WHIR, sigma-rs) use it. Jolt's protocol flow is deterministic, so positional order already provides domain separation that per-call labels would redundantly provide.
- **Trait impls directly on `spongefish::ProverState<Sponge>` / `VerifierState<Sponge>`.** Orphan rule allows it; no wrapper structs.
- **`JoltProof` collapses to a NARG byte string** plus whatever public inputs the verifier doesn't already know. Today's cfg-gated `opening_claims` / `blindfold_proof` fields become different prover-message sequences inside the NARG.
- **Downstream verifier byte layouts change as a consequence.** Solidity verifier (`transpilable_verifier.rs`), gnark verifier (`transpiler/`), and Lean extractor (`zklean-extractor/`) are explicit non-goals here and become coordinated follow-ups.
- **Dory migration is a follow-up PR on `a16z/dory`.** This PR updates the `JoltToDoryTranscript` bridge to wrap the new traits while dory keeps its current `DoryTranscript`. The bridge is removed once dory consumes `jolt-transcript`.
- **Transcript-internal tests are deleted, not adapted.** They tested generic Fiat-Shamir properties (determinism, domain separation, etc.) that spongefish owns upstream. Correctness of new code (Poseidon sponge impl, 128-bit decoder) is exercised by the `muldiv` e2e matrix across 3 sponges × 2 modes.

Open question flagged in the spec for maintainer input: how CI should treat `transpiler/go/e2e_test.go` and any Solidity tests that pin current byte layouts, since those will fail as a direct consequence.

Context: this is the redirect from a16z/dory PR #17 (closed by the maintainer) — per that direction, spongefish work lives in `jolt-transcript`, and dory will consume it.

## Test plan

- [ ] Run \`/analyze-spec\` remotely via \`claude-spec-review-request\` label to surface any remaining ambiguities
- [ ] Maintainer review and \`claude-spec-approved\` label
- [ ] Run \`/implement-spec\` from the approved spec in a follow-up implementation PR

Closes none — this is a spec PR.


0xAndoroid:
**Spec Analysis: Port jolt-transcript to spongefish**

| Dimension | Score | Gap |
|-----------|-------|-----|
| Goal | 0.80 | `JoltProof = narg + "public inputs the verifier doesn't already know"` — the "public inputs" set is unspecified |
| Constraints | 0.65 | Open CI question unresolved; perf bar soft; 254-bit challenge path not addressed |
| Success Criteria | 0.75 | Perf regression threshold has no concrete % / time budget |
| Context | 0.85 | "`crates/jolt-transcript` is currently parked" doesn't match the crate's actual state |
| **Ambiguity** | | **~24%** |

**Status: Questions remain** — 5 ambiguities to resolve before one-shot implementation.

The spec is in strong shape overall. Goal, abstractions, acceptance checklist, and alternatives are precise; file:line citations throughout make the scope easy to trace. Ambiguities cluster around (a) the proof wire format's non-NARG tail, (b) the challenge-width contract for Poseidon, and (c) pass/fail thresholds that the implementer would otherwise have to guess.

**Questions:**

**1. [Goal]** `JoltProof` is described as collapsing to `narg: Vec<u8>` "plus any public inputs the verifier doesn't already know." What concretely lives outside the NARG on the wire? Today's `fiat_shamir_preamble` (`jolt-core/src/zkvm/mod.rs:204`) binds preprocessing digest, memory layout, `max_input/output`, heap, inputs, outputs, panic flag, `ram_K`, `trace_length`, `entry_address`, `ReadWriteConfig` phase counts, `log_k_chunk`, and `DoryLayout` — are all of these "already known" to the verifier via preprocessing and guest I/O, or does some subset still travel on `JoltProof` alongside the NARG? An enumerated list (or a pointer to the preprocessing-vs-proof split rule) makes this mechanical to implement.

**2. [Constraints]** `transcript-poseidon` currently force-enables `challenge-254-bit` (`jolt-core/Cargo.toml:25`). The spec specifies a custom "128-bit-truncating decoder" to preserve the `challenge_*_optimized` performance profile (63 callsites verified). Under the new API, does Poseidon still produce 254-bit challenges where today's `challenge-254-bit` feature mandates it, or is the port unifying every sponge onto the 128-bit-truncating decoder? If the latter, that's a soundness-adjacent semantic change and contradicts Non-Goal 4 ("mechanical port of protocol semantics"). Please specify the per-sponge challenge-width contract (Blake2b/Keccak: 128-bit; Poseidon: 254-bit vs 128-bit?) and, if Poseidon changes, call it out explicitly.

**3. [Constraints]** The spec flags the CI treatment of `transpiler/go/e2e_test.go` and Solidity byte-layout tests as an open question for maintainers (options a/b/c). This is a real merge-gate dependency — without a decision, the implementer either stops to ask or picks arbitrarily (likely option a, `#[ignore]`). Please pick one so implementation can proceed one-shot. Suggested default: option (a) with a `FIXME(jolt-transcript-spongefish)` marker tying each ignore to the downstream follow-up PRs.

**4. [Success Criteria]** The performance bar reads "no observable regression beyond benchmark noise on `transcript_ops` and on `muldiv` e2e wall-clock under `--features host` with `transcript-blake2b`." Two gaps:
  - What is "benchmark noise" in practice — ≤3%? ≤5%? A Criterion p-value threshold? Without a number the criterion is unfalsifiable.
  - The bar names Blake2b only. Is `transcript-keccak` / `transcript-poseidon` held to any regression threshold, or is per-sponge perf non-binding beyond "muldiv passes"?

**5. [Context]** Minor — the Architecture section says `crates/jolt-transcript` is "currently parked; this spec activates it." The crate currently contains a live hand-rolled implementation: `src/{lib,transcript,blake2b,keccak,poseidon,blanket,digest}.rs`, populated `tests/`, `fuzz/`, and `benches/`. Acceptance Criteria correctly enumerate these as deletions, but the "parked" framing is misleading — an implementer expecting an empty scaffold will be surprised. Suggest rewording to "currently a hand-rolled Fiat-Shamir implementation that this spec replaces in-place, and `jolt-core/src/transcripts/` is deleted in favor of depending on it."

---

moodlezoup:
> How should CI treat transpiler/go/e2e_test.go and any Solidity integration tests that pin the current Poseidon / Keccak absorb byte layout? These will fail as a direct consequence of this port, since spongefish's domain separator and absorb semantics differ from the current hand-rolled layout. Options considered: (a) mark as #[ignore] until the downstream gnark / Solidity follow-ups land; (b) leave them failing, gating merge on coordinated downstream PR readiness; (c) maintainer's preference.

We don't have any Solidity integration tests. With regard to the transpiler tests, I think we should keep the jolt-core transcript implementation in place for now. We can deal with activating the new `jolt-transcript` crate (and updating the gnark transpiler) at a later date.

shreyas-londhe:
> We don't have any Solidity integration tests. With regard to the transpiler tests, I think we should keep the jolt-core transcript implementation in place for now. We can deal with activating the new `jolt-transcript` crate (and updating the gnark transpiler) at a later date.

Thanks for the context, @moodlezoup. I've tailored the spec to a narrower goal — internal port of `crates/jolt-transcript` only — with the jolt-core migration, dory consolidation, and gnark transpiler update each deferred to their own follow-up PRs.

Ready for re-analysis when convenient.

0xAndoroid:
## Spec Analysis

| Dimension | Score | Weight | Gap |
|-----------|-------|--------|-----|
| Goal Clarity | 0.88 | 0.35 | Trait method generic bounds & `narg_string()` return type not spelled out; `label`→`DomainSeparator` mapping implied |
| Constraint Clarity | 0.90 | 0.20 | Clear |
| Success Criteria | 0.87 | 0.30 | Invariant `Setup` associated type unspecified; sequence-validity (e.g. `Challenge` before any absorb) not stated |
| Context Clarity | 0.92 | 0.15 | Clear — all `file:line` anchors validate against HEAD (`wrappers.rs:336`, `Cargo.toml:45-48`/`44-46`/`26-28`, `field/ark.rs:2`, `README.md:229-239`) |
| **Ambiguity** | | | **11.3%** |

**Status: Approved** — the spec is clear enough for a one-shot implementation.

### Summary of what will be built
- Replace `crates/jolt-transcript`'s hand-rolled digest-based `Transcript` with spongefish-native `ProverTranscript` / `VerifierTranscript` traits, implemented directly on `spongefish::ProverState<Sponge>` / `spongefish::VerifierState<Sponge>` (no wrappers, via the orphan rule).
- Three feature-gated sponges: `transcript-blake2b` → `spongefish::Blake2b512`, `transcript-keccak` → `spongefish::Keccak`, `transcript-poseidon` → a new `PoseidonSponge` adapter over `light-poseidon` (Circom-compatible BN254).
- Per-sponge challenge-width contract preserved: Blake2b/Keccak expose both 128-bit-optimized and full-field decoders; Poseidon exposes full-field only (the 128-bit decoder is simply not defined on `PoseidonSponge`, so calling `_optimized` against it is a compile error).
- Delete `crates/jolt-transcript/tests/` and `crates/jolt-transcript/fuzz/` (rationale: they verify generic-sponge properties, not adapter parametrization). Adapt `benches/transcript_ops.rs` to the new positional API.
- Add `jolt-eval/src/invariant/transcript_symmetry.rs` (`TranscriptProverVerifierConsistency<Sponge>`) with `#[invariant(Test, Fuzz, RedTeam)]`, one instantiation per sponge. `Input` enum: `PublicBytes | PublicScalar | ProverBytes | ProverScalar | Challenge`. `check` runs the sequence on prover, replays on verifier against the NARG, asserts `receive_prover_message` round-trip equality and `verifier_message` equality. Seed corpus: empty, single-per-variant, 10-mixed, 1000-mixed.
- Add `jolt-transcript` dep to `jolt-eval/Cargo.toml`; register module in `invariant/mod.rs`; run `sync_targets.sh` to sync fuzz-target manifests.

### Key invariants
1. Prover/verifier sponge symmetry within the new crate (mechanized by the new invariant across all three sponges).
2. No external behavior change: `jolt-core/src/transcripts/`, `JoltToDoryTranscript`, `transpiler/`, `zklean-extractor/`, and `JoltProof` remain byte-identical; the three identically-named `transcript-*` features in `jolt-core`/`jolt-sdk`/`transpiler` continue to drive the old `jolt-core/src/transcripts/` unchanged.

### Critical evaluation criteria
- `cargo nextest run -p jolt-eval` (exercises the new invariant for all three sponge instantiations).
- `cargo nextest run -p jolt-core muldiv --features host` and `--features host,zk` continue to pass (workspace sanity only — jolt-core does not reach the new crate in this PR).
- `transcript_ops` Criterion benchmark stays within 5% of `main_run` baseline per `BenchmarkId` (enforced via `critcmp main_run pr_run --threshold 5` in `.github/workflows/bench-crates.yml`).

### Validation of spec anchors against HEAD
- `crates/jolt-transcript/src/{transcript,digest,blake2b,keccak,poseidon,blanket}.rs` present and hand-rolled digest-based ✓
- `crates/jolt-transcript/tests/` (4 files) and `fuzz/` populated ✓
- `crates/jolt-transcript/benches/transcript_ops.rs` present with `BenchmarkId::new("Blake2b"|"Keccak"|"Poseidon", ...)` groups ✓
- `jolt-eval/src/invariant/split_eq_bind.rs` serves as close structural model (same `#[invariant(Test, Fuzz, RedTeam)]` pattern, `type Setup = ()`, `type Input = ...`, `check(&self, _setup, input)`) ✓
- `jolt-eval/Cargo.toml` does **not** currently depend on `jolt-transcript` ✓
- No existing `transcript_symmetry.rs` or `TranscriptProverVerifierConsistency` ✓
- `jolt-core/src/poly/commitment/dory/wrappers.rs:336` hosts `JoltToDoryTranscript` ✓
- Features at `jolt-core/Cargo.toml:45-48`, `jolt-sdk/Cargo.toml:44-46`, `transpiler/Cargo.toml:26-28` ✓
- `challenge-254-bit` gate at `jolt-core/src/field/ark.rs:2` ✓
- `jolt-eval/README.md:229-239` describes `sync_targets.sh` ✓
- `.github/workflows/bench-crates.yml` uses Criterion `--save-baseline` and `critcmp` ✓
- `spongefish` is **not** yet a workspace dependency; `light-poseidon` already is ✓ (implementer needs to add `spongefish` to root `Cargo.toml`)

### Minor, non-blocking notes for the implementer (judgement calls, not ambiguities that would block a one-shot impl)
- Generic bounds on `public_message<T>` / `prover_message<T>` / `verifier_message<T>` are not spelled out — follow spongefish's native `CommonUnitToBytes`/`Decoding` trait bounds (the spec names `spongefish::Decoding<[H::U]>` for challenge decoders).
- `narg_string()` return shape not specified — use spongefish's native `&[u8]` return.
- The invariant's `Setup` associated type is not named — default `type Setup = ()` matches `SplitEqBindLowHighInvariant`.
- Ordering constraints on the `Input` sequence (e.g., starting with `Challenge`) are not stated — spongefish's `DomainSeparator` defines the alphabet; the `check` method should construct both transcripts from the same `DomainSeparator` derived from the sequence.

**Next step:** Run `/implement-spec` to implement this spec.

---

crates/jolt-transcript/src/setup.rs
```diff
@@ -0,0 +1,61 @@
...
+    let mut i = 0;
+    while i < src.len() {
+        buf[i] = src[i];
+        i += 1;
+    }
+    buf
+}
+
+/// Empty `instance` value so the spongefish builder reaches the `to_prover`
+/// stage. Encodes to zero bytes.
```
moodlezoup: I'm not sure I understand –– are `to_prover` and `to_verifier` exclusively used for testing (to dump the transcript state)? If so, let's rename them to `dump_prover_state` and `dump_verifier_state`

crates/jolt-transcript/src/codec.rs
```diff
@@ -0,0 +1,188 @@
...
+    type Repr = [u8; FR_TRUNCATED_BYTES];
+    fn decode(buf: Self::Repr) -> Self {
+        FieldElOptimized(Fr::from(u128::from_le_bytes(buf)))
+    }
+}
+
+/// Length-prefixed byte string. 8-byte LE length keeps `BytesMsg(a) ; BytesMsg(b)`
+/// distinguishable from `BytesMsg(a||b)`.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct BytesMsg(pub Vec<u8>);
```
moodlezoup, line 18: Can we implement `Encoding` and `NargDeserialize` for `Vec<u8>` upstream, so we can get rid of this wrapper?

crates/jolt-transcript/src/codec.rs
moodlezoup, line 1: Instead of implementing these traits for `ark_bn254::Fr`, can we do a blanket implementation over `F: jolt_field::CanonicalBytes` like we had in `blanket.rs`?

moodlezoup (commented):
great work! just a couple of comments

crates/jolt-transcript/src/codec.rs
```diff
@@ -0,0 +1,188 @@
...
+
+impl NargDeserialize for BytesMsg {
+    fn deserialize_from_narg(buf: &mut &[u8]) -> VerificationResult<Self> {
+        if buf.len() < 8 {
+            return Err(VerificationError);
+        }
+        let mut len_bytes = [0u8; 8];
+        len_bytes.copy_from_slice(&buf[..8]);
+        let len = u64::from_le_bytes(len_bytes) as usize;
+        if buf.len() < 8 + len {
```
moodlezoup: `8 + len` overflows when an attacker-supplied NARG declares `len = u64::MAX`. On 64-bit release builds the sum wraps to 7, the length guard passes, and `buf[8..8+len]` then panics on slice OOB instead of returning `VerificationError`. Any service that verifies untrusted proofs gets DoSed via a single 8-byte length prefix.

```suggestion
        let total = 8usize.checked_add(len).ok_or(VerificationError)?;
        if buf.len() < total {
            return Err(VerificationError);
        }
        let body = buf[8..total].to_vec();
        *buf = &buf[total..];
```

---

crates/jolt-transcript/src/compat.rs
```diff
@@ -0,0 +1,252 @@
...
+        // BytesMsg of the same body. The length prefix keeps
+        // `append_bytes(a) ; append_bytes(b)` distinct from
+        // `append_bytes(a || b)`.
+        const APPEND_MARKER: u8 = 0x9B;
+        let mut buf = Vec::with_capacity(9 + bytes.len());
+        buf.push(APPEND_MARKER);
+        buf.extend_from_slice(&(bytes.len() as u64).to_le_bytes());
+        buf.extend_from_slice(bytes);
+        let _ = self.sponge.absorb(&buf);
+        self.state = peek_state(&self.sponge);
```
moodlezoup: `peek_state` clones the entire sponge after every `append_bytes` and `challenge` call. For `PoseidonTranscript` the clone calls `fresh_hasher()` (poseidon.rs:95), which re-runs `Poseidon::new_circom(3)` and rebuilds the BN254 x5 round constants from scratch on every transcript op. No current consumer of `jolt-transcript` (`jolt-sumcheck`, `jolt-openings`, `jolt-crypto`) actually calls `state()`, so this work is pure overhead. Consider lazy refresh: store `Cell<Option<[u8; 32]>>`, invalidate on absorb/squeeze, populate on `state()` access.

---

crates/jolt-transcript/src/legacy.rs
```diff
@@ -0,0 +1,252 @@
...
+        buf.extend_from_slice(bytes);
+        let _ = self.sponge.absorb(&buf);
+        self.state = peek_state(&self.sponge);
+    }
+
+    fn challenge(&mut self) -> F {
+        let mut buf = [0u8; 16];
+        let _ = self.sponge.squeeze(&mut buf);
+        self.state = peek_state(&self.sponge);
+        F::from_challenge_bytes(&buf)
```
moodlezoup, line 273: The compat facade squeezes 16 bytes for *every* sponge including Poseidon, then maps via `from_challenge_bytes`. That contradicts the doc at prover.rs:53-55 which says calling a 128-bit decoder on a Poseidon-backed state is "a compile error" — Poseidon-via-compat does produce 128-bit-truncated challenges, just through a different entry point. Either add a `WARNING:` comment here noting the deliberate inconsistency pending the jolt-core migration, or specialize the squeeze width per-sponge so both surfaces agree.

---

crates/jolt-transcript/src/codec.rs
```diff
@@ -0,0 +1,188 @@
...
+
+/// 128-bit-truncating challenge wrapper. Decodes 16 squeezed bytes via
+/// `Fr::from(u128)`. Used only as a verifier message; the `Encoding` impl
+/// is the same 32-byte LE form as [`FieldEl`] so that absorbing one of
+/// these symmetrically with the other type stays a code error rather than
+/// a wire-format hazard.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct FieldElOptimized(pub Fr);
+
+impl Encoding<[u8]> for FieldElOptimized {
```
moodlezoup: Spongefish's blanket `impl<T: Encoding<[u8]>> NargSerialize for T` (spongefish-0.7.0/src/io.rs:62) means `prover.prover_message(&FieldElOptimized(f))` compiles silently, writes 32 bytes to the NARG, and the verifier has no way to read them back (no `NargDeserialize` impl). The data orphans in NARG and `check_eof` fails later, far from the misuse. The intended compile-error guard described in the doc comment doesn't fire on `prover_message`. Consider dropping the `Encoding` impl entirely (or wrapping `FieldElOptimized` in a `pub(crate)` newtype that only implements `Decoding`) so misuse is caught at the callsite.

---

moodlezoup (commented):
Spongefish port looks solid overall — known-vector tests pin the wire format and the new symmetry invariant is a good defense-in-depth. A few items worth fixing before this lands.

---

crates/jolt-transcript/src/setup.rs
shreyas-londhe: Not test-only — mirrors spongefish's own `DomainSeparator::to_prover` / `to_verifier` (we delegate at `setup.rs:48,60`). Used by `jolt-eval/src/invariant/transcript_symmetry.rs` and the planned jolt-core migration.

crates/jolt-transcript/src/codec.rs
shreyas-londhe: Upstream = external spongefish crate. Wrapper enforces an 8-byte LE u64 length prefix so `BytesMsg(a); BytesMsg(b)` ≠ `BytesMsg(a‖b)` — a jolt-side choice.

@mmaker — thoughts on adding `Encoding<[u8]>` + `NargDeserialize` for `Vec<u8>` upstream?

crates/jolt-transcript/src/codec.rs
shreyas-londhe: Done — `FieldEl<F>` / `FieldElOptimized<F>` now generic with per-impl bounds (`CanonicalBytes` for Encoding, `ReducingBytes`/`FromPrimitiveInt` for Decoding). 64-byte uniform constant stays hardcoded with a doc note (BN254-tuned; sound for any ≤254-bit field).

crates/jolt-transcript/src/codec.rs
shreyas-londhe: Fixed — `8usize.checked_add(len)` per suggestion. Added regression test feeding `u64::MAX` length.

crates/jolt-transcript/src/compat.rs
shreyas-londhe: Fixed — dropped the per-op cache. `state()` now returns owned `[u8; 32]` and calls `peek_state` on demand. No production caller exists (debug-only path), so the trade is fine.

crates/jolt-transcript/src/codec.rs
shreyas-londhe: Fixed — dropped the `Encoding` impl entirely. `FieldElOptimized` is verifier-message-only (squeeze → Decoding), so no caller breaks. Misuse via `prover_message` now fails at the call site instead of orphaning bytes in the NARG.

crates/jolt-transcript/src/legacy.rs
shreyas-londhe: Added WARNING comment. Divergence is intentional (compat preserves legacy width; not unsound — 128-bit truncation of Poseidon output is fine, just leaves entropy on the table). No external crate consumes compat `PoseidonTranscript`; goes away once jolt-sumcheck/openings/crypto migrate.

crates/jolt-transcript/src/codec.rs
mmaker: I agree with you and @moodlezoup. This wrapper should be in spongefish and `derive(Encoding)` should work for `Vec<u8>`.

mmaker (commented):
This PR is great progress. Thank you, @shreyas-londhe!

Here is some review/feedback, sorted by priority.

## Constructing the sponge

**The instance is not specified!** It is always set to `EmptyInstance`, which is going to be a big footgun. I would suggest removing this struct, and requesting an instance from the caller. 

Would it make sense for the constructor to be part of the prover and verifier transcripts, or perhaps to have a constructor specifically for the domain separation part?

Secondly, the **session ID** is not really used in this PR, and I’m assuming it will come from the application layer. It would be useful for me to have an example that I can check.

Third, `Transcript` requires `Clone`, and the legacy transcript structs derive it. This defeats the “prover state cannot be copied/restored” mitigation at the API level that we had in spongefish.

Top-level proof bytes are malleable. Are you checking the NARG string is fully read? `valid_proof || garbage` appears to deserialize as the same proof. This also affects generated verifier paths that call it.

## Poseidon

[This is not necessarily related to this specific PR; it is a problem in `main` too.]

The Poseidon implementation is not using DSFS, so it does not benefit from the provable security results we have. If I understand correctly, it is doing something like this for absorb:

```rust
state_0 = 0
state_1 = P(0, block_0_a, block_0_b)[0]
state_2 = P(0, state_1, block_1_a, block_1_b)[0]
...
````

And something like this for squeezing:

```rust
state_next = PoseidonPermutation([0, state_old, 0, 0])[0]
output state_next
```

For absorb: you are throwing away 1/4 of the rate. That is, you can have 25% more hash throughput with a small change.

For squeeze: this is not a sponge. It seems closer to the hash-chain Fiat-Shamir variant. I am not able to see an attack, as long as the protocol ID does not create confusion there.
Also here you're wasting all the rate, and **losing at least 50% of the throughput**.
It is also **not streaming-friendly**, which I believe is one of Jolt’s targets.

I would suggest implementing a true permutation function, in the sense of `spongefish::Permutation`, and then simply using `DuplexSponge<P: Permutation>` instead of writing a custom absorb/squeeze implementation.

I have not checked the Poseidon implementation in detail, but it seems to go through another library. I didn't check yet if that library has test vectors to match the spec.

## Redundant structures

Some structures can be removed. I will go through them in order, though I may be missing something for some of them.

1. `FieldEl`, `FieldElOptimized`, `UniformFrBytes`

   Field elements mod $p$ have historically been represented using the smallest representation in big-endian order [[RFC8017](https://www.rfc-editor.org/rfc/rfc8017.html#section-4)].

   You can patch the `spongefish` codec for `ark-ff` using an appropriate [`[replace]` block](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#the-replace-section). This is already done for you in my fork here: [https://github.com/mmaker/jolt/commit/168ab7421c32cfbefe88abbe46d01f63f85dba68](https://github.com/mmaker/jolt/commit/168ab7421c32cfbefe88abbe46d01f63f85dba68).

2. `OptimizedChallenge`

   Once item 1 is done, this can be a function with a default implementation in the trait. I would suggest moving the following into the trait, since it is such a simple function:

   ```rust
   fn challenge_128(&mut self) -> Fr {
       Fr::from(VerifierState::verifier_message::<u128>(self))
   }
   ```

3. `BytesMsg`

   I agree with Michael that we should handle this at the `spongefish` level.

## Minor

Some wording/miscellaneous advice after having spent a lot of time thinking about this:

* *challenge*: I would rather call this a *verifier message*.
* *transcript* generally refers to the list of prover and verifier messages. Here, it is a bit misused.
* *compat*: why not `legacy`?
* Add a `deprecated` annotation to all the structures that are not going to be used/needed?
* Some consistency tests are listed as executables. Why not move them into unit tests? They do not seem to be testing the full API (all the different transcript traits you have), are they?

shreyas-londhe:
Thanks for the review, @mmaker. Status per item below.

## Sponge construction

> **The instance is not specified!** It is always set to `EmptyInstance`…

Native API redesigned. `EmptyInstance` deleted. Construction goes through two factory functions plus an escape hatch:

```rust
prover_transcript(session: &[u8], instance: [u8; 32], sponge) -> ProverState<H, StdRng>
verifier_transcript(session: &[u8], instance: [u8; 32], sponge, narg) -> VerifierState<'_, H>
transcript_builder() -> DomainSeparator<WithoutInstance, WithoutSession>  // escape hatch
```

`instance: [u8; 32]` is positional and required. `transcript_builder()` exposes spongefish's `DomainSeparator` with `PROTOCOL_ID` pre-bound for composite verifiers (e.g. `BlindFold`). `legacy::Transcript` (formerly `compat`) unchanged — slated for deletion when callers migrate.

> Secondly, the **session ID** is not really used in this PR…

Doc-test on `prover_transcript` shows the wiring. `instance` in jolt-core will be `Blake2b(CanonicalSerialize(JoltDevice))` — preprocessing digest, memory layout, IO, panic flag, `ram_K`, `trace_length`, `entry_address`.

> Third, `Transcript` requires `Clone`…

Dropped from both `legacy::Transcript` and jolt-core's local `Transcript` trait. Only consumer was a dead `#[cfg(test)]` field — removed.

> Top-level proof bytes are malleable. Are you checking the NARG string is fully read?

`check_eof` docstring strengthened: soundness consequence + contract (call at end of top-level verify, success path only). WIRE-FORMAT.md marks it mandatory. Added `tests/narg_eof_tests.rs` proving rejection of trailing bytes and unread messages. When jolt-core migrates, its `verify(self)` adds one line: `transcript.check_eof()?`.

`JoltProof`'s ark `CanonicalDeserialize` at the jolt-core boundary has its own trailing-bytes case (different serialization layer). Tracked as follow-up.

## Poseidon

> Not DSFS; wastes 25% absorb rate, ≥50% squeeze rate, non-streaming.

Out of scope for this PR. `light-poseidon` only exposes `hash([Fr; 3]) → Fr`, not the round-permutation needed for `spongefish::Permutation<4>`.

@moodlezoup — three paths: (a) swap to `ark-crypto-primitives::Poseidon`, (b) fork `light-poseidon` to expose the inner permutation, (c) hand-roll the round function with Circom constants. Each needs test-vector verification + downstream gnark/Solidity coordination. Which direction?

## Redundant structures

> 1. `FieldEl`, `FieldElOptimized`, `UniformFrBytes` — patch `spongefish` for `ark-ff` via `[replace]`.

Deleted `FieldElOptimized` and `UniformFrBytes`. `OptimizedChallenge` now uses spongefish's built-in `u128` codec — wire format unchanged (also LE).

`FieldEl<F>` kept. spongefish 0.7's `ark-ff` feature pulls in `ark-ff ^0.6`; the workspace's pinned fork (`a16z/arkworks-algebra@dev/twist-shout`) is at `0.5`. Neither `[patch.crates-io]` nor `[replace]` bridges a major-version skew. Real blocker: the fork needs to bump to 0.6 — a workspace-wide change requiring patch rebase + perf re-test + audit. Worth its own RFC.

> 2. `OptimizedChallenge` becomes a default trait method.

Body collapsed to `Fr::from(verifier_message::<u128>(self))`. Kept per-backend impl pattern — Poseidon must remain a compile error for `challenge_128()` (128-bit truncation isn't sound for a field-native sponge).

@moodlezoup — flag if Poseidon should be able to opt into truncated challenges eventually; would switch to a marker-trait design then.

> 3. `BytesMsg` should be at the spongefish level.

Agreed. spongefish 0.7 has `Encoding<[u8]> for [u8]` (raw) but no length-prefixed variant. Will file an upstream issue.

## Minor

> _challenge_ → _verifier_message_; _compat_ → _legacy_; `#[deprecated]`; consistency tests → unit tests.

- Native trait already uses `verifier_message`. Legacy `challenge()` kept for BC.
- `compat` → `legacy` rename landed.
- `#[deprecated]` skipped — workspace runs `-D warnings`; flagging ~136 in-flight sites would fail CI. Will land in the migration PR.
- Integration-test layout kept (the `transcript_tests!` macro parametrizes over backends). Added `tests/native_traits_tests.rs` covering the native split-trait surface — round-trip, 128-bit truncation, session/instance domain separation.

---

moodlezoup:
> @moodlezoup — three paths: (a) swap to ark-crypto-primitives::Poseidon, (b) fork light-poseidon to expose the inner permutation, (c) hand-roll the round function with Circom constants. Each needs test-vector verification + downstream gnark/Solidity coordination. Which direction?

iirc the reason we went with light-poseidon was that it seemed relatively mature and has received an audit from Veridise. What's the status of the arkworks implementation? cc @mmaker

moodlezoup (approved):
great work, this lgtm. fyi @markosg04 @0xAndoroid I think merging this would break compatibility between `crates/` and `jolt-core`, but once we follow-up and integrate `jolt-core` with `jolt-transcript` we should be ok again

crates/jolt-transcript/tests/narg_eof_tests.rs
```diff
@@ -0,0 +1,102 @@
...
+const SESSION: &[u8] = b"narg-eof-test";
+const INSTANCE: [u8; 32] = [0x42; 32];
+
+fn build_valid_narg(messages: &[&[u8]]) -> Vec<u8> {
+    let mut prover = prover_transcript(SESSION, INSTANCE, Blake2b512::default());
+    for m in messages {
+        ProverTranscript::<Blake2b512>::prover_message(&mut prover, &BytesMsg(m.to_vec()));
+    }
+    ProverTranscript::<Blake2b512>::narg_string(&prover).to_vec()
+}
```
mmaker, line 23:
```suggestion
fn build_valid_narg(messages: &[&[u8]]) -> Vec<u8> {
    let mut prover = prover_transcript(SESSION, INSTANCE, Blake2b512::default());
    for m in messages {
        prover.prover_message(&BytesMsg(m.to_vec()));
    }
    prover.narg_string().to_vec()
}
```

crates/jolt-transcript/tests/narg_eof_tests.rs
```diff
@@ -0,0 +1,102 @@
...
+    let msgs: &[&[u8]] = &[b"alpha", b"beta", b"gamma"];
+    let narg = build_valid_narg(msgs);
+
+    let mut verifier = build_verifier(&narg);
+    for expected in msgs {
+        let got: BytesMsg = VerifierTranscript::<Blake2b512>::prover_message(&mut verifier)
+            .expect("valid prover message must deserialize");
+        assert_eq!(got.as_slice(), *expected);
+    }
+    VerifierTranscript::<Blake2b512>::check_eof(verifier).expect("exact narg must pass check_eof");
```
mmaker: same as above

crates/jolt-transcript/tests/narg_eof_tests.rs
```diff
@@ -0,0 +1,102 @@
...
+    let original_len = narg.len();
+    narg.extend_from_slice(&[0xFFu8; 7]);
+
+    let mut verifier = build_verifier(&narg);
+    for expected in msgs {
+        let got: BytesMsg = VerifierTranscript::<Blake2b512>::prover_message(&mut verifier)
+            .expect("valid prefix must deserialize");
+        assert_eq!(got.as_slice(), *expected);
+    }
+    let result = VerifierTranscript::<Blake2b512>::check_eof(verifier);
```
mmaker: same as above

crates/jolt-transcript/tests/native_traits_tests.rs
```diff
@@ -0,0 +1,113 @@
...
+        VerifierTranscript::<Blake2b512>::public_message(&mut verifier, &BytesMsg(b"pub".to_vec()));
+        let got: BytesMsg = VerifierTranscript::<Blake2b512>::prover_message(&mut verifier)
+            .expect("prover_message must deserialize");
+        assert_eq!(got.as_slice(), b"private");
+        let _c2: Fr =
+            <spongefish::VerifierState<'_, Blake2b512> as OptimizedChallenge>::challenge_128(
+                &mut verifier,
+            );
+        VerifierTranscript::<Blake2b512>::check_eof(verifier).expect("eof");
+    }
```
mmaker, line 36:
```suggestion
    #[test]
    fn prover_verifier_round_trip() {
        let mut prover = prover_transcript(SESSION, INSTANCE, Blake2b512::default());
        prover.public_message(&BytesMsg(b"pub".to_vec()));
        prover.prover_message(&BytesMsg(b"private".to_vec()));
        let _c1: Fr = prover.challenge_128();
        let narg = prover.narg_string();

        let mut verifier = verifier_transcript(SESSION, INSTANCE, Blake2b512::default(), &narg);
        verifier.public_message(&BytesMsg(b"pub".to_vec()));
        let got: BytesMsg = verifier
            .prover_message()
            .expect("prover_message must deserialize");
        assert_eq!(got.as_slice(), b"private");
        let _c2: Fr = verifier.challenge_128();
        verifier.check_eof().expect("eof");
    }
```

mmaker (commented):
there's some sloppy tests, perhaps we can make them a bit more clear and reduce the size of this PR?

PR #1458, opened 2026-04-21
fix: use independent per-point Shamirs in P-256 Fake GLV ECDSA verify

sagar-a16z:
```
P-256 Fake-GLV advice can be forged to prove an invalid signature end-to-end
```

The previous 4-scalar combined Shamir check (a1*G + a2*Q - b1*R1 - b2*R2 = O) had a cross-cancellation vulnerability.
Replace with two independent 2-scalar 128-bit Shamirs that bind each point individually, matching the original Fake GLV construction.

Performance: ~497K cycles vs 3.8M native p256 crate. Still 7.7x faster than native.

Reference: https://ethresear.ch/t/fake-glv-you-dont-need-an-efficient-endomorphism-to-implement-glv-like-scalar-multiplication-in-snark-circuits/20394

### DIFF
--- a/jolt-inlines/p256/src/sdk.rs
+++ b/jolt-inlines/p256/src/sdk.rs
@@ -928,17 +928,17 @@ pub(crate) fn verify_ecdsa_inner(
     // Reference: https://ethresear.ch/t/fake-glv-you-dont-need-an-efficient-endomorphism-to-implement-glv-like-scalar-multiplication-in-snark-circuits/20394
 
     // Check R1: a1*G - b1*R1 = O  (binds R1 = u1*G)
-    let r1_neg = if b1_sign { r1.clone() } else { r1.neg() };
+    let r1_adj = if b1_sign { r1.clone() } else { r1.neg() };
     let g_adj = if a1_sign { g.neg() } else { g };
-    let check1 = shamir_2x128([a1_val, b1_val], [g_adj, r1_neg]);
+    let check1 = shamir_2x128([a1_val, b1_val], [g_adj, r1_adj]);
     if !check1.is_infinity() {
         spoil_proof();
     }
 
     // Check R2: a2*Q - b2*R2 = O  (binds R2 = u2*Q)
-    let r2_neg = if b2_sign { r2.clone() } else { r2.neg() };
+    let r2_adj = if b2_sign { r2.clone() } else { r2.neg() };
     let q_adj = if a2_sign { q.neg() } else { q.clone() };
-    let check2 = shamir_2x128([a2_val, b2_val], [q_adj, r2_neg]);
+    let check2 = shamir_2x128([a2_val, b2_val], [q_adj, r2_adj]);
     if !check2.is_infinity() {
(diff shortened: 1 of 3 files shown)


jolt-inlines/p256/src/sdk.rs
```diff
@@ -925,36 +910,40 @@ pub(crate) fn verify_ecdsa_inner(
...
+    // Combining both into a single 4-scalar check would give the prover a
+    // cross-cancellation degree of freedom (one equation, two free point
+    // variables). Independent checks eliminate this.
+    //
+    // Reference: https://ethresear.ch/t/fake-glv-you-dont-need-an-efficient-endomorphism-to-implement-glv-like-scalar-multiplication-in-snark-circuits/20394
+
+    // Check R1: a1*G - b1*R1 = O  (binds R1 = u1*G)
+    let r1_neg = if b1_sign { r1.clone() } else { r1.neg() };
+    let g_adj = if a1_sign { g.neg() } else { g };
+    let check1 = shamir_2x128([a1_val, b1_val], [g_adj, r1_neg]);
```
0xAndoroid: Regressed from `r1_adj` to `r1_neg`, but when `b1_sign == true` the variable isn't negated (`r1.clone()`). Now inconsistent with `g_adj`/`q_adj` and the symmetric `r2_neg` at L939. Sign-flip code in a soundness-critical path shouldn't have misleading names.

```suggestion
    let r1_adj = if b1_sign { r1.clone() } else { r1.neg() };
    let g_adj = if a1_sign { g.neg() } else { g };
    let check1 = shamir_2x128([a1_val, b1_val], [g_adj, r1_adj]);
```

Same rename applies to `r2_neg` at L939–L941.

---

0xAndoroid (commented):
Fix closes the cross-cancellation bug cleanly and the new test directly exercises the exploit. One naming regression worth cleaning up (inline comment below). Separately, the crate doc at `jolt-inlines/p256/src/lib.rs:9` still says "enabling 4-scalar Shamir's trick" — no longer accurate after this change.

---

0xAndoroid (commented):
Soundness fix checks out. Per-point `a_i*P - b_i*R_i = O` together with the existing `b_i*u_i ≡ a_i (mod n)` and `b_val != 0` guards tightly binds `R_i = u_i*P`, eliminating the cross-cancellation degree of freedom the combined 4-scalar check had. Sign adjustment (`g_adj`/`r_adj`) is consistent with computing `a*G - b*R` via unsigned u128 scalars, and `shamir_2x128` edge cases (equal/inverse operands, mid-ladder infinity) are handled by `AffinePoint::add` and the `res.is_infinity()` branch. The regression test is well-constructed: the old combined check evaluates to `O` while the new check1 evaluates to `-G`, so it correctly distinguishes old-vs-new behavior.

One minor nit (out of diff): the crate-level doc at `jolt-inlines/p256/src/lib.rs:9` still advertises "4-scalar Shamir's trick" — worth updating while you're here.

---

PR #1460, opened 2026-04-22
feat: migrate architectural tests from RISCOF to ACT4

moodlezoup:
## Summary

Implements the ACT4 migration described in `specs/act4-tests.md`:

- Re-points `third-party/riscv-arch-test` to `riscv/riscv-arch-test` (ACT4), pinned at `a7c99303...` (main HEAD w/ v4.0 release notes).
- Replaces the RISCOF plugin tree with an ACT4-shaped config under `tests/arch-tests/jolt/` (test_config.yaml, UDB config, rvmodel_macros.h, link.ld, sail.json, rvtest_config.{h,svh}).
- New shell runner (`tests/arch-tests/run.sh`) iterates Sail-generated ELFs, skips entries in `skip.txt`, runs each through `jolt-emu`, and uses the process exit status as the pass/fail signal.
- `tracer/src/emulator/mod.rs::run_test` now surfaces the HTIF endcode; `tracer/src/main.rs` propagates it as the process exit status (previously always 0).
- `scripts/bootstrap`: xpack GCC 15.2 (ACT4 requires GCC 15+), mise (Ruby 3.4.9 + uv 0.11.6), sail_riscv_sim from upstream `riscv/sail-riscv` weekly release. No more RISCOF/CTG/ISAC/Spike.
- `Makefile`: single `arch-tests-64imac` target; drops the old multi-ISA targets; adds `arch-tests-smoke` for deliberate-failure plumbing coverage.
- `.github/workflows/arch-tests.yml`: single ubuntu:24.04 job running `make arch-tests-64imac` + `make arch-tests-smoke`.
- Deletes `scripts/apply-patches`, `patches/`, `tests/arch-tests/spike/`, all 32-bit/GC `.ini` configs, `jolt_platform.yaml`, all `jolt_isa_*.yaml`, `riscof_jolt.py`, and `tests/arch-tests/jolt/env/`.

Two real tracer bugs surfaced and were fixed:

1. **C.SLLI with `rd=x0`** was rejected as reserved; per the RISC-V C standard it's a HINT (decode as no-op). ACT4's `Zca-c.slli-00.S` literally contains `c.slli x0, 52` — prior behavior panicked with exit 101.
2. **SC.W after LR.D** was forced to fail (strict width-match); the RISC-V A spec is permissive (implementation-defined), and Sail's reference picks "succeeds if the reservation set contains the 4 bytes SC.W writes." ACT4 bakes Sail's choice into signatures, so the tracer must match. Fix: new `Cpu::reservation_covers(addr, min_bytes)`; SC.W uses 4-byte coverage, SC.D uses 8-byte; LR.D records the reservation in both word and doubleword slots for the ZK trace path.

Also ships `tests/arch-tests/run-in-docker.sh` — a macOS-friendly wrapper that runs the full suite in the same ubuntu:24.04 container image CI uses. Named Docker volumes persist the Cargo build, the xpack toolchain, and Sail across runs.

End-to-end verification: `make arch-tests-64imac` → **125 passed, 0 failed, 8 skipped (of 133)**. The 8 skips are CSR-family tests (Zicsr/Zicntr/Zihpm `csrrc*` and `csrrs` over counter CSRs) that would need new instruction decoders and counter-CSR mappings in the tracer — out of scope per the spec's Non-Goals; tracked in `skip.txt` with per-group reasons.

## Test plan

- [ ] `make arch-tests-64imac` on ubuntu:24.04 (CI) exits 0.
- [ ] `make arch-tests-smoke` — deliberate-failure ELF causes `jolt-emu` to exit non-zero and the target to report "smoke OK".
- [ ] `cargo nextest run -p tracer` — 110/112 pass (`test_checkpoints` and `test_trace_length` are sandbox-only failures unrelated to these changes; they build the muldiv guest and hit /tmp write restrictions).
- [ ] `cargo clippy -p tracer --features host,zk --all-targets -- -D warnings` — clean.
- [ ] `cargo nextest run -p jolt-core muldiv --features host` and `--features host,zk` — should still pass (only tracer public API change is `run_test` return type, and there's only one in-tree caller).


0xAndoroid (commented):
A few findings on the new exit-code plumbing and the `has_reservation`/`reservation_covers` split. Tracer fixes (C.SLLI HINT, SC.W/LR.D relaxation) look spec-correct and the ZK constraint path (`v_reservation_w`/`v_reservation_d` vs `reservation_covers`) stays consistent across all four LR/SC combinations.

---

tracer/src/main.rs
```diff
@@ -69,4 +69,16 @@ fn main() {
...
+    // Propagate the HTIF endcode as the process exit status so the ACT4 shell
+    // runner (tests/arch-tests/run.sh) can detect pass/fail reliably. ACT4
+    // self-checking ELFs set endcode=0 on pass (RVMODEL_HALT_PASS) and non-zero
+    // on fail (RVMODEL_HALT_FAIL, derived from gp). Clamp to i32 range while
+    // preserving the non-zero property.
+    let code: i32 = if endcode == 0 {
+        0
+    } else {
+        i32::try_from(endcode).unwrap_or(1).max(1)
+    };
```
0xAndoroid: The comment claims this preserves the non-zero property, but on Unix `std::process::exit` truncates to the low 8 bits. `RVMODEL_HALT_FAIL` in `rvmodel_macros.h` writes `endcode = gp` (the failing test number), and any ACT4 test whose `gp` is a multiple of 256 will return shell exit 0 — so `run.sh` records **PASS on a failing test**. `max(1)` only covers the `try_from` fallback, not the kernel-level truncation.

```suggestion
    let code: i32 = if endcode == 0 { 0 } else { 1 };
    exit(code);
```

---

tracer/src/emulator/mod.rs
```diff
@@ -131,7 +131,7 @@ impl Emulator {
     /// * Disassembles every instruction and dumps to terminal
     /// * The emulator stops when the test finishes
     /// * Displays the result message (pass/fail) to terminal
-    pub fn run_test(&mut self, trace: bool, disassemble: bool) {
+    pub fn run_test(&mut self, trace: bool, disassemble: bool) -> u64 {
```
0xAndoroid, line 143: Doc still describes the old void behavior ("Displays the result message to terminal") — worth adding a line for the new return contract: 0 = pass / clean exit, non-zero = HTIF endcode from `tohost`, so callers know what to propagate.

---

tracer/src/instruction/scw.rs
```diff
@@ -36,7 +36,10 @@ impl SCW {
         let address = cpu.x[self.operands.rs1 as usize] as u64;
         let value = cpu.x[self.operands.rs2 as usize] as u32;
 
-        if cpu.has_reservation(address, ReservationWidth::Word) {
+        // Per RISC-V A spec, SC.W succeeds if the reservation set covers the 4
+        // bytes being written. An LR.D reservation (8 bytes) at the same
+        // address qualifies; an LR.W reservation (4 bytes) does too.
+        if cpu.reservation_covers(address, 4) {
```
0xAndoroid: `reservation_covers(addr, 4)` / `reservation_covers(addr, 8)` in scd.rs are the only two valid `min_bytes` values. Taking a `ReservationWidth` (with `Ord` so the check collapses to `self.reservation_width >= width`) would make call sites read as intent and prevent accidental `reservation_covers(addr, 2)` etc.

---

tracer/src/emulator/cpu.rs
```diff
@@ -439,6 +439,26 @@ impl Cpu {
...
+    /// Returns true if a reservation is held at `address` whose reservation set
+    /// covers at least `min_bytes` starting at that address. Per the RISC-V A
+    /// spec (2024 ratified, §8.3): "SC succeeds only if the reservation is
+    /// still valid and the reservation set contains the bytes being written."
+    /// Concretely:
+    ///   - LR.W (4-byte reservation) + SC.W (4-byte write) → succeeds
+    ///   - LR.D (8-byte reservation) + SC.W (4-byte write) → succeeds (spec)
+    ///   - LR.W (4-byte reservation) + SC.D (8-byte write) → fails (out of range)
+    ///   - LR.D (8-byte reservation) + SC.D (8-byte write) → succeeds
+    pub fn reservation_covers(&self, address: u64, min_bytes: u64) -> bool {
```
0xAndoroid: `has_reservation` (just above) has no remaining callers after this PR — both SC.W and SC.D migrated to `reservation_covers`. Leaving the old strict-width predicate in the API makes it easy for the next author to call `has_reservation(addr, Word)` in a new SC variant and silently re-introduce the bug this PR just fixed. Consider deleting it.

---

0xAndoroid (commented):
Tracer fixes (C.SLLI HINT, SC.W/LR.D relaxation) look spec-correct, and the ZK constraint path (`v_reservation_w`/`v_reservation_d` vs `reservation_covers`) stays consistent across all four LR/SC combinations. A few findings on the new exit-code plumbing and the `has_reservation`/`reservation_covers` split inline.

---

PR #1475, opened 2026-04-29
refactor(jolt-lookup-tables): port jolt-core's per-instruction fuzz tests

moodlezoup:
## Why

`jolt-lookup-tables` was extracted from `jolt-core` over the previous commits, but the per-instruction fuzz tests in `jolt-core/src/zkvm/instruction/test.rs` weren't carried along — so the new crate had no harness for catching divergence between \`LookupQuery\` and the lookup tables / R1CS constraints / RISC-V semantics. This PR brings the new crate to test-coverage parity with \`jolt-core\` and lands the supporting refactors that became necessary to let the tests run there in isolation.

\`master..HEAD\`: 22 commits, 288 files, +6.5k/−4.8k.

## Tests ported

All three jolt-core harnesses are now wired up per-instruction in \`jolt-lookup-tables\`, exposed as macros so each call site is a single line:

- \`materialize_entry_test!\` — fuzz-checks \`to_lookup_output\` agrees with \`materialize_entry(to_lookup_index)\` (57 instructions).
- \`instruction_inputs_match_constraint_test!\` — fuzz-checks \`to_instruction_inputs\` matches the operand-routing R1CS constraint (\`left = LeftOperandIsRs1Value · rs1 + LeftOperandIsPC · pc\`, etc.) (57 instructions).
- \`lookup_output_matches_trace_test!\` — runs tracer's CPU emulator on a random cycle and asserts the resulting register write (or PC, for \`JAL\`/\`JALR\`) matches \`LookupQuery::to_lookup_output\` (41 instructions, matching jolt-core's coverage).

302 tests pass on the branch (vs. 0 instruction tests in the new crate at \`master\`).

## Bugs surfaced by the new fuzz tests (also fixed here)

- \`change_divisor\` overflowed at \`i64::MIN\` for XLEN=64.
- \`sign_extend_word\` / \`zero_extend_word\` referenced the wrong lookup tables.
- \`rotriw\` used \`% half\` instead of \`.min(half)\` for the rotation amount.
- \`bge\` / \`blt\` sign-extended in \`to_instruction_inputs\` instead of \`to_lookup_output\`, breaking the operand constraint for negative \`rs2\`.
- \`assert_halfword_alignment\` / \`assert_word_alignment\` masked their immediate through \`as u64\`, zero-extending negatives.
- \`Pow2\` and \`ShiftRightBitmask\` declared a spurious \`RightOperandIsImm\` flag.

## Supporting refactors

So the tests can live in \`jolt-lookup-tables\`:

- Extract \`jolt-riscv\` from \`jolt-trace\`: per-instruction newtypes, \`Flags\`, and the \`JoltInstructions\` enum (now wrapping concrete tracer structs per variant).
- Replace the \`jolt-riscv-derive\` proc-macro with a declarative \`jolt_instruction!\` macro that emits the struct, \`Flags\`, and \`JoltInstruction\` impls in one place.
- Define \`JoltInstruction\` / \`JoltCycle\` traits so \`LookupQuery\` impls can be parameterized over \`C: JoltCycle\` instead of tying directly to \`tracer::RISCVCycle<T>\`. Implementations for tracer's \`RISCVCycle<T>\` live in \`jolt-trace\`; the trait + blanket \`impl<T: RISCVInstruction> JoltInstruction for T\` live in \`jolt-riscv\`.
- Collapse the four ~40-line match expressions in \`LookupTableKind\` to a shared \`dispatch!\` macro.

## Test plan

- [x] \`cargo nextest run -p jolt-lookup-tables\` — 302/302 pass locally
- [x] \`cargo clippy -p jolt-lookup-tables -p jolt-riscv -p jolt-trace --all-targets -- -D warnings\` clean
- [x] \`cargo nextest run -p jolt-core muldiv --features host\` (e2e correctness)
- [x] \`cargo nextest run -p jolt-core muldiv --features host,zk\` (ZK mode)
- [x] CI green on the full workspace


0xAndoroid:
## Code review (Claude)

Read the PR body, the file inventory (288 files), and a sample of ~25 representative files (test harness, macros, lookup-table dispatch, R1CS constraints, several `LookupQuery` impls including the bug-fix targets).

### Overview
Ports `jolt-core`'s three per-instruction fuzz harnesses (`materialize_entry`, `instruction_inputs_match_constraint`, `lookup_output_matches_trace`) into `jolt-lookup-tables`, plus the supporting refactors that make the new crate self-contained: extracts `jolt-riscv` from `jolt-trace`, replaces the `jolt-riscv-derive` proc-macro with a declarative `jolt_instruction!` macro, generalizes `LookupQuery` over `C: JoltCycle`, and collapses `LookupTableKind` dispatch with a `dispatch!` macro. Surfaces and fixes six concrete bugs along the way and drops one R1CS input by replacing `IsLastInSequence` with `Jump` as the `NextPCEqPCPlusOneIfInline` guard.

### What's good
- **Bug-fix justification is clear.** Each of the six fixes (`change_divisor` `i64::MIN`, `sign_extend_word`/`zero_extend_word` table mismatch, `rotriw` `% half` vs `.min(half)`, `bge`/`blt` sign-extension placement, alignment masking, `Pow2`/`ShiftRightBitmask` spurious `RightOperandIsImm`) ties to a specific test. Comments at the fix sites (e.g. `signed_min = i64::MIN >> shift` in `change_divisor.rs:24`, `.min(half)` in `rotriw.rs:20`) explain *why*.
- **R1CS optimization is sound.** Swapping `(VirtualInstruction − IsLastInSequence)` for `(VirtualInstruction − Jump)` as the constraint-17 guard is supported by the per-cycle invariants and saves one circuit flag + one R1CS input. The new `rv64.rs:324-329` form is clearer.
- **Test harness design.** Three exported macros (`materialize_entry_test!`, `instruction_inputs_match_constraint_test!`, `lookup_output_matches_trace_test!`) keep call sites to a single line per instruction and enforce the `Jolt$Foo<RISCVCycle<$Tracer>>`/`RISCVCycle<$Tracer>` type pairing centrally — prevents drift between cycle-wrapper and tracer types at each call site.
- **`dispatch!` macro** in `tables/mod.rs:151` cleanly collapses the four ~40-line match expressions; variants are listed once.
- **`#[repr(u8)] LookupTableKind::index()`** has a defensible safety comment with a Rust reference link.
- **Parameterization over `XLEN`** is consistent (both XLEN=64 prod and XLEN=8 hypercube tests pass).

### Issues / suggestions

1. **Missing `instruction_inputs_match_constraint_test` for 6 of 8 XOR-rot variants** (`xor_rot.rs`, `xor_rotw.rs`). Only `VirtualXorRot32` and `VirtualXorRotW16` get the constraint test; the other six (`...24/16/63`, `...12/8/7`) have only `materialize_entry_test` and `lookup_output_matches_trace_test`. Since the four variants in each file have identical `to_instruction_inputs` and identical operand flags, the omitted tests are technically redundant, but the PR body claims "57 instructions" of constraint-test coverage — actual coverage is 51. Either add the six tests for symmetry or drop a one-line comment in each file explaining that the four variants share `to_instruction_inputs` so a single test suffices.

2. **JAL / JALR `to_instruction_inputs` may diverge from the constraint test for negative imm.** `jal.rs:22` returns `(pc & mask, imm & mask as i128)` — the masked low-XLEN-bits form. The constraint test (`test.rs:64-83`) computes `right_expected = imm` (raw signed `i128` from `instr.imm()`). These differ for any `imm < 0` (the BGE/BLT bug fix is the same shape, fixed by masking instead of sign-extending). If `tracer::instruction::jal::JAL::random` ever yields a negative immediate, this test should fail — but the PR claims green CI, so either:
   - JAL random gens are bounded to non-negative, in which case it'd be worth either generating negatives explicitly in this test, or documenting the assumption, or
   - There's a reconciliation I'm missing — please confirm by hand-running e.g. `cargo nextest run -p jolt-lookup-tables instruction_inputs_match_constraint_jal --release` after seeding the rng to force a negative imm.

3. **`impl_lookup_table!` (`traits.rs:32-51`) emits an unconstrained `T`.** `impl<const XLEN: usize, T> InstructionLookupTable<XLEN> for $instr<T>` accepts any `T`; the matching `LookupQuery` impls require `C: JoltCycle`. Probably intentional so that `Add<()>` (the marker form used in `JoltInstructions`) can satisfy `InstructionLookupTable`, but the asymmetry is worth a one-line comment in the macro.

4. **`JoltInstructions` enum drift risk** (`instructions/mod.rs:267-374`). 102+ variants are listed three times: in the `enum`, in the `impl_jolt_instructions_flags!` invocation, and exercised once-per-iter via `strum::EnumIter` in the four exclusivity tests. Adding a new instruction requires touching the first two — a future maintainer is likely to add it to the enum and forget the macro list, which would silently fall through to no `Flags` impl on that variant. Consider asserting `JoltInstructions::iter().count() == EXPECTED` in the existing test module to catch divergence.

5. **`(1u128 << XLEN).wrapping_sub(1) as u64`** is repeated in nearly every `to_instruction_inputs` impl. A `const fn xlen_mask<const XLEN: usize>() -> u64` (or `XlenMask::<XLEN>::VALUE` const) would centralize the pattern and make the code in each `LookupQuery` impl smaller. Cosmetic, can be a follow-up.

6. **`lookup_output_matches_trace_test_fn` reads `cpu.x[rd]` directly** (`test.rs:145`) rather than going through any `JoltCycle` accessor — that's fine since it's the ground truth, but worth noting in the doc-comment that the test deliberately bypasses the cycle abstraction.

### Risk
- **Blast radius is large** (instruction layer, R1CS witness layout, prover hot path) but the new fuzz harness exercises every per-instruction code path that lookup-tables touches. The dropped `IsLastInSequence` flag is the only externally-visible behavioral change to the constraint system; e2e `muldiv` tests pass per the PR description.
- **Mechanical refactor** of `jolt-riscv` extraction looks well-scoped — no behavior changes there beyond reorganization.

### Bottom line
LGTM modulo (1) and (2). (1) is a small follow-up; (2) deserves a confirmation that JAL/JALR negative-imm cases are actually fuzzed, since that's the exact failure mode the BGE/BLT fix addressed.

— Generated by [Claude Code](https://claude.com/claude-code)

---

0xAndoroid:
claude is being retarded, review is still running.

crates/jolt-r1cs/src/constraints/rv64.rs
```diff
@@ -319,12 +318,12 @@ pub fn rv64_constraints<F: Field>() -> crate::ConstraintMatrices<F> {
...
 
     // 17: NextPCEqPCPlusOneIfInline
-    //     guard = VirtualInstruction − IsLastInSequence
+    //     guard = VirtualInstruction − Jump
     //     left  = NextPC
     //     right = PC + 1
     a_rows.push(row::<F>(&[
         (V_FLAG_VIRTUAL_INSTRUCTION, 1),
-        (V_FLAG_IS_LAST_IN_SEQUENCE, -1),
+        (V_FLAG_JUMP, -1),
```
0xAndoroid: This swap isn't equivalent to the old `(VirtualInstruction − IsLastInSequence)` guard. For real (non-virtual) JAL/JALR, `VI=0` and `Jump=1`, so the guard evaluates to `−1` (not 0) and the eq-conditional fires, forcing `NextPC = PC + 1`. But `next_pc = bytecode_preprocessing.get_pc(next_cycle)` (jolt-core/src/zkvm/r1cs/inputs.rs:312) is the dense bytecode row of the jump target, not `PC + 1` — so any program with a real jump fails completeness.

The old `IsLastInSequence` was set only for `JALR` with `virtual_sequence_remaining == Some(0)`, which made the guard 0 for non-virtual jumps. The predicate `VI ∧ ¬Jump` isn't expressible as the linear difference `VI − Jump` over boolean inputs; you'd need a product factor (e.g. `VI · (1 − Jump)`) or to reinstate `IsLastInSequence`.

These matrices aren't yet wired into the production prover (`jolt-core/src/zkvm/r1cs/constraints.rs` still uses the old guard against its own `CircuitFlags` enum that retains `IsLastInSequence`), so the bug is dormant — but it'll fail every program with a JAL/JALR the moment the cutover lands. The PR description's claim that the two expressions "cover the same set of cycles" is incorrect.

---

crates/jolt-riscv/src/instructions/mod.rs
```diff
@@ -0,0 +1,489 @@
...
+    VirtualXorRot63(VirtualXorRot63<VirtualXORROT63>),
+    VirtualXorRotW16(VirtualXorRotW16<VirtualXORROTW16>),
+    VirtualXorRotW12(VirtualXorRotW12<VirtualXORROTW12>),
+    VirtualXorRotW8(VirtualXorRotW8<VirtualXORROTW8>),
+    VirtualXorRotW7(VirtualXorRotW7<VirtualXORROTW7>),
+    VirtualAdvice(VirtualAdvice<TracerVirtualAdvice>),
+    VirtualAdviceLen(VirtualAdviceLen<TracerVirtualAdviceLen>),
+    VirtualAdviceLoad(VirtualAdviceLoad<TracerVirtualAdviceLoad>),
+    VirtualHostIO(VirtualHostIO<TracerVirtualHostIO>),
+}
```
0xAndoroid, line 381: The enum and `impl_jolt_instructions_flags!` list (lines 402–427) are missing ~20 instruction kinds that ARE declared and re-exported above (lines 228–258): `Csrrs`, `Csrrw`, `Mret`, the entire A-extension (`Amo*`, `LrD`/`LrW`, `ScD`/`ScW`), the advice-load family (`AdviceLb`/`AdviceLd`/`AdviceLh`/`AdviceLw`), and `VirtualLw`/`VirtualSw`. The four `iter()`-based exclusivity tests below silently skip them, and any consumer using `JoltInstructions` for static dispatch will hit a missing variant. Either add them (and the matching macro entries) or document the intentional omission with a compile-time assertion that all declared structs are covered.

---

crates/jolt-trace/src/jolt_cycle.rs
```diff
@@ -0,0 +1,116 @@
...
+        let normalized: NormalizedInstruction = instruction.into();
+        let register_state =
+            <<T::Format as InstructionFormat>::RegisterState as InstructionRegisterState>::random(
+                rng,
+                &normalized.operands,
+            );
+        Self {
+            instruction,
+            register_state,
+            ram_access: T::RAMAccess::default(),
```
0xAndoroid, line 122: `ram_access: T::RAMAccess::default()` means the three new test macros never exercise RAM-dependent paths in `LookupQuery` — every load/store/AMO is fuzzed with all-zero RAM access, masking sign-extension and width bugs in those impls (exactly the class of bug this PR is meant to surface). Either generate random RAM accesses via a `RAMAccess::random` helper, or skip those instruction kinds with an explicit gate so the coverage gap is visible to a future maintainer.

---

crates/jolt-lookup-tables/src/instructions/test.rs
```diff
@@ -0,0 +1,206 @@
...
+            let cpu_pc = cpu.read_pc();
+            assert_eq!(cpu_pc, lookup_result, "{raw:?}");
+        } else if let Some(rd) = rd_idx {
+            // x0 is hardwired to zero; writes are discarded so the CPU
+            // result is always 0 regardless of the lookup output.
+            if rd != 0 {
+                let cpu_result = cpu.x[rd as usize] as u64;
+                assert_eq!(cpu_result, lookup_result, "{raw:?}");
+            }
+        }
```
0xAndoroid, line 165: When `rd_idx` is `None` and the instruction isn't JAL/JALR, this loop iterates 10,000 times with no assertion — a silent pass for stores, branches, ecall/ebreak, fence, asserts, and any virtual instruction without `rd`. Per the doc comment those kinds are expected callers, so the macro will be invoked from those modules' tests and look like coverage when it isn't.

Add an explicit panic for the unhandled case, or extend the oracle: asserts must produce 1, branches must produce the taken bit, etc.

---

crates/jolt-lookup-tables/src/instructions/virt/xor_rot.rs
```diff
@@ -0,0 +1,160 @@
...
+        );
+    }
+
+    #[test]
+    fn lookup_output_matches_trace_virtualxorrot63() {
+        lookup_output_matches_trace_test!(
+            VirtualXorRot63,
+            tracer::instruction::virtual_xor_rot::VirtualXORROT63
+        );
+    }
```
0xAndoroid, line 183: Two suggestions on this file (and `xor_rotw.rs` which has the same shape):

1. Only `VirtualXorRot32` (and `VirtualXorRotW16`) gets `instruction_inputs_match_constraint_test!` — the other six variants are missing it. Since their `to_instruction_inputs` and operand flags are identical across the four rotation amounts, the missing tests are technically redundant — but the omission contradicts the PR description's "57 instructions" claim. Either add the six missing tests for symmetry or drop a one-line comment in each file noting why one variant suffices.

2. The four `LookupQuery` impls at lines 13–78 differ only in the rotation literal (32/24/16/63). The table side `VirtualXORROTTable<XLEN, ROT>` is already const-generic on the rotation; consider parameterizing the wrapper as `VirtualXorRot<C, const ROT: u32>` so changing rotations becomes a one-line edit instead of touching four impl blocks.

---

0xAndoroid (commented):
Re-running this as the proper multi-agent flow (semantic / deep-bug / tech-debt / security) — supersedes the single-pass review I posted earlier as an issue comment.

Top finding: a dormant soundness/completeness bug in the new R1CS constraint 17 — see the inline on `crates/jolt-r1cs/src/constraints/rv64.rs:326`. The new matrices aren't yet wired into the prover, so this isn't breaking CI today, but it will the moment they replace `jolt-core/src/zkvm/r1cs/constraints.rs`. Five other findings inline.

— Generated by [Claude Code](https://claude.com/claude-code)

---

PR #1488, opened 2026-05-01
feat(jolt-hyperkzg): add HyperKZG multilinear commitment scheme

0xAndoroid:
## Summary
- Introduces `jolt-hyperkzg`, a pairing-based multilinear polynomial commitment scheme implementing the `jolt-openings` trait surface
- Generic over `P: PairingGroup`; BN254 is the production instantiation
- Builds on univariate KZG primitives (commit, evaluate, witness polynomial division) reused internally by the multilinear protocol

## Changes
- `src/scheme.rs` — `HyperKZGScheme` implementing `CommitmentScheme` and `AdditivelyHomomorphic`; prove/verify built on KZG univariate openings
- `src/kzg.rs` — univariate KZG primitives used internally
- `src/types.rs` — `HyperKZGCommitment`, `HyperKZGProof`, `HyperKZGProverSetup`, `HyperKZGVerifierSetup`
- `src/error.rs` — `HyperKzgError` enum
- `tests/commit_open_verify.rs` — end-to-end commit/open/verify coverage
- `benches/hyperkzg.rs` — perf benchmarks
- `fuzz/` — three fuzz targets: `commit_open_verify`, `tampered_proof`, `wrong_eval`

## Testing
- `cargo nextest run -p jolt-hyperkzg`
- Fuzz targets exercise tampering and wrong-eval paths

### DIFF
--- a/crates/jolt-hyperkzg/fuzz/fuzz_targets/tampered_proof.rs
+++ b/crates/jolt-hyperkzg/fuzz/fuzz_targets/tampered_proof.rs
@@ -1,0 +1,65 @@
+#![no_main]
+
+//! Fuzz: tamper with proof evaluation bytes and verify that verification rejects.
+//!
+//! We generate a valid proof, then corrupt an evaluation entry using fuzzer-chosen bytes.
+
+use jolt_crypto::Bn254;
+use jolt_field::{Field, Fr};
+use jolt_hyperkzg::HyperKZGScheme;
+use jolt_openings::CommitmentScheme;
+use jolt_poly::Polynomial;
+use jolt_transcript::{Blake2bTranscript, Transcript};
+use libfuzzer_sys::fuzz_target;
+use rand_chacha::ChaCha20Rng;
+use rand_core::SeedableRng;
+
+type TestScheme = HyperKZGScheme<Bn254>;
+
+fuzz_target!(|data: &[u8]| {
+    if data.len() < 10 {
+        return;
+    }
+
+    let num_vars = 3;
+    let n = 1usize << num_vars;
+
+    let mut rng = ChaCha20Rng::seed_from_u64(0xfade);
+    let g1 = Bn254::g1_generator();
+    let g2 = Bn254::g2_generator();
+    let pk = TestScheme::setup(&mut rng, n, g1, g2);
+    let vk = TestScheme::verifier_setup(&pk);
+
+    let poly = Polynomial::<Fr>::random(num_vars, &mut rng);
+    let point: Vec<Fr> = (0..num_vars).map(|_| Fr::random(&mut rng)).collect();
+    let eval = poly.evaluate(&point);
+
+    let (commitment, ()) = TestScheme::commit(poly.evaluations(), &pk);
+
+    let mut pt = Blake2bTranscript::new(b"fuzz-tamper");
+    let proof =
+        <TestScheme as CommitmentScheme>::open(&poly, &point, eval, &pk, None, &mut pt);
+
+    let tamper_row = (data[0] as usize) % proof.v.len();
+    let tamper_col = (data[1] as usize) % proof.v[tamper_row].len();
+    let tamper_val = Fr::from_bytes(&data[2..]);
+
+    // Skip if the corruption is a no-op
+    if tamper_val == proof.v[tamper_row][tamper_col] {
+        return;
+    }
+
+    let mut tampered = proof.clone();
+    tampered.v[tamper_row][tamper_col] = tamper_val;
+
+    let mut vt = Blake2bTranscript::new(b"fuzz-tamper");
+    let result = <TestScheme as CommitmentScheme>::verify(
+        &commitment,
+        &point,
+        eval,
+        &tampered,
+        &vk,
+        &mut vt,
+    );
+    assert!(result.is_err(), "tampered proof must be rejected");
+});
--- a/crates/jolt-hyperkzg/fuzz/fuzz_targets/tampered_proof.rs
+++ b/crates/jolt-hyperkzg/fuzz/fuzz_targets/tampered_proof.rs
@@ -1,10 +1,11 @@
 #![no_main]
 
-//! Fuzz: tamper with proof evaluation bytes and verify that verification rejects.
+//! Fuzz: tamper with proof fields and verify that verification rejects.
 //!
-//! We generate a valid proof, then corrupt an evaluation entry using fuzzer-chosen bytes.
+//! We generate a valid proof, then corrupt either an evaluation entry,
+//! an intermediate commitment, or a witness commitment using fuzzer-chosen bytes.
 
-use jolt_crypto::Bn254;
+use jolt_crypto::{Bn254, JoltGroup};
 use jolt_field::{Field, Fr};
 use jolt_hyperkzg::HyperKZGScheme;
 use jolt_openings::CommitmentScheme;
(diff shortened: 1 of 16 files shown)


crates/jolt-hyperkzg/src/kzg.rs
```diff
@@ -0,0 +1,259 @@
...
+/// The remainder is `f(u)`, but we don't need it since the verifier
+/// can derive it from the evaluation vectors.
+pub(crate) fn compute_witness_polynomial<F: Field>(f: &[F], u: F) -> Vec<F> {
+    let d = f.len();
+    let mut h = vec![F::zero(); d];
+    for i in (1..d).rev() {
+        h[i - 1] = f[i] + h[i] * u;
+    }
+    h
+}
```
moodlezoup, line 44: `h[d-1]` is always 0 (the loop writes `h[0..=d-2]`), but the returned vec has length `d` and the caller at line 109 MSMs `&setup.g1_powers[..h.len()]` over the trailing zero. The MSM backend (arkworks `msm_bigint`) still pays the `into_affine`/`into_bigint` conversion for that term. Return a length-`d-1` vec and slice the SRS accordingly — three witness commits per open in a hot path.

---

crates/jolt-hyperkzg/src/scheme.rs
```diff
@@ -0,0 +1,701 @@
...
+{
+    fn combine(commitments: &[Self::Output], scalars: &[Self::Field]) -> Self::Output {
+        assert_eq!(commitments.len(), scalars.len());
+        let combined = commitments
+            .iter()
+            .zip(scalars.iter())
+            .map(|(c, s)| c.point.scalar_mul(s))
+            .fold(P::G1::identity(), |acc, x| acc + x);
+        HyperKZGCommitment { point: combined }
+    }
```
moodlezoup, line 342: Naive `scalar_mul + fold` here loses to `P::G1::msm(points, scalars)` for any non-trivial `commitments.len()`. The rest of the crate (kzg.rs:26, :109, :196) uses `msm` directly.

```suggestion
    fn combine(commitments: &[Self::Output], scalars: &[Self::Field]) -> Self::Output {
        assert_eq!(commitments.len(), scalars.len());
        let bases: Vec<P::G1> = commitments.iter().map(|c| c.point).collect();
        HyperKZGCommitment {
            point: P::G1::msm(&bases, scalars),
        }
    }
```

---

crates/jolt-hyperkzg/src/scheme.rs
```diff
@@ -0,0 +1,701 @@
...
+    /// $P_i[j] = (1 - x_{\ell-i}) \cdot P_{i-1}[2j] + x_{\ell-i} \cdot P_{i-1}[2j+1]$
+    fn fold_polynomials(
+        evals: &[P::ScalarField],
+        point: &[P::ScalarField],
+    ) -> Vec<Vec<P::ScalarField>> {
+        let ell = point.len();
+        let mut polys = Vec::with_capacity(ell);
+        polys.push(evals.to_vec());
+
+        for i in 0..ell - 1 {
```
moodlezoup, line 96: `for i in 0..ell - 1` underflows `usize` when `ell == 0`. The trait-level `commit`/`open` API makes this hard to reach in normal use (a 0-var poly has 1 eval), but the public `HyperKZGScheme::open(setup, evals, point, ...)` is also pub and would crash on an empty `point`. Either return `HyperKZGError::InvalidProof` up front (this function's caller already returns `Result`) or guard with an early `if ell == 0` return so a malformed caller doesn't panic.

---

crates/jolt-hyperkzg/src/scheme.rs
```diff
@@ -0,0 +1,701 @@
...
+            source.g1_powers.len() > capacity,
+            "SRS has {} G1 powers, need at least {} (capacity + 1 for blinding)",
+            source.g1_powers.len(),
+            capacity + 1,
+        );
+        let message_generators = source.g1_powers[..capacity].to_vec();
+        let blinding_generator = source.g1_powers[capacity];
+        PedersenSetup::new(message_generators, blinding_generator)
+    }
+}
```
moodlezoup, line 243: Using KZG SRS powers as Pedersen generators couples Pedersen binding to the same trapdoor `beta` that breaks KZG binding — once `beta` is destroyed both are sound, but you lose defense-in-depth (a typical Pedersen setup uses independent hash-to-curve generators). Worth a sentence in the doc here so an integrator doesn't assume the two schemes have independent security.

---

crates/jolt-hyperkzg/src/error.rs
```diff
@@ -0,0 +1,14 @@
...
+#[derive(Debug, thiserror::Error)]
+pub enum HyperKZGError {
+    #[error("SRS too small: have {have} powers, need {need}")]
+    SrsTooSmall { have: usize, need: usize },
+
+    #[error("proof verification failed")]
+    VerificationFailed,
+
+    #[error("invalid proof structure: {0}")]
+    InvalidProof(&'static str),
```
moodlezoup: `InvalidProof(&'static str)` and the context-free `VerificationFailed` lose information that would help debugging. The three call sites in scheme.rs:159-178 are semantically distinct (wrong com length, wrong v shape, wrong w length), and `VerificationFailed` covers both "folding consistency failed at level i" and "pairing check failed" — those have very different implications when a downstream regression triggers them. Consider splitting into typed variants.

---

crates/jolt-hyperkzg/src/kzg.rs
```diff
@@ -0,0 +1,259 @@
...
+            P::G1::msm(&setup.g1_powers[..h.len()], &h)
+        })
+        .collect();
+
+    // Absorb witness commitments and derive one more challenge to keep
+    // prover/verifier transcripts in sync
+    for wi in &w {
+        transcript.append(wi);
+    }
+    let _: P::ScalarField = transcript.challenge();
```
moodlezoup: Naming this `_` (and the "keep transcripts in sync" framing) obscures that this is the verifier's `d_0`. The prover doesn't use it, but it must advance the transcript identically to the verifier (kzg.rs:161).

```suggestion
    // Absorb witness commitments and mirror the verifier's `d_0` challenge
    // to keep prover/verifier transcripts in sync.
    for wi in &w {
        transcript.append(wi);
    }
    let _d_0: P::ScalarField = transcript.challenge();
```

---

crates/jolt-hyperkzg/fuzz/fuzz_targets/tampered_proof.rs
```diff
@@ -0,0 +1,65 @@
...
+    let tamper_col = (data[1] as usize) % proof.v[tamper_row].len();
+    let tamper_val = Fr::from_bytes(&data[2..]);
+
+    // Skip if the corruption is a no-op
+    if tamper_val == proof.v[tamper_row][tamper_col] {
+        return;
+    }
+
+    let mut tampered = proof.clone();
+    tampered.v[tamper_row][tamper_col] = tamper_val;
```
moodlezoup: Fuzz only perturbs `proof.v`. A target that tampers `proof.com[i]` or `proof.w[i]` (e.g., scalar-multiplying by a fuzzer-supplied value) would exercise the pairing-check side of the verifier, which is currently uncovered.

---

moodlezoup (commented):
Protocol math (folding consistency + 3-point batched KZG pairing) checks out, and the dimension validation before transcript mutation is good. A few items below — mostly perf in hot paths plus one trapdoor-leakage hardening for `setup_from_secret`.

---

PR #1492, opened 2026-05-03
fix(tracer): use wrapping_add in alignment assertions

wstran:
## Summary

`VirtualAssertWordAlignment::exec` and `VirtualAssertHalfwordAlignment::exec`
compute the effective address via `cpu.x[rs1] + imm`. Both operands are
`i64`, so the bare `+` panics in debug builds when the signed sum falls
outside `[i64::MIN, i64::MAX]`. The other 30+ memory-touching tracer
instructions (LB/LBU/LH/LHU/LW/LWU/LD, SB/SH/SW/SD, VirtualLW/VirtualSW,
AMO*) all use `wrapping_add`; these two are the only outliers.

## Reproducer

```rust
let mut cpu = Cpu::new(Box::new(DummyTerminal::default()));
cpu.x[1] = i64::MAX;
let instr = VirtualAssertWordAlignment {
    address: 0,
    operands: FormatAssert { rs1: 1, imm: 1 },
    virtual_sequence_remaining: None,
    is_first_in_sequence: false,
    is_compressed: false,
};
let mut ram_access = ();
instr.execute(&mut cpu, &mut ram_access);
// thread '...' panicked at virtual_assert_word_alignment.rs:21:23:
// attempt to add with overflow
```

## Why this is correct

- **RV64 spec**: effective addresses are `rs1 + sign_ext(imm)` mod 2^64
  (per the audit Finding 1 framing in #1442).
- **Codebase convention**: every other memory-touching tracer instruction
  uses `wrapping_add` for the same operation. Verified via
  `grep -rn 'as usize\] +' tracer/src/instruction/` — only these two
  files match before this fix.
- **No release-mode change**: signed `+` already wraps in release, and
  the alignment check operates on the same low bits in both forms. The
  fix only stops the spurious debug-build panic on overflow.

## Test plan

- [x] `cargo fmt -q`
- [x] `cargo clippy -p jolt-core --features host --message-format=short -q --all-targets -- -D warnings`
- [x] `cargo clippy -p jolt-core --features host,zk --message-format=short -q --all-targets -- -D warnings`
- [x] `cargo nextest run -p tracer` — 114/114 pass (includes 2 new regression tests, both red on the un-fixed code)
- [x] `cargo nextest run -p jolt-core muldiv --cargo-quiet --features host` — pass
- [x] `cargo nextest run -p jolt-core muldiv --cargo-quiet --features host,zk` — pass

### DIFF
--- a/tracer/src/instruction/virtual_assert_halfword_alignment.rs
+++ b/tracer/src/instruction/virtual_assert_halfword_alignment.rs
@@ -18,7 +18,7 @@ impl VirtualAssertHalfwordAlignment {
         cpu: &mut Cpu,
         _: &mut <VirtualAssertHalfwordAlignment as RISCVInstruction>::RAMAccess,
     ) {
-        let address = cpu.x[self.operands.rs1 as usize] + self.operands.imm;
+        let address = cpu.x[self.operands.rs1 as usize].wrapping_add(self.operands.imm);
         assert!(
             address & 1 == 0,
             "RAM access (LH or LHU) is not halfword aligned: {address:x}"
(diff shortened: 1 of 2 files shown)


moodlezoup (approved):
thanks!

PR #1493, opened 2026-05-03
feat(jolt-sumcheck): add fuzz crate for verifier panic coverage

wstran:
## Summary

Adds a `fuzz/` sub-workspace to `jolt-sumcheck` with two libFuzzer targets,
matching the pattern already established for `jolt-crypto`, `jolt-field`,
`jolt-poly`, and `jolt-transcript`. `jolt-sumcheck` was the only protocol-
critical modular crate without fuzz coverage.

## Targets

- **`sumcheck_verifier`** — drives `SumcheckVerifier::verify` with attacker-
  controlled `SumcheckClaim` (`num_vars` 0..=8, `degree` 1..=6, arbitrary
  `claimed_sum`) and a fuzzer-chosen sequence of `UnivariatePoly<Fr>` round
  polynomials whose lengths are independently controlled per round
  (0..=degree+1). The verifier must return `Ok(_)` or a typed
  `SumcheckError` and never panic.

- **`batched_sumcheck_verifier`** — same panic-guard for
  `BatchedSumcheckVerifier::verify`, exercising the front-loaded batching
  path: variable `num_claims` (including 0, which must surface as
  `EmptyClaims`), per-claim `num_vars` mismatches, and `mul_pow_2`
  scaling.

## CI

Adds `jolt-sumcheck` to the matrix in `.github/workflows/fuzz-crates.yml`,
so it runs on every PR alongside the four existing fuzz crates.

## Why

`jolt-sumcheck` is **verifier-only** and protocol-critical: a panic on a
malformed proof would let a malicious prover crash the verifier rather
than receive a clean rejection. This is the same threat model that
- #1408 hardened in `jolt-core` (`catch_unwind` around the verifier),
- `verify_tampered` in `jolt-dory` (#1487) covers for Dory, and
- `tampered_proof` in `jolt-hyperkzg` (#1488) covers for HyperKZG.

The existing unit tests in `crates/jolt-sumcheck/src/tests.rs` cover
honest-prover soundness; this crate covers the dual property —
adversarial-input robustness — with coverage-guided exploration.

## Notes

The fuzz sub-workspace re-applies the root workspace's `[patch.crates-io]`
for the arkworks fork, mirroring the existing `jolt-transcript/fuzz` and
`jolt-crypto/fuzz` setup. Without the patch, `light-poseidon` (transitive
through `jolt-transcript`) pulls unpatched `ark-ff` and the build fails.

## Local validation

- `cargo +nightly fuzz build` (apple-darwin) — clean
- `cargo +nightly fuzz run sumcheck_verifier -- -max_total_time=15` — 2.66M runs, no crashes
- `cargo +nightly fuzz run batched_sumcheck_verifier -- -max_total_time=15` — 3.06M runs, no crashes
- `cargo nextest run -p jolt-sumcheck` — 40/40 pass
- `cargo fmt -q` clean

CI's `Fuzz crates` workflow caps each target at `FUZZ_DURATION = 60s`, matching the existing crates.

### DIFF
--- a/.github/workflows/fuzz-crates.yml
+++ b/.github/workflows/fuzz-crates.yml
@@ -28,6 +28,7 @@ jobs:
           - jolt-crypto
           - jolt-field
           - jolt-poly
+          - jolt-sumcheck
           - jolt-transcript
     name: Fuzz ${{ matrix.crate }}
     concurrency:
--- a/crates/jolt-sumcheck/fuzz/Cargo.toml
+++ b/crates/jolt-sumcheck/fuzz/Cargo.toml
@@ -35,3 +35,8 @@ doc = false
 name = "batched_sumcheck_verifier"
 path = "fuzz_targets/batched_sumcheck_verifier.rs"
 doc = false
+
+[[bin]]
+name = "valid_prefix_proof"
+path = "fuzz_targets/valid_prefix_proof.rs"
+doc = false
--- a/crates/jolt-sumcheck/fuzz/Cargo.toml
+++ b/crates/jolt-sumcheck/fuzz/Cargo.toml
@@ -1,0 +1,37 @@
+[workspace]
+
+# The fuzz sub-workspace does not inherit the root workspace's `[patch.crates-io]`.
+# `light-poseidon` (via jolt-transcript) pulls unpatched `ark-ff` from crates.io;
+# jolt-sumcheck's types are built against the patched fork. Apply the same patches
+# here so both sides see the same `ark-ff`.
+[patch.crates-io]
+ark-bn254 = { git = "https://github.com/a16z/arkworks-algebra", branch = "dev/twist-shout" }
+ark-ec = { git = "https://github.com/a16z/arkworks-algebra", branch = "dev/twist-shout" }
+ark-ff = { git = "https://github.com/a16z/arkworks-algebra", branch = "dev/twist-shout" }
+ark-serialize = { git = "https://github.com/a16z/arkworks-algebra", branch = "dev/twist-shout" }
+
+[package]
+name = "jolt-sumcheck-fuzz"
+version = "0.0.0"
+publish = false
+edition = "2021"
+
+[package.metadata]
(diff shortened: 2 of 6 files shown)


0xAndoroid (commented):
Can we also add fuzzing of proofs that have several first (potentially all) rounds be valid polys, just to test fiat-shamir as well. Otherwise looks great

wstran:
@0xAndoroid thanks for the review! Pushed `be5e90b0` adding a third fuzz target `valid_prefix_proof` that covers exactly the case you described.

The new target lets the fuzzer pick `valid_rounds ∈ [0, num_vars]` and constructs the first `valid_rounds` round polynomials so the sum-check invariant `s(0) + s(1) = running_sum` holds — `c_1` is derived from the running sum and the fuzzer's high-order coefficients, mirroring the standard compressed-unipoly format. It then mirrors the verifier's `append_to_transcript / challenge / evaluate` pipeline prover-side so the running sum the verifier will check against is known.

When `valid_rounds == num_vars` the proof is fully valid by construction; the target asserts the verifier accepts and the returned `EvaluationClaim::value` matches the prover-side running sum (and `point.len() == num_vars`). When `valid_rounds < num_vars` it's a panic-guard only — the tail uses raw random bytes.

This way every iteration where `valid_rounds > 0` exercises Fiat-Shamir past round 0, including all challenge derivations and the final claim, which the original two targets only reach for round 0 before the round-0 sum check rejects.

Local runs:
- `cargo +nightly fuzz run valid_prefix_proof -- -max_total_time=15` → 742k runs, no crash
- `cargo +nightly fuzz run sumcheck_verifier -- -max_total_time=8` → 1.46M runs, no crash (regression-checked)
- `cargo +nightly fuzz run batched_sumcheck_verifier -- -max_total_time=8` → 1.22M runs, no crash (regression-checked)
- `cargo nextest run -p jolt-sumcheck` — 42/42 pass

0xAndoroid (approved):
lgtm, thanks

PR #1501, opened 2026-05-05
Optimize Stage 6 bytecode read-RAF

markosg04:
Summary:
- Route Stage 6 bytecode read-RAF through sparse bytecode RA indices plus chunk lengths instead of dense witness chunks.
- Stop materializing dense bytecode read-RAF chunks in normal jolt-witness output while keeping dense fallback coverage.
- Document current SHA2-chain perf state, remaining Stage 6 gaps, and next attack plan in crates/jolt-equivalence/GOAL.md.

Measured perf:
- 2^16 after sparse bytecode path: prove ratio 1.174x, inside the 20% target.
- 2^20 after sparse bytecode path: prove ratio about 1.97x, down from about 2.29-2.30x.
- Final allocation-only cleanup was not remeasured; the attempted rerun was stopped for laptop handoff.

Validation:
- cargo nextest run -p jolt-witness --cargo-quiet
- cargo nextest run -p jolt-kernels stage6 --cargo-quiet
- cargo clippy -p jolt-witness -p jolt-kernels --lib -- -D warnings
- cargo fmt --check
- git diff --check
- env MLIR_SYS_220_PREFIX=/opt/homebrew/opt/llvm PATH=/opt/homebrew/opt/llvm/bin:/Users/mgeorghiades/.cargo/bin:/opt/homebrew/bin:/usr/bin:/bin:/usr/sbin:/sbin SDKROOT=/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk BINDGEN_EXTRA_CLANG_ARGS=-isysroot/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk CARGO_NET_OFFLINE=true cargo nextest run -p jolt-equivalence --cargo-quiet --offline --test-threads 1

Note: normal git commit hooks were bypassed because lefthook's check-clippy target references missing package jolt-compiler in this checkout.


markosg04:
cherry-picked into `refactor/crates`

PR #1504, opened 2026-05-06
fix(jolt-dory): blind tier-2 commitment in ZK mode

ainta:
## Problem

`DoryScheme::commit()` produced a transparent, deterministic GT element even when the `zk` Cargo feature was enabled. `open_zk()` then passed a **zero** `commit_blind` to `dory::prove`, so the tier-2 Dory commitment was never blinded — the same class of bug that PR #1497 fixed in `jolt-core`'s old Dory adapter.

Concretely:
- `commit()` called `multi_pair_g2_setup(...)` and returned the raw GT element with no mask applied.
- `open_zk()` hard-coded `<ArkFr as DoryField>::zero()` as the `commit_blind` argument to `dory::prove`.
- `combine_hints()` ignored the blind entirely when combining per-polynomial hints for batched openings.

## Fix

Mirrors the approach from #1497, adapted for the new `crates/jolt-dory` abstraction layer:

- **`Cargo.toml`**: Add `[features] zk = []` to `jolt-dory`.
- **`maybe_blind()`** (cfg-gated helper): in ZK mode calls `ZK::sample::<ArkFr>()` and `ZK::mask(tier_2, setup.ht, blind)`; in non-ZK mode is a no-op that returns `Fr::default()`.
- **`DoryHint`**: changed from a tuple-struct wrapping `Vec<Bn254G1>` to a named struct `{ row_commitments, commit_blind: Fr }`. `commit()` stores the sampled blind there.
- **`open()` / `open_zk()`**: extract `commit_blind` from the hint and forward it to `dory::prove`. `open()` selects `dory::mode::ZK` in ZK feature (required for internal consistency with a non-zero tier-2 blind; `Transparent` + non-zero blind produces an unverifiable proof).
- **`combine_hints()`**: accumulates blinds as `Σ scalar_i * blind_i` — correct for the homomorphic GT commitment.
- **Streaming**: `DoryPartialCommitment` gains `commit_blind`; the new `finish_with_hint()` public method returns `(DoryCommitment, DoryHint)` so streaming callers can retrieve the blind. `StreamingCommitment::finish()` delegates to it and discards the hint.

## Tests

- `zk_commit_is_non_deterministic`: asserts two calls to `commit()` return different GT elements in ZK mode.
- `zk_streaming_open_zk_round_trip`: full prove/verify cycle using the streaming path + `open_zk`/`verify_zk`.
- Tests that assume transparent/deterministic commitments are gated `#[cfg(not(feature = "zk"))]` and still run in non-ZK mode.
- Clippy passes clean in both modes:
  ```
  cargo clippy -p jolt-dory -q --all-targets -- -D warnings
  cargo clippy -p jolt-dory -q --all-targets --features zk -- -D warnings
  ```

## Test plan

- [ ] `cargo nextest run -p jolt-dory --cargo-quiet` (non-ZK)
- [ ] `cargo nextest run -p jolt-dory --cargo-quiet --features zk` (ZK)
- [ ] `cargo clippy -p jolt-dory -q --all-targets -- -D warnings`
- [ ] `cargo clippy -p jolt-dory -q --all-targets --features zk -- -D warnings`

### DIFF
--- a/crates/jolt-dory/Cargo.toml
+++ b/crates/jolt-dory/Cargo.toml
@@ -5,6 +5,9 @@ edition = "2021"
 license = "MIT"
 description = "Dory commitment scheme implementation for the Jolt zkVM"
 
+[features]
+zk = []
+
 [lints]
 workspace = true
 
--- a/crates/jolt-dory/src/scheme.rs
+++ b/crates/jolt-dory/src/scheme.rs
@@ -8,6 +8,7 @@
 )]
 
 use dory::backends::arkworks::{ArkworksProverSetup, G1Routines, G2Routines};
+#[cfg(not(feature = "zk"))]
 use dory::mode::Transparent;
 use dory::primitives::arithmetic::{
     DoryRoutines, Field as DoryField, Group as DoryGroup, PairingCurve,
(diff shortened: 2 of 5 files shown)


ainta:
Closing this version. The fix needs a cleaner API split: keep CommitmentScheme as the public-evaluation path, and add an explicit ZK commitment/opening path for blinded commitments.

PR #1512, opened 2026-05-08
feat(jolt-witness): add witness crate and modular helpers

markosg04:
Adds the jolt-witness crate and the supporting modular crate APIs needed before landing the Bolt and generated-code stack.

The new witness crate centralizes Bolt-facing witness/oracle helpers for commitment trace inputs, one-hot chunk materialization, sparse stage 4/5 trace data, stage 6 bytecode/read-RAF/booleanity inputs, and small deterministic oracle helpers used by generated tests.

This also moves the primitive modular crates onto the semantic surface expected by Bolt-generated code: a simplified jolt-field trait bundle, transcript byte-length accounting and MockTranscript, reusable polynomial/source helpers, R1CS row-dot helpers, trace extraction over CycleRow, canonical lookup-table ordering/names, and sumcheck verifier helper modules.


crates/jolt-transcript/src/lib.rs
```diff
@@ -59,6 +60,7 @@ pub use blake2b::Blake2bTranscript;
 pub use digest::DigestTranscript;
 pub use domain::{Label, LabelWithCount, U64Word};
 pub use keccak::KeccakTranscript;
+pub use mock::MockTranscript;
```
moodlezoup, line 63: `MockTranscript` is exposed as an unconditional `pub` re-export. Its `append_*` are no-ops, `state()` is a fixed seed, and the default `new()` ignores the label — anything that constructs it instead of `Blake2bTranscript`/`KeccakTranscript` silently collapses Fiat-Shamir and produces unsound proofs. Worth gating both `mod mock` (line 54) and this re-export behind `#[cfg(any(test, feature = "test-utils"))]` so misuse is impossible at link time.

---

crates/jolt-witness/src/lib.rs
```diff
@@ -0,0 +1,1119 @@
...
+    }
+}
+
+/// Evaluates a `u64`-valued multilinear extension at `point`.
+pub fn mle_eval_u64<F: Field>(values: &[u64], point: &[F]) -> F {
+    EqPolynomial::<F>::evals(point, None)
+        .iter()
+        .zip(values)
+        .map(|(&weight, &value)| weight * F::from_u64(value))
+        .sum()
```
moodlezoup, line 406: `.iter().zip(values)` silently stops at the shorter side. If a caller passes `values` whose length isn't exactly `1 << point.len()`, the result is a partial MLE with no warning — and `stage4_ram_val_init_opening` just below relies on the caller matching them. An explicit length check would surface miscalibrations instead of returning a wrong field element.

```suggestion
/// Evaluates a `u64`-valued multilinear extension at `point`.
pub fn mle_eval_u64<F: Field>(values: &[u64], point: &[F]) -> F {
    assert_eq!(
        values.len(),
        1 << point.len(),
        "mle_eval_u64 expected {} values, got {}",
        1 << point.len(),
        values.len()
    );
    EqPolynomial::<F>::evals(point, None)
        .iter()
        .zip(values)
        .map(|(&weight, &value)| weight * F::from_u64(value))
        .sum()
}
```

---

crates/jolt-witness/src/lib.rs
```diff
@@ -0,0 +1,1119 @@
...
+        .collect::<Vec<_>>();
+
+    Stage6WitnessPolynomials {
+        instruction_ra_indices: instruction_indices,
+        bytecode_ra_indices: bytecode_indices,
+        ram_ra_indices: ram_indices,
+        instruction_ra_booleanity: Vec::new(),
+        bytecode_ra_booleanity: Vec::new(),
+        ram_ra_booleanity: Vec::new(),
+        bytecode_ra_read_raf: Vec::new(),
```
moodlezoup, line 714: `stage6_witness_polynomials` leaves `instruction_ra_booleanity`, `bytecode_ra_booleanity`, `ram_ra_booleanity`, and `bytecode_ra_read_raf` as `Vec::new()`, but `Stage6WitnessPolynomials::slices()` happily folds those empties into `booleanity_chunks` and `bytecode_ra_read_raf_chunks`. A caller that takes the returned struct and immediately calls `slices()` gets misleadingly-aligned empty slices with no type-level signal that the witness is half-built. Worth either documenting the two-phase contract on the struct/function or splitting into `Stage6WitnessPartial` + `Stage6WitnessComplete` so the type system enforces ordering.

---

crates/jolt-poly/src/split_eq.rs
```diff
@@ -0,0 +1,446 @@
...
+                    }
+                    outer_step(x_out, e_out[x_out], inner_acc)
+                })
+                .reduce_with(merge);
+            if let Some(result) = result {
+                result
+            } else {
+                assert!(!e_out.is_empty(), "split eq e_out invariant");
+                std::process::abort();
+            }
```
moodlezoup, line 320: `std::process::abort()` here is unreachable in practice (the `assert!` on the line above panics first) and, if it ever did run, would kill the whole process without unwinding or a backtrace — worse than `panic!`/`unreachable!`, and inconsistent with the `.expect("E_in_vec is never empty")` pattern in `jolt-core/src/poly/split_eq_poly.rs`.

```suggestion
            } else {
                unreachable!("split eq e_out invariant violated: e_out is empty");
            }
```

---

crates/jolt-transcript/src/mock.rs
```diff
@@ -0,0 +1,109 @@
...
+    fn append_bytes(&mut self, _bytes: &[u8]) {}
+
+    fn append<A: AppendToTranscript>(&mut self, _value: &A) {}
+
+    fn challenge(&mut self) -> F {
+        <F as jolt_field::TranscriptChallenge>::from_challenge_bytes(&self.next_bytes32())
+    }
+
+    fn state(&self) -> &[u8; 32] {
+        &self.seed
```
moodlezoup, line 77: `state()` returns the original seed regardless of how many challenges have been squeezed, so two mock transcripts that have advanced to different counters still compare equal — which defeats the trait's documented "synchronization-detection" purpose. Either fold `counter` into the returned digest (e.g. cache `H(seed || counter)` on each squeeze) or document the limitation explicitly so callers don't trust `state()` for divergence checks.

---

moodlezoup (commented):
Reviewed the new `jolt-witness` crate and the supporting helpers in `jolt-poly`/`jolt-r1cs`/`jolt-transcript`. The migrated `GruenSplitEqPolynomial` and `ExpandingTable` look algorithmically faithful to the jolt-core originals. A few items below — the biggest one is that `MockTranscript` is exposed unconditionally as a public re-export, which is a soundness footgun if anything ever instantiates it outside tests.

---

crates/jolt-witness/src/lib.rs
```diff
@@ -0,0 +1,1119 @@
...
+    }
+}
+
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
+pub struct CommitmentTraceSources {
+    pub rd_inc: Vec<i128>,
+    pub ram_inc: Vec<i128>,
+    pub instruction_keys: Vec<Option<u128>>,
+    pub ram_addresses: Vec<Option<u128>>,
+    pub bytecode_indices: Vec<Option<u128>>,
```
moodlezoup, line 39: can't these be u64?

crates/jolt-witness/src/lib.rs
moodlezoup, line 1: this file is a bit bloated, let's split it up

crates/jolt-witness/src/lib.rs
```diff
@@ -0,0 +1,1119 @@
...
+#[derive(Clone, Copy, Debug)]
+pub struct CycleInput {
+    pub dense: [i128; NUM_DENSE_TRACE_COLUMNS],
+    pub one_hot: [Option<u128>; NUM_ONE_HOT_TRACE_SOURCES],
+}
+
+impl CycleInput {
+    pub const PADDING: Self = Self {
+        dense: [0; NUM_DENSE_TRACE_COLUMNS],
+        one_hot: [Some(0), Some(0), None],
```
moodlezoup, line 23: add a comment here

crates/jolt-witness/src/lib.rs
```diff
@@ -0,0 +1,1119 @@
...
+
+pub const NUM_DENSE_TRACE_COLUMNS: usize = 2;
+pub const NUM_ONE_HOT_TRACE_SOURCES: usize = 3;
+
+/// Per-cycle primitive inputs consumed by Bolt oracle generation.
+#[derive(Clone, Copy, Debug)]
+pub struct CycleInput {
+    pub dense: [i128; NUM_DENSE_TRACE_COLUMNS],
+    pub one_hot: [Option<u128>; NUM_ONE_HOT_TRACE_SOURCES],
+}
```
moodlezoup, line 18: why this shape, as opposed to:


```suggestion
pub struct CycleInput {
    pub rd_inc: i128,
    pub ram_inc: i128,
    pub instruction_keys: Option<u128>,
    pub ram_addresses: Option<u128>,
    pub bytecode_indices: Option<u128>,
}
```

crates/jolt-witness/src/lib.rs
```diff
@@ -0,0 +1,1119 @@
...
+        }
+    }
+}
+
+pub fn commitment_trace_sources(cycle_inputs: &[CycleInput]) -> CommitmentTraceSources {
+    CommitmentTraceSources::from_cycle_inputs(cycle_inputs)
+}
+
+/// Returns a dense trace source by its generated oracle source name.
+pub fn dense_cycle_source(cycle_inputs: &[CycleInput], source: &str) -> Vec<i128> {
```
moodlezoup, line 59: I'm lacking context here, but why `&str` as opposed to enum variants for source labels?

crates/jolt-witness/src/lib.rs
```diff
@@ -0,0 +1,1119 @@
...
+        "field oracle has {} values, target length is {target_len}",
+        values.len()
+    );
+    let mut output = values.to_vec();
+    output.resize(target_len, F::zero());
+    output
+}
+
+/// Deterministic placeholder data for optional advice oracles in synthetic tests.
+pub fn deterministic_oracle_data<F: Field>(oracle: &str, num_vars: usize) -> Vec<F> {
```
moodlezoup, line 124: put this behind `#[cfg(test)]` if it's only used for tests

crates/jolt-witness/src/lib.rs
```diff
@@ -0,0 +1,1119 @@
...
+        rd_write_addresses,
+    }
+}
+
+/// Evaluates a `u64`-valued multilinear extension at `point`.
+pub fn mle_eval_u64<F: Field>(values: &[u64], point: &[F]) -> F {
+    EqPolynomial::<F>::evals(point, None)
+        .iter()
+        .zip(values)
+        .map(|(&weight, &value)| weight * F::from_u64(value))
```
moodlezoup, line 405:
```suggestion
        .map(|(&weight, &value)| weight.mul_u64(value))
```

PR #1513, opened 2026-05-08
feat(bolt): add Jolt protocol emitter

markosg04:
Adds the Bolt compiler/emitter crate and its Jolt protocol lowering, IR, Rust artifact emission, MLIR dialect definitions, templates, and unit tests.

Validation:

- cargo check -p bolt -q was attempted but is blocked locally by missing llvm-config required by mlir-sys/tblgen.

- Previous downstack modular package validation remains: clippy all-targets and nextest for touched primitive crates.

PR #1515, opened 2026-05-08
test(equivalence): add Bolt parity oracle

markosg04:
Add the jolt-equivalence crate for comparing the Bolt-generated path against jolt-core, including generated-role parity, real-data/tamper gates, and perf oracle tests.


0xAndoroid:
Closing as stale: both sides of the parity comparison are gone — the Bolt-generated crates never landed (#1513/#1514 closed) and jolt-core was stripped into jolt-prover-legacy (#1632). Branch left in place.

PR #1517, opened 2026-05-08
remove CODEOWNERS

moodlezoup:
imo was mostly just adding noise to PR assignment. now that none of our CI workflows rely on CODEOWNERS, we might as well delete


markosg04 (approved):
yes pls

PR #1514, opened 2026-05-08
feat(generated-roles): add generated Jolt role crates

markosg04:
Add the Bolt-generate prover, and verifier crates. Also adds `jolt-kernels` for coarse but efficient prover side cpu code

Validation: cargo check -p jolt-kernels -p jolt-verifier -p jolt-prover -q; cargo clippy -p jolt-kernels -p jolt-verifier -p jolt-prover -q --all-targets -- -D warnings; cargo nextest run -p jolt-kernels -p jolt-verifier -p jolt-prover --cargo-quiet.

PR #1467, opened 2026-04-23
refactor(openings): split PCS traits into verifier/prover halves and fuse batched openings

quangvdao:
## Summary

Splits the PCS trait family into role-based verifier/prover halves and fuses batched openings into a single `prove_batch` / `verify_batch` pair. Verifier-side code (`jolt-verifier`, `JoltProof`, `JoltVerifyingKey`, `OpeningClaim`) bounds on the new verifier-only base trait, removing prover-only associated types from verifier compilation paths. Spec: `specs/1467-pcs-prover-verifier-split.md`.

Two commits, both under `crates/jolt-openings/` and the call sites that consume it:

1. `cbb47baf7` — full trait split + fused batching + ISA cutover
2. `ba8136ef3` — file rename `verification.rs` → `homomorphic.rs` (the file holds both prover- and verifier-side bodies for the homomorphic batched-opening protocol; old name was a leftover from an interim design)

## Trait shape

- **`CommitmentSchemeVerifier`** (new base): `Field`, `VerifierSetup`, `Proof`, `BatchProof`, `VerifierSetupParams`, `verifier_setup`, `verify_batch`, `bind_opening_inputs`
- **`CommitmentScheme: CommitmentSchemeVerifier`** (prover extension): `ProverSetup`, `Polynomial`, `OpeningHint`, `SetupParams`, `setup`, `project_verifier_setup`, `commit`, `prove_batch`
- **`AdditivelyHomomorphicVerifier: CommitmentSchemeVerifier`** (verifier extension): `combine`, `verify`
- **`AdditivelyHomomorphic: AdditivelyHomomorphicVerifier + CommitmentScheme`** (prover extension): `combine_hints`, `open`
- **`ZkOpeningSchemeVerifier`** + **`ZkOpeningScheme`**: same split for ZK openings

Single-claim `open` / `verify` moved off the core PCS traits onto `AdditivelyHomomorphic(Verifier)` so non-homomorphic / fused-batch schemes never have to implement them. They survive as RLC primitives consumed by `homomorphic_prove_batch` / `homomorphic_verify_batch` (which are the only production callers).

## Fused batched openings

[...]

PR #1521, opened 2026-05-12
jolt-openings crate API cutover

quangvdao:
## Summary

This PR now scopes the cutover to the extracted PCS crates only. The earlier `jolt-core` integration experiment was useful for finding the right abstraction boundary, but the merge target leaves legacy `jolt-core` untouched so it can remain the end-to-end reference implementation.

The intended consumer for this API is the upcoming Bolt-generated prover/verifier split, such as #1514. Generated role crates should be able to use `jolt-openings` / `jolt-dory` directly without going through the old in-core PCS trait family.

## What Changed

- Defines the verifier/prover PCS trait split in `crates/jolt-openings`.
- Adds backend-neutral commitment and opening source abstractions, including `natural_chunk_len` and separate linear-combination opening support.
- Moves ordinary homomorphic batch helpers into `jolt-openings`.
- Implements the extracted trait family for `jolt-dory`, preserving source-backed streaming behavior and ZK openings.
- Keeps `DoryScheme::BatchProof = Vec<DoryProof>`, with one proof per opening-point group.
- Keeps `jolt-hyperkzg` aligned with the same base trait family.
- Rewrites the spec around the crate-level API and future generated-role integration path.
- Removes the previous `jolt-core`, SDK, example, transpiler, CLI, book, and extractor diff surface from the PR.

## Validation

- `cargo fmt -q --check`
- `cargo check -p jolt-openings -q --features test-utils`
- `cargo check -p jolt-dory -q`
- `cargo check -p jolt-hyperkzg -q`
- `cargo nextest run -p jolt-openings --cargo-quiet --features test-utils`
- `cargo nextest run -p jolt-dory --cargo-quiet`
- `cargo nextest run -p jolt-hyperkzg --cargo-quiet`
- `cargo clippy -p jolt-openings -q --features test-utils --all-targets -- -D warnings`
- `cargo clippy -p jolt-dory -q --all-targets -- -D warnings`
- `cargo clippy -p jolt-hyperkzg -q --all-targets -- -D warnings`

Posted by Cursor assistant (model: GPT-5.5) on behalf of the user (Quang Dao) with approval.


specs/jolt-openings-crate.md
```diff
@@ -0,0 +1,468 @@
...
+4. Current BlindFold opening proof data flow and ZK verification behavior.
+5. Current `muldiv` behavior in standard and ZK modes.
+
+### Invariants
+
+1. `jolt-openings` remains backend-neutral and must not depend on `jolt-core`, `jolt-dory`, `dory`, arkworks, `common`, `tracer`, `jolt-sdk`, Akita, or Hachi.
+2. `jolt-openings` depends only on reusable leaf crates plus generic dependencies: `jolt-field`, `jolt-poly`, `jolt-transcript`, `jolt-crypto`, `serde`, `thiserror`, and `tracing`.
+3. The base verifier trait does not expose prover-only associated types such as `ProverSetup`, `Polynomial`, `OpeningHint`, or `SetupParams`.
+4. The base PCS traits expose fused batched openings through `prove_batch` and `verify_batch`.
+5. Single-claim `open` and `verify` are not on the base PCS trait.
```
moodlezoup: why? seems like they would make more sense on the base triat

specs/jolt-openings-crate.md
```diff
@@ -0,0 +1,468 @@
...
+### Dory Implementation
+
+`DoryScheme` should follow #1467's impl split while preserving current `main` correctness work.
+
+`CommitmentSchemeVerifier for DoryScheme`:
+
+1. `Field = Fr`.
+2. `VerifierSetup = DoryVerifierSetup`.
+3. `Proof = DoryProof`.
+4. `BatchProof = Vec<DoryProof>`.
```
moodlezoup: the relationship between `BatchProof` and the `AdditivelyHomomorphic` traits is a little unclear –– below it says that `prove_batch` and `verify_batch` delegate to their homomorphic counterparts, but then shouldn't `BatchProof = DoryProof`?

moodlezoup:
**Spec Analysis: `jolt-openings` PCS API Cutover**

| Dimension | Score | Gap |
|-----------|-------|-----|
| Goal | 0.90 | Clear — one-sentence goal stated in Intent; trait hierarchy and dependency direction fully spelled out in Design. |
| Constraints | 0.85 | Clear — 15 explicit invariants, 10 non-goals, dependency-tree checks. |
| Success Criteria | 0.85 | Clear — ~25 checkbox acceptance criteria with exact cargo invocations; testing strategy split across `jolt-openings`, `jolt-dory`, and `jolt-core`. |
| Context | 0.90 | Clear — names #1467 + branch `layerzero/quang/pcs-prover-verifier-split` as transplant source, references concrete files in current `main` (Stage 8, BlindFold, Dory layout). |
| **Ambiguity** | | **≈ 12.5%** |

**Verification against the current tree:**
- `crates/jolt-openings/src/schemes.rs` today defines `CommitmentScheme`, `AdditivelyHomomorphic`, `ZkOpeningScheme`, `StreamingCommitment`, but has no `CommitmentSchemeVerifier` and keeps `open`/`verify` on the base trait — matching the spec's premise.
- `crates/jolt-openings/src/reduction.rs` exposes the `reduce_prover` / `reduce_verifier` API the spec proposes to remove or demote.
- `crates/jolt-dory/src/scheme.rs` already implements the existing `jolt-openings` trait family, but `jolt-core` does **not** depend on `jolt-openings`/`jolt-dory` and still imports `jolt-core/src/poly/commitment/commitment_scheme.rs` (associated types: `Commitment`, `Proof`, `BatchedProof`, `OpeningProofHint`, plus `ZkEvalCommitment` and `StreamingCommitmentScheme` companions).
- `JoltProof.joint_opening_proof` is `PCS::Proof` today (`jolt-core/src/zkvm/proof_serialization.rs`); the spec's swap to `PCS::BatchProof` is well-defined.
- `DoryOpeningState::build_streaming_rlc` lives in `jolt-core/src/poly/opening_proof.rs`; Stage 8's BlindFold path (`OpeningProofData`, `y_com`, `y_blinding`) lives in `jolt-core/src/subprotocols/blindfold/mod.rs` and `jolt-core/src/poly/commitment/dory/commitment_scheme.rs`. The spec preserves each explicitly.

- New verifier-first trait hierarchy in `crates/jolt-openings/src/schemes.rs`: `CommitmentSchemeVerifier` → `CommitmentScheme`; `AdditivelyHomomorphicVerifier` → `AdditivelyHomomorphic`; `ZkOpeningSchemeVerifier` → `ZkOpeningScheme`; `StreamingCommitment` unchanged.
- `open`/`verify` removed from the base PCS trait, retained only on homomorphic extension traits.
- Fused `prove_batch` / `verify_batch` on the base traits; `homomorphic_prove_batch` / `homomorphic_verify_batch` in `homomorphic.rs` provide the default group-by-point RLC implementation; `reduce_prover` / `reduce_verifier` removed or demoted.
- `crates/jolt-dory` re-implements the split trait family, preserving current Dory hardening, ZK `y_com`/`y_blinding` behavior, transcript bridge, streaming commitment, and bounded proof deserialization.
- `jolt-core` adds `jolt-openings` + `jolt-dory` deps, swaps `PCS::Commitment` → `PCS::Output`, `joint_opening_proof: PCS::Proof` → `PCS::BatchProof`, and routes Stage 8 through `prove_batch` / `verify_batch` (one-group batch wrapping the current streaming RLC).

**Key invariants to preserve (all named in the spec):**
- `jolt-openings` stays leaf-crate-only (no `jolt-core`, `jolt-dory`, `dory`, arkworks, `common`, `tracer`, `jolt-sdk`, Akita, Hachi).
- Fiat–Shamir byte-identity between prover and verifier in the homomorphic batch helper (claim count, evaluations, per-point `rho` schedule).
- Stage 8 keeps current dense increment scaling, RA ordering, advice Lagrange scaling, `opening_ids`, `constraint_coeffs`, and `joint_claim` semantics.
- ZK mode still extracts and binds the Dory evaluation commitment for BlindFold.

**Critical evaluation criteria:**
- `cargo nextest run -p jolt-core muldiv` passes with both `--features host` and `--features host,zk`.
- `cargo nextest run -p jolt-openings --features test-utils` and `cargo nextest run -p jolt-dory` pass.
- `cargo clippy --all --features host -- -D warnings` and the matching `host,zk` invocation pass with zero warnings.
- `cargo tree -d` shows no duplicate Jolt leaf crates; `cargo tree -p jolt-openings` shows no edge into `jolt-core`/`jolt-dory`/`dory`/`tracer`/Akita.

**Minor soft spots (not blocking):**
- Performance is stated qualitatively ("should not regress", "no obvious runtime blowups"). For a mechanical refactor that explicitly preserves the existing streaming-RLC hot path, this is acceptable — `muldiv` timing in CI is the implicit guardrail.
- "`Claims<F>` … should remain unless a strictly mechanical replacement falls out" and "`dory_layout` should remain in this PR unless a small PCS config type already exists" are intentionally hedged. Both have a clear default (keep) and a bounded escape hatch — fine for a one-shot.

---

moodlezoup (commented):
this looks good to me, just a couple of minor comments. 
Also, while you're refactoring the PCS traits, perhaps you could also take a look at the relationship between the streaming and ZK traits? There's an awkward interaction there for Dory which I flagged [here](https://github.com/a16z/jolt/pull/1511#discussion_r3210291213)

quangvdao:
@moodlezoup thanks for the comments! I updated the spec to reflect the resolution, and the spec now has a redesign of the CommitmentScheme trait that bakes in streaming via a generic "polynomial source" rather than having a separate trait. Below is GPT-5.5's response after discussion:

__

First, agreed on single-claim openings: `open` / `verify` now live on the base PCS traits. The homomorphic traits are only responsible for the extra linear-combination operations (`combine`, `combine_hints`) needed by batch/RLC helpers. That makes the base trait a complete single-opening PCS interface, which is the cleaner split.

Second, the Dory batch-proof shape is now spelled out more explicitly. For Dory, `BatchProof = Vec<DoryProof>`, where each `DoryProof` corresponds to one opening-point group after the homomorphic helper RLC-combines that group’s claims. In the first Stage 8 cutover, Jolt currently forms one joint group, so this is a one-element `Vec<DoryProof>`. That should explain why `prove_batch` / `verify_batch` can delegate to the homomorphic helpers while still keeping the batch proof as a vector.

On the streaming/ZK interaction: I think the right fix is to remove `StreamingCommitment` as a public commitment-scheme abstraction rather than trying to make `finish` branch on ZK-ness. The new direction is source-oriented:

- `CommitmentSource` describes where polynomial data comes from.
- `CommitmentSourceBatch` describes a batch of such sources.
- `commit_batch` and `commit_batch_zk` are the PCS-level entry points.
- Dory overrides those methods to preserve the existing CycleMajor streaming behavior exactly: same padded trace scan, same row order, same one-hot row handling, same row-commitment hint structure, and no materialization of all witness polynomials.

So the ZK distinction moves to the PCS method boundary (`commit_batch` vs `commit_batch_zk`) rather than being hidden inside a streaming commitment object.

One subtlety: Stage 8 has a different “streaming” path from witness commitment streaming. Stage 8 currently builds a joint RLC polynomial directly from the trace/hints before opening, and we should preserve that optimized construction for this PR. The new API still applies there: Stage 8 should feed the resulting joint polynomial/claim through `prove_batch` / `verify_batch`. In other words, the public API changes, but we do not need to deeply redesign the Stage 8 opening accumulator or BlindFold constraint wiring as part of this cutover.

Net-net: I think this addresses the awkward Dory ZK/streaming interaction you flagged by making streaming a property of sources and batch commitment, not a separate commitment scheme trait.

moodlezoup:
> On the streaming/ZK interaction: I think the right fix is to remove `StreamingCommitment` as a public commitment-scheme abstraction rather than trying to make `finish` branch on ZK-ness. The new direction is source-oriented:
> 
> * `CommitmentSource` describes where polynomial data comes from.
> * `CommitmentSourceBatch` describes a batch of such sources.
> * `commit_batch` and `commit_batch_zk` are the PCS-level entry points.
> * Dory overrides those methods to preserve the existing CycleMajor streaming behavior exactly: same padded trace scan, same row order, same one-hot row handling, same row-commitment hint structure, and no materialization of all witness polynomials.

I guess I'm not convinced that streaming can be fully quarantined to the source implementation. It seems to me like streaming tends to have implications on the commitment algorithm itself, e.g. how we parallelize

quangvdao:
@moodlezoup yeah, I agree with this. I overstated it by saying streaming lives in the source implementation.

The distinction I want is closer to the compute/backend split in the Bolt work on `refactor/crates`: protocol-level code should be oblivious to the streaming/parallelization strategy, but the backend absolutely still owns that strategy.

So the intended split is:

- `CommitmentSource` / `BatchCommitmentSource` describe the data and traversal shape: what sources exist, how to evaluate them, and how to expose row-shaped access without materializing all witness polynomials.
- `PCS::commit_batch` / `commit_batch_zk` are the backend execution boundary.
- Dory overrides those methods precisely because streaming affects the commitment algorithm: row MSMs, one-hot grouping, row/source/aggregation parallelism, hint construction, and transparent-vs-ZK finishing.

So no, streaming is not fully quarantined inside the source. The source gives the PCS enough structure to avoid materialization; the PCS/backend decides how to consume that structure.

This also seems compatible with the Bolt direction. Bolt has compute-level obligations like `compute.oracle_dense_trace`, `compute.oracle_one_hot_chunk`, `compute.oracle_family_append`, and `compute.pcs_commit_batch`. Those describe oracle data/families and batch commitment obligations, while the eventual backend/provider decides whether to materialize, stream, use sparse one-hot paths, or target a different execution backend.

I updated the spec to call this out explicitly. The goal is that this PR gives Bolt a Rust PCS boundary it can lower into, without baking Dory’s current row-major streaming schedule into `jolt-openings`.

crates/jolt-openings/src/schemes.rs
```diff
@@ -78,23 +219,8 @@ where
     }
```
moodlezoup: Silent default that returns `OpeningHint::default()` will silently produce wrong opening proofs for any backend that implements `AdditivelyHomomorphic` but forgets to override `combine_hints`. `homomorphic_prove_batch` (homomorphic.rs:80) consumes the result without checks, and the verifier just sees a bad proof. Verifier-side `combine` has no default (good); making this required matches.

```suggestion
    fn combine_hints(
        hints: Vec<Self::OpeningHint>,
        scalars: &[Self::Field],
    ) -> Self::OpeningHint;
```

---

crates/jolt-dory/src/scheme.rs
```diff
@@ -277,166 +1060,600 @@ impl AdditivelyHomomorphic for DoryScheme {
...
             .collect();
 
-        DoryHint::new(combined, combined_blind)
+        let chunk_len = hints
+            .iter()
+            .map(|hint| hint.chunk_len)
+            .max()
+            .unwrap_or_default();
+        DoryHint::new(combined, combined_blind, chunk_len)
     }
```
moodlezoup, line 1048: Body is textually identical to `combine_hint_refs` at line 657 (modulo `Vec<&DoryHint>` vs `Vec<DoryHint>`). Any future change to hint combination has to be made twice. Express one via the other — either have the trait impl call the helper with `.iter().collect()`, or take `impl IntoIterator<Item: Borrow<DoryHint>>` in the helper.

---

crates/jolt-hyperkzg/src/scheme.rs
```diff
@@ -342,10 +375,86 @@ where
...
+                            dense[hot_index.get()] = F::from_u64(1);
+                        }
+                        evaluations.extend(dense);
+                    }
+                }
+            }
+        }
+    });
+    evaluations
+}
```
moodlezoup: This re-implements the same `SourceRow` enum dispatch as `materialize_source_evaluations` in `jolt-openings/src/sources.rs:438`, but with a different one-hot layout (column-major inline here vs hot-index-major accumulated in jolt-openings). Adding a new `SourceRow` variant requires updating two places, and the divergent one-hot handling means a multi-row one-hot source produces a different polynomial through HyperKZG vs the canonical helper. Easiest fix is making `materialize_source_evaluations` `pub` and deleting this copy; if the column-major shape was intentional for HyperKZG, leave a comment explaining why.

---

jolt-core/src/zkvm/prover.rs
```diff
@@ -194,18 +190,162 @@ pub struct JoltCpuProver<
...
+                poly_ids,
+                &coeffs,
+                advice_polys,
+            ));
+            *cache = Some(Stage8CachedRlc {
+                terms: terms.to_vec(),
+                chunk_len,
+                source,
+            });
+        }
```
moodlezoup: `advice_polys.take()` permanently moves the advice map out of the `Mutex<Option<...>>` on the first rebuild. If `fold_linear_rows` is ever invoked a second time with a different `terms` or `chunk_len` (cache miss), the second rebuild gets `None` and `unwrap_or_default()` silently substitutes an empty `HashMap` — advice contributions vanish from the RLC with no error.

Current Dory path is safe because `prove_source_backed_batch_with_mode` calls `vector_matrix_product` with a single fixed `(terms, chunk_len)`. But `Stage8OpeningSource::fold_rows` at line 326 builds a single-element `terms` per source, so anything that routes through the default `prove_batch_opening` and hits this from multiple sources will silently drop advice starting on the second call. Worth either making the advice map cloneable so rebuilds remain consistent, or `debug_assert!(cache.is_none())` here so a second rebuild fails loudly.

---

jolt-core/src/zkvm/prover.rs
```diff
@@ -194,18 +190,162 @@ pub struct JoltCpuProver<
...
+    fn evaluate(&self, _point: &[F]) -> F {
+        panic!("Stage8OpeningSource is only used through source-backed batch opening")
+    }
+
+    fn for_each_row<V>(&self, _chunk_len: usize, _visit: V)
+    where
+        V: for<'row> FnMut(usize, SourceRow<'row, F>),
+    {
+        panic!("Stage8OpeningSource is only used through source-backed batch opening")
+    }
```
moodlezoup: Panicking `evaluate`/`for_each_row` means any PCS that uses the default `prove_batch_opening` against `Stage8OpeningSourceBatch` (which routes per-id through `source(id)`) will crash at runtime instead of failing at compile time. Dory overrides `prove_batch_opening`, so this is latent today, but it's a non-obvious coupling between this source impl and the chosen backend. Worth a doc comment noting the requirement that the backend implement source-backed batch opening directly.

---

moodlezoup (commented):
Reviewed the jolt-openings cutover. The Stage 8 prover/verifier symmetry holds, transcript binding order is preserved, and the ZK/standard mode split looks consistent end-to-end. Most flags below are footguns introduced by the new trait surface (silent defaults and duplicated helpers) that will bite a future PCS backend or a future SourceRow variant rather than the current Dory-only path.

---

crates/jolt-crypto/src/ec/bn254/gt.rs
```diff
@@ -209,3 +211,33 @@ impl<'de> serde::Deserialize<'de> for Bn254GT {
         Ok(Self(inner))
     }
 }
+
+impl CanonicalSerialize for Bn254GT {
```
moodlezoup, line 215: can we just `#[derive(CanonicalSerialize, CanonicalDeserialize)]` on `Bn254GT` instead of manually implementing these?

crates/jolt-openings/src/claims.rs
```diff
@@ -13,8 +15,179 @@ pub struct ProverClaim<F: Field, P = Polynomial<F>> {
...
+
+/// Opening point pair for protocols whose public point order differs from the
+/// backend's proof point order.
+///
+/// `public` is the point that belongs to the surrounding protocol transcript
+/// and output relation. `proof` is the coordinate order consumed by the PCS
+/// proof algorithm. Most schemes set both fields equal; Dory uses `proof` for
+/// its private row-major opening order while Jolt keeps `public` in protocol
+/// order.
+#[derive(Clone, Debug, PartialEq, Eq)]
```
moodlezoup, line 36: I don't really like this; I feel like the PCS implementation should just expect one coordinate order, and it should be responsible for any reordering it might need

crates/jolt-openings/src/claims.rs
```diff
@@ -13,8 +15,179 @@ pub struct ProverClaim<F: Field, P = Polynomial<F>> {
...
+    pub point: Vec<F>,
+    pub value: BatchOutputValue<F, HidingCommitment>,
+}
+
+/// Expression describing how a PCS output is derived from raw claim terms.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum BatchOutputExpression<F: Field, ClaimId> {
+    /// Linear relation `output = Σ coefficient_i * claim_i`.
+    Linear(Vec<(ClaimId, F)>),
+}
```
moodlezoup, line 126: can we fold this into `BatchOutputRelation` instead of introducing a new type?

crates/jolt-openings/src/homomorphic.rs
```diff
@@ -0,0 +1,337 @@
...
+
+fn source_evaluations<F, S>(source: &S) -> Vec<F>
+where
+    F: Field,
+    S: CommitmentSource<F> + ?Sized,
+{
+    materialize_source_evaluations(source)
+}
+
+type ProverPointGroup<F, PCS, S> = Vec<(Vec<F>, Vec<ProverClaimWithHint<F, PCS, S>>)>;
```
moodlezoup, line 244: i feel like a map makes more sense here?

crates/jolt-openings/src/homomorphic.rs
```diff
@@ -0,0 +1,337 @@
...
+            group.push((claim, hint));
+        } else {
+            let point = claim.point.clone();
+            groups.push((point, vec![(claim, hint)]));
+        }
+    }
+    groups
+}
+
+type OpeningPointGroup<F, PCS> = Vec<(Vec<F>, Vec<OpeningClaim<F, PCS>>)>;
```
moodlezoup, line 267: same here

crates/jolt-openings/src/schemes.rs
```diff
@@ -1,100 +1,230 @@
...
+}
+
+/// Verifier setup derivable from public parameters without prover setup.
+///
+/// Transparent schemes such as Dory can build verifier setup from a size
+/// parameter alone. Structured-reference-string schemes such as KZG generally
+/// cannot: their verifier setup contains trapdoor-derived elements generated
+/// during setup, so verifier-only code should receive the verifier setup as an
+/// input rather than pretending it can derive it from public generators.
+pub trait PublicVerifierSetup: CommitmentSchemeVerifier {
```
moodlezoup: why have a separate trait here?

crates/jolt-openings/src/schemes.rs
```diff
@@ -1,100 +1,230 @@
...
+    type ProverSetup: Clone + Send + Sync;
+    type OpeningHint: Clone + Send + Sync + Default;
     type SetupParams;
 
+    /// Builds prover and verifier setup.
     fn setup(params: Self::SetupParams) -> (Self::ProverSetup, Self::VerifierSetup);
 
-    fn verifier_setup(prover_setup: &Self::ProverSetup) -> Self::VerifierSetup;
+    /// Projects prover setup down to verifier setup.
+    fn project_verifier_setup(prover_setup: &Self::ProverSetup) -> Self::VerifierSetup;
```
moodlezoup:
```suggestion
    fn prover_to_verifier_setup(prover_setup: &Self::ProverSetup) -> Self::VerifierSetup;
```

crates/jolt-openings/src/sources.rs
```diff
@@ -0,0 +1,712 @@
...
+    /// This preserves the common compact-polynomial benchmark and commitment
+    /// path without paying the cost of first converting every row entry into a
+    /// full-width field element.
+    U64(&'a [u64]),
+
+    /// A strided unsigned-integer row embedded canonically into the field.
+    StridedU64 {
+        values: &'a [u64],
+        column_stride: usize,
+    },
```
moodlezoup: I think I slightly prefer to collapse this into a single variant, since `U64` is just the special case of `StridedU64` with a stride of 1. Same for `FieldElements` and `I128`. wdyt?

0xAndoroid:
Closing as superseded: the refactor/audit-prep stack (#1544, #1635, #1654) landed the crate-level PCS API on main with a different design (kept StreamingCommitment over the CommitmentSource abstraction, verifier split via jolt-crypto's Commitment trait + the jolt-verifier crate, batching via BatchOpeningScheme/HomomorphicBatch + packed openings), and the intended consumers now exist as jolt-prover / jolt-verifier. Verified nothing here remains to cherry-pick — the branch is 122 commits behind with all its unique surface superseded.

PR #1518, opened 2026-05-08
Refactor RV64 bytecode expansion pipeline

quangvdao:
Posted by Cursor assistant (model: GPT-5) on behalf of the user (Quang Dao) with approval.

## Summary
- make the program pipeline consistently RV64-only after the previous tracer cleanup and ELF32/RV32 rejection
- split instruction identity by phase with `RiscvInstructionKind`, `JoltInstructionKind`, and `LookupInstructionKind`
- refactor `jolt-program::expand` into a compiler-style lowering pass: concrete lowerers record readable `ExpansionOp` recipes, and owned `ExpansionState` materializes allocator state, recursive helper expansion, temp lifetimes, capacity checks, and metadata stamping centrally
- keep concrete instruction lowerings readable as ordered `asm.emit_*`, `asm.expand_*`, `asm.allocate`, and `asm.release` snippets while removing allocator borrows from the builder
- feature-gate serde/ark serialization behind default `serialization` features so the provider-free expansion core can be extracted with `--no-default-features`
- add compact parity validation for provider-free expansion without checking in giant expanded-row fixtures

## Extraction status
- Hax extraction now succeeds for `jolt_program::expand` with `cargo hax -C -p jolt-program --no-default-features`, producing Lean under `/tmp/jolt-program-hax-lean-refactor`
- the extracted core no longer pulls in serde/ark serialization, the old generic `Into<RegisterOperand>` helper noise, the CSR helper name collision, or the dynamic Vec-backed temp-table update shape
- the generated Lean still does not compile end-to-end; remaining blockers are concentrated in Hax/Lean support for generated `while_loop_return` forms, very large nested lowerer definitions, and helper definitions that fail to enter the environment after earlier Lean errors

## Bug fixes from parity validation
- fixed word AMO min/max temp allocation so recursive helper expansion reuses the same virtual registers as main
- fixed `ECALL` temp allocation/materialization order so its final expanded rows match main exactly

## Verification
- `cargo fmt -q`
- `cargo clippy -p jolt-program --no-default-features -q --all-targets -- -D warnings`
- `cargo clippy -p jolt-program -q --all-targets -- -D warnings`
- `cargo nextest run -p jolt-program --cargo-quiet`
- `cargo clippy --all --features host -q --all-targets -- -D warnings`
- `cargo clippy --all --features host,zk -q --all-targets -- -D warnings`
- `cargo nextest run -p jolt-core muldiv --cargo-quiet --features host`
[...]

PR #1522, opened 2026-05-12
Split source/Jolt instruction catalogs

quangvdao:
Posted by Cursor assistant (model: GPT-5.5) on behalf of the user (Quang Dao) with approval.

## Summary

This PR implements the source/final instruction split for Jolt. Decoded RV64/source-program instructions now flow through source row types, expansion emits final Jolt bytecode rows, and proving/lookup metadata is only queried from final `JoltInstructionRow` values.

The PR started as the implementation spec, but it is now an implementation PR: the spec remains in-tree as the design record and checklist for the cutover.

## What Changed

- Added phase-specific instruction rows and enums in `jolt-riscv`: `SourceInstructionRow`, `SourceInstruction<T>`, `SourceInstructionKind`, `JoltInstructionRow`, `JoltInstruction<T>`, and `JoltInstructionKind`.
- Removed the old shared `NormalizedInstruction` / broad mirrored instruction boundary.
- Made `SourceInstructionKind` and `JoltInstructionKind` distinct universes. Source includes decode/expansion inputs such as `ADDW`, narrow loads/stores, atomics, CSRs, traps, and `Inline`; final Jolt includes only rows admitted into expanded bytecode and proving tables.
- Added canonical names, stable final-row tags, profile-local dense indexes, shipped profiles (`RV64IM_JOLT`, `RV64IMAC_JOLT`, `RV64IMAC_JOLT_ALL_INLINES`), and profile/catalog fingerprints.
- Updated RV64 decode and `jolt-program::expand` so decode produces source rows and expansion/materialization emits validated final `JoltInstructionRow` bytecode.
- Reworked expansion recipes to distinguish source helper expansion from final-row emission while preserving readable instruction-sequence style.
- Moved inline dispatch identity to source rows with `SourceInlineKey { opcode, funct3, funct7 }`; final `JoltInstructionKind::Inline` is gone.
- Updated inline provider boundaries so registered inlines accept source inline rows, declare `InlineExtension`s, and return validated final Jolt rows.
- Threaded selected `JoltInstructionProfile`s through decode, expansion, bytecode preprocessing, execution rows, and serialized preprocessing metadata.
- Updated tracer conversions so source conversion and final-row execution paths are phase-aware while tracer remains the owner of concrete execution semantics.
- Added `JoltTraceCycle<'a>` in `jolt-core` as the current prover-side adapter from raw tracer `Cycle` values to final-row proof metadata.
- Removed source/tracer proof-metadata leakage: aggregate `Flags for Instruction`, `InstructionLookup for Instruction`, and `InstructionLookup for Cycle` are gone; lookup-table metadata is constrained to final-row payloads.
- Updated R1CS, Spartan instruction-input, instruction lookup, ZkLean extraction, and Z3 extraction call sites to use final rows for flags/table routing.
- Added the `jolt-eval` source-to-expanded-Jolt equivalence invariant and updated tests around phase separation, profile legality, canonical names/tags, dense indexes, inline expansion, and proof-metadata boundaries.
- Updated `specs/compiler-native-bytecode-expansion.md` and added/maintained `specs/source-jolt-instruction-split.md` as the implementation record.

## Current Follow-Ups

- `JoltTraceCycle` is intentionally an interim boundary. The cleaner endpoint is to normalize once after tracing into a proof trace row equivalent to `jolt_program::execution::TraceRow`, then make prover internals consume that instead of `Arc<Vec<tracer::instruction::Cycle>>`.
- `jolt-riscv` still owns some final-row flag bitfield machinery. Lookup/flag queries are now final-row-only, but longer term we may want the RV64 catalog crate to know even less about Jolt proving metadata.
- `z3-verifier` package tests still expose existing incomplete/weak constraints in that crate; the main Jolt proof checks pass.

## Validation

Passed locally:

- `cargo fmt -q`
- `cargo nextest run -p jolt-program --cargo-quiet`
- `cargo nextest run -p tracer --cargo-quiet --features test-utils`
- `cargo nextest run -p jolt-lookup-tables --cargo-quiet`
- focused Jolt-core proof-metadata boundary tests
- ZkLean extraction smoke run


moodlezoup:
**Spec Analysis: Split source/Jolt instruction catalogs** (`specs/source-jolt-instruction-split.md`)

| Dimension | Score | Gap |
|-----------|-------|-----|
| Goal | 0.78 | Type-level relationship between `SourceInstructionKind` variants and the `SourceInline { opcode, funct3, funct7 }` payload is contradictory |
| Constraints | 0.65 | Compile-time profile selection mechanism + `inventory` interplay is undefined |
| Success Criteria | 0.70 | Fixture-hash regeneration is gated on an unspecified equivalence check; no `jolt-eval` invariant proposed |
| Context | 0.85 | Strong — cites #1518, names crates, lists each extension's members |
| **Ambiguity** | | **~26%** |

**Status: Questions remain** — 4 ambiguities to resolve before a one-shot implementation. The spec is thorough and well-grounded in the post-#1518 code (`crates/jolt-riscv/src/{kind,normalized}.rs`, `crates/jolt-program/src/expand/mod.rs:46`, `tracer/src/instruction/inline.rs:77`), but a few load-bearing decisions are still open.

---

**Questions:**

**1. [Goal] How does `SourceInstructionKind` represent registered inlines — one `Inline` variant + payload, or one variant per registered inline op?**

The Design section is internally inconsistent:

- The acceptance criterion `Generated SourceInstructionKind contains decoded RV64, Jolt custom source opcodes, **and registered inline opcodes** selected by the source catalog/profile` reads as "one `SourceInstructionKind` variant per registered inline" (e.g., `Sha256Inline`, `Keccak256Inline`, …).
- The struct sketch `SourceInstruction { kind: SourceInstructionKind, … inline: Option<SourceInline> }` plus `SourceInline { opcode, funct3, funct7 }` reads as "one universal `SourceInstructionKind::Inline` variant carrying dispatch metadata", matching today's `Instruction::INLINE { opcode, funct3, funct7 }` in `tracer/src/instruction/inline.rs`.
- The current generated `SourceInstructionKind::Inline` variant (`crates/jolt-riscv/src/kind.rs:25`) supports the second reading, but the catalog framing (`InlineExtension::Sha2`, `InlineExtension::Keccak256`, …) implies the first.

Which is it? If per-inline variants, what is `SourceInline`'s role — only for unregistered inlines decoded from the custom opcode space, or always present and redundant with `kind`? If one universal `Inline`, what does `InlineExtension::Sha2` actually gate — only `inventory` registration availability, or `SourceInstructionKind` membership?

**2. [Constraints] What is the compile-time profile selection mechanism, and does it shrink the generated enums?**

The spec says profiles must be `compile-time selections … visible to the Rust type system and Cargo feature/build configuration`, and shows `RV64IMAC_JOLT` as a `const JoltInstructionProfile`. But it doesn't specify:

- **Generated enum scope.** Does selecting `RV64IM_JOLT` (no `Rv64A`, no `Rv64C`) generate a *smaller* `JoltInstructionKind` enum than `RV64IMAC_JOLT` (no `LR/SC/AMO*` variants), or is the enum the union of all `JoltTargetExtension` members and the profile only gates legality at preprocessing? The acceptance criterion `Generated JoltInstructionKind contains only target-legal rows` is ambiguous between these. The former breaks cross-profile proof artifact compatibility and downstream `match` exhaustiveness; the latter contradicts "generated identifiers over those selected catalog entries".
- **API entry point.** What consumes the chosen `&'static JoltInstructionProfile`? Cargo feature gates on `jolt-program` (e.g. `--features profile-rv64imac`), a const generic threaded through `JoltProgram`/preprocessing keys, a type-level marker like `profile::Rv64imacJolt`, or a value passed at `JoltProgram::new`?
- **Inline inventory interplay.** Today `jolt-inlines-*` crates self-register via `inventory`. If a workspace links `jolt-inlines-sha2` but the active profile is `RV64IMAC_JOLT` (empty `inline_extensions`), is `SHA256_INLINE` excluded by (a) refusing to compile the crate, (b) validating registered inventory against the profile at startup, or (c) decode rejection?

Step 2 of Execution ("Split the instruction macro/catalog so it generates source and target enums from explicit source/target membership") doesn't disambiguate.

**3. [Success Criteria] What concretely gates regenerating the expansion parity fixture?**

`crates/jolt-program/src/expand/fixtures/main_expand_parity_hashes.json` hashes `NormalizedInstruction` byte serialization. After this PR, `JoltInstruction` will at minimum lose the `Inline` variant and likely reorder discriminants (Rv64-only kinds split off, virtuals partitioned across extensions), so every hash will change for purely-mechanical reasons. The spec says `update the expansion fixture/hash if the row type serialization changes but semantic output is unchanged` — but what concretely proves "semantic output is unchanged" before the hash is overwritten?

Options: a manual one-time `cargo run` against `main` capturing pre-PR sequences and a Rust-level structural diff (modulo kind renaming) committed as a test; a new equivalence harness invoked from the test; or an unstated reviewer trust step. Without a concrete check, the fixture update becomes the load-bearing correctness gate and silently weakens what was a tight regression test.

**4. [Success Criteria] Should source→final expansion equivalence become a `jolt-eval` invariant?**

`jolt-eval` already hosts the `soundness` red-team invariant and reference-vs-optimized invariants like `split_eq_bind_low_high`. This refactor changes the type representation of every instruction lowering and is therefore a natural fit for an invariant of the form "for every program in the corpus, the `JoltInstruction` stream produced by the new pipeline matches the pre-refactor `NormalizedInstruction` stream modulo `JoltInstructionKind::Inline` removal and discriminant renaming". The Evaluation section lists fixture parity but doesn't engage `jolt-eval` at all. Is the position that fixture-hash parity is sufficient and no `jolt-eval` invariant is warranted — and if so, why? If a `jolt-eval` invariant is in scope, please name it and where it lives.

---

specs/source-jolt-instruction-split.md
```diff
@@ -0,0 +1,694 @@
...
+
+```rust
+pub struct InstructionCatalogEntry {
+    pub name: InstructionName,
+    pub phase: InstructionPhase,
+    pub source: Option<SourceInstructionSpec>,
+    pub target: Option<JoltInstructionSpec>,
+    pub side_effects: bool,
+    pub lookup: LookupSupport,
+}
```
moodlezoup: I don't really understand this struct. Static metadata like `side_effects` and `lookup` should be declared at compile time

specs/source-jolt-instruction-split.md
```diff
@@ -0,0 +1,694 @@
...
+deliberately and consistently.
+
+Use these names unless implementation uncovers a concrete conflict:
+
+```rust
+pub enum InstructionPhase {
+    SourceOnly,
+    TargetOnly,
+    SourceAndTarget,
+}
```
moodlezoup: I don't understand the point of this enum

specs/source-jolt-instruction-split.md
```diff
@@ -0,0 +1,694 @@
...
+```
+
+The exact operand/spec type names can change, but the ownership relationship
+should not: catalog entries define what exists, selected profiles define what is
+enabled, and `SourceInstructionKind` / `JoltInstructionKind` are generated
+identifiers over those selected catalog entries. Source operands may continue to
+use normalized register fields for ordinary rows; inline dispatch metadata
+should not be stored in an immediate field.
+
+### Catalog And Profiles
```
moodlezoup: This part seems is cool but seems unnecessary for now, consider cutting from the scope

specs/source-jolt-instruction-split.md
```diff
@@ -0,0 +1,694 @@
...
+Registered inline providers may emit ordinary target rows plus virtual helper
+rows, but provider output must be validated against the selected
+`JoltTargetExtension` set before preprocessing.
+
+The long-term shape is profile-driven:
+
+```rust
+pub struct JoltInstructionProfile {
+    pub source_extensions: &'static [SourceExtension],
+    pub target_extensions: &'static [JoltTargetExtension],
```
moodlezoup: In theory, the target profile can be computed from source + inlines, right? it's just the minimal set cover of instructions that can be emitted from the given source + inline extensions

moodlezoup (commented):
Somewhat related to the change proposed here, one thing I'd like to address is the dispersal of per-instruction implementations/metadata across the codebase:

- `crates/jolt-program/src/image/decode.rs`: instruction format and operands dispatch
- `crates/jolt-riscv/src/kind.rs`: has_side_effects dispatch
- `crates/jolt-program/src/expand/mod.rs`: virtual sequence expansion dispatch
- `crates/jolt-program/src/expand/grammar.rs`: is_source_only

this is I think in part due to the gap between the `jolt_instruction!` structs declared in `jolt-riscv` and the "kind" enums (e.g. `JoltInstructionKind`). imo all of this per-instruction behavior should be implemented on the former, with the enum variants of the latter dispatching to its associated struct as necessary. I have some ideas about how to consolidate and colocate more of the instruction-related code, which I can try to implement on this branch

quangvdao:
Thanks for the comments! I workshopped your concern with GPT-5.5 and synthesized the (AI-generated) answer below:

Posted by Cursor assistant (model: GPT-5.5) on behalf of the user (Quang Dao) with approval.

I agree with the main concern: we currently have too many independent places that each “know” something about the same instruction:

- decode knows `ADD` is R-format;
- `kind.rs` knows whether `ADD` has side effects;
- expansion knows whether `ADDW` is source-only and how to lower it;
- lookup/proving knows whether a final row participates in lookup metadata;
- tracer knows how to actually execute the instruction.

The part I want to be careful about is **where those facts live**.

I don’t think the right end state is one giant `jolt_instruction!` declaration in `jolt-riscv` that contains decode facts, expansion facts, lookup/proof facts, circuit flags, and tracer execution facts. That would make `jolt-riscv` know too much about downstream crates. In particular, I’d like `jolt-riscv` to stay mostly about stable instruction identity and row shape, not about proving-system or tracer implementation details.

The model I’m leaning toward is:

> The instruction marker struct is the shared identity, but each crate owns the facts for its own layer.

So `Add`, `AddW`, `VirtualHostIO`, etc. are the common anchor points. But the metadata is not all centralized in one place.

For example, `jolt-riscv` would define the stable identities and generated enums:

```rust
// jolt-riscv
pub struct Add<T = ()>(pub T);
pub struct AddW<T = ()>(pub T);
pub struct VirtualHostIO<T = ()>(pub T);

pub enum SourceInstructionKind {
    ADD,
    ADDW,
    VirtualHostIO,
    Inline,
}

pub enum JoltInstructionKind {
    ADD,
    VirtualHostIO,
}
```

There still needs to be a small registry somewhere, because Rust macros cannot discover all instruction declarations across files automatically. But I think that registry should be minimal: marker name, enum variant, stable discriminant, maybe module path. It should not duplicate semantic metadata.

Something like:

```rust
instruction_registry! {
    ADD => Add, discriminant = 0x0001;
    ADDW => AddW, discriminant = 0x0002;
    VirtualHostIO => VirtualHostIO, discriminant = 0x0100;
}
```

Then each crate decorates those marker structs with the facts it owns.

For decode, `jolt-program` owns the encoding and operand format:

```rust
// jolt-program
trait DecodeSpec {
    const SOURCE_KIND: SourceInstructionKind;
    const FORMAT: SourceFormat;
    const ENCODING: SourceEncoding;
}

impl DecodeSpec for Add<()> {
    const SOURCE_KIND: SourceInstructionKind = SourceInstructionKind::ADD;
    const FORMAT: SourceFormat = SourceFormat::R;
    const ENCODING: SourceEncoding = SourceEncoding::R {
        opcode: 0b0110011,
        funct3: 0b000,
        funct7: 0b0000000,
    };
}

impl DecodeSpec for AddW<()> {
    const SOURCE_KIND: SourceInstructionKind = SourceInstructionKind::ADDW;
    const FORMAT: SourceFormat = SourceFormat::R;
    const ENCODING: SourceEncoding = SourceEncoding::R {
        opcode: 0b0111011,
        funct3: 0b000,
        funct7: 0b0000000,
    };
}
```

Then `decode_instruction` can be generated or table-driven from `jolt-program`’s decode specs:

```rust
// jolt-program
pub fn decode_instruction(
    word: u32,
    address: u64,
    is_compressed: bool,
) -> Result<SourceInstruction, ProgramError> {
    let kind = match decode_key(word) {
        Add::<()>::ENCODING_KEY => SourceInstructionKind::ADD,
        AddW::<()>::ENCODING_KEY => SourceInstructionKind::ADDW,
        // ...
        _ => return Err(ProgramError::InvalidInstruction),
    };

    let operands = parse_operands(kind, word)?;
    Ok(SourceInstruction { kind, address, operands, ... })
}
```

For expansion, `jolt-program` also owns the lowering behavior:

```rust
// jolt-program
trait SourceExpansion {
    fn expand_source(
        source: &SourceInstruction,
        allocator: &mut ExpansionAllocator,
    ) -> Result<Vec<JoltInstruction>, ExpansionError>;
}

impl SourceExpansion for AddW<()> {
    fn expand_source(
        source: &SourceInstruction,
        allocator: &mut ExpansionAllocator,
    ) -> Result<Vec<JoltInstruction>, ExpansionError> {
        expand_addw(source, allocator)
    }
}
```

And the dispatcher can be generated from the registry, but still call the marker-struct implementations:

```rust
// jolt-program
pub fn expand_instruction(
    source: &SourceInstruction,
    allocator: &mut ExpansionAllocator,
) -> Result<Vec<JoltInstruction>, ExpansionError> {
    match source.kind {
        SourceInstructionKind::ADD => emit_direct(source),
        SourceInstructionKind::ADDW => AddW::expand_source(source, allocator),
        SourceInstructionKind::Inline => inline_provider.expand_inline(source, allocator),
        _ => Err(ExpansionError::UnsupportedInstruction),
    }
}
```

For tracer execution, tracer remains the owner:

```rust
// tracer
impl Execute for tracer::instruction::ADD {
    fn execute(&self, cpu: &mut Cpu, cycle: &mut Cycle) {
        // mutate registers / memory / PC here
    }
}
```

I would not move this into `jolt-riscv`.

For lookup/proof metadata, I also don’t think `jolt-riscv` should necessarily own everything. A lookup/proving crate can attach its own metadata to the same marker structs:

```rust
// lookup/proving crate
trait LookupMetadata {
    fn lookup_table(kind: JoltInstructionKind) -> Option<LookupTableKind>;
    fn circuit_flags(kind: JoltInstructionKind) -> CircuitFlagSet;
}
```

Or, if we want marker-struct dispatch:

```rust
impl LookupMetadata for Add<()> {
    const LOOKUP: Option<LookupTableKind> = Some(LookupTableKind::Add);
}

impl LookupMetadata for VirtualHostIO<()> {
    const LOOKUP: Option<LookupTableKind> = None;
}
```

The important part is that this metadata lives in the crate that owns the concern, not necessarily in `jolt-riscv`.

So I think I agree with your direction, but I’d phrase the design slightly differently:

- Use the instruction marker structs as the stable identity.
- Keep a minimal registry for closed enum generation and stable discriminants.
- Let each crate provide crate-local metadata/behavior for those marker structs.
- Generate dispatch from the registry where possible, so we don’t keep hand-maintained parallel `match JoltInstructionKind` lists.
- Do not make `jolt-riscv` a central dumping ground for decode, expansion, lookup, flags, and tracer semantics.

This also answers the profile point: the final target set should be computed from the selected source profile plus enabled inline extensions. But that computation should produce a positive set of target-legal rows. It should not be `!is_source_only`, because that makes forgotten source-only rows accidentally target-legal.

For inlines, I think the split should be:

```rust
pub enum SourceInstructionKind {
    // ...
    Inline,
}

pub struct SourceInline {
    pub opcode: u32,
    pub funct3: u32,
    pub funct7: u32,
}
```

`InlineExtension::Sha2` should not create a new `SourceInstructionKind` variant. It should gate whether the `(opcode, funct3, funct7)` key is accepted and whether the provider is expected to exist.

One concrete edge case: if `jolt-inlines-sha2` is linked but the selected profile does not enable `InlineExtension::Sha2`, the provider might exist in inventory, but decode/expansion should reject that inline key. Conversely, if the profile enables Sha2 but the provider crate is not linked, startup validation should fail early with a clear “profile enables Sha2 but provider is not registered” error.

So the spec should probably say: registered provider existence and profile permission are separate checks; expansion requires both.

That’s the design I think is both closer to your concern and still preserves the crate boundaries I care about.

moodlezoup:
I generally agree with the updated approach, the only part that I'm not sure about is the registry macro. 
Instead of having the registry own the enum variant -> struct mapping, why not just have the enum variants carry the struct themselves?

So something like:

```rust
pub enum SourceInstructionKind<T = SourceInstruction> {
    ADD(Add<T>),
    ADDW(Addw<T>),
    // ... 
}

// Similar for JoltInstructionKind
``` 

This would also allow us to flip the relationship between the `SourceInstructionKind` enum and the `SourceInstruction` struct. Instead of having a `kind: SourceInstructionKind` field on the struct, we could instead use `SourceInstructionKind<SourceInstruction>` as the concrete, universal source-instruction type. 

This is similar to the pattern I used for the `LookupInstruction` enum in order to achieve the "impl on instruction struct" + "enum dispatch" behavior that we want (see e.g. the `impl_jolt_instructions_flags` macro)

quangvdao:
Thanks, I think this is the right direction. I revised the spec to adopt the `LookupInstruction`-style shape you suggested.

Posted by Cursor assistant (model: GPT-5.5) on behalf of the user (Quang Dao) with approval.

```rust
pub enum SourceInstruction<T = SourceRow> {
    ADD(Add<T>),
    ADDW(AddW<T>),
    Inline(Inline<T>),
    // ...
}

pub enum JoltInstruction<T = JoltRow> {
    ADD(Add<T>),
    LD(Ld<T>),
    VirtualSignExtendWord(VirtualSignExtendWord<T>),
    // ...
}
```

So the enum variants now carry the marker structs directly, instead of having a separate registry own the variant -> struct mapping. That seems cleaner and closer to the “impl on instruction struct + enum dispatch” pattern already used by `LookupInstruction`.

The one nuance I added is around profiles. I think profiles should be MLIR-style legality layers over a stable operation universe, rather than generating different Rust enum shapes per profile. So `SourceInstruction<T>` / `JoltInstruction<T>` stay universal shipped enums, and a selected profile determines:

- which source rows decode;
- which inline keys are accepted;
- which final rows are legal after expansion.

That means `RV64IM_JOLT` rejects atomics through source legality and computes a smaller target legality closure, but it does not produce a different Rust enum from `RV64IMAC_JOLT`.

This keeps the marker-struct dispatch you want, while avoiding profile-dependent exhaustive matches downstream. It also keeps the ownership boundaries we discussed: `jolt-program` owns decode/expansion metadata, tracer owns execution semantics, and lookup/proving metadata stays out of `jolt-riscv`.

crates/jolt-program/src/expand/mod.rs
```diff
@@ -62,50 +66,72 @@ pub struct NoInlineExpansionProvider;
...
     } else {
         instruction
     };
+    let source = *instruction.row();
 
-    let result = if instruction.instruction_kind == JoltInstructionKind::Inline {
-        let rows = inline_provider.expand_inline(instruction, allocator)?;
-        finalize_inline_provider_rows(*instruction, allocator, rows)
+    let result = if instruction.kind() == SourceInstructionKind::Inline {
+        let instructions = inline_provider.expand_inline(instruction, allocator, profile)?;
```
moodlezoup: If `expand_inline` errors here, `?` short-circuits the whole function and the rd=x0 cleanup at lines 138–140 is skipped, leaking the placeholder virtual register in the caller's allocator. The non-Inline branch below binds the result first and runs the cleanup unconditionally — same pattern is needed here.

```suggestion
        let instructions = match inline_provider.expand_inline(instruction, allocator, profile) {
            Ok(instructions) => instructions,
            Err(err) => {
                if let Some(register) = allocated_rd_zero_register {
                    let _ = allocator.release(register);
                }
                return Err(err);
            }
        };
```

---

crates/jolt-program/src/execution/mod.rs
```diff
@@ -31,9 +36,14 @@ pub fn build_jolt_program(elf_bytes: &[u8]) -> Result<JoltProgram, ProgramError>
 pub fn build_jolt_program_with_inline_provider<P: InlineExpansionProvider + ?Sized>(
     elf_bytes: &[u8],
     inline_provider: &mut P,
+    profile: JoltInstructionProfile,
 ) -> Result<JoltProgram, ProgramError> {
-    let image = decode_elf(elf_bytes)?;
-    let expanded_bytecode = expand_program_with_provider(&image.instructions, inline_provider)?;
+    let image = decode_elf(elf_bytes, RV64IMAC_JOLT)?;
```
moodlezoup: The decoder is pinned to `RV64IMAC_JOLT` while the expander uses the caller-supplied `profile`. A caller that passes a restrictive profile (e.g. `RV64IM_JOLT`) still gets an `RV64IMAC_JOLT`-shaped image; in the other direction, an `RV64IMAC_JOLT_ALL_INLINES` expansion sees a decoder that may have rejected inline source rows. Thread `profile` into `decode_elf` so both halves agree.

---

jolt-core/src/zkvm/instruction/mod.rs
```diff
@@ -303,15 +343,54 @@ impl<const XLEN: usize> InstructionLookup<XLEN> for NormalizedInstruction {
             }
             JoltInstructionKind::LD
             | JoltInstructionKind::SD
-            | JoltInstructionKind::EBREAK
-            | JoltInstructionKind::ECALL
             | JoltInstructionKind::FENCE
             | JoltInstructionKind::VirtualHostIO => return None,
             _ => return None,
```
moodlezoup: The `LD | SD | FENCE | VirtualHostIO => return None` arm is redundant with the wildcard below, and the wildcard itself suppresses the exhaustiveness check. If a future `JoltInstructionKind` is added that needs a lookup table, the compiler will not flag the omission — it silently routes through `_ => return None` and the row's lookup table is skipped. Spell out `NoOp` explicitly and drop the wildcard so adding a kind is a compile error.

---

jolt-core/src/zkvm/instruction/mod.rs
```diff
@@ -175,21 +212,24 @@ pub trait Flags {
...
             .unwrap_or([false; NUM_CIRCUIT_FLAGS])
     }
 
+    #[inline(always)]
     fn instruction_flags(&self) -> [bool; NUM_INSTRUCTION_FLAGS] {
-        LookupInstruction::try_from(*self)
+        JoltInstruction::try_from(*self)
             .map(|instruction| instruction_flags_from_riscv(instruction.instruction_flags()))
             .unwrap_or([false; NUM_INSTRUCTION_FLAGS])
     }
```
moodlezoup, line 236: `unwrap_or([false; …])` silently returns empty flags when `JoltInstruction::try_from(*self)` errs. The `try_from` impl is exhaustive today, so this is unreachable — but the all-false fallback is the wrong policy for a proof system. An unflagged row counts as not-Add/Sub/Mul/Load/Store/Branch/Virtual/Assert, i.e. a free-form row. If a future `JoltInstructionKind` is added without a matching `try_from` arm, this masks the wiring failure rather than surfacing it. Prefer `.expect("JoltInstructionRow kinds are exhaustive over JoltInstruction")`.

---

crates/jolt-program/src/preprocess/bytecode.rs
```diff
@@ -39,14 +48,15 @@ impl BytecodePreprocessing {
             bytecode,
             pc_map,
             entry_address,
+            profile_fingerprint: profile.fingerprint(),
```
moodlezoup: `profile_fingerprint` is stored on `BytecodePreprocessing` (and on `JoltProgramPreprocessing`) but never compared anywhere outside this file's self-test. It is not appended to the Fiat-Shamir transcript, not validated against a verifier-side profile, and not checked against the trace generator's profile. As written it is a defensive byte that costs serialization weight without buying any cross-party agreement. Either bind it into the transcript / verifier preprocessing, or drop the field until it has a consumer.

---

jolt-core/src/zkvm/instruction_lookups/read_raf_checking.rs
```diff
@@ -1246,17 +1250,15 @@ impl<F: JoltField> InstructionReadRafSumcheckProver<F> {
                         let e_lo_unreduced = E_lo[c_lo].to_unreduced();
 
                         // Accumulate table flag
-                        if let Some(table) = cycle.lookup_table() {
+                        let jolt_cycle = JoltTraceCycle::try_new(cycle)
```
moodlezoup, line 1253: Inner-loop perf concern: `JoltTraceCycle::try_new` runs the giant `Instruction::try_jolt_instruction_row` match on every cycle here, and the resulting `lookup_table()` / `circuit_flags()` calls each trigger another match. With T cycles × many sumcheck rounds this adds up — and `lookup_table` / `circuit_flags` on the underlying `Cycle` were direct methods on main. CLAUDE.md flags poly/sumcheck inner loops as performance-critical. Consider precomputing a `Vec<JoltInstructionRow>` (or `Vec<(LookupTable, CircuitFlags)>`) at trace-load time so this loop is a `[i]` index rather than a per-cycle decode.

---

crates/jolt-riscv/src/profile.rs
```diff
@@ -0,0 +1,529 @@
...
+                hash = hash_str(hash_byte(hash, 0x03), kind.canonical_name());
+            }
+        }
+        for kind in JoltInstructionKind::ALL {
+            if self.supports_jolt(*kind) {
+                hash = hash_tag(hash_byte(hash, 0x04), kind.tag());
+            }
+        }
+        hash
+    }
```
moodlezoup, line 150: The hash mixes source kinds by canonical name but jolt kinds by `u16` tag. Renaming a final kind's canonical name (e.g. `jolt.virtual.rotri` → `jolt.virtual.rotr_i`) does not change the fingerprint, while renaming a source kind does. If the fingerprint ever gains a real consumer (see the unrelated `profile_fingerprint` comment), this asymmetry is a foot-gun. Hash both canonical name and tag for jolt kinds, and prefix the whole computation with a `FINGERPRINT_SCHEMA_VERSION` byte so future encoding changes can be deliberately bumped.

---

moodlezoup (commented):
Reviewed the source/Jolt instruction catalog split. The phase boundary is enforced consistently at the prover via `JoltTraceCycle::try_new(...).expect(...)` — no silent skip-on-error patterns in the hot paths. Flagged one real bug (allocator leak when an inline provider errors on an rd=x0 source), one decoder/expander profile inconsistency, two exhaustiveness-suppression patterns that could mask future kinds, an unused `profile_fingerprint`, an inner-loop perf regression from the new `JoltTraceCycle` adapter, and an asymmetric `fingerprint()` hash.

---

crates/jolt-riscv/src/kind.rs
```diff
@@ -70,74 +51,31 @@ macro_rules! define_instruction_kind {
                 }
             }
 
-            pub const fn jolt_kind(self) -> JoltInstructionKind {
+            pub const fn canonical_name(self) -> &'static str {
```
moodlezoup, line 1364: is this for Lean extraction purposes?

moodlezoup:
> I generally agree with the updated approach, the only part that I'm not sure about is the registry macro. Instead of having the registry own the enum variant -> struct mapping, why not just have the enum variants carry the struct themselves?
> 
> So something like:
> 
> ```rust
> pub enum SourceInstructionKind<T = SourceInstruction> {
>     ADD(Add<T>),
>     ADDW(Addw<T>),
>     // ... 
> }
> 
> // Similar for JoltInstructionKind
> ```
> 
> This would also allow us to flip the relationship between the `SourceInstructionKind` enum and the `SourceInstruction` struct. Instead of having a `kind: SourceInstructionKind` field on the struct, we could instead use `SourceInstructionKind<SourceInstruction>` as the concrete, universal source-instruction type.
> 
> This is similar to the pattern I used for the `LookupInstruction` enum in order to achieve the "impl on instruction struct" + "enum dispatch" behavior that we want (see e.g. the `impl_jolt_instructions_flags` macro)

I may have been unclear, but the point of this change is to deduplicate `enum SourceInstruction<T>` vs `enum SourceInstructionKind`, and `enum JoltInstruction<T>` vs `enum JoltInstructionKind`

PR #1532, opened 2026-05-15
feat(trace): proof trace row layout spec + JoltTraceRow foundation

quangvdao:
## Summary

Follow-up to #1522 (source/Jolt instruction split). This PR contains both the
**proof-trace-row-layout spec** and the **foundation implementation** of
`JoltTraceRow` — a compact, proof-facing trace row that replaces repeated
prover-side adaptation of `tracer::instruction::Cycle` (via `JoltTraceCycle`).

`specs/proof-trace-row-layout.md` covers the design (Quang); this PR also lands
the foundation slices (the type, the builder, and tests). It does **not** yet
wire `JoltTraceRow` into the prover — the consumer cutover and the performance
gate are deferred follow-ups.

## Implementation (foundation)

Lands in the new `crates/` (not the deprecating `jolt-core`):

- **`crates/jolt-riscv/src/trace_row.rs`** — `JoltTraceRow`, a 64-byte
  `#[repr(C)]` row (Option C "balanced packed"; checked `size_of`). Per-cycle
  values are a typed `CapturedState { NonMemory | Load | Store }`, so the final
  memory-row aliasing (`LD`: RamRead = RamWrite = RdWrite; `SD`: RamWrite = Rs2)
  is enforced by the type rather than by runtime checks. Logical accessors only,
  private physical storage; register indices come from the instruction operands;
  `TraceRowError` via `thiserror`.
- **`crates/jolt-lookup-tables/src/traits.rs`** — `InstructionLookupTable` for
  `JoltTraceRow`, routed from the cached instruction tag (this crate owns
  `LookupTableKind`, so the accessor can't live in `jolt-riscv`).
- **`tracer/src/trace_row.rs`** — `cycle_to_trace_row` / `build_trace_rows`: the
  `Cycle → JoltTraceRow` conversion, and the place that verifies the memory-row
  contract equalities before they collapse into the typed state. This flips the
  dependency to `tracer → jolt-riscv`.
- **`jolt-core`** — a `host`-gated real-trace parity test comparing every
  accessor (built via the tracer conversion) against `R1CSCycleInputs::from_trace`
  over a real fibonacci trace.

## Deferred (not in this PR)

- **Consumer cutover:** `R1CSCycleInputs`/`ProductCycleInputs`, Spartan outer,
  RAM/register/instruction-lookup/bytecode phases; removing `JoltTraceCycle::try_new`
  from hot paths.
- **Performance gate:** the ≤2% end-to-end prover-time regression bound
  (`prover_time_fibonacci_100`, `prover_time_sha2_chain_100`) + phase-level
  measurements.
- **`CompactBytecodeRow`** co-design.
- The **`jolt-eval` `trace_row_accessor_parity` invariant**.

Intended cutover approach: `impl JoltCycle for JoltTraceRow` + the
`JoltCycle`-generic `jolt_lookup_tables::LookupQuery`, which lets the row replace
the `Cycle` outright (rather than coexisting with it).

## Validation

- `cargo clippy --all` clean in both `host` and `host,zk`.
- `cargo nextest run -p jolt-core muldiv` passes in both modes.
- jolt-riscv `trace_row` unit tests + jolt-core real-trace parity test pass.

> Note: `muldiv` passes trivially today because the prover does not yet consume
> `JoltTraceRow`; it becomes a meaningful gate once the cutover lands.


moodlezoup:
**Spec Analysis: Proof Trace Row Layout**

| Dimension | Score | Gap |
|-----------|-------|-----|
| Goal | 0.75 | Logical accessor / `JoltTraceRow` boundary is clear, but 5+ candidate row layouts (40B/48B/64B/72B/80B + SoA) without commitment |
| Constraints | 0.85 | Non-Goals and Invariants are thorough; minor stale C++ reference |
| Success Criteria | 0.60 | Acceptance checklist is concrete but layout-choice ambiguity bleeds in; no `jolt-eval` invariants; no performance pass/fail bar |
| Context | 0.85 | Strongly grounded — all cited current-state details verified against `main` |
| **Ambiguity** | | **26%** |

**Codebase claims that I verified against `main`:**
- `Cycle` is 96B; static assertion `rv64imac_cycle_size` exists at `tracer/src/instruction/mod.rs:1827`.
- `JoltTraceCycle<'a>` at `jolt-core/src/zkvm/instruction/mod.rs:59-85`, with ~32 call sites across `r1cs/`, `spartan/`, RAM, registers, bytecode, instruction-lookup phases (10 in `spartan/outer.rs` alone).
- `BytecodePreprocessing` / `JoltInstructionRow` / `NormalizedOperands` / `BytecodePCMapper` fields in `crates/jolt-program/src/preprocess/bytecode.rs` and `crates/jolt-riscv/src/row.rs` match the spec exactly.
- LD/SD R1CS contract (`RamReadValue == RamWriteValue == RdWriteValue` for LD; `Rs2Value == RamWriteValue` for SD) is enforced in `jolt-core/src/zkvm/r1cs/constraints.rs:250-270`.
- `R1CSCycleInputs::from_trace` at `jolt-core/src/zkvm/r1cs/inputs.rs:258-268` currently takes `&[Cycle]` and wraps via `JoltTraceCycle::try_new`.
- PR #1522 has merged (`cb1e464`).

**Status: Questions remain** — 5 ambiguities to resolve before implementation.

**Questions:**

**1. [Goal / Success Criteria]** The spec is internally contradictory about scope of the implementation PR.
   - Acceptance Criteria says "**The** selected layout has a checked size budget" (singular).
   - Execution step 8 says "Benchmark **40B, 48B, 64B, and 80B** candidate layouts behind the same accessor API" (plural).
   - Five options (A–F) are presented without a default.

   Which is the deliverable?
   - **(a)** Ship the accessor API + ONE conservative default (e.g., Option B / 48B), defer the other layouts to follow-up PRs.
   - **(b)** Ship the accessor API + all 4 layouts behind a `cfg` or feature flag, ship a chosen default after benchmarking inside the same PR.
   - **(c)** Ship the accessor API + benchmark scaffolding only; commit to a default in a follow-up after measurement.

   Without an answer, the implementer has to choose; that's a multi-day scope swing.

**2. [Success Criteria]** No `jolt-eval` invariant is proposed, but several Invariants in the spec are textbook binary properties suitable for the framework — particularly "for every cycle index `t`, the logical values exposed by `JoltTraceRow` must match the values currently derived from `Cycle`/`JoltTraceCycle`." That is a reference-vs-optimized comparison directly analogous to `split_eq_bind_low_high` in `jolt-eval/src/invariant/`. Should this be captured as a new `jolt-eval` invariant (so it gets `#[test]` + fuzz + red-team coverage automatically), or is `jolt-eval` explicitly out of scope here? If out of scope, the Invariants section should say so.

**3. [Success Criteria]** The Performance section lists 8 measurements (peak RSS, Spartan outer, RAM read-write/RA/hamming/val-check, register, bytecode, instruction lookup read-RAF) but no pass/fail threshold. What is the bar?
   - Neutral-or-better wall-clock on existing Criterion benches `prover_time_fibonacci_100` and `prover_time_sha2_chain_100` (the closest existing jolt-eval bench targets)?
   - Some absolute regression tolerance (e.g., ≤2% e2e prover time)?
   - Purely informational — performance is recorded but not gating?

   Without this, "benchmark before committing to a default layout" is unimplementable as a one-shot.

**4. [Constraints]** Acceptance Criteria entry 11 says "a compact bytecode-row design is implemented or **explicitly deferred with benchmark evidence** showing the current 48-byte `JoltInstructionRow` table is not the bottleneck." What concretely qualifies as benchmark evidence?
   - A specific delta on a named benchmark, or
   - An absolute timing budget, or
   - A profiling observation (e.g., bytecode lookup < X% of Spartan outer time)?

**5. [Context — minor]** The References section cites `jolt-cpp/src/tracer/trace_owner.hpp` as prior art, but this path does not exist in `a16z/jolt`. Is `jolt-cpp` a separate repo that reviewers can be pointed to, an internal/private repo, or a stale reference? The substantive design isn't blocked on this — it's only cited as inspiration for the 80B "Expanded Row" option — but reviewers will be confused.

---

specs/proof-trace-row-layout.md
```diff
@@ -0,0 +1,1082 @@
...
+Introduce a proof-facing `JoltTraceRow` representation that replaces direct
+prover use of `tracer::instruction::Cycle` while preserving the existing logical
+proof columns and prover/verifier semantics.
+
+The follow-up should introduce or clarify these boundaries:
+
+- `JoltTraceRow`: a compact, copyable row used by prover-side witness and
+  sumcheck code after source tracing has been converted to final Jolt rows.
+- `JoltTraceRows` or an equivalent builder: constructs rows once from
+  `Vec<Cycle>`, final bytecode preprocessing, and selected profile metadata.
```
moodlezoup: I wonder if we should avoid materializing a `Vec<Cycle>` and instead directly materialize `Vec<JoltTraceRow>` in tracer?

moodlezoup (commented):
just one question

specs/proof-trace-row-layout.md
quangvdao: Yes, I think that is the right longer-term target. I updated the spec to avoid making `Vec<Cycle>` the permanent producer contract: the first implementation can adapt the existing `Vec<Cycle>` boundary to keep the cutover small and preserve an easy parity reference, but the row builder/API should be shaped so a future tracer path can emit `JoltTraceRow` directly once final bytecode metadata is available, without materializing both vectors.

quangvdao:
## Suggested tighter 64B layout (alternative to current `register_pack` + scattered flags)

Posted by Cursor assistant (model: GPT-5.4) on behalf of Quang Dao — sharing a layout direction we discussed offline; very open to pushback.

The foundation slice (`JoltTraceRow`, tracer conversion, parity test) looks good. One thing I'd tweak before cutover: the physical row can use its 64 bytes more deliberately without changing the accessor API or row size.

### Register indices: don't pack into `u32`

Current `register_pack: u32` is `[rs1, rs2, rd, 0]` — byte 3 is dead. REGISTER_COUNT is 128, so each operand is naturally a `u8` with `0xFF` = None.

Prefer four explicit bytes (same 4B, clearer, byte 4 is useful):

```rust
rs1_id: u8,            // 0xFF = None
rs2_id: u8,
rd_id: u8,
lookup_table_id: u8,   // 0 = none; avoids instruction_kind() → dispatch in lookup paths
```

No pack/unpack helpers, no opaque `to_le_bytes()[1]`. Only reason to keep `u32` pack would be if profiling shows a single load of all three registers wins in sumcheck loops (seems unlikely).

Avoid `Option<u8>`×3 in the struct — niche layout gets worse, not better.

### Keep `imm_abs: u64`

`NormalizedOperands.imm` is `i128` and the prover already bounds magnitude to `u64`. Today's expansion recipes likely fit `u32`, but narrowing to `u32` in the row is a storage-format commitment for only 4 bytes saved, with footgun risk if a future final/virtual instruction needs a larger constant. I'd keep `u64` unless we land a catalog-wide invariant + jolt-eval guard first.

### `bytecode_pc`: `u32` storage is fine (matches spec + current impl)

Yes — both the spec (Options A/B/C) and the landed code use `u32` physical storage with a logical `pc()` / `BytecodeIndex(u64)` accessor and loud reject at build if `pc > u32::MAX`. That limit is ~4B expanded bytecode rows; memory blows up long before that (e.g. 4B × 48B bytecode table ≈ 192 GiB). So `u32` saves 4 bytes vs `u64` with no practical downside today. Going `u64` only makes sense if we explicitly want to spend those 4 bytes on something else *less* important than index headroom — I'd keep `u32` in the 64B budget.

### Separate `jolt_tag: u16`, pack the rest into `meta: u32`

Stable instruction identity stays explicit. Flags + imm sign go in one word:

| `meta` bits | Field |
|-------------|-------|
| 0..15 | `circuit_flags` |
| 16..21 | `instruction_flags` |
| 22 | `imm_is_negative` |
| 23..31 | spare (9 bits) |

### Proposed `#[repr(C)]` layout (~64B)

```rust
#[repr(C)]
pub struct JoltTraceRow {
    values: TraceValueSlots,   // 32 — unchanged aliasing (NonMemory / Load / Store)
    unexpanded_pc: u64,        //  8
    imm_abs: u64,              //  8
    bytecode_pc: u32,          //  4
    meta: u32,                 //  4 — flags + imm sign (bit map above)
    jolt_tag: u16,             //  2
    rs1_id: u8,
    rs2_id: u8,
    rd_id: u8,
    lookup_table_id: u8,
    _pad: u16,                 //  2 — tail align; could repurpose e.g. next_pc_delta u16 if useful
}
// size_of == 64
```

All witness/metadata accessors unchanged; only physical storage rearranged.

### What this doesn't solve (honest scope)

With `imm_abs: u64` + cached `unexpanded_pc: u64`, there's not room for a full `next_bytecode_pc: u32` without dropping something else (e.g. cache `unexpanded_pc` only in bytecode table) or going wider than 64B. That can stay deferred — same as the cutover/perf gate.

### Trace memory @ 2^27

64B × 2^27 = 8 GiB for the row vector alone (vs 12 GiB at today's 96B `Cycle`).

---

Opinion, not a blocker — happy to iterate on field order, whether `lookup_table_id` belongs in the row vs derived from tag, and whether the 2-byte tail should stay padding or fund a bounded next-PC delta. Main ask: kill the dead byte in `register_pack` and pack flags without burying `jolt_tag`.

quangvdao:
## Follow-up: revised 64B layout (packed meta + next-row flags + lookup table id)

Posted by Cursor assistant (model: GPT-5.4) on behalf of Quang Dao — amending the earlier layout comment with bitfield packing we discussed offline.

Building on the prior proposal (four named register bytes, separate `jolt_tag: u16`, `imm_abs: u64`, `bytecode_pc: u32`, `meta: u32`), we can also cache the remaining **boolean metadata** cutover still derives from `rows[t+1]` or from `instruction_kind()` dispatch — using slack that is already paid for in the 64B budget, not new bytes.

### What to pack (zero extra bytes)

| Field | Bits | Source at build time |
|-------|------|----------------------|
| `lookup_table_id` | 8 | `InstructionLookupTable::lookup_table(row)` at materialization (store `id_plus_one`, 0 = none) |
| `next_is_noop` | 1 | row `t+1` `instruction_flags.IsNoop` |
| `next_is_virtual` | 1 | row `t+1` `circuit_flags.VirtualInstruction` |
| `next_is_first_in_sequence` | 1 | row `t+1` `circuit_flags.IsFirstInSequence` |

These are denormalized (especially the `next_*` trio — same pattern as optionally caching `next_bytecode_pc`, but only 3 bits). Fill at `build_trace_rows` time; parity/invariant should assert they match `rows[t+1]`.

**Not proposing** cached `next_bytecode_pc` — `rows[t+1].pc()` is enough once `bytecode_pc` is on every row.

### Proposed `#[repr(C)]` layout (64B exact)

```rust
#[repr(C)]
pub struct JoltTraceRow {
    values: TraceValueSlots,   // 32
    unexpanded_pc: u64,        //  8  → 40
    imm_abs: u64,              //  8  → 48
    bytecode_pc: u32,          //  4  → 52
    meta: u32,                 //  4  → 56
    jolt_tag: u16,             //  2  → 58
    rs1_id: u8,                //  1  → 59
    rs2_id: u8,                //  1  → 60
    rd_id: u8,                 //  1  → 61
    lookup_table_id: u8,       //  1  → 62  (0 = none; else dense table id)
    _pad: u8,                  //  1  → 63  (spare / future instr-flag growth)
}
// size_of == 64
```

### `meta: u32` bit map

| Bits | Field |
|------|-------|
| 0..15 | `circuit_flags` (16 flags — full, no spare) |
| 16..21 | `instruction_flags` (6 today; 2 spare for enum growth) |
| 22 | `imm_is_negative` |
| 23 | `next_is_noop` |
| 24 | `next_is_virtual` |
| 25 | `next_is_first_in_sequence` |
| 26..31 | spare (6 bits) |

Accessor sketch: `next_is_noop()` etc. read from `meta`; `lookup_table()` can use cached `lookup_table_id` instead of `instruction_kind()` → dispatch (keep dispatch as debug/parity fallback during migration).

### Still not in the row (unchanged)

- `NextPC` / `NextUnexpandedPC` — use `rows[t+1].pc()` / `.unexpanded_pc()`
- `lookup_output`, lookup operands — execution witness; stay on `Cycle` until cutover adds accessors or derivation
- `imm_abs: u64` — keep (don't narrow to u32 without a catalog invariant)

### Ask

Before cutover lands, consider folding this into the foundation struct so we don't bake in `register_pack` byte 3 = 0 and scattered `bool`/padding. Happy to iterate on field order or moving `lookup_table_id` entirely into `meta` bits 26..31 if table count stays ≤ 64.

Opinion / refinement of earlier comment, not a blocker.

moodlezoup:
Thanks @quangvdao — applied most of this in 97f6fd55c.

**Taken as suggested:**
- Register indices are now explicit `rs1_id` / `rs2_id` / `rd_id: u8` (`0xFF` = None) — dropped the `register_pack: u32` + `to_le_bytes()` indexing and the pack/unpack helpers.
- Flags + immediate sign packed into one `meta: u32` (circuit flags `0..16`, instruction flags `16..22`, imm sign bit `22`, 9 spare); `jolt_tag` stays a separate `u16`. `is_load` / `is_store` / `is_noop` are now single masked bit-tests on `meta`. Added `CircuitFlagSet` / `InstructionFlagSet::from_bits` for reconstruction.
- Kept `imm_abs: u64` and `bytecode_pc: u32` — no narrowing.

Row is still 64 B (checked `size_of` assertion); clippy clean in both `host` and `host,zk`, `muldiv` green in both modes.

**One deviation — `lookup_table_id` left reserved.** I couldn't populate a cached table id from where the row is built: `JoltTraceRow::from_components` lives in `jolt-riscv` and the `Cycle → row` conversion lives in `tracer`, and both sit *below* `jolt-lookup-tables` in the dep graph (`jolt-lookup-tables → jolt-riscv`), so neither can compute a `LookupTableKind` discriminant without a circular or new dependency. I reserved the byte and kept `lookup_table()` deriving from `jolt_tag` on demand. Wiring the cache feels like cutover work — at that point either `tracer` takes a `jolt-lookup-tables` dep and passes the id into `from_components`, or a higher crate fills it post-build. Happy to do whichever you prefer.

(Minor: the `BytecodeIndex(u64)` accessor you referenced was dropped earlier in review; `pc() -> u64` is the equivalent logical-width accessor over the `u32` storage.)

_Drafted with Claude Code on behalf of Michael Zhu._

crates/jolt-riscv/src/trace_row.rs
```diff
@@ -0,0 +1,707 @@
...
+const MAX_REGISTER_ID: u8 = u8::MAX - 1;
+
+/// Sentinel stored in a register-id byte for an absent (`None`) operand.
+const REGISTER_NONE: u8 = u8::MAX;
+
+/// `meta` bit layout: `circuit_flags` occupy the low 16 bits; instruction flags
+/// the next 6; the immediate sign one more; the top 9 bits are spare.
+const META_INSTRUCTION_FLAGS_SHIFT: u32 = 16;
+const META_INSTRUCTION_FLAGS_MASK: u32 = (1u32 << (crate::NUM_INSTRUCTION_FLAGS as u32)) - 1;
+const META_IMM_NEGATIVE_SHIFT: u32 = 22;
```
moodlezoup: Nothing enforces that the instruction-flags region (`16..16+NUM_INSTRUCTION_FLAGS`) stays below `META_IMM_NEGATIVE_SHIFT`. Today `NUM_INSTRUCTION_FLAGS = 6` exactly meets `META_IMM_NEGATIVE_SHIFT = 22`, so any 7th `InstructionFlags` variant silently widens `META_INSTRUCTION_FLAGS_MASK` to `0x7F`, collides with bit 22, and corrupts both `instruction_flags()` and `imm()` reads (negative imms would set a spurious flag bit, and an unrelated flag would flip imm sign). Worth a sibling to the 64-byte `const _` assert:

```suggestion
const META_INSTRUCTION_FLAGS_SHIFT: u32 = 16;
const META_INSTRUCTION_FLAGS_MASK: u32 = (1u32 << (crate::NUM_INSTRUCTION_FLAGS as u32)) - 1;
const META_IMM_NEGATIVE_SHIFT: u32 = 22;
const _: () = assert!(
    META_INSTRUCTION_FLAGS_SHIFT + (crate::NUM_INSTRUCTION_FLAGS as u32) <= META_IMM_NEGATIVE_SHIFT,
    "instruction-flags region overlaps the imm-sign bit in `meta`"
);
```

---

moodlezoup (commented):
Foundation looks solid — the typed `CapturedState`, the contract checks pushed into the tracer side, and the 64-byte layout assertion all land cleanly. One layout-fragility note inline; otherwise nothing actionable.

---

PR #1533, opened 2026-05-15
Move inline expansion to jolt-program recipes

quangvdao:
## Summary

- Moves registered inline static expansion off tracer instruction construction and onto the `jolt-program` recipe/materializer path.
- Adds `InlineExpansionBuilder`/recipe APIs, centralizes materialization through `ExpansionState`, and keeps runtime advice generation tracer-owned.
- Ports shipped inline packages to emit static recipes instead of `InstrAssembler`/`tracer::Instruction` sequences.
- Centralizes inline virtual-register allocation, reset-row generation, metadata stamping, and `rd = x0` handling in the expansion materializer.
- Updates parity fixtures, extraction notes, and `specs/inline-expansion-grammar.md` to match the implemented cutover.

## Scope Notes

- This PR preserves the existing guest bytecode trust model.
- It intentionally removes the partial inline admissibility/raw custom-opcode rejection machinery from this branch; arbitrary custom-opcode hardening should be handled as a separate design.
- Runtime advice remains host/tracer-owned; static bytecode expansion is the tracer-free part moved into `jolt-program`.

## Validation

Latest focused validation:

- `cargo fmt -q`
- `cargo check -p jolt-inlines-fixtures --features host -q`
- `cargo nextest run -p jolt-inlines-fixtures --features host --cargo-quiet`
- `cargo clippy -p jolt-inlines-sdk --features host -q --all-targets -- -D warnings`
- `cargo clippy -p tracer --features test-utils -q --all-targets -- -D warnings`
- `git diff --check && git diff --cached --check`


moodlezoup:
**Spec Analysis: Inline Expansion Grammar**

| Dimension | Score | Gap |
|-----------|-------|-----|
| Goal | 0.95 | Clear |
| Constraints | 0.85 | Clear (minor: lifetime model is "implementer choice") |
| Success Criteria | 0.88 | Clear |
| Context | 0.85 | Clear (minor: `TempId` vs. `InlineTempId` unification left open) |
| **Ambiguity** | | **11%** |

**Summary:**

- **What will be built**: Move static registered-inline expansion off the tracer-owned `InstrAssembler -> tracer::Instruction -> try_jolt_instruction_row()` path onto the existing `crates/jolt-program/src/expand/` recipe/materializer grammar (`ExpansionBuilder` + `ExpansionState` + `ExpansionAllocator`). Split `InlineOp` into a tracer-free `StaticInlineOp` (recipe construction) plus a host-owned `InlineAdviceProvider` (runtime advice). Port all 8 shipped inline crates in one cutover.

- **Key invariants**:
  - Static path is tracer-free at the type boundary — enforced by an explicit ripgrep denylist (`InstrAssembler|inline_helpers|try_jolt_instruction_row|tracer::instruction|FormatInline|RISCVInstruction|RISCVTrace|RISCVCycle`) over `jolt-inlines/` and a `cargo tree -p jolt-program` check that drops `tracer`.
  - Final `JoltInstructionRow` stream is byte-identical to the old path, locked by golden fixtures generated before the API break, one per `(opcode, funct3, funct7, InlineExtension)` key.
  - Register layout preserved: `vr32..39` reserved / `vr40..47` source-expansion temps / `vr48..` inline temps. Inline resets emitted by `ExpansionAllocator::take_registers_for_reset()` as `ADDI rd, x0, 0`, not by tracer RAII.
  - Two-stage profile gating: `profile.supports_source(SourceInstructionKind::Inline)` AND `profile.supports_inline(registration.extension)` — both checks required, neither implies the other.
  - Metadata stamped exactly once over flattened sequence (functional rows + recursive helper rows + reset rows). Compressed-tail marker on final reset row when the source row is compressed.
  - Runtime advice arity must match the count of `VirtualAdvice` rows in the materialized static sequence; mismatch produces a structured error, not silent miscount.

- **Critical evaluation criteria**:
  - `cargo nextest run -p jolt-core muldiv --features host` and `--features host,zk` both pass.
  - Mechanical denylist (`! rg '...' jolt-inlines jolt-inlines/sdk/src`) returns clean.
  - Old `InstrAssembler` path deleted or made private to runtime/test utilities — unreachable from static builders.
  - Static parity tests compare final `JoltInstructionRow`s (post-metadata-stamp) against golden fixtures for all 8 shipped inline crates: `sha2`, `blake2`, `blake3`, `keccak256`, `bigint`, `secp256k1`, `grumpkin`, `p256`.
  - Perf smoke: wall-time regression < 5% on the all-shipped-inlines expansion microbench.

The substrate already exists in the repo (`ExpansionBuilder`, `ExpansionOp`, `TempId`, `RowTemplate`, `ExpansionAllocator::take_registers_for_reset`, `SourceInstructionRow.inline: Option<SourceInlineKey>`, `profile.supports_source/supports_inline`), so this is a refactor + port, not a greenfield design.

---

specs/inline-expansion-grammar.md
```diff
@@ -37,254 +46,578 @@ registered inline expansion ────┘
...
+### Goals
+
+- Replace the static registered-inline `InstrAssembler -> tracer::Instruction`
+  path with a `jolt-program` recipe builder and materializer path.
+- Make static inline sequence construction independent of tracer instruction
+  types, concrete tracer cycles, CPU state, memory devices, advice tapes, and
+  runtime trace formatting.
+- Preserve the final `JoltInstructionRow` stream emitted by the old static path,
+  except for intentional documented bug fixes locked by fixtures.
+- Keep runtime inline advice generation separate and host/tracer-owned.
```
moodlezoup: I'm not sure what this means/how it would work

specs/inline-expansion-grammar.md
```diff
@@ -37,254 +46,578 @@ registered inline expansion ────┘
...
+  boundary that currently hides static inline recipes from extraction, but
+  extractor backend and standard-library limitations remain separate follow-up
+  work.
+- Do not move CPU, RAM, trace-cycle, or advice execution semantics into
+  `jolt-program`.
+- Do not replace inventory registration if the registration surface calls a
+  tracer-free static recipe builder. Inventory is a host discovery mechanism,
+  not the place where static expansion semantics may be hidden.
+- Do not remove tracer's execution-time virtual register allocator if concrete
+  tracing still needs it after static expansion is cut over.
```
moodlezoup: afaict it shouldn't still need it, but not 100% sure

moodlezoup (commented):
it's a little unclear to me how advice would work with inlines, but otherwise this lgtm

quangvdao:

One audit concern we should fold into this design: today, registered inline opcodes are accepted from decoded guest bytecode, but some of those opcodes are only sound when reached through their SDK wrappers.

The concrete issue is provenance. The verifier/preprocessing path sees a raw custom inline opcode in the ELF and dispatches by `(opcode, funct3, funct7)`. It generally cannot tell whether that opcode came from the intended Rust SDK wrapper or from hand-written `.insn`. For several inlines, the SDK wrapper performs important checks after invoking the raw opcode: canonical output checks, nonzero divisor checks, field relation checks such as `b * q == a`, GLV sign-word validation, and decomposition/recomposition checks. If a guest encodes the raw opcode directly, those wrapper checks are skipped while the inline expansion/advice path still runs.

So I think the generic fix should be registration-level admissibility, not a one-off denylist for today’s opcodes.

Each inline registration should declare whether its raw opcode is verifier-admissible from decoded guest bytecode:

- `Public { requirements }`: may appear as raw bytecode, but any safety preconditions must be satisfied by verifier-visible expansion rows.
- `InternalOnly { reason }`: rejected from decoded guest bytecode before proving.

For public inlines with requirements, the static builder should emit both the actual check rows and a small safety-evidence token through certified helper methods. Finalization then checks `requirements ⊆ evidence`. SDK wrapper checks, Rust constructors, comments, or post-inline `spoil_proof()` calls should not count unless the corresponding check appears in the static expansion.

This gives us a mechanical contract: the verifier does not infer that an inline is safe because the host has a recipe; it accepts only public registrations whose declared requirements are evidenced by the recipe. Advice-only or SDK-checked helper opcodes default to `InternalOnly` until their full relation/canonicality/sign/recomposition checks are moved into verifier-visible expansion.

jolt-inlines/grumpkin/src/sequence_builder.rs
```diff
@@ -3,37 +3,36 @@ use std::collections::VecDeque;
...
-        GrumpkinDivAdv {
-            asm,
-            vr,
-            operands,
-            is_base_field,
-        }
+    fn new(
+        mut asm: InlineExpansionBuilder,
+        operands: InlineOperands,
+        _is_base_field: bool,
```
moodlezoup: delete this parameter

moodlezoup:
**Automated code review (Claude)**

Reviewed the cutover from tracer-side inline expansion to the jolt-program recipe/materializer path. The new abstractions (`InlineExpansionBuilder`, `ExpansionState`, `InlineExpansionProvider`, separated inline/instruction register pools, reset-row centralization) are internally consistent and the determinism boundary between the preprocessor's shared allocator and the tracer's per-inline fresh allocator holds — `take_registers_for_reset` clears `pending_clearing_inline` on every successful `materialize_inline`, leak/recursion checks force error termination on bad recipes, and both static and runtime paths use `RV64IMAC_JOLT_ALL_INLINES`. No bugs, soundness gaps, or panic-on-valid-input paths found across the four review passes.

A few minor observations that don't rise to blocking:
- `rotri32`/`rotr64`/`rotl64` in `crates/jolt-program/src/expand/inline.rs` underflow on shift amounts > 32/64; no current caller exercises this, but the public API would benefit from a `debug_assert`.
- `TempBindings` and `InlineTempBindings` in `materialize.rs` are near-identical modulo const N and id type — generic over `<Id, const N: usize>` would remove ~70 lines.
- `materialize` vs `materialize_inline` could share a body via an options struct; the hand-rolled reset-row capacity check duplicates `ExpansionBuffer::push`.
- `lookup_inline`/`find_inline` do a linear `inventory::iter` per inline expansion. Fine at today's 23 registrations; a lazy `OnceLock<HashMap>` would collapse to O(1) and unify the two helpers.

_(Posted as an issue comment because a pending review draft already exists on this PR.)_

---

tracer/src/instruction/inline.rs
```diff
@@ -52,79 +63,94 @@ fn find_inline(opcode: u32, funct3: u32, funct7: u32) -> &'static InlineRegistra
...
+        let inline = instruction
+            .row()
+            .inline
+            .ok_or(ExpansionError::MalformedInstruction(
+                "missing inline source metadata",
+            ))?;
+
+        if !profile.supports_source(SourceInstructionKind::Inline) {
+            return Err(ExpansionError::UnsupportedInstruction);
+        }
```
moodlezoup: move this to the top of the function body. Also, I think we should still be checking that `instruction` is in fact an `INLINE`, no?

crates/jolt-program/src/expand/materialize.rs
```diff
@@ -82,11 +96,61 @@ impl ExpansionState {
...
+    /// here, after all provider-owned inline registers have been released, so
+    /// virtual-register cleanup is centralized and independent of tracer RAII.
+    pub(super) fn materialize_inline(
+        &mut self,
+        sequence: ExpandedInstructionSequence,
+    ) -> Result<Vec<JoltInstructionRow>, ExpansionError> {
+        let source = sequence.source;
+        let mut rows = self.materialize_unstamped(sequence, MAX_INLINE_ROWS_PER_SOURCE)?;
+        for register in self.allocator.take_registers_for_reset()? {
+            if rows.len() == MAX_INLINE_ROWS_PER_SOURCE {
```
moodlezoup:
```suggestion
            if rows.len() >= MAX_INLINE_ROWS_PER_SOURCE {
```

moodlezoup (commented):
lgtm, just a couple of minor comments

jolt-inlines/sdk/Cargo.toml
moodlezoup, line 1: Actually, I have a question here -- `jolt-inlines-sdk` (and therefore, the inline crates) still depends on `tracer`. Is the eventual goal to completely remove the inline crates' dependence on `tracer`? Is there a clear path to that?

quangvdao: Posted by Codex assistant (model: GPT-5) on behalf of the user (Quang Dao) with approval.

Yes, the eventual direction should be to remove the inline crates’ static-expansion dependency on `tracer`; this PR gets the static recipe construction onto `jolt-program`, but it intentionally leaves the host/runtime advice side behind the `host` feature.

The remaining `tracer` dependency is there for runtime-only pieces: `Cpu`, `FormatInline`, and the inventory registration bridge used while tracing patches `VirtualAdvice` rows. There is a clear follow-up path, but I think it is a separate API migration: move inline operand formatting into `jolt-riscv`/`jolt-program`, replace the `Cpu` dependency with a narrow advice-context trait, and move the inventory registration/dispatch bridge out of tracer. That would make the inline crates fully tracer-independent, while keeping this PR scoped to the static materialization cutover and preserving the current runtime advice behavior.

PR #1536, opened 2026-05-16
test(jolt-openings): regression tests for failing Fiat--Shamir transform

mmaker:
In jolt-openings today, the opening points and the polynomial commitments are not part of the Fiat--Shamir transformation. 
As a result, the challenge can be predicted on some batching protocols. 
This commit adds regression tests highlighting this issue.

/cc @shreyas-londhe


0xAndoroid:
So, the previous claims are indeed part of the FS. They get appended at the end of the sumcheck that produces the claim, instead of being appended right before `gamma` is sampled. We've already investigated this and this was deemed to be enough for soundness purposes, and this simplified blindfold efforts. Lmk if you think there's still a soundness issue.

PR #1569, opened 2026-05-24
Fix batched sumcheck padding for non-Boolean domains

quangvdao:
## Summary

- Derive batched sumcheck front-padding scale from the active `SumcheckDomain` instead of assuming Boolean-hypercube scaling.
- Add domain-level padding helpers so `BooleanHypercube` keeps the original `mul_pow_2` fast path while non-Boolean domains use their own constant-round scale.
- Add a regression covering a heterogeneous batched proof over `CenteredIntegerDomain::new(3)`.

## Bug Context

PR #1547 generalized `BatchedSumcheckVerifier::verify` to accept any `D: SumcheckDomain`, but the combined input claim still scaled shorter instances with `2^(max_num_vars - claim.num_vars)`. That factor is correct for the Boolean hypercube because each front-padding variable ranges over two points.

For a non-Boolean domain, the padding multiplier must be the domain's constant-round sum raised to the number of inactive front-padding rounds. For example, over `CenteredIntegerDomain::new(3)`, a zero-variable claim with value `1` embedded into a one-round batch contributes `3 * alpha`, not `2 * alpha`. With the old code, an honest constant round polynomial over that three-point domain is rejected with `RoundCheckFailed` because the verifier computes the wrong combined claimed sum before checking the round.

## Fix

This PR adds `SumcheckDomain::padding_scale()` and `SumcheckDomain::scale_padding()`. The batched clear verifier now delegates front-padding scaling to the domain instead of doing the scaling inline.

The performance-sensitive Boolean path is restored by overriding `BooleanHypercube::scale_padding()` to call the previous `value.mul_pow_2(padding_rounds)` implementation directly. `CenteredIntegerDomain` overrides `padding_scale()` to return its validated domain size without allocating or computing all power sums. The default implementation remains available for future domains.

The compressed batched verifier remains Boolean-specific and continues using its existing Boolean-hypercube path.

## Validation

- `cargo nextest run -p jolt-sumcheck batched_verify_uses_domain_padding_scale --cargo-quiet`
- `cargo nextest run -p jolt-sumcheck --cargo-quiet`
- `cargo nextest run -p jolt-sumcheck --features r1cs --cargo-quiet`
- `cargo clippy -p jolt-sumcheck --features r1cs -q --all-targets -- -D warnings`
- `git diff --check`

After the performance follow-up:

- `cargo nextest run -p jolt-sumcheck batched_verify_uses_domain_padding_scale --cargo-quiet`
- `cargo nextest run -p jolt-sumcheck --cargo-quiet`
- `cargo nextest run -p jolt-sumcheck --features r1cs --cargo-quiet`
- `cargo clippy -p jolt-sumcheck --features r1cs -q --all-targets -- -D warnings`
- `git diff --check`

## Attribution

Posted by the assistant on behalf of Quang Dao. AI-authored by Codex (GPT-5).

### DIFF
--- a/crates/jolt-sumcheck/src/batched_verifier.rs
+++ b/crates/jolt-sumcheck/src/batched_verifier.rs
@@ -200,12 +200,14 @@ impl BatchedSumcheckVerifier {
         }
 
         let alpha: F = transcript.challenge();
+        let padding_scale = domain.padding_scale()?;
 
         // Running power of alpha: alpha^j for j = 0, 1, 2, …
         let mut alpha_pow = F::one();
         let mut combined_sum = F::zero();
         for claim in claims {
-            let scaled = claim.claimed_sum.mul_pow_2(max_num_vars - claim.num_vars);
+            let scaled =
+                claim.claimed_sum * pow_usize(padding_scale, max_num_vars - claim.num_vars);
             combined_sum += alpha_pow * scaled;
             alpha_pow *= alpha;
         }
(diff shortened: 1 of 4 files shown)


0xAndoroid:
Thanks

PR #1570, opened 2026-05-24
Add Poseidon2 Goldilocks inline

jay-clarke:
## Summary

Adds a new `jolt-inlines-poseidon2-goldilocks` crate implementing the canonical 8-wide Poseidon2 permutation over the Goldilocks field as a Jolt inline.

Spec: `specs/1570-poseidon2-goldilocks-inline.md`

The inline exposes a small guest API:

- `poseidon2_permute(&mut [u64; 8])`
- RISC-V guest builds emit one custom inline instruction.
- Host builds use the standalone reference implementation.

The host side registers the inline with the Jolt tracer/prover through the existing `register_inlines!` path and adds a `Poseidon2Goldilocks` entry to `InlineExtension`.

## Why

Poseidon2 over Goldilocks is commonly used inside proof systems for Merkle trees, commitments, and other ZK-native hashing paths. Running that permutation as ordinary guest Rust creates a much larger trace than an inline expansion. This adds a dedicated inline for the Plonky3-compatible Poseidon2-Goldilocks-8 instance.

## Validation

- `cargo check -p jolt-inlines-poseidon2-goldilocks`
- `cargo check -p jolt-inlines-poseidon2-goldilocks --features host`
- `cargo test -p jolt-inlines-poseidon2-goldilocks --features host`
- `cargo test -p jolt-riscv`

The new tests cover:

- Goldilocks multiplication reduction against `u128` arithmetic.
- Permutation parity with Plonky3's default `Poseidon2Goldilocks<8>`.
- Permutation parity with an explicitly constructed generic Plonky3 `Poseidon2` path.
- Round-constant and internal-diagonal parity with Plonky3 constants.
- Deterministic sequence emission.
- Inline emulator execution against the host reference for fixed and randomized states.

## Notes

The crate intentionally covers only the 8-wide Goldilocks instance. Other widths or fields can be added separately if they are useful.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3003,6 +3003,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "jolt-inlines-poseidon2-goldilocks"
+version = "0.1.0"
+dependencies = [
+ "jolt-inlines-sdk",
+ "p3-field",
+ "p3-goldilocks",
+ "p3-poseidon2",
+ "p3-symmetric",
+ "tracer",
+]
+
+[[package]]
 name = "jolt-inlines-sdk"
 version = "0.1.0"
 dependencies = [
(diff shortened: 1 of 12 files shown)


0xAndoroid:
Thanks for opening the PR. Please mark it as ready for review when its ready.

PR #1586, opened 2026-06-03
spec: jolt-core transcript migration to spongefish split-trait surface

shreyas-londhe:
## Summary

Spec for follow-up PR (b) of the jolt-transcript spongefish port (#1455). Migrates jolt-core — and all remaining legacy-facade consumers (`jolt-verifier`, `jolt-dory`, `jolt-sumcheck`, `jolt-openings`, `jolt-crypto`) — onto jolt-transcript's spongefish-native split-trait surface (`ProverTranscript` / `VerifierTranscript` / `OptimizedChallenge`), deletes jolt-core's duplicate transcript stack (~1,370 LOC) and the legacy facade (`legacy.rs`), so a single spongefish-backed Fiat-Shamir surface remains workspace-wide.

## Key decisions captured

- **Split-trait surface** (positional + NARG), not the facade — the facade is retired.
- **Clean break** on Fiat-Shamir bytes; correctness gate is internal prover/verifier consistency (`muldiv` host+zk), not byte-equality. Downstream verifier regeneration (Solidity / gnark / Lean) stays a coordinated follow-up.
- **`OptimizedChallenge` for Poseidon** added, with a pinned 128-bit squeeze rule + entropy-audit requirement.
- **Single big-bang PR**; two distinct migrations (jolt-core's own `Transcript` trait + the differently-shaped facade `Transcript` trait) each mapped per-method.

## Notes

- Planning/spec PR — no code changes; only `specs/jolt-core-transcript-migration.md`.
- Self-reviewed via three independent zero-context analysis passes (15.8% → 28.4% → 9.5% ambiguity); the second pass surfaced the facade-consumer scope, now mapped.
- Run `/analyze-spec` or add the spec-review label for an external pass.


moodlezoup:
**Spec Analysis: jolt-core Transcript Migration to Spongefish Split-Trait Surface**

| Dimension | Score | Gap |
|-----------|-------|-----|
| Goal | 0.92 | Clear |
| Constraints | 0.78 | One explicit deferred design decision (`transcript-poseidon` ↔ `challenge-254-bit` coupling) and one conditional fallback on a property the implementer must audit |
| Success Criteria | 0.85 | New `jolt_core_transcript_roundtrip` invariant is described conceptually — the "representative sequence" of calls is not pinned to an artifact or hand-written sequence |
| Context | 0.90 | Mostly accurate citations; minor drift (preamble at `zkvm/mod.rs:312-355`, not `:277-320`; declared 4 `RV64IMAC*` pairs but 2 resolve to identical Blake2b types) |
| **Ambiguity** | | **~13%** |

- Delete `jolt-core/src/transcripts/` (1,370 LOC confirmed) and `jolt-transcript/src/legacy.rs` (287 LOC confirmed), plus the `lib.rs` re-exports of `SpongeTranscript`, `MAX_LABEL_LEN`, `domain::{Label, LabelWithCount, U64Word}`, and the `Blake2bTranscript`/`KeccakTranscript`/`PoseidonTranscript` aliases.
- Re-point every transcript callsite (jolt-core: 27 files importing `crate::transcripts`; facade consumers `jolt-verifier`, `jolt-dory`, `jolt-sumcheck`, `jolt-openings`, `jolt-crypto`: 54 `T: Transcript<Challenge = …>` bound sites and ~55 `AppendToTranscript` impls per exploration) onto `jolt_transcript::{ProverTranscript, VerifierTranscript, OptimizedChallenge}`. Per-method mapping tables are in §Architecture for both migrations (jolt-core trait and facade trait have different shapes — explicit).
- Add `OptimizedChallenge for PoseidonSponge` (confirmed currently unimplemented in `crates/jolt-transcript/src/prover.rs`) with a pinned 128-bit squeeze rule: squeeze one `Fr` via `verifier_message::<Fr>`, truncate to low-128 bits of its canonical LE bigint representation, re-embed via `Fr::from(u128)`. Sponge advances by exactly one squeeze on both sides.
- Rewire jolt-core's `transcript-*` Cargo features to forward to jolt-transcript's; keep `challenge-254-bit` (it gates `JoltField::Challenge` in `field/ark.rs:44-49`, not `transcripts/`).

**Key invariants pinned:**
1. Prover/verifier challenge agreement under clean-break FS bytes — `muldiv` (host + zk) is the e2e gate; downstream Solidity/gnark/Lean verifiers are explicit non-goals (coordinated follow-ups).
2. Existing `jolt-eval` `soundness` RedTeam invariant must continue to hold post-swap.
3. `OptimizedChallenge::challenge_128` agreement across all three sponges, including the newly-added Poseidon impl.
4. Workspace-wide single transcript implementation: grep gate on `crate::transcripts::*` and `jolt_transcript::{Transcript, AppendToTranscript, SpongeTranscript, …}`.

**Critical evaluation criteria:**
- `cargo nextest run -p jolt-core muldiv --features host` and `--features host,zk` both pass.
- `cargo nextest run -p jolt-verifier -p jolt-blindfold -p jolt-eval` passes (model crate, BlindFold, invariants).
- `cargo nextest run -p jolt-sumcheck -p jolt-openings -p jolt-crypto` passes (facade-consumer migration validation).
- Clippy clean in both `host` and `host,zk` modes.
- Extended `transcript_symmetry` invariant (new shared `Op::OptimizedChallenge` variant) plus new `jolt_core_transcript_roundtrip` invariant pass under `cargo nextest run -p jolt-eval`.

**Residual items to resolve during implementation (flagged in spec, not blocking approval):**
- The "open question" on whether `transcript-poseidon` should retain its `challenge-254-bit` enable in `Cargo.toml:46`. The spec correctly notes that `OptimizedChallenge for PoseidonSponge` truncates to 128 bits regardless of `JoltField::Challenge` width, so the historical coupling looks vestigial — but the spec explicitly defers the decision. Recommend: keep the coupling unless the audit below proves it inert, since `Mont254BitChallenge` vs `MontU128Challenge` may affect non-transcript callers.
- The Poseidon low-128-bit min-entropy audit (audit (b) in §Architecture). If it fails for the Circom-compatible BN254 parameters, fall back to `Fr::from(s) mod 2^128` via the field's own reduction. Document the chosen rule inline at the impl.
- The `state() → spongefish state peek` mapping for facade consumers is conditional on "verify no production use" — grep first; if any production callsite is found, the mapping needs to be defined upfront.

> Note: the `claude-spec-approved` label should be added to this PR. (The remote review environment lacks the GitHub label-management MCP tool; @shreyas-londhe please add it manually, or the reviewing maintainer can.)


---

specs/jolt-core-transcript-migration.md
```diff
@@ -0,0 +1,174 @@
...
+
+### Non-Goals
+
+1. **Byte-identical Fiat-Shamir output.** This is an explicit clean break. The migrated transcript uses spongefish's duplex-sponge + NARG layout; serialized proofs produced before this PR will not verify after it, and that is acceptable. No backward-compatibility shim for old proof bytes.
+2. **Downstream verifier regeneration.** The Solidity verifier (`jolt-core/src/zkvm/transpilable_verifier.rs`), the gnark transpiler (`transpiler/`), and the Lean extractor (`zklean-extractor/`) consume the transcript byte layout and will need regeneration — these are coordinated follow-up PRs, named as non-goals in the parent spec, and are out of scope here.
```
moodlezoup: There is no Solidity verifier

specs/jolt-core-transcript-migration.md
```diff
@@ -0,0 +1,174 @@
...
..._bytes_le()[..16].try_into()?))`. This keeps the 128-bit challenge width identical across all three sponges. **Sponge-advancement pin:** the Poseidon `challenge_128` consumes exactly **one** `Fr`-worth of squeeze on both prover and verifier (the truncation happens on the squeezed value, not by squeezing fewer elements), so both sides advance the sponge state identically and a `challenge_128` at a given position leaves the sponge where a `verifier_message::<Fr>()` at that position would (one squeeze). This is what makes prover/verifier agreement (Invariant #3) hold despite Blake2b/Keccak squeezing a 16-byte `u128` while Poseidon squeezes a full field element. **Audit requirements (must be checked, not assumed):** (a) the prover and verifier `OptimizedChallenge` impls squeeze in lock-step so they derive the same value (Invariant #3); (b) the low-128-bit truncation preserves ≥128 bits of min-entropy for the Poseidon sponge output (the canonical low limbs must be uniform). If (b) fails for the Circom-compatible BN254 parameters, fall back to `Fr::from(s) ` reduced mod `2^128` via the field's own reduction. The chosen rule and its entropy justification are documented inline at the impl.
```
moodlezoup: I think we can leave challenge_128 `unimplemented!` for Poseidon. For recursion, the truncation itself is costly and somewhat defeats the purpose of using Poseidon in the first place. So to answer the earlier open question, `transcript-poseidon` should still enable `challenge-254-bit`.

transpiler/src/narg_parser.rs
```diff
@@ -0,0 +1,219 @@
...
+            return Err(NargParseError::TruncatedLengthPrefix { offset, remaining });
+        }
+        #[expect(clippy::unwrap_used)] // 8-byte slice into [u8; 8] is infallible
+        let len = u64::from_le_bytes(
+            narg[offset..offset + FRAME_LEN_PREFIX_BYTES]
+                .try_into()
+                .unwrap(),
+        );
+        let body_start = offset + FRAME_LEN_PREFIX_BYTES;
+        let body_remaining = (narg.len() - body_start) as u64;
```
moodlezoup: `parse_narg` re-implements the 8-byte LE length-prefix framing that `crates/jolt-transcript/src/codec.rs::read_length_prefixed_body` declares "the one authoritative parser ... so the accept/reject behavior cannot drift." Two independent copies of the NARG framing is exactly the drift that doc set out to prevent.

Not a correctness bug today — the `as u64`/`as usize` casts here are safe because the `len > body_remaining` check (line 119) runs in `u64` space before the `len as usize` at line 126, so a length above `usize::MAX` is rejected just as `read_length_prefixed_body`'s `usize::try_from` rejects it. But consider making `read_length_prefixed_body` (or a thin `split_frame` wrapper) `pub` and looping over it here, so the framing logic lives in one place.

---

moodlezoup (commented):
Reviewed the transcript migration across four parallel passes (semantic consistency, deep-bug, soundness/security, simplification), focused on the Fiat-Shamir-critical surface.

**No soundness or correctness issues found.** Specifically verified and clean:
- Absorb/squeeze ordering preserved 1:1 vs. the pre-PR code across the preamble, all 8 stages, sumcheck/uni-skip round loops, BlindFold, opening proofs, and the Dory bridge — every commitment/round-poly/opening is bound before the challenge that depends on it.
- `public_message` vs `prover_message` classification is correct: prover-only values (witness/advice commitments, round polys, openings) go into the NARG; shared/recomputable values are absorbed on both sides.
- Challenge embeddings preserved (no plain↔optimized swaps); Poseidon's `challenge_u128`/`challenge_128` `unimplemented!()` and `parse_scalar_units`' asserts are unreachable from attacker-controlled proof bytes.
- `fiat_shamir_instance` digest binds the same field set/order the old `fiat_shamir_preamble` did, computed byte-identically on both sides; NARG framing is length-prefixed and `usize::try_from`-guarded, with staged cursors and a forward-progress check; `check_eof` runs on the success path.

One minor maintainability note inline. Also: the PR description still says "Planning/spec PR — no code changes; only `specs/jolt-core-transcript-migration.md`," but this is the full 173-file implementation (the spec header reads `Status: implemented`) — worth updating so reviewers aren't misled about scope.

---

moodlezoup (commented):
We should try to keep the new crates decoupled from `ark-serialize`.
Also, can you split this into multiple PRs to make it easier to review?

transpiler/src/narg_parser.rs
Vishalkulkarni45: Thx for the review will fix it

Vishalkulkarni45:
> We should try to keep the new crates decoupled from `ark-serialize`. Also, can you split this into multiple PRs to make it easier to review?

sure, do you want this in two PRs (one for the spongefish migration and cleanup, and one for the new transpiler), or should I split it up even further?

Vishalkulkarni45:
> We should try to keep the new crates decoupled from `ark-serialize`. Also, can you split this into multiple PRs to make it easier to review?

do you mean we should remove `ark-serialize` usage entirely from the new crates, rather than just hiding it behind a local trait/API boundary and tests?

0xAndoroid:
Closing: the spongefish migration was put on pause (see moodlezoup's note on #1657) — transcript parity between jolt-core and the new crates was restored from the other direction via the byte-compatible DigestTranscript (#1617), and jolt-core has since been frozen as jolt-prover-legacy (#1632). If the migration is revived, it would target jolt-prover/jolt-verifier and need a fresh spec against those crates. Branch left in place.

PR #1596, opened 2026-06-09
feat: add field-inline tracing support

markosg04:

Base: `main`
Head: `prover-stack/01-field-inline-tracing`

Adds field-inline guest/tracer/program wiring and the SDK feature hooks needed by later prover work.

Validation before submission:
- `cargo fmt -q -- --check`


crates/jolt-riscv/src/lib.rs
```diff
@@ -156,6 +157,22 @@ macro_rules! for_each_instruction_kind {
                 VirtualXORROTW12 => VirtualXorRotW12 => "jolt.virtual.xorrotw12",
                 VirtualXORROTW8 => VirtualXorRotW8 => "jolt.virtual.xorrotw8",
                 VirtualXORROTW7 => VirtualXorRotW7 => "jolt.virtual.xorrotw7",
+                #[cfg(feature = "field-inline")]
+                FIELD_ADD => FieldAdd => "field.add",
```
0xAndoroid, line 162: These `#[cfg(feature = "field-inline")]` entries are forwarded as raw tokens to every callback macro, but `impl_inline_instruction!` in `crates/jolt-program/src/expand/inline.rs` still matches `$($instr:ident => $marker:ident => $canonical_name:expr),*` with no attribute support, so jolt-program fails with ``error: no rules expected `#` `` regardless of feature state. This is the root cause of the red CI; that callback needs the same `$(#[$meta:meta])*` treatment the tracer and jolt-riscv callbacks got.

crates/jolt-riscv/src/lib.rs
```diff
@@ -255,9 +296,6 @@ pub use profile::{
 pub use row::{JoltInstructionRow, NormalizedOperands, SourceInlineKey, SourceInstructionRow};
 pub use row_data::JoltInstructionRowData;
 pub use trace::JoltCycle;
```
0xAndoroid, line 300: Removing the `trace_row` re-exports breaks `tracer/src/trace_row.rs:16-19`, which still imports `CapturedState`, `JoltTraceRow`, `LoadState`, `NonMemoryState`, `StoreState`, and `TraceRowError` from `jolt_riscv` and is still declared in tracer's `lib.rs`. The tracer crate can't compile once the macro error in jolt-program is fixed.

crates/jolt-riscv/src/lib.rs
```diff
@@ -6,14 +6,15 @@
...
 
+#[cfg(feature = "field-inline")]
+pub mod field_inline;
 mod flags;
 pub mod instructions;
 mod kind;
 mod profile;
 mod row;
 mod row_data;
 pub mod trace;
```
0xAndoroid, line 17: Dropping `pub mod trace_row;` leaves `crates/jolt-riscv/src/trace_row.rs` orphaned on disk — no longer compiled but still present. If the #1532 JoltTraceRow foundation is being unwound here, the file (and its tracer-side consumer) should be deleted in the same change; otherwise the module declaration needs to stay.

jolt-sdk/macros/src/lib.rs
```diff
@@ -559,74 +563,15 @@ impl MacroBuilder {
                 };
                 let memory_layout = MemoryLayout::new(&memory_config);
 
-                let program_data =
-                    jolt::ProgramPreprocessing::preprocess(bytecode, memory_init, e_entry)?;
-                Ok(JoltSharedPreprocessing::new(
-                    program_data,
+                let preprocessing = JoltSharedPreprocessing::new(
```
0xAndoroid: This generates `JoltSharedPreprocessing::new(bytecode, memory_layout, memory_init, max_trace_length, e_entry)?`, but jolt-core's `JoltSharedPreprocessing::new` is the infallible 3-arg `new(program: ProgramPreprocessing<PCS>, memory_layout, max_padded_trace_length)` (zkvm/verifier.rs). Every `#[jolt::provable]` expansion fails to compile — this looks like it targets a jolt-core API from a later PR in the stack.

crates/jolt-program/src/image/decode.rs
```diff
@@ -262,6 +281,17 @@ fn operands(instruction_kind: SourceInstructionKind, word: u32) -> NormalizedOpe
         | SourceInstructionKind::AdviceLH
         | SourceInstructionKind::AdviceLW
         | SourceInstructionKind::AdviceLD => format_advice_load_operands(word),
+        #[cfg(feature = "field-inline")]
+        SourceInstructionKind::FIELD_ADD
+        | SourceInstructionKind::FIELD_SUB
+        | SourceInstructionKind::FIELD_MUL
+        | SourceInstructionKind::FIELD_ASSERT_EQ => format_r_operands(word),
```
0xAndoroid: `FIELD_ASSERT_EQ` decodes through `format_r_operands`, which populates `rd: Some(rd(word))`, but its operand shape declares no rd and the tracer's `FormatFieldInline::parse` yields `rd: None`. Since FIELD_ASSERT_EQ isn't in `field_inline_handles_rd_zero` and has side effects, every SDK-emitted `field_assert_eq!` (rd bits = 0) gets its bytecode-row rd rewritten to a fresh virtual register during expansion while the trace row carries `rd: None` — committed bytecode and trace operands diverge for the same PC once the prover consumes them. A dedicated formatter with `rd: None` keeps decode consistent with the shape.

tracer/src/utils/inline_helpers.rs
```diff
@@ -0,0 +1,538 @@
+//! InstrAssembler
```
0xAndoroid: This file isn't declared in `tracer/src/utils/mod.rs` and nothing references it, so it never compiles — and it resurrects the tracer `InstrAssembler` that #1533 deliberately removed in favor of `jolt-program/expand`. Suggest deleting it; a later PR that needs an assembler should build on the jolt-program recipes.

crates/jolt-program/src/expand/operands.rs
```diff
@@ -56,5 +56,23 @@ pub(super) const fn handles_rd_zero_internally(instruction_kind: SourceInstructi
             | SourceInstructionKind::EBREAK
             | SourceInstructionKind::CSRRW
             | SourceInstructionKind::CSRRS
+    ) || field_inline_handles_rd_zero(instruction_kind)
+}
+
+#[cfg(feature = "field-inline")]
+const fn field_inline_handles_rd_zero(instruction_kind: SourceInstructionKind) -> bool {
```
0xAndoroid, line 63: `FIELD_STORE_TO_X` is missing from this list. Its rd is the destination x-register, so a store to x0 takes the rd-zero rewrite path: the bytecode row gets a virtual register >= 32, which then fails `x_register()`'s `< RISCV_REGISTER_COUNT` bound during metadata construction (`InvalidXRegister`) — a confusing hard error for a legal (if pointless) store-to-x0, and the rewritten bytecode diverges from the un-expanded trace. If store-to-x0 should be a no-op like the fr-writing ops, it belongs in this list.

tracer/src/emulator/cpu.rs
```diff
@@ -93,12 +93,45 @@ use alloc::{boxed::Box, format, rc::Rc, string::String, vec::Vec};
...
+    fn default() -> Self {
+        Self {
+            registers: [FieldEncodedValue::zero(); jolt_riscv::FIELD_REGISTER_COUNT as usize],
+        }
+    }
+}
+
+#[cfg(feature = "field-inline")]
+impl FieldRegisterFile {
+    pub fn read(&self, register: u8) -> FieldEncodedValue {
```
0xAndoroid, line 121: The 5-bit encoding admits register indices 16-31, which `read` silently maps to zero and `write` silently drops, while `FieldInlineBytecodeRow::from_instruction` rejects them at preprocess time. A guest using fr16+ traces "successfully" with wrong semantics and only fails when proving. Rejecting out-of-range indices here (or at parse) keeps the emulator and the proving pipeline in agreement.

0xAndoroid (commented):
This slice doesn't build: the `#[cfg]` attributes added to jolt-riscv's instruction-list macros break the un-updated `impl_inline_instruction!` callback in jolt-program, the removed `JoltTraceRow` re-exports break tracer's `trace_row.rs`, and the SDK macro calls a `JoltSharedPreprocessing::new` signature that doesn't exist in this slice's jolt-core (details inline — this accounts for every red build job, plus machete; taplo also wants `crates/jolt-lookup-tables/Cargo.toml`, `jolt-core/Cargo.toml`, and `tracer/Cargo.toml` reformatted).

On the soundness side: I diffed all ~45 ported prefix `prefix_mle`/`update_prefix_checkpoint` implementations against the jolt-core originals — zero semantic divergences — and `required_prefixes()` matches every table's `combine()` usage. The remaining comments are mostly trace/preprocess strictness mismatches and witness-format ambiguities worth settling before the prover PRs consume these artifacts.

tracer/src/instruction/field_inline.rs
```diff
@@ -0,0 +1,330 @@
...
+fn execute_store_to_x(
+    op: FieldInlineOp,
+    operands: FormatFieldInline,
+    cpu: &mut Cpu,
+) -> FieldInlineTraceData {
+    let field_register = operands.rs1.unwrap_or(0);
+    let x_register = operands.rd.unwrap_or(0);
+    let field_value = cpu.field_registers.read(field_register);
+    let x_value = decode_field(field_value)
+        .to_canonical_u64_checked()
```
0xAndoroid, line 288: For field values >= 2^64 this silently truncates to the low 8 canonical bytes, and the trace records both `field_value` and the truncated `x_value` with no indication truncation happened. That freezes "x = field mod 2^64" into the witness format; whatever constraint later binds this bridge must either adopt that semantics or range-check, and nothing in the artifact forces the choice. Consider trapping (like assert_eq) or recording a fits-in-u64 flag so the semantics are explicit.

tracer/src/instruction/field_inline.rs
```diff
@@ -0,0 +1,330 @@
...
+    op: FieldInlineOp,
+    operands: FormatFieldInline,
+    cpu: &mut Cpu,
+) -> FieldInlineTraceData {
+    let rs1_register = operands.rs1.unwrap_or(0);
+    let rd_register = operands.rd.unwrap_or(0);
+    let rs1_value = cpu.field_registers.read(rs1_register);
+    let pre_value = cpu.field_registers.read(rd_register);
+    let inverse = decode_field(rs1_value)
+        .inverse()
```
0xAndoroid, line 200: inv(0) produces rd = 0 with `inv_product` = 0 (not 1), and nothing constrains `inv_product` at the artifact level. If the eventual constraint is `rs1 * rd == 1`, inv(0) traces become unprovable; if it's `rs1 * rd == inv_product`, rd is unconstrained for zero inputs. Worth deciding and documenting the inv(0) protocol semantics now, since this trace data is the contract later constraints build on.

crates/jolt-riscv/src/field_inline.rs
```diff
@@ -0,0 +1,293 @@
...
+#[cfg_attr(
+    feature = "serialization",
+    derive(
+        CanonicalSerialize,
+        CanonicalDeserialize,
+        serde::Serialize,
+        serde::Deserialize
+    )
+)]
+pub struct FieldRegister(pub u8);
```
0xAndoroid, line 151: The derived `CanonicalDeserialize` has a no-op `Valid`, so deserialization can produce `FieldRegister(200)`, bypassing the `new() < FIELD_REGISTER_COUNT` invariant — and `FieldInlineBytecodeMetadata::validate()` is only called in `from_bytecode`, never on a deserialize path. Nothing deserializes this metadata yet, but once preprocessing is loaded from disk the rows are trusted unvalidated. A real `Valid::check` here (and re-running `validate()` after deserialize) closes the gap.

jolt-core/src/host/program.rs
```diff
@@ -272,7 +307,7 @@ impl Program {
         jolt_program::build_jolt_program_with_inline_provider(
             &elf_contents,
             &mut inline_provider,
-            RV64IMAC_JOLT_ALL_INLINES,
+            self.instruction_profile,
```
0xAndoroid, line 310: `decode()`/`trace()` now honor `self.instruction_profile`, but the `JoltProgramSource::decode` trait default still goes through `guest::program::decode`, which hardcodes `RV64IMAC_JOLT` (as does `tracer::decode`). A field-inline ELF through those entry points panics on `IllegalSourceInstruction` via the `expect`. Worth plumbing the profile through all decode entry points so the feature doesn't depend on which API the caller picked.

crates/jolt-lookup-tables/Cargo.toml
```diff
@@ -11,16 +11,23 @@ categories = ["cryptography"]
 [lints]
 workspace = true
 
+[features]
+core-abi-tests = []
+field-inline = ["jolt-core/field-inline", "jolt-riscv/field-inline", "tracer/field-inline"]
```
0xAndoroid: `jolt-core` and `tracer` are dev-dependencies here, so these two feature edges only fire for this crate's own test builds. In a normal build, `jolt-program/field-inline` activates jolt-riscv's field-inline variants without tracer's, so a consumer mixing jolt-program (feature on) with jolt-core (feature off) gets a tracer built against a jolt-riscv whose enums it doesn't fully match. Consider keeping library feature edges to real dependencies and using `required-features` for the test-only activation (like `core-abi-tests` above).

crates/jolt-riscv/Cargo.toml
```diff
@@ -15,13 +15,13 @@ workspace = true
 default = ["serialization"]
 serialization = ["dep:ark-serialize", "dep:serde"]
 test-utils = ["dep:rand"]
+field-inline = []
 
 [dependencies]
 ark-serialize = { workspace = true, optional = true }
 rand = { workspace = true, optional = true }
```
0xAndoroid: With `JoltCycle::random` and the rest of the fuzz machinery removed from this crate, the optional `rand` dependency and the `test-utils` feature that enables it are dangling — this is what's failing the machete CI job.

crates/jolt-program/src/field_inline.rs
```diff
@@ -0,0 +1,398 @@
...
+    }
+}
+
+fn encoded_immediate(value: i128) -> Result<FieldEncodedValue, FieldInlineMetadataError> {
+    let value =
+        u64::try_from(value).map_err(|_| FieldInlineMetadataError::InvalidImmediate(value))?;
+    Ok(FieldEncodedValue::from_u64(value))
+}
+
+fn shape_for_op(op: FieldInlineOp) -> jolt_riscv::FieldInlineOperandShape {
```
0xAndoroid: This match is a byte-for-byte copy of the one inside jolt-riscv's `field_inline_operand_shape`. Exposing an op-keyed `pub const fn` in jolt-riscv (with `field_inline_operand_shape` delegating to it) removes this duplicate and also lets `FormatFieldInline::x_register_role` in the tracer drop its 8-arm op-to-kind-to-shape roundtrip.

crates/jolt-program/src/execution/trace.rs
```diff
@@ -189,6 +233,112 @@ pub struct TraceRow {
...
+    #[inline]
+    fn ram_write_value(&self) -> Option<u64> {
+        match self.ram_access {
+            RamAccess::Write(write) => Some(write.post_value),
+            RamAccess::Read(_) | RamAccess::NoOp => None,
+        }
+    }
+}
+
+impl JoltCycle for &TraceRow {
```
0xAndoroid: Identical body to the `TraceRow` impl above. A blanket `impl<T: JoltCycle> JoltCycle for &T` next to the trait (or one-line delegations) avoids maintaining two copies of all seven methods.

crates/jolt-program/src/lookup.rs
```diff
@@ -0,0 +1,119 @@
...
+
+use crate::execution::TraceRow;
+
+#[derive(Clone, Debug, Error, PartialEq, Eq)]
+pub enum LookupIndexError {
+    #[error("unsupported XLEN for Jolt lookup index derivation: {xlen}")]
+    UnsupportedXlen { xlen: usize },
+}
+
+pub fn instruction_lookup_index<const XLEN: usize>(
```
0xAndoroid, line 12: Neither `instruction_lookup_index` nor `JoltLookupQuery` has a production caller in this PR, and the per-call `validate_xlen` `Result` diverges from jolt-core's infallible `LookupQuery::to_lookup_index` pattern on what will be a hot path. If this is staging for the next PR, consider landing it with its consumer; either way the XLEN guard can be a const assertion instead of a runtime check.

crates/jolt-lookup-tables/src/tables/prefixes/and.rs
```diff
@@ -15,4 +16,47 @@ impl<F: Field> SparseDensePrefix<F> for AndPrefix {
...
     }
+
+    fn prefix_mle(
+        checkpoints: &[PrefixCheckpoint<F>],
+        r_x: Option<F>,
+        c: u32,
+        mut b: LookupBits,
+        j: usize,
+    ) -> F {
+        let _ = (checkpoints, r_x, c, b, j);
```
0xAndoroid: `let _ = (checkpoints, r_x, c, b, j);` — these parameters are all used directly below. This line (repeated across ~40 prefix files in both `prefix_mle` and `update_prefix_checkpoint`) looks like leftover codegen scaffolding and can be deleted everywhere.

tracer/src/instruction/field_inline.rs
```diff
@@ -0,0 +1,330 @@
...
+    }
+}
+
+fn execute_load_imm(
+    op: FieldInlineOp,
+    operands: FormatFieldInline,
+    cpu: &mut Cpu,
+) -> FieldInlineTraceData {
+    let rd_register = operands.rd.unwrap_or(0);
+    let value = u64::try_from(operands.imm)
```
0xAndoroid, line 315: A negative `imm` silently becomes zero here, while the preprocess-side `encoded_immediate` rejects it with `InvalidImmediate`. Unreachable today (both decoders zero-extend a 12-bit field), but the two sides should agree on the failure mode before a non-decode construction path appears.

tracer/src/instruction/field_inline.rs
```diff
@@ -0,0 +1,330 @@
...
+            fn execute(&self, cpu: &mut Cpu, ram_access: &mut Self::RAMAccess) {
+                ram_access.trace = Some(execute_field_inline($op, self.operands, cpu));
+            }
+        }
+
+        impl RISCVTrace for $name {}
+
+        impl From<super::SourceInstructionRow> for $name {
+            fn from(row: super::SourceInstructionRow) -> Self {
+                let mut operands = FormatFieldInline::from(row.operands);
```
0xAndoroid, line 76: This sets `operands.op` on the `SourceInstructionRow` path, but the `From<JoltInstructionRow>` conversion generated by `impl_final_jolt_row_data!(@from_row ...)` leaves `op: None`. Execution still works (the macro passes `$op` directly), but `x_register_role()` returns `None`, so x-register state capture silently disappears for LoadFromX/StoreToX if instructions are ever materialized from final bytecode rows. Currently unreachable for field ops, but worth setting `op` in both conversions to kill the asymmetry.

crates/jolt-lookup-tables/src/tables/mod.rs
```diff
@@ -239,6 +239,62 @@ impl<const XLEN: usize> LookupTableKind<XLEN> {
         dispatch!(self, t => PrefixSuffixDecomposition::suffixes(t))
     }
 
+    pub fn required_prefixes(&self) -> &'static [Prefixes] {
```
0xAndoroid: I'd rather not have this centralized table — the per-table prefix lists belong in the individual table files, next to each table's `combine()`/`suffixes()`, so adding a table touches one file. Also worth noting jolt-core has no equivalent list at all (its `update_checkpoints` updates every prefix unconditionally), so this is a new structure introduced by the port, and it has no consumers in this PR yet — good time to reshape or drop it.

crates/jolt-lookup-tables/src/tables/prefixes/andn.rs
```diff
@@ -16,4 +17,49 @@ impl<F: Field> SparseDensePrefix<F> for AndnPrefix {
...
+    }
+
+    fn update_prefix_checkpoint(
+        checkpoints: &[PrefixCheckpoint<F>],
+        r_x: F,
+        r_y: F,
+        j: usize,
+        suffix_len: usize,
+    ) -> PrefixCheckpoint<F> {
+        let _ = (checkpoints, r_x, r_y, j, suffix_len);
```
0xAndoroid: Forward-looking note: I want to redesign the prefix-suffix mechanism to materialize the entire small 2^16 prefix table up front, which would eliminate `update_prefix_checkpoint` entirely and shrink this code surface. Not blocking for this port, but worth keeping the checkpoint machinery thin — it may not survive the redesign.

crates/jolt-program/src/field_inline.rs
```diff
@@ -0,0 +1,398 @@
...
+        limb_count: 4,
+        canonical: true,
+    };
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(
+    feature = "serialization",
+    derive(
+        CanonicalSerialize,
```
0xAndoroid: Why are we still deriving `CanonicalSerialize`/`CanonicalDeserialize` here — I thought we were moving off ark serialization in the new stack? It does match the existing `serialization`-feature pattern in jolt-riscv/jolt-program (kind.rs, row.rs, preprocess/*), so it's consistent with what's landed; flagging the direction question.

0xAndoroid (commented):
A few design-direction notes on the lookup-tables port, follow-up to the earlier review.

crates/jolt-riscv/src/field_inline.rs
0xAndoroid: CanonicalSerialize / eDeserialize should be removed regardless

crates/jolt-riscv/src/lib.rs
0xAndoroid: the trace row is still declared here? wdym?

0xAndoroid:
**Stack maintenance** — #1596 is the base of the prover stack and targets `main` directly.

Pika validates it against the latest `main` and fixes CI as needed; it stays as the foundation the rest of the stack builds on.

📚 **Stack A:** **#1596 (this PR)** → #1597 → #1598 → #1599 → #1600 → #1601 → #1602 → #1603
📚 **Stack B:** **#1596 (this PR)** → #1597 → #1604 → #1605 → #1606

0xAndoroid (approved):
lgtm

PR #1597, opened 2026-06-09
feat: extend Jolt claim protocol formulas

markosg04:
Part of the draft Jolt prover stack.

Base: `prover-stack/01-field-inline-tracing`
Head: `prover-stack/02-jolt-claims-protocol-formulas`

Extends the Jolt claim formulas, including field-inline protocol formulas and supporting specs.

Validation before submission:
- `cargo fmt -q -- --check`
- `cargo metadata --no-deps --format-version 1`
- local `gh stack push` simulation against a temporary bare remote
- forbidden-path scan for `handoffs/`, old `STACK.md`, `stack/`, and old stack workflow

PR #1598, opened 2026-06-09
feat: add R1CS extension foundations

markosg04:

Base: `prover-stack/02-jolt-claims-protocol-formulas`
Head: `prover-stack/03-r1cs-extensions`

Adds shared R1CS, transcript, polynomial, field, BlindFold, and crypto foundations used by protocol wrappers.

Validation before submission:
- `cargo fmt -q -- --check`


0xAndoroid:
**Stack maintenance** — Pika is rebasing this PR onto #1597 (`prover-stack/02-jolt-claims-protocol-formulas`) and fixing CI as part of the prover stack (#1596–#1606). Rebase + CI summary to follow.

0xAndoroid:
**Stack maintenance** — rebased onto #1597 (`prover-stack/02-jolt-claims-protocol-formulas`); CI fixed.

📚 **Stack A:** #1596 → #1597 → **#1598 (this PR)** → #1599 → #1600 → #1601 → #1602 → #1603
📚 **Stack B:** #1596 → #1597 → #1604 → #1605 → #1606

**Conflict resolution**
- `crates/jolt-claims/src/protocols/jolt/formulas/bytecode.rs` — kept #1597's committed-program / two-phase bytecode formulas; #1598's overlapping single-phase formula expansion was superseded by the parent stack.
- `crates/jolt-claims/src/protocols/jolt/formulas/claim_reductions/advice.rs` — kept #1597's shared precommitted reduction layout/scheduling; #1598's older local split helpers were superseded.
- `crates/jolt-claims/src/protocols/jolt/formulas/error.rs` — kept #1597's richer precommitted/bytecode error variants and retained the evaluation-domain overflow variant.
- `crates/jolt-poly/src/{dense.rs,eq.rs}` — kept #1597's current optimized/equivalent helpers; #1598's overlapping helper edits were already present or equivalent.
- `crates/jolt-transcript/{Cargo.toml,src/digest.rs,src/lib.rs}` — kept #1597's spongefish-compatible transcript surface, added only the R1CS transcript feature/module from #1598, and resolved the add/add on `digest.rs` by keeping #1597's core-compatible digest transcript implementation.

crates/jolt-field/src/arkworks/small_scalar_accumulator.rs
```diff
@@ -0,0 +1,120 @@
+use crate::{Limbs, SignedScalarAccumulator};
+
+use super::{bn254::Fr, bn254_ops};
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct FrSmallScalarAccumulator {
+    pos: Limbs<5>,
+    neg: Limbs<5>,
+}
```
moodlezoup, line 9: `pos`/`neg` are only `Limbs<5>` (320 bits), but `fmadd_magnitude` adds terms up to `value (~254 bits) × scalar (~64 bits) ≈ 2^318`, and `add_assign_trunc::<5>` truncates with no overflow check or `debug_assert`. Summing only a few full-width terms wraps silently and `reduce()` returns a wrong value with no signal (confirmed empirically). Nothing calls this yet so it's dormant, but `fmadd_u64`/`add` accept arbitrary `Fr` values, so a caller passing full field elements gets silent corruption. Unlike `WideAccumulator`/`FrSignedProductAccumulator` (u128 slots, documented headroom), the capacity here is undocumented. Either widen to u128-based slots or document the small-value precondition on `SignedScalarAccumulator` and add a `debug_assert` on the carry-out limb.

---

crates/jolt-transcript/src/r1cs/poseidon.rs
```diff
@@ -0,0 +1,568 @@
...
+    let mut lc = LinearCombination::zero();
+    let mut coefficient = Fr::from_u64(1);
+    let radix = Fr::from_u64(256);
+    for byte in bytes {
+        value += byte.value * coefficient;
+        lc = lc + byte.lc.clone().scale(coefficient);
+        coefficient *= radix;
+    }
+    AssignedScalar::new(value, lc)
+}
```
moodlezoup, line 208: `pack_bytes` computes `Σ byte_i · 256^i` over the byte witnesses' `lc` but never range-constrains each byte to `[0,256)`. The native `from_le_bytes_mod_order` has no such freedom, so for the gadget to bind to a canonical byte string the caller must guarantee each `Byte` is in range — otherwise a malicious witness (e.g. `byte_0 = 256, byte_1 = 0` vs `byte_0 = 0, byte_1 = 1`) produces the same packed scalar and the same transcript state from different "byte strings", breaking the binding of `absorb_bytes`. This obligation is currently undocumented on `R1csByteTranscript::Byte`/`absorb_bytes`. Either range-check internally or document the `[0,256)` invariant prominently.

---

crates/jolt-blindfold/src/verify.rs
```diff
@@ -14,481 +14,433 @@ const OUTER_SUMCHECK_DEGREE: usize = 3;
...
+            actual: inner.value,
+        });
     }
+
+    append_vector_opening(
+        transcript,
+        b"bf_witness_opening",
+        b"bf_witness_blind",
+        &proof.witness_opening,
+    );
```
moodlezoup, line 441: Transcript asymmetry: the verifier appends `bf_witness_opening`/`bf_witness_blind` here, but the prover (`prove.rs`, after `open_rows` for the folded witness) never appends it — it only appends `bf_error_opening`. Harmless today since this is the last transcript op and no challenge is derived afterward, but it's a real prover/verifier divergence. If this BlindFold transcript is ever chained into a larger protocol (the stated direction of this stack), the verifier's transcript state will diverge from the prover's. Either append it on the prover side too, or drop this append.

---

crates/jolt-r1cs/src/nonnative.rs
```diff
@@ -0,0 +1,918 @@
...
+
+fn constant_mul_terms(lhs: [u64; NUM_LIMBS], rhs: [u64; NUM_LIMBS]) -> [BigUint; MUL_LIMBS] {
+    let mut terms = std::array::from_fn(|_| BigUint::zero());
+    for (lhs_index, lhs_limb) in lhs.into_iter().enumerate() {
+        for (rhs_index, rhs_limb) in rhs.into_iter().enumerate() {
+            terms[lhs_index + rhs_index] += BigUint::from(lhs_limb) * BigUint::from(rhs_limb);
+        }
+    }
+    terms
+}
```
moodlezoup, line 539: `constant_mul_terms` is a byte-for-byte copy of `convolution_terms` above (lines 521-529). Both callers (the product convolution at line 300 and the modulus·quotient reduction at line 314) can share a single helper — drop one.

---

crates/jolt-field/src/arkworks/signed_product_accumulator.rs
```diff
@@ -0,0 +1,121 @@
...
+        let mut out = [0u64; 9];
+        let mut carry = 0u128;
+        for (index, slot) in slots.into_iter().enumerate() {
+            let (sum, overflow) = slot.overflowing_add(carry);
+            out[index] = sum as u64;
+            carry = (sum >> 64) + ((overflow as u128) << 64);
+        }
+        out[8] = carry as u64;
+        Limbs(out)
+    }
```
moodlezoup, line 47: The 4×4 `u128`-slot product loop and the `normalize` carry-propagation pass here are identical to `wide_accumulator.rs` (`fmadd`/`normalize`), differing only in `Limbs<9>` vs `BigInt<9>` return type (which interconvert via `arkworks/mod.rs`). This signed-product accumulator is essentially two wide accumulators (pos/neg) plus a sign branch in `reduce`. Extract a shared `fmadd_into(slots, a, b)` + `normalize(slots)` so the carry logic lives in one place.

---

moodlezoup (commented):
Reviewed the R1CS foundations across the nonnative-field, EC-gadget, Poseidon-transcript, polynomial, and BlindFold/sumcheck additions. No soundness-critical issues found — the nonnative `Fq`-in-`Fr` arithmetic (range checks, carry chains, modular-reduction uniqueness), the Grumpkin EC gadgets, and the Nova-fold/Spartan structure all check out, and the split-eq/Lt/Lagrange/one-hot ports are validated against dense references. The inline comments are mostly latent (these foundations have few or no callers yet), but worth pinning down before they get wired up.

A few smaller notes not worth inline comments:
- `poseidon.rs`: the in-circuit transcript only self-validates against a hand-rolled `NativeTranscript` in the same file, not against the actual `jolt-core` `PoseidonTranscript` it's meant to reproduce bit-for-bit. There are two divergent native transcripts in the repo (LE/32-byte/`n_rounds` vs BE/31-byte/length-prefixed); the gadget matches the former. A pinned known-answer test against the real prover transcript would catch silent desync.
- `split_eq.rs:275-315`: `e_out_in_for_window`/`e_active_for_window` are caller-less and untested, with index math guarded only by `debug_assert!`. Add tests against a dense reference or defer to the PR that introduces a caller.
- `nonnative.rs`: `CARRY_BITS = 68` is load-bearing for both completeness and the no-wrap soundness argument (honest max carry is 66 bits) — worth a WHY comment.
- `prove.rs` BlindFold outer/inner sumcheck accumulation loops are serial in the dominant first rounds (unlike the `par_iter` hot sumchecks in jolt-core); fine for a small verifier R1CS but worth revisiting if the circuit grows.

---

0xAndoroid:
Closing the prover-stack/03–08 recursion/wrapper chain (#1598–#1603) as stale: dormant since June 17, and the cheap-verification goal moved to the Akita lattice PCS line (#1655/#1675/#1676, ports in #1718/#1732), with the ZK need served by BlindFold in the modular prover (#1690). The stack's merge-base predates ~7 weeks of main (incl. #1632 jolt-core→jolt-prover-legacy, #1654, #1665), so merging as-is would silently revert shared files. The recursion content itself never landed elsewhere — branches left in place in case the wrapper direction revives.

PR #1600, opened 2026-06-09
feat: add HyperKZG zero-knowledge openings

markosg04:
Part of the draft Jolt prover stack.

Base: `prover-stack/04-hyrax`
Head: `prover-stack/05-hyperkzg-zk`

Adds HyperKZG ZK setup/opening support, tests, and protocol spec.

Validation before submission:
- `cargo fmt -q -- --check`
- `cargo metadata --no-deps --format-version 1`
- local `gh stack push` simulation against a temporary bare remote
- forbidden-path scan for `handoffs/`, old `STACK.md`, `stack/`, and old stack workflow

PR #1599, opened 2026-06-09
feat: add Hyrax row commitments

markosg04:
Part of the draft Jolt prover stack.

Base: `prover-stack/03-r1cs-extensions`
Head: `prover-stack/04-hyrax`

Adds the Hyrax row commitment crate used by later assist and wrapper protocols.

Validation before submission:
- `cargo fmt -q -- --check`
- `cargo metadata --no-deps --format-version 1`
- local `gh stack push` simulation against a temporary bare remote
- forbidden-path scan for `handoffs/`, old `STACK.md`, `stack/`, and old stack workflow

PR #1601, opened 2026-06-09
feat: add PCS assist Dory opening formulas

markosg04:

Base: `prover-stack/05-hyperkzg-zk`
Head: `prover-stack/06-pcs-assist-dory-openings`

Adds Dory PCS-assist opening formulas, verifier protocol hooks, and the Dory assist protocol spec.

Validation before submission:
- `cargo fmt -q -- --check`


0xAndoroid:
**Stack maintenance** — Pika is rebasing this PR onto #1600 (`prover-stack/05-hyperkzg-zk`) and fixing CI as part of the prover stack (#1596–#1606). Rebase + CI summary to follow.

0xAndoroid:
**Stack maintenance** — rebased onto #1600 and retargeted this PR's base to `prover-stack/05-hyperkzg-zk`; CI green.

📚 **Stack A:** #1596 → #1597 → #1598 → #1599 → #1600 → **#1601 (this PR)** → #1602 → #1603
📚 **Stack B:** #1596 → #1597 → #1604 → #1605 → #1606

**Conflict resolution**
- `crates/jolt-dory/src/{scheme.rs,streaming.rs,types.rs}` — integrated the Dory-assist opening artifact/transcript extraction with the existing stack's Dory streaming and transcript interfaces; fixed transcript replay to consume an owned transcript and removed unnecessary clones on copyable field values.
- `crates/jolt-verifier/**` — the PR's verifier wiring targeted older precommitted/advice/program-image APIs; preserved #1600's verifier stage architecture and kept the new Dory-assist formulas/artifacts without reintroducing obsolete verifier wiring.
- `crates/jolt-claims/src/protocols/dory_assist/formulas/miller_loop.rs` — renamed `G2HomState` to `G2MapState` to satisfy the repo typos check; no semantic protocol change.

crates/jolt-dory/src/scheme.rs
```diff
@@ -217,16 +385,7 @@ impl CommitmentScheme for DoryScheme {
...
-            || proof.0.y_com.is_some()
-            || proof.0.sigma1_proof.is_some()
-            || proof.0.sigma2_proof.is_some()
-            || proof.0.scalar_product_proof.is_some()
-        {
-            return Err(OpeningsError::VerificationFailed);
-        }
-
-        dory::verify::<ArkFr, InnerBN254, G1Routines, G2Routines, _>(
+        dory::verify::<ArkFr, InnerBN254, JoltG1Routines, JoltG2Routines, _>(
```
moodlezoup, line 388: This drops the guard that rejected proofs carrying ZK artifacts (`e2`/`y_com`/`sigma1_proof`/`sigma2_proof`/`scalar_product_proof`) on the clear path. `dory` is built with `zk` enabled unconditionally (root `Cargo.toml`), so the `#[cfg(feature = "zk")]` branch in `verify_evaluation_proof` is always live: when those fields are `Some`, it takes `e2` from the proof and the caller-supplied `eval` is never used (only the `(None, None)` arm binds `e2 = g2_0^eval`). It then only checks the sigma proofs, i.e. that `y_com`/`e2` open to *some* value — not that that value equals `eval`.

`stage8_clear` calls `PCS::verify(&joint_commitment, point, joint_claim, &proof.joint_opening_proof, …)`. A prover who submits a ZK-shaped `joint_opening_proof` gets `Ok(())` regardless of `joint_claim`, severing the binding between the sumcheck-derived claim and the committed witness. The clear stage-8 path is internally reachable in this crate, so the binding is unsound here even though `jolt-verifier` isn't the production verifier yet.

`types.rs` already has `has_transparent_opening_artifacts()` — restore the rejection (call it and bail otherwise) on the clear path. `verify_zk` is fine since it requires `y_com.is_some()` and dory rejects mixed shapes.

---

crates/jolt-dory/src/scheme.rs
```diff
@@ -270,28 +429,32 @@ impl AdditivelyHomomorphic for DoryScheme {
...
-            "combine_hints: ragged hint lengths",
-        );
+        // Hints may be ragged: dense (e.g. increment) commitments use a smaller
+        // matrix than one-hot commitments, so their row-commitment vectors are
+        // shorter. Resize to the maximum row count (matching core, which resizes
+        // to the global max), treating absent rows as the identity commitment.
+        let num_rows = hints
+            .iter()
+            .map(|hint| hint.row_commitments.len())
+            .max()
```
moodlezoup, line 439: The comment says this resizes "to the global max" like core, but it uses the per-call batch max (`hints.iter().map(len).max()`), whereas `jolt-core/src/poly/commitment/dory/commitment_scheme.rs:311` uses `DoryGlobals::get_max_num_rows()`. These agree only when the batch happens to contain a hint at the global row count. If `combine_hints` is ever called on a subset whose local max is smaller than `2^nu`, the combined hint is shorter than the matrix shape `open`/`dory::prove` derive from `point.len()`. Latent today (the live prover still uses the core impl), but worth aligning before this crate replaces it — either resize to the global max or document why the batch always contains the max-width hint.

---

crates/jolt-verifier/src/pcs_assist.rs
```diff
@@ -0,0 +1,100 @@
+//! Generic PCS-assist verifier boundary.
```
moodlezoup, line 1: This file isn't declared anywhere — `lib.rs` lists only `compat, config, error, preprocessing, proof, stages, verifier`, and there's no `#[path]`/`build.rs` pulling it in. As an orphan module it's excluded from compilation, so the `PcsProofAssist` trait, `NoPcsAssist` impl, and the input structs get no `cargo build`/`clippy`/test coverage and can silently rot. If it's intended scaffolding for a later stack PR, either add `pub mod pcs_assist;` now so CI checks it, or hold the file until the consumer lands. (Same applies to `crates/jolt-verifier/src/stages/stage8/final_openings.rs`, which `stage8/mod.rs` doesn't declare.)

---

crates/jolt-verifier/src/stages/stage8/final_openings.rs
```diff
@@ -0,0 +1,595 @@
+#[cfg(feature = "field-inline")]
```
moodlezoup, line 1: `stage8/mod.rs` declares only `inputs`, `outputs`, `verify` — there's no `mod final_openings;`, so these 595 lines (and their `#[cfg(test)]` module) are never compiled. None of `stage8_final_opening_*` are referenced from `stage8/verify.rs`. Add the `mod` declaration so it's type-checked/linted, or defer the file until it's wired into the verify flow.

---

moodlezoup (commented):
Reviewed the Dory PCS-assist additions. The formula/relation tree under `dory_assist/` is the bulk of the diff but isn't wired into live verification yet (only the no-op `NoPcsAssist` exists), so I focused on the code that compiles and runs: the dory scheme/streaming/types changes and the stage-8 opening path.

Main thing to look at: the clear `DoryScheme::verify` no longer rejects ZK-artifact proofs, and with the `zk` dory feature always on this lets a ZK-shaped proof pass `PCS::verify` while ignoring the claimed evaluation — a binding gap in the stage-8 clear path. Details inline. Also two new verifier files (`pcs_assist.rs`, `stage8/final_openings.rs`) are orphan modules that never get compiled, and a `combine_hints` row-count divergence from core that's latent for now.

The reduce-round transcript replay, fold-factor coordinate math, one-hot commitment fast paths, the `repr(transparent)` transmutes, and the streaming bounds checks all read correctly against the upstream dory implementation.

---

0xAndoroid:
Closing the prover-stack/03–08 recursion/wrapper chain (#1598–#1603) as stale: dormant since June 17, and the cheap-verification goal moved to the Akita lattice PCS line (#1655/#1675/#1676, ports in #1718/#1732), with the ZK need served by BlindFold in the modular prover (#1690). The stack's merge-base predates ~7 weeks of main (incl. #1632 jolt-core→jolt-prover-legacy, #1654, #1665), so merging as-is would silently revert shared files. The recursion content itself never landed elsewhere — branches left in place in case the wrapper direction revives.

PR #1603, opened 2026-06-09
feat: add wrapper verifier

markosg04:
Part of the draft Jolt prover stack.

Base: `prover-stack/07-dory-assist-verifier`
Head: `prover-stack/08-wrapper-verifier`

Adds the wrapper verifier crate, wrapper claim formulas, and wrapper protocol specs.

Validation before submission:
- `cargo fmt -q -- --check`
- `cargo metadata --no-deps --format-version 1`
- local `gh stack push` simulation against a temporary bare remote
- forbidden-path scan for `handoffs/`, old `STACK.md`, `stack/`, and old stack workflow

PR #1602, opened 2026-06-09
feat: add Dory assist verifier

markosg04:
Part of the draft Jolt prover stack.

Base: `prover-stack/06-pcs-assist-dory-openings`
Head: `prover-stack/07-dory-assist-verifier`

Adds the concrete Dory assist verifier crate and its completeness/soundness fixtures.

Validation before submission:
- `cargo fmt -q -- --check`
- `cargo metadata --no-deps --format-version 1`
- local `gh stack push` simulation against a temporary bare remote
- forbidden-path scan for `handoffs/`, old `STACK.md`, `stack/`, and old stack workflow

PR #1604, opened 2026-06-09
feat: add Jolt witness crate

markosg04:

Base: `prover-stack/08-wrapper-verifier`
Head: `prover-stack/09-jolt-witness`

Adds the standalone Jolt witness crate and protocol-specific witness providers.

Validation before submission:
- `cargo fmt -q -- --check`

### DIFF
--- a/crates/jolt-witness/src/protocols/jolt_vm/stage2.rs
+++ b/crates/jolt-witness/src/protocols/jolt_vm/stage2.rs
@@ -1,0 +1,106 @@
+//! Stage 2 trace rows.
+
+use super::*;
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct JoltVmStage2TraceRow {
+    pub remapped_ram_address: Option<usize>,
+    pub ram_read_value: u64,
+    pub ram_write_value: u64,
+    pub ram_increment: i128,
+    pub left_instruction_input: u64,
+    pub right_instruction_input: i128,
+    pub lookup_output: u64,
+    pub left_lookup_operand: u64,
+    pub right_lookup_operand: u128,
+    pub branch_flag: bool,
+    pub jump_flag: bool,
+    pub write_lookup_output_to_rd_flag: bool,
+    pub virtual_instruction_flag: bool,
+    pub next_is_noop: bool,
+}
+
+pub trait JoltVmStage2Rows {
+    fn stage2_rows(&self) -> Result<Vec<JoltVmStage2TraceRow>, WitnessError>;
+
+    fn initial_ram_state_words(&self) -> Result<Vec<u64>, WitnessError>;
+
+    fn final_ram_state_words(&self) -> Result<Vec<u64>, WitnessError>;
+}
+
+impl<T: TraceSource + Clone> JoltVmStage2Rows for TraceBackedJoltVmWitness<'_, T> {
+    fn stage2_rows(&self) -> Result<Vec<JoltVmStage2TraceRow>, WitnessError> {
+        let rows = checked_pow2(self.config.log_t)?;
+        let mut values = Vec::with_capacity(rows);
+        let mut trace = self.trace.trace.clone();
+        let mut current = trace.next_row().unwrap_or_default();
+        for index in 0..rows {
+            let next = (index + 1 < rows).then(|| trace.next_row().unwrap_or_default());
+            let remapped_ram_address = ram_access_address(current.ram_access)
+                .map(|address| self.remapped_ram_address(address))
+                .transpose()?
+                .flatten();
+            values.push(stage2_trace_row(
+                &current,
+                next.as_ref(),
+                remapped_ram_address,
+            )?);
+            if let Some(row) = next {
+                current = row;
+            }
+        }
+        Ok(values)
+    }
+
+    fn initial_ram_state_words(&self) -> Result<Vec<u64>, WitnessError> {
+        self.initial_ram_state()
+    }
+
+    fn final_ram_state_words(&self) -> Result<Vec<u64>, WitnessError> {
+        self.final_ram_state()
+    }
+}
+
+pub(crate) fn stage2_trace_row(
+    row: &TraceRow,
+    next: Option<&TraceRow>,
+    remapped_ram_address: Option<usize>,
+) -> Result<JoltVmStage2TraceRow, WitnessError> {
+    let instruction = JoltInstruction::try_from(row.instruction).map_err(|kind| {
+        WitnessError::InvalidWitnessData {
+            namespace: JOLT_VM_NAMESPACE.name,
+            reason: format!("unsupported Jolt instruction kind in trace row: {kind:?}"),
+        }
+    })?;
+    let circuit_flags = instruction.circuit_flags();
+    let instruction_flags = instruction.instruction_flags();
+    let query = JoltLookupQuery::new(row.instruction.instruction_kind, row);
+    let (left_instruction_input, right_instruction_input) =
+        LookupQuery::<RV64_XLEN>::to_instruction_inputs(&query);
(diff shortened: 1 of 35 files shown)


0xAndoroid:
**Stack maintenance** — Pika is rebasing this PR onto #1597 (`prover-stack/02-jolt-claims-protocol-formulas`) and fixing CI as part of the prover stack (#1596–#1606). Rebase + CI summary to follow.

0xAndoroid:
**Stack maintenance** — rebased onto #1597 (`prover-stack/02-jolt-claims-protocol-formulas`); CI green. Ready for review.

📚 **Stack A:** #1596 → #1597 → #1598 → #1599 → #1600 → #1601 → #1602 → #1603
📚 **Stack B:** #1596 → #1597 → **#1604 (this PR)** → #1605 → #1606

**Conflict resolution**
- `crates/jolt-claims/src/protocols/jolt/formulas/{bytecode.rs,claim_reductions/advice.rs,error.rs}` — kept #1597's committed-program / precommitted scheduling structure; preserved #1604's formula additions where compatible.
- `crates/jolt-poly/src/{dense.rs,eq.rs}` — kept #1597's helper implementations; #1604's overlapping doc/helper changes were already covered.
- `crates/jolt-transcript/{Cargo.toml,src/{digest.rs,lib.rs,poseidon.rs},tests/*}` — kept #1597's spongefish/legacy-compatible transcript surface; added only the R1CS transcript hook needed by #1604.
- `crates/jolt-verifier/**` — resolved broad PCS-assist vs #1597 committed-program verifier conflicts by keeping #1597's verifier semantics for this PR and removing premature Dory-assist verifier workspace wiring; later Chain B PRs reintroduce it on the now-green base.
- CI fixes — removed unavailable workspace crates for this PR boundary, restored missing Jolt IDs / embedding handling, and fixed typos/machete/clippy issues.

**Local validation**
- `cargo fmt -q`
- `cargo build -q`

crates/jolt-witness/src/protocols/jolt_vm/stage2.rs
```diff
@@ -0,0 +1,106 @@
...
+        LookupQuery::<RV64_XLEN>::to_instruction_inputs(&query);
+    let (left_lookup_operand, right_lookup_operand) =
+        LookupQuery::<RV64_XLEN>::to_lookup_operands(&query);
+    let lookup_output = LookupQuery::<RV64_XLEN>::to_lookup_output(&query);
+    let (ram_read_value, ram_write_value) = match row.ram_access {
+        RamAccess::Read(read) => (read.value, read.value),
+        RamAccess::Write(write) => (write.pre_value, write.post_value),
+        RamAccess::NoOp => (0, 0),
+    };
+    let next_is_noop = next.is_none_or(row_is_noop);
```
moodlezoup: `next_is_noop` here returns `true` at the last cycle (`next == None`), but the canonical definition in `jolt-core/src/zkvm/r1cs/inputs.rs:339` and this crate's own `spartan_outer.rs:109` / `trace.rs:214` all use `is_some_and` (→ `false` at the boundary). Since this feeds `ShouldJump = Jump·(1 − NextIsNoop)`, the boundary value is inverted relative to every other path.

```suggestion
    let next_is_noop = next.is_some_and(row_is_noop);
```

---

crates/jolt-witness/src/protocols/jolt_vm/stage6.rs
```diff
@@ -0,0 +1,61 @@
...
+            let ram_address = ram_access_address(row.ram_access);
+            let remapped_ram_address = ram_address
+                .and_then(|address| {
+                    self.preprocessing
+                        .memory_layout
+                        .remap_word_address(address)
+                        .ok()
+                })
+                .flatten()
+                .map(|address| address as usize);
```
moodlezoup: This computes `remapped_ram_address` inline without the `>= ram_k` bound that the shared `remapped_ram_address()` helper (ram.rs:437) enforces — and which stage2 and every path in ram.rs route through. An address that remaps beyond `ram_k` passes silently here and would index out of the `ram_k`-sized one-hot domain downstream. Mirror stage2:

```suggestion
            let remapped_ram_address = ram_access_address(row.ram_access)
                .map(|address| self.remapped_ram_address(address))
                .transpose()?
                .flatten();
```

---

moodlezoup (commented):
Reviewed the **jolt-witness** crate that this PR actually introduces. To get a tractable diff I reviewed the incremental range `81a7c62..57390c6` (the 6 witness commits) rather than the full GitHub diff — heads up that the PR base is set to `prover-stack/02` while the description says `08`, which inflates the displayed diff to 251 files / 64k lines. Re-targeting the base to `08` would make this reviewable.

**No soundness issues.** The booleanity address/cycle phase split is constraint-preserving (the combined output is reused verbatim, the degree-3 bound and round counts partition correctly, and the intermediate claim is bound across phases via the shared `BooleanityAddrClaim` opening). The crate is prover-side only, with no `unsafe`, no attacker-reachable panics, and dimension/overflow guards (`checked_pow2`, `power_of_two_log_rows`) used consistently.

Two latent correctness divergences flagged inline — both currently dormant (their `stageN_rows` have no consumer yet) but will produce wrong/malformed witness once wired into the prover.

Lower-priority notes (non-blocking):
- **Missing-PC handling diverges across paths.** `stage6.rs:39` (`pc_for_row_optional(...).unwrap_or(0)`) and `streams.rs:493` (`get_pc(...)` → silent non-hot row) silently swallow a missing bytecode PC for a real instruction, whereas the fast path `ra.rs::bytecode_pc_for_row` errors via `missing_pc_mapping`. Unreachable for valid traces (every executed instruction is mapped; NoOp→`Some(0)`), so defensive consistency rather than a live bug — but worth unifying on the erroring path.
- **Speculative surface.** `MaterializationPolicy::Streaming` is never constructed, and neither `ViewRequirement.materialization` nor `.retention` is read by any non-test logic (`PolynomialView::Deferred` is only constructed in tests). Per the repo's "abstractions must pay for themselves within the PR" guidance, consider trimming until a consumer exists — though this is forward-looking API for the stack, so reasonable to keep if a near-term consumer is planned.
- **Duplication of jolt-core.** `RaChunkSelector` (ra.rs), `JoltVmIncrementStreamKind::value_from_row` (streams.rs:452), and the one-hot/compact expansion in `streams.rs` re-derive `OneHotParams` chunk math and `witness.rs::stream_witness_and_commit_rows` increment logic. Presumably intentional for the modular-crate migration; worth a note in the PR that the jolt-core versions are slated for deletion so this doesn't become permanent duplication.
- **Perf (when wired up).** `materialize_ram_val`/`materialize_ram_ra`/RegistersVal are sequential O(K·T) loops; jolt-core parallelizes the equivalent generation. `field_inline` already uses rayon, so the gap is confined to the dense-materialization fallbacks.

---

crates/jolt-witness/src/protocols/jolt_vm/stage2.rs
moodlezoup: Fixed in 98b6a4cdb — switched to `is_some_and`, so `next_is_noop` is `false` at the final cycle, matching `jolt-core` r1cs/inputs.rs, spartan_outer.rs, and trace.rs.

crates/jolt-witness/src/protocols/jolt_vm/stage6.rs
moodlezoup: Fixed in 98b6a4cdb — `remapped_ram_address` now routes through the `remapped_ram_address()` helper (with its `>= ram_k` bound), mirroring stage2. One deviation from the suggestion: I kept the `let ram_address = …` binding, since `ram_access_nonzero` still reads it, and applied the helper to the remap step only rather than fully inlining `ram_access_address(…)`.

PR #1605, opened 2026-06-09
feat: add Jolt prover and CPU backend

markosg04:

Base: `prover-stack/09-jolt-witness`
Head: `prover-stack/10-jolt-prover`

Adds the Jolt prover crate and bundled CPU backend implementation.

Validation before submission:
- `cargo fmt -q -- --check`


0xAndoroid:
**Stack maintenance** — Pika is rebasing this PR onto #1604 (`prover-stack/09-jolt-witness`) and fixing CI as part of the prover stack (#1596–#1606). Rebase + CI summary to follow.

0xAndoroid:
**Stack maintenance** — rebased onto #1604 (`prover-stack/09-jolt-witness`); CI green. Ready for review.

📚 **Stack A:** #1596 → #1597 → #1598 → #1599 → #1600 → #1601 → #1602 → #1603
📚 **Stack B:** #1596 → #1597 → #1604 → **#1605 (this PR)** → #1606

**Conflict resolution**
- `Cargo.lock` — reconciled the new prover/backend dependencies with #1604's lockfile.
- `crates/jolt-blindfold/src/protocol.rs` — kept the #1604-compatible `crate::r1cs::build_with_sources(...)` call and removed the now-unused import.
- `tracer/Cargo.toml` — removed the duplicate `thiserror.workspace = true`.
- `tracer/src/emulator/cpu.rs` — kept the lint-clean guarded interrupt-mask match while preserving the existing privilege/interrupt gating behavior.
- `crates/jolt-claims/src/protocols/jolt/formulas/*`, `crates/jolt-poly/src/*`, `crates/jolt-transcript/*`, `crates/jolt-verifier/**` — skipped the duplicate already-applied #1604 foundation commits instead of reapplying conflicting changes.

**Local validation**
- `cargo fmt -q`
- `cargo build -q`

crates/jolt-prover/src/stages/stage6/io.rs
```diff
@@ -0,0 +1,241 @@
...
+        Self {
+            input_claims,
+            challenges,
+        }
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct Stage6ProofComponent<F: Field, Proof> {
+    pub stage6_sumcheck_proof: Proof,
```
moodlezoup: The prover emits a single `stage6_sumcheck_proof`, but the in-tree verifier expects a two-phase split: `jolt-verifier/src/proof.rs:221-222` declares both `stage6a_sumcheck_proof` and `stage6b_sumcheck_proof`, and `jolt-verifier/src/stages/stage6/mod.rs` exposes `verify_a`/`verify_b` with a `gamma`/`eta` squeeze between the 6a and 6b sub-batches. A single batch with all gammas derived up front in `derive_regular_batch_prefix` will diverge from the verifier's Fiat-Shamir at the 6a/6b boundary. The proof shapes alone don't match.

---

crates/jolt-prover/src/stages/stage7/prove.rs
```diff
@@ -0,0 +1,809 @@
...
+
+    let verifier_output = stage7_clear_output(Stage7ClearOutputRequest {
+        hamming_dimensions: dimensions,
+        hamming_gamma: prefix.hamming_gamma,
+        public_challenges: hamming_point.clone(),
+        output_claims: &claims,
+        input_claims: &prefix.input_claims,
+        batching_coefficients: &batch.batching_coefficients,
+        sumcheck_challenges: batch.challenges.clone(),
+        sumcheck_final_claim: batch.output_claim,
```
moodlezoup: Stage 7 computes `expected_final_claim` but never compares it to the running `batch.output_claim` — both are just stored into the output struct. Stages 5 and 6 fail fast here (`if running_claim != expected_final_claim { return Err(...) }`, e.g. `stage5/prove.rs:876`, `stage6/prove.rs:531`). Combined with the per-round invariant at line 537 being gated behind `#[cfg(any(test, debug_assertions))]` (stages 5/6 run it unconditionally), Stage 7 has no prover-side integrity check at all in release builds, so a miscomputation only surfaces at the verifier.

---

crates/jolt-backends/src/cpu/sumcheck/kernels/regular_batch.rs
```diff
@@ -0,0 +1,912 @@
...
+    let has_field_registers = match rest.first() {
+        Some(instance) => {
+            let _field_registers_shape =
+                stage2_linear_reduction_shape(instance, "field-registers claim-reduction")?;
+            true
+        }
+        None => false,
+    };
+
+    if product.polynomials.len() != 3
```
moodlezoup, line 568: This bind fast-path activates on polynomial *counts* only (3/2/2), but the matching evaluate fast-path at lines 142-152 additionally requires every reduction instance's element length to equal the product's half. Since evaluate and bind dispatch independently per round, if a reduction instance ever has fewer rounds than the product, evaluate falls back to the per-instance generic path while bind still uniformly binds all instances here — binding the shorter instruction poly too early and diverging the two. Latent under the current stage-2 config (shared round counts), but the two fast-paths should share one activation predicate.

---

crates/jolt-prover/CLEANUP_SPEC.md
```diff
@@ -0,0 +1,550 @@
+# jolt-prover Readability Cleanup Spec
```
moodlezoup, line 1: This is an in-progress refactor journal (dated progress log, per-stage LOC quotas, "next pressure point", benchmark table), not crate documentation — it'll rot immediately and documents debt that's still present (the stage `prove.rs` files are still 1500-2400 LOC, well over the targets here). Keep it in the PR description or an issue rather than committing it into the crate.

---

crates/jolt-backends/src/lib.rs
```diff
@@ -0,0 +1,142 @@
...
+    STAGE5_REGISTERS_VAL_EVALUATION_OPENING_SLOT_START,
+    STAGE6_BOOLEANITY_BYTECODE_RA_OPENING_SLOT_START,
+    STAGE6_BOOLEANITY_INSTRUCTION_RA_OPENING_SLOT_START,
+    STAGE6_BOOLEANITY_RAM_RA_OPENING_SLOT_START, STAGE6_BYTECODE_RA_OPENING_SLOT_START,
+    STAGE6_INC_RAM_OPENING_SLOT, STAGE6_INC_RD_OPENING_SLOT,
+    STAGE6_INSTRUCTION_RA_VIRTUALIZATION_OPENING_SLOT_START,
+    STAGE6_RAM_HAMMING_BOOLEANITY_OPENING_SLOT, STAGE6_RAM_RA_VIRTUALIZATION_OPENING_SLOT_START,
+};
+#[cfg(feature = "prover-harness")]
+pub use timing::{reset_backend_timings, take_backend_timings, BackendTiming};
```
moodlezoup, line 120: `take_backend_timings`/`reset_backend_timings` (the only readers of the timing data) have zero callers anywhere, including benches and tests. `record_backend_timing` writes into a `static Mutex<Vec<_>>` that's never drained — dead code, and a process-global mutex is a test-isolation hazard. Remove the module or wire up a consumer.

---

moodlezoup (commented):
Reviewed the new `jolt-backends` + `jolt-prover` crates (~61k lines). This is prover-only code (no verifier in the diff), so cryptographic soundness can't regress here directly; I focused on compilation/wiring, prover↔verifier consistency, Fiat-Shamir ordering, and the CPU sumcheck kernels. Transcript ordering inside each stage's self-contained sumcheck core is correct and mirrors the verifier. The dominant issue is structural:

**These crates are not built or tested by CI, and don't compile against the in-tree verifier.**
- `jolt-prover` and `jolt-backends` are absent from the root `Cargo.toml` `members` list, and no workspace member depends on them. So `cargo build`/`cargo clippy --all` and the `-p jolt-core muldiv` validation in the PR description never compile this code — that's why CI is green despite the issues below. `cargo build -p jolt-prover` fails with "did not match any packages".
- The prover is out of sync with the in-tree `jolt-verifier` API on this branch. Examples: it imports `jolt_verifier::stages::stage6::{stage6_transcript_challenges, stage6_validate_dependencies, …}`, none of which `jolt-verifier/src/stages/stage6/mod.rs` exports (it exports only `deps`/`Deps`/outputs/`verify`); and it emits a single `stage6_sumcheck_proof` while `jolt-verifier/src/proof.rs:221-222` expects `stage6a_sumcheck_proof` + `stage6b_sumcheck_proof`. As written it cannot compile against this tree.

If this is intentionally staged (wiring lands in a later stack PR), please say so in the description — but as it stands none of these 61k lines are exercised by CI, and the "CI is green / muldiv passes" validation doesn't cover them. I'd wire the crates into the workspace and reconcile the verifier API first, then assert transcript equality via `muldiv` and `advice` e2e in both standard and zk modes (the stage-6 6a/6b boundary and stage-4 committed-program opening-claim absorb are the spots most likely to diverge but aren't covered by muldiv).

Inline comments cover: the stage-6 single-vs-two-phase proof mismatch, a missing stage-7 prover-side claim check (release builds have no integrity check at all), an evaluate/bind fast-path predicate asymmetry in the stage-2 regular batch, the committed `CLEANUP_SPEC.md` working tracker, and dead timing code behind a global mutex.

A few more quality items not worth individual threads, all in `jolt-backends`: `cpu/eq`, `cpu/lagrange`, `cpu/split_eq` are pure pass-through shims over `jolt-poly` (delete the indirection, ~30 call sites can use `jolt-poly` directly); `BackendKernelMetadata.optimization_ids` is set via ~40 builders but only read in test assertions (no production dispatch consumes it); `README.md:10-13` links to four spec/source files that don't exist; and the `SumcheckBackend`/`*Backend` trait hierarchy is ~30 methods defaulting to `UnsupportedTask` with a single production implementor (`CpuBackend`) — defensible if a second backend is on the roadmap, but worth a note in the PR if so.

---

PR #1584, opened 2026-06-01
feat: integrate bytecode and program image reductions

RadNi:
Generated stack PR from `amir/bytecode-commitment-merged` after PR03 merged into `a16z/main`.

Stack position: `04`
Base branch: `a16z/main`
Replaces: https://github.com/LayerZero-Research/jolt/pull/19

Owned paths:
```text
jolt-core/src/zkvm/claim_reductions/precommitted.rs
jolt-core/src/zkvm/claim_reductions/mod.rs
jolt-core/src/zkvm/bytecode/read_raf_checking.rs
jolt-core/src/zkvm/bytecode/mod.rs
jolt-core/src/zkvm/ram/mod.rs
jolt-core/src/zkvm/ram/val_check.rs
jolt-core/src/zkvm/claim_reductions/hamming_weight.rs
jolt-core/src/poly/opening_proof.rs
jolt-core/src/poly/rlc_polynomial.rs
jolt-core/src/utils/errors.rs
jolt-core/src/zkvm/config.rs
jolt-core/src/zkvm/mod.rs
jolt-core/src/zkvm/program.rs
jolt-core/src/zkvm/prover.rs
jolt-core/src/zkvm/transpilable_verifier.rs
jolt-core/src/zkvm/verifier.rs
jolt-core/src/zkvm/witness.rs
jolt-core/src/guest/prover.rs
jolt-core/src/guest/verifier.rs
jolt-core/benches/e2e_profiling.rs
jolt-sdk/macros/src/lib.rs
jolt-sdk/src/host_utils.rs
src/build_wasm.rs
transpiler/src/main.rs
crates/jolt-verifier/src/compat/codec.rs
```

This PR is expected to be updated manually when `amir/bytecode-commitment-merged` is resliced.

Made with [Cursor](https://cursor.com)

PR #1571, opened 2026-05-27
refactor: split stage 6 sumchecks

RadNi:
Generated stack PR from `amir/bytecode-commitment-merged`.

Spec PR: https://github.com/a16z/jolt/pull/1565

Stack position: `01`
Base branch: `a16z/main`
Head branch: `LayerZero-Research:amir/bytecode-stack/01-stage6-split`

Owned paths:
```text
jolt-core/src/subprotocols/booleanity.rs
jolt-core/src/subprotocols/mod.rs
jolt-core/src/poly/opening_proof.rs
jolt-core/src/zkvm/bytecode/read_raf_checking.rs
jolt-core/src/zkvm/claim_reductions/advice.rs
jolt-core/src/zkvm/proof_serialization.rs
jolt-core/src/zkvm/prover.rs
jolt-core/src/zkvm/verifier.rs
jolt-core/src/zkvm/transpilable_verifier.rs
jolt-core/src/zkvm/witness.rs
transpiler/src/symbolic_proof.rs
```

## Test plan
- Full stack branch CI was run before pushing the stack branches.
- PR01 branch was previously validated as part of the stack update flow.

Made with [Cursor](https://cursor.com)

PR #1610, opened 2026-06-10
feat: port precommitted polynomial reductions to modular verifier

0xAndoroid:
## Summary
- Ports jolt-core #1572 (precommitted polynomial reductions) into the modular verifier stack — step 02 of porting the bytecode-commitment series (#1571–#1584) into `crates/`.
- Advice claim reductions now run on a shared `PrecommittedClaimReduction` schedule: a reference domain spanning the main trace and all precommitted candidates, a Dory opening-round permutation projected per polynomial, full-phase round counts, and distinct cycle-only vs full `(1/2)^gap` skip scales.
- Stage 8 mirrors core's `compute_final_opening_point`: the unified opening point is assembled from the stage-6 inc-reduction cycle challenges and stage-7 hamming address challenges (trace-order dependent), with dominant-precommitted anchoring and a uniform per-claim Lagrange embedding scale replacing the previous special-cased scales.

## Changes
- `jolt-claims`: new `formulas/claim_reductions/precommitted.rs` (scheduling reference + per-poly round schedule; core's Dory globals are parameter-passed — `joint_col_vars` reduces to `ceil(rtv/2)` after the stage-6 re-embedding); `advice.rs` rewritten on top of it; `committed_openings.rs` adds `final_opening_point` and generalizes `advice_commitment_embedding_scale` → `commitment_embedding_scale`; new structured error variants.
- `jolt-verifier`: shared `stages::advice_layouts` helper (both advice kinds must share one scheduling reference), stage-6b/7 wiring, stage-8 clear+ZK rewrite (single point used for both PCS verify and transcript binding), blindfold layout plumbing, tampering-test helpers.
- Tests: jolt-claims layout tests re-pinned to the new schedule; new unit tests for skip scales, dominance permutation, and final-point assembly (passthrough, AddressMajor ordering, dominant-anchor agreement/mismatch).

Committed bytecode chunks and the program image (#1583/#1584) plug into this same scheduling as additional candidates in the follow-up port steps; this PR keeps the candidate list advice-only.

## Testing
Validated against jolt-core pinned at #1572 (`7f97cbad1`), where the regression is isolated (current main additionally diverges at stage 1 from #1583/#1584 — out of scope here, next port step):
- `standard_advice_consumer_core_proof_is_accepted`: failing (`StageClaimOutputMismatch { HammingWeightClaimReduction }`) → **passing**; muldiv stays green.
- Full `-p jolt-verifier --features core-fixtures` suite: **110/110** (incl. stage-4/6/7 advice tampering rejections).
- `--features core-fixtures,zk` suite: **71/71**.
- At HEAD: 170 jolt-claims/jolt-verifier unit tests pass; `cargo clippy -D warnings` clean for `host`, `host,zk`, `core-fixtures`, `core-fixtures,zk`.

Spec: `specs/1344-committed-bytecode-program-image.md`. Core counterpart: #1572.

### DIFF
--- a/crates/jolt-claims/src/protocols/jolt/formulas/claim_reductions/advice.rs
+++ b/crates/jolt-claims/src/protocols/jolt/formulas/claim_reductions/advice.rs
@@ -1,4 +1,4 @@
-use std::{cmp::min, ops::Range};
+//! Two-phase advice claim reduction (Stage 6 cycle -> Stage 7 address).
 
 use jolt_field::{Field, RingCore};
 use jolt_poly::EqPolynomial;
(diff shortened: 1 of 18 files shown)


crates/jolt-claims/src/protocols/jolt/formulas/claim_reductions/advice.rs
```diff
@@ -11,265 +11,181 @@ use super::super::super::{
...
-        };
-        point.reverse();
-        Ok(point)
+        self.precommitted
+            .address_phase_opening_point(cycle_var_challenges, challenges)
     }
 
+    /// `FinalScale` value when the reduction completes in the cycle phase
+    /// (i.e. no active address-phase rounds remain).
     pub fn cycle_phase_final_output_scale<F: Field>(
```
moodlezoup, line 101: In ZK mode this is called from blindfold's `add_advice_cycle_publics` (zk/blindfold/mod.rs:1116), which is gated only on advice presence — so advice with an address phase hits `CyclePhaseNotFinal` during BlindFold public-input assembly, even though the cycle-phase relation only references `FinalScale` when `!has_address_phase()`. The old code failed the same way (length mismatch), so this isn't a regression, but since this PR makes address-phase advice pass in clear mode, the ZK counterpart of `standard_advice_consumer_core_proof_is_accepted` would still fail. Worth gating the blindfold stage-6 call on `!has_address_phase()` (mirroring the stage-7 gating), here or in the next port step.

---

crates/jolt-claims/src/protocols/jolt/formulas/committed_openings.rs
```diff
@@ -72,14 +74,17 @@ pub fn final_opening_relation(polynomial: JoltCommittedPolynomial) -> JoltRelati
     }
 }
 
-pub fn advice_commitment_embedding_scale<F: Field>(
+/// Lagrange factor for embedding a smaller polynomial's opening into the
+/// top-left block of the unified final opening point: `1` on variables the
+/// embedded point binds, `1 - r` on the rest.
+pub fn commitment_embedding_scale<F: Field>(
```
moodlezoup, line 80: Core's `compute_lagrange_factor` carries a `#[cfg(test)]` assertion that every embedded coordinate is contained in the unified point; that check didn't make it into the port. Now that this scale applies to every polynomial in the final batch (not just advice), a future change that breaks the slice-identity invariants (e.g. a stage-6b instance growing beyond `log_T` rounds) would silently produce a wrong-but-nonzero scale instead of failing loudly. A containment check here — even debug-only — is cheap insurance.

---

crates/jolt-claims/src/protocols/jolt/formulas/committed_openings.rs
```diff
@@ -88,6 +93,94 @@ pub fn advice_commitment_embedding_scale<F: Field>(
         .product()
 }
 
+/// Inputs to [`final_opening_point`], gathered from earlier verification
+/// stages.
+pub struct FinalOpeningPointInputs<'a, F: Field> {
+    /// `log_T + log_k_chunk` for the native trace domain.
+    pub native_main_vars: usize,
```
moodlezoup: `native_main_vars` is always `log_t + log_k_chunk` (the only caller computes it exactly that way), so the struct admits inconsistent states where the dominance threshold and the address split disagree. Passing `log_t` and deriving the sum internally would close that off.

---

crates/jolt-claims/src/protocols/jolt/formulas/committed_openings.rs
```diff
@@ -88,6 +93,94 @@ pub fn advice_commitment_embedding_scale<F: Field>(
...
+                        return Err(JoltFormulaPointError::IncompatibleDominantAnchors {
+                            first,
+                            second: index,
+                        });
+                    }
+                }
+            }
+        }
+        let (_, dominant_point) =
+            dominant.ok_or(JoltFormulaPointError::OpeningPointLengthMismatch {
```
moodlezoup: This `ok_or` is unreachable — `max_anchor_len` is attained by at least one anchor — and `OpeningPointLengthMismatch { got: 0 }` mislabels the condition if it ever did fire. Taking the dominant from the same pass that computed the max would remove the fallible step.

---

crates/jolt-verifier/src/stages/mod.rs
```diff
@@ -9,3 +15,58 @@ pub mod stage6;
...
+}
+
+impl PrecommittedSchedule {
+    pub fn new(
+        trace_order: TracePolynomialOrder,
+        log_t: usize,
+        log_k_chunk: usize,
+        memory_layout: &MemoryLayout,
+        trusted_advice_present: bool,
+        untrusted_advice_present: bool,
```
moodlezoup: Two adjacent bools that callers fill from `.is_some()` — swapping them compiles and only misbehaves when exactly one advice kind is present (layouts built from the wrong `max_*_advice_size`). They're immediately converted to `Option&lt;usize&gt;` max-bytes anyway; taking `trusted_max_bytes: Option&lt;usize&gt;, untrusted_max_bytes: Option&lt;usize&gt;` directly would make call sites self-checking and drop the `MemoryLayout` coupling.

---

crates/jolt-claims/src/protocols/jolt/formulas/claim_reductions/precommitted.rs
```diff
@@ -0,0 +1,452 @@
...
+        let nu_full = total_full.saturating_sub(sigma_full);
+        assert_eq!(
+            dory_opening_round_permutation_be.len(),
+            total_full,
+            "reference dory round permutation length mismatch",
+        );
+        assert!(
+            poly_row_vars <= nu_full && poly_col_vars <= sigma_full,
+            "top-left projection requires poly dims <= full dims (poly row/col vars={poly_row_vars}/{poly_col_vars}, full row/col vars={nu_full}/{sigma_full})"
+        );
```
moodlezoup: These asserts run inside `validate_inputs` on every proof. They're unreachable today because advice shapes are always `balanced(v)` with `v &lt;= reference_total_vars`, but once committed bytecode chunks / the program image join the candidate list (possibly with non-balanced shapes), this becomes a panic reachable from proof-controlled config (`one_hot_config`, `trace_length`). Returning a `JoltFormulaPointError` now is cheaper than rediscovering this in the follow-up port.

---

crates/jolt-verifier/src/stages/stage7/verify.rs
```diff
@@ -439,6 +460,55 @@ where
 
     append_stage7_opening_claims(transcript, claims);
 
+    let mut precommitted_final_openings = Vec::new();
+    for (kind, layout, address_phase, cycle_phase) in [
```
moodlezoup, line 464: This block and its ZK twin at line 225 are near-identical copies of the trusted/untrusted tuple pattern, and this is the exact spot where the bytecode-chunk/program-image candidates will plug in. A single loop over the kinds shared by both modes (per-kind accessors on the stage-6/7 outputs plus a closure producing `AdviceFinalSource`) would leave one place to extend instead of four.

---

moodlezoup (commented):
Reviewed with focus on the soundness of the stage-8 rewrite. The port checks out against jolt-core #1572 in the load-bearing details: permutation construction and per-poly projection, skip-round scales, the unified final-opening-point assembly (including dominant anchoring), embedding-scale equivalence with the old positional code in both trace orders, clear/ZK batch order and gamma alignment, and the transcript-binding change — which actually fixes a latent AddressMajor Fiat-Shamir divergence from the core prover in the old modular stage 8.

The comments below are mostly hardening and extensibility ahead of the committed-bytecode follow-ups. The one behavioral item: the ZK BlindFold path still can't handle address-phase advice (stage-6 public-input assembly computes `FinalScale` eagerly), so the clear-mode fix this PR lands doesn't yet have a ZK counterpart. Also worth noting that the dominance path and AddressMajor ordering don't appear to be exercised by current fixtures.

---

crates/jolt-verifier/src/stages/stage7/verify.rs
0xAndoroid: Deferring this one: unifying the clear/ZK gather loops needs per-kind accessors across four output types, which pays off once the bytecode-chunk/program-image candidates land — planned for that port step. The other six comments are addressed in 61945c3d6.

0xAndoroid:
📚 **PR Stack**

1. ✅ #1610 — merged
2. #1616 — rebased onto main ✔
3. #1617 — rebased ✔

_Stack managed by Pika (auto-rebase on squash-merge)._

PR #1616, opened 2026-06-11
feat: port committed bytecode foundation to jolt-claims

0xAndoroid:
## Summary

- Ports jolt-core #1583 (committed bytecode foundation) into the modular verifier stack — formulas and geometry only, no stage wiring yet (that mirrors #1584 in the next step).
- Stacked on #1610 (precommitted polynomial reductions); base branch is `feat/bytecode-commit`.
- Committed vs full program mode is a runtime property of preprocessing: the new formulas are additive and mode-free, so the full-mode path is untouched.

## Changes

- `formulas/claim_reductions/bytecode.rs` (new): committed bytecode lane layout (`BYTECODE_LANE_LAYOUT`, `committed_lanes()`), chunking validation, `BytecodeClaimReductionLayout` on the shared precommitted schedule, `lane_weights()` folding the five read-RAF stage gamma vectors into per-lane weights, trace-order-aware `eq_combined` factorization, and the cycle/address-phase relation formulas (input `Σ ηˢ·BytecodeValStage(s)`, output intermediate claim or `Σ ChunkOutputWeight(i)·BytecodeChunk(i)`).
- `formulas/claim_reductions/program_image.rs` (new): `ProgramImageClaimReductionLayout`, carry-propagation DP `eval_shifted_eq_poly_at_opening_point` (evaluates the start-index-shifted eq slice without materializing it), and the formulas binding `ProgramImageInitContributionRw` to `ProgramImageInit` openings.
- `precommitted.rs`: shared `PrecommittedReductionDimensions` replaces the advice-specific dimensions struct (same bookkeeping for all precommitted reductions).
- `ids.rs`: `BytecodeClaimReductionChallenge::Eta`, `BytecodeClaimReductionPublic::ChunkOutputWeight(i)` (value `chunk_rbc_weight · eq_combined · skip_scale`, matching core's BlindFold challenge values), `ProgramImageClaimReductionPublic::FinalScale`.
- Existing deliberately-exhaustive challenge matches in formula tests extended with the new variant.

## Testing

- `cargo nextest run -p jolt-claims` — 144 passed (18 new), including two cross-invariants: lane weights applied to a row's sparse lane values reproduce `Σ ηˢ·StageValue(s)` from the fixture-validated `read_raf_public_values` (the spec invariant that the lane layout encodes exactly what bytecode read-RAF reads), and the shifted-eq DP matches naive eq-slice MLE evaluation across start indices.
[...]

PR #1617, opened 2026-06-11
feat: port committed program verification to modular verifier

0xAndoroid:
## Summary

- Ports jolt-core #1584 (committed program mode) into the modular verifier: wires the committed bytecode and program-image reductions from #1616 through the stage pipeline and adds the committed preprocessing mode end to end.
- Stacked on #1616 (committed bytecode foundation); base branch is `feat/bytecode-commit-2`.
- Program mode is a runtime property of the deserialized preprocessing (`ProgramPreprocessing::{Full, Committed}`), matching jolt-core; full-mode verification paths are unchanged.

## Changes

- `preprocessing.rs`: `ProgramPreprocessing` enum and `CommittedProgramPreprocessing` — trusted bytecode-chunk and program-image commitments plus `ProgramMetadata`, absorbed into the transcript alongside proof commitments.
- Stage 4: RAM init in committed mode uses inputs-only public segments with the program-image contribution bound as a staged opening instead of public image words.
- Stage 6: committed bytecode reduction cycle phase and program-image reduction cycle phase (`verify_b.rs`), with reduction weights derived from the read-RAF stage gammas.
- Stage 7: bytecode and program-image address phases, resolving final per-chunk openings from whichever phase completed the reduction; stage 8 verifies them against the trusted commitments via the canonical precommitted schedule.
- BlindFold (`stages/zk/blindfold/`): publics and constraints for the committed reductions in stages 4/6/7, so ZK mode covers both program modes.
- `jolt-claims`: `read_raf_cycle_phase_committed` — per-stage Val factors come from `BytecodeValStage(s)` openings with public `StageCycleEq(s)` coefficients (in full mode the factor folds into `StageValue(s)`); committed read-RAF public-value evaluation.
- `jolt-program`: `ProgramMetadata` (verifier-side shape for committed mode), `compute_min_ram_k` decoupled from `RAMPreprocessing`, `inputs_only` initial-RAM segments.
- `jolt-transcript`: `DigestTranscript` — generic digest-based Fiat-Shamir transcript byte-compatible with jolt-core's hash transcripts, used at the core-compat boundary.

## Testing

- `cargo nextest run -p jolt-claims -p jolt-transcript -p jolt-program -p jolt-blindfold -p jolt-verifier` — 359 passed.
- New core-fixture cases: committed muldiv standard + ZK acceptance, precompat case, BlindFold shape audit against the modular protocol; tamper manifest extended for the committed-mode proof fields.
- `cargo clippy --all-targets -- -D warnings` clean on the changed crates, including `jolt-verifier --features core-fixtures`.


0xAndoroid:
📚 Stack fully merged ✅
~~#1610~~ ✅ → ~~#1616~~ ✅ → ~~#1617 (this PR)~~ ✅

All three bytecode-commit PRs landed on main.

crates/jolt-verifier/tests/soundness/fixtures.rs
```diff
@@ -58,7 +58,8 @@ pub fn metadata(id: FixtureId) -> FixtureMetadata {
         | FixtureId::Sha2Small
         | FixtureId::MulDivZkSmall
         | FixtureId::ZkStage1Prefix
-        | FixtureId::AdviceConsumer => FixtureMetadata {
+        | FixtureId::AdviceConsumer
+        | FixtureId::CommittedMulDivSmall => FixtureMetadata {
```
moodlezoup, line 62: `CommittedMulDivSmall` is registered completeness-only, and every new committed-mode tamper target in `tamper_manifest.rs` is `TamperCoverage::IgnoredUntilFixture` — so none of the rejection paths this PR adds (val-stage presence cross-checks, chunk-count checks at stage6/stage7, the `UnexpectedOpeningClaim` guards, the stage-8 trusted-commitment mapping) are exercised by any negative test. That's the entire soundness surface of committed mode shipping with acceptance-only coverage; a regression in any of those checks would pass CI. Consider promoting the committed fixture to a soundness fixture and activating those tamper targets in this PR or an immediate follow-up.

---

crates/jolt-program/src/preprocess/program.rs
```diff
@@ -34,3 +34,19 @@ impl JoltProgramPreprocessing {
...
+    feature = "serialization",
+    derive(serde::Serialize, serde::Deserialize)
+)]
+pub struct ProgramMetadata {
+    pub entry_address: u64,
+    pub min_bytecode_address: u64,
+    pub entry_bytecode_index: usize,
+    pub program_image_len_words: usize,
+    pub bytecode_len: usize,
+}
```
moodlezoup, line 52: This is a field-for-field copy of `ProgramMetadata` in `jolt-core/src/zkvm/program.rs:353`, and jolt-core already depends on jolt-program (and re-exports its types, e.g. `BytecodePreprocessing`/`RAMPreprocessing`) — the struct could live here once and be consumed by core, which would also delete the manual field-by-field conversion in `tests/support/core_fixtures.rs`. Agreement between the two copies is load-bearing for committed mode (this metadata is the trust root the verifier never cross-checks against the commitments). If touching jolt-core is out of scope for this stack, at least add a must-stay-in-sync warning pointing at the core definition.

---

crates/jolt-verifier/src/stages/zk/blindfold/mod.rs
```diff
@@ -1100,6 +1159,168 @@ where
     Ok(())
 }
 
+fn bytecode_reduction_weights<PCS, VC, ZkProof>(
```
moodlezoup, line 1159: This duplicates `bytecode_reduction_weights` in `stage6/verify_b.rs:183` — same `split_address_point` + `lane_weights` sequence, same error attribution, differing only in where the inputs are plumbed from. Making the verify_b helper `pub(crate)` and calling it here with a `BytecodeReductionWeightInputs` built from `BlindFoldInputs` would keep the clear/ZK weight derivation in one place — this is exactly the claim/constraint divergence class CLAUDE.md warns about.

---

crates/jolt-claims/src/protocols/jolt/formulas/bytecode.rs
```diff
@@ -274,13 +316,81 @@ impl<F: Field> BytecodeReadRafPublicValues<F> {
                 .get(index)
                 .copied()
                 .unwrap_or_else(F::zero),
+            BytecodeReadRafPublic::StageCycleEq(_) => F::zero(),
```
moodlezoup: The cross-mode arms (`StageCycleEq(_)` here, `StageValue(_)` in the committed struct below) silently evaluate to zero. If a formula edit ever references the wrong-mode public, this fails far downstream as an opaque output-claim mismatch instead of a "missing public" error at the source — weakening the desync safety net the explicit `Err` paths in the `try_evaluate` closures provide. Worth surfacing the wrong-mode case as an error rather than zero.

---

crates/jolt-claims/src/protocols/jolt/formulas/bytecode.rs
```diff
@@ -274,13 +316,81 @@ impl<F: Field> BytecodeReadRafPublicValues<F> {
...
 }
 
+pub struct BytecodeReadRafCommittedEvaluationInputs<'a, F> {
+    pub r_address: &'a [F],
+    pub r_cycle: &'a [F],
+    pub stage_cycle_points: [&'a [F]; NUM_BYTECODE_VAL_STAGES],
+    pub entry_bytecode_index: usize,
+}
+
+pub fn read_raf_committed_public_values<F>(
```
moodlezoup, line 359: The `identity`/`spartan_outer_raf`/`spartan_shift_raf`/`entry_bits`/`entry` block in this function is verbatim identical to the tail of `read_raf_public_values` below (~line 455). A small shared helper over `(r_address, r_cycle, stage cycle eqs, entry_bytecode_index)` would keep the full and committed variants from drifting; both call sites are in this PR.

---

moodlezoup (commented):
Reviewed with four parallel deep-analysis passes (soundness/Fiat-Shamir, semantic consistency, bug analysis, code quality), validating the committed-mode binding chains against the jolt-core reference.

**No soundness or correctness bugs found.** The high-risk surfaces all check out: the `eta` squeeze happens only after the val-stage claims are bound (cleartext-appended in clear mode, commitment-absorbed in ZK); `bytecode_val_stages` is a single proof field shared by the read-RAF factors and the reduction input claim, so the two uses cannot diverge; all dimensions and weights (chunk counts, lane weights, `ChunkOutputWeight`, `FinalScale`) derive from trusted preprocessing and transcript challenges, never the proof; the count-only ZK output-claim checks at stages 4/6a/6b/7 exactly match the BlindFold-registered opening ids; transcript orderings (commitment absorption, opening-claim appends, eta position) byte-match jolt-core; and the `DigestTranscript` implementation is bit-compatible with core's Blake2b transcript. Mode cross-checks fail closed in both directions, and no panic/unbounded allocation is reachable from malformed proof bytes.

The five inline comments cover the main gap — committed-mode rejection paths have zero negative-test coverage (all tamper targets `IgnoredUntilFixture`) — plus duplication/hardening points: the `ProgramMetadata` core/modular twin, the `bytecode_reduction_weights` clear/BlindFold twin, silent-zero cross-mode publics, and a duplicated RAF/entry computation in jolt-claims.

---

crates/jolt-verifier/src/stages/zk/blindfold/mod.rs
0xAndoroid: Fixed in 349f6ff1b: blindfold `bytecode_reduction_weights` now adapts `BlindFoldInputs` into `BytecodeReductionWeightInputs` and calls `verify_b::bytecode_reduction_weights` (promoted to `pub(crate)`), so the clear/ZK weight derivation lives in one place.

crates/jolt-claims/src/protocols/jolt/formulas/bytecode.rs
0xAndoroid: Fixed in 349f6ff1b: both `value()` methods now return `Option<F>`; wrong-mode publics and out-of-range stage indices surface as `MissingStageClaimPublic` in the stage-6 `try_evaluate` closures instead of a silent zero.

crates/jolt-claims/src/protocols/jolt/formulas/bytecode.rs
0xAndoroid: Fixed in 349f6ff1b: extracted a shared `read_raf_raf_entry_publics` helper over `(r_address, r_cycle, outer/shift stage-cycle eqs, entry index)`, used by both the full and committed variants.

crates/jolt-program/src/preprocess/program.rs
0xAndoroid: Dropping this: jolt-core is being deleted, so neither unifying the struct into jolt-program (no core consumer left) nor a must-stay-in-sync warning (constraint expires with core) pays off. The core-fixtures tests catch any divergence while core still exists.

crates/jolt-verifier/tests/soundness/fixtures.rs
0xAndoroid: Acknowledged — real gap, but promoting the committed fixture into the soundness registry and activating the tamper targets is scoped work matching the established IgnoredUntilFixture lifecycle. Tracking as a follow-up rather than expanding this PR; leaving open.

Issue #16, opened 2023-07-11
Verifier benchmarks


sragss:
Add verifier time and size to benchmarks

sragss:
Outdated

PR #1621, opened 2026-06-17
fix: repair jolt CLI install resolution

0xAndoroid:
## Summary
- use released crates.io `allocative = =0.3.6`, which includes the `i128`/`u128` `Allocative` impl fix previously pinned via git
- drop the stale `[replace]` git pin to `facebookexperimental/allocative`
- update install docs to use stable cargo commands and remove the `+nightly` install instruction

## Root cause
`cargo install` ignores the repository `Cargo.lock` unless `--locked`, so the install graph was freshly resolved. Before this PR, Jolt directly used a git-pinned `allocative` 0.3.4 while transitive arkworks dependencies accepted crates.io `allocative` `^0.3.4` and resolved to a different published version. That created two distinct `Allocative` traits in the graph; arkworks field types implemented one, while `jolt-core` required the other, producing the `Fp<MontBackend<FrConfig, 4>, 4>: Allocative` failure.

The original git pin targeted `facebookexperimental/allocative` rev `85b773d85d526d068ce94724ff7a7b81203fc95e` (`Implement Allocative for i128 and u128 (#16)`). That fix is now published in crates.io `allocative` 0.3.6, so this PR removes the git pin and uses the released crate directly.

`facebookexperimental/allocative` is now archived; its README says maintenance has moved to `https://github.com/facebook/buck2`. Since the fix is released, no replacement git URL is needed.

## Verification
- `cargo install --path . --force` on stable toolchain 1.95: succeeds
- `cargo install --git https://github.com/a16z/jolt --branch fix-install-flow --force --bins jolt` on stable: succeeds
- `jolt --version`: `jolt 0.1.0 (9dadb855f 2026-06-17)`
- `cargo fmt -q --message-format=short`
- `cargo clippy --all --features host -q --all-targets --message-format=short -- -D warnings`
- `cargo clippy --all --features host,zk -q --all-targets --message-format=short -- -D warnings`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1407,7 +1407,7 @@ checksum = "b05b61dc5112cbb17e4b6cd61790d9845d13888356391624cbe7e41efeac1e75"
 name = "common"
 version = "0.2.0"
 dependencies = [
- "allocative 0.3.4 (git+https://github.com/facebookexperimental/allocative?rev=85b773d85d526d068ce94724ff7a7b81203fc95e)",
+ "allocative 0.3.4 (registry+https://github.com/rust-lang/crates.io-index)",
  "ark-serialize 0.5.0 (git+https://github.com/a16z/arkworks-algebra?branch=dev%2Ftwist-shout)",
  "serde",
  "syn 2.0.117",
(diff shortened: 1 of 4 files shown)


moodlezoup (commented):
Reviewed — no issues found. The fix is sound: pinning the workspace `allocative` to crates.io `=0.3.4` unifies with the arkworks fork's `^0.3.4` requirement to exactly `0.3.4`, and the existing `[replace] "allocative:0.3.4"` entry redirects that to the git rev, so the whole graph shares one `Allocative` trait. The lockfile correctly retains the git package as the replacement target, and dropping `+nightly` from the docs matches the `1.95` pin in `rust-toolchain.toml`. No Rust source changed, so no soundness/correctness/perf surface.

---

PR #1630, opened 2026-06-18
refactor: move Stage 6 relation state out of jolt-claims

0xAndoroid:
## Summary
- move the prover-owned advice cycle-phase relation state into private jolt-prover Stage 6 code
- remove public prover execution state structs from jolt-claims
- drop tests that only covered those concrete state structs
- remove unused manifest deps caught by cargo-machete on the PR1605 tree

## Verification
- cargo check -p jolt-claims -q --message-format=short
- cargo check -p jolt-prover -q --message-format=short
- cargo clippy -p jolt-claims -p jolt-prover -q --message-format=short --all-targets -- -D warnings
- cargo nextest run -p jolt-claims -p jolt-prover --cargo-quiet
- cargo check -p jolt-prover -q --message-format=short --features zk
- cargo clippy -p jolt-claims -p jolt-prover -q --message-format=short --all-targets --features zk -- -D warnings
- cargo machete --with-metadata
- taplo fmt --check


moodlezoup (commented):
Clean refactor — no issues found. Verified the 6 deleted relation-state structs (booleanity, bytecode read-raf, inc claim reduction, instruction ra-virtualization, ram ra-virtualization, ram hamming booleanity) have zero usages on the head branch, so they were genuinely dead code only exercised by their own tests. `AdviceCyclePhaseRelationState` is the one type with a real call site; its move into the private `stages/stage6/relation_state.rs` is behavior-identical (only the visibility modifiers and a doc comment changed), and the single use in `stage6/batch.rs` was updated correctly. The removed dev-deps (`jolt-core` from jolt-prover, `criterion` from jolt-backends) are unreferenced — the remaining "jolt-core" mentions are only bench string labels and docs. Net effect is reduced public surface in jolt-claims with no protocol change.

---

PR #1634, opened 2026-06-23
Reorganize opcode 0x5B to allow for more custom instructions

mathmasterzach:
## Changes

- Modified `0x5B` opcode so that `funct3=0b000` is further split into up to 128 instructions based on `funct7`
- Moved advice instructions and `rev8w` to use `funct3=0b000`
- Modified `FormatAdviceLoadI` to remove unused `imm`
- Added `FormatT` instruction type for use in custom instructions that don't use `imm` or `rs2`
- Migrated `rev8w` to be `FormatT`

### DIFF
--- a/crates/jolt-program/src/image/decode.rs
+++ b/crates/jolt-program/src/image/decode.rs
@@ -180,20 +180,22 @@ fn decode_system(word: u32) -> Result<SourceInstructionKind, ProgramError> {
 }
 
 fn decode_custom(word: u32) -> Result<SourceInstructionKind, ProgramError> {
-    match funct3(word) {
-        0b000 => Ok(SourceInstructionKind::VirtualRev8W(
+    let funct3 = funct3(word);
+    let funct7 = funct7(word);
+    match (funct3, funct7) {
+        (0b000, 0x00) => Ok(SourceInstructionKind::AdviceLB),
+        (0b000, 0x01) => Ok(SourceInstructionKind::AdviceLH),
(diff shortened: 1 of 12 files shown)


crates/jolt-program/src/image/decode.rs
```diff
@@ -180,21 +180,23 @@ fn decode_system(word: u32) -> Result<SourceInstructionKind, ProgramError> {
 }
 
 fn decode_custom(word: u32) -> Result<SourceInstructionKind, ProgramError> {
-    match funct3(word) {
-        0b000 => Ok(SourceInstructionKind::VirtualRev8W(
+    let funct3 = funct3(word);
+    let funct7 = funct7(word);
+    match (funct3, funct7) {
+        (0b000, 0x00) => Ok(SourceInstructionKind::AdviceLB),
```
0xAndoroid, line 186: These new advice-load arms still use `format_advice_load_operands`, which leaves `imm` set to the encoded instruction word. `FormatAdviceLoadI` now normalizes advice loads as `rd`-only with `imm: 0`, and bytecode commitments include `operands.imm`, so decoded program images diverge from tracer source rows for the new encoding.

0xAndoroid (commented):
One validated issue from review.

crates/jolt-program/src/image/decode.rs
mathmasterzach: Good catch. Pushed a fix that modifies `format_advice_load_operands` to set `imm` to 0.

PR #1635, opened 2026-06-23
feat: modular verifier refactor + supporting crate updates

moodlezoup:
Part of splitting #1605 into a reviewable, individually-CI-passing stack (1/3).

**Base:** `prover-stack/09-jolt-witness`

Everything in #1605 **except** the two new crates (jolt-backends, jolt-prover). The bulk is the jolt-verifier refactor plus the supporting workspace changes it depends on.

### jolt-verifier
- Port stages 1–7 to self-contained `SumcheckInstance` relations; route per-stage `verify()` through shared helpers
- Collapse/unify output structs; fold `inputs.rs` into `outputs.rs`; delete dead per-stage code
- Drop the `field-inline` feature; depend on the new `jolt-verifier-derive`

### Supporting crates
- **jolt-verifier-derive** (new): proc-macro for opening-claim plumbing
- **jolt-claims / jolt-witness / jolt-lookup-tables / jolt-field / jolt-poly / jolt-dory / jolt-blindfold / jolt-openings**: shared relation helpers and the Challenge→Public id rename consumed by the refactored verifier
- **jolt-core**: stop seeding dory-pcs's global prepared-point cache
- Workspace + CI: register `jolt-verifier-derive`; drop `-p jolt-verifier` from the field-inline CI jobs

### Validation
- `cargo clippy --all --all-targets` in `allocative,host` / `allocative,host,zk` / `--no-default-features`, plus `-p jolt-verifier --features core-fixtures{,zk}` — all clean
- `cargo fmt --all --check`
- `cargo nextest` across jolt-verifier + supporting crates (~1,650 tests) and the field-inline group


crates/jolt-poly/src/lib.rs
```diff
@@ -70,6 +72,7 @@ pub use mle::{
     block_selector_mle_msb, range_mask_mle_msb, sparse_mle_msb, sparse_segments_mle_msb, MleError,
 };
 pub use multilinear::{MultilinearBinding, MultilinearEvaluation, MultilinearPoly, RlcSource};
-pub use one_hot::OneHotPolynomial;
+pub use one_hot::{OneHotIndexOrder, OneHotPolynomial};
 pub use point::{Endianness, Point, HIGH_TO_LOW, LOW_TO_HIGH};
+pub use split_eq::{GruenSplitEqPolynomial, TensorEqTable};
```
moodlezoup, line 77: `GruenSplitEqPolynomial`/`TensorEqTable` have no consumer anywhere in this PR's crate stack — the only references are this re-export and `split_eq.rs`'s own tests. The 637-line `split_eq.rs` is a re-port of the still-present `jolt-core/src/poly/split_eq_poly.rs`. If the consumer lands in a later stack PR (jolt-backends/jolt-prover), consider moving the file there; otherwise it's dead duplication in this PR.

---

crates/jolt-field/src/lib.rs
```diff
@@ -79,4 +88,10 @@ pub mod arkworks;
 #[cfg(feature = "bn254")]
 pub use arkworks::bn254::Fr;
 #[cfg(feature = "bn254")]
+pub use arkworks::bn254_fq::Fq;
```
moodlezoup, line 91: This `Fq` newtype (479-line `arkworks/bn254_fq.rs`) has no consumer in this PR — the would-be user, `jolt-crypto/src/ec/bn254/glv/decomp_2d.rs`, imports `ark_bn254::Fq` directly and isn't touched here. Drop it from this PR, or land it together with the consumer that needs it.

---

moodlezoup (commented):
Reviewed the verifier refactor in depth across all 8 stages (old-vs-new Fiat-Shamir ordering, cross-input consistency checks, the boolean-hypercube guards, and the `jolt-verifier-derive` macro codegen). No soundness regression found: the opening-claim transcript order is preserved, several cross-relation checks were *relocated* rather than dropped (e.g. the rs1/rs2 and lookup-output alias checks into `stage2`/`stage3` `validate()`), and a couple of paths were actually *tightened*. The `field-inline` verifier removal matches the PR's stated scope. The derive macro's internal consistency (`opening_values`/`opening_count`/`append_openings`/`resolve_*` all walking the same field order) holds up.

Two non-blocking items inline (dead/duplicated additions with no consumer in this PR), plus a few minor notes below.

**Dropped defensive count checks on RA opening-claim vectors (robustness + test gap).** The old monolith hard-rejected a wire/layout length mismatch on the one-hot RA claim vectors — old stage6 `verify.rs` (`booleanity.{instruction,bytecode,ram}_ra.len() != ra_layout.*()`) and old stage7 `ensure_hamming_output_claim_counts`. The new per-relation code drops these for booleanity, the two virtualization families, and hamming (only `bytecode_read_raf.bytecode_ra` and the committed chunks keep an explicit check). This is **not** exploitable: `zip_openings` zips wire-against-derived-points and stops at the shorter, so an under-count surfaces as `MissingOpeningClaim` and an over-count's extras never reach `expected_output` or the stage-8 PCS batch (they only get absorbed into the transcript, self-consistently). But it's a strictness regression, and it's currently untested — the tamper manifest only applies `OffsetScalar` to these claims; `TruncateVector`/`ExtendVector` are exercised only on `round_polynomials`. Worth either restoring the explicit count checks (or making `zip_openings` error on unequal `Vec` lengths) and adding length-mutation tamper coverage.

Minor:
- `crates/jolt-field/src/{signed_product_accumulator,small_scalar_accumulator}.rs` re-implement the deferred-Montgomery folded-accumulator technique already in `jolt-core/src/field/folded_accum.rs`; both copies now coexist. Lower confidence since they're consumed via the `Field` trait bound rather than orphaned, but flagging the duplication.
- The macro's `append_openings` (its single-sourced FS order) is only invoked in a unit test; the live verifier still appends via hand-written aggregate `append_to_transcript`/`append_opening_claims`. They currently match the derived per-field order, but the single-source guarantee isn't actually wired into the transcript path, leaving latent drift risk.
- `ProductRemainder::resolve_public` (`stage2/product_remainder.rs`) handles `UniskipLagrangeWeight`, which that relation's formula never references — dead arm.

---

crates/jolt-verifier/src/stages/stage2/outputs.rs
```diff
@@ -1,33 +1,268 @@
...
+        }
+    }
+}
+
+/// Combines the stage 1 product values against the uni-skip Lagrange `weights`
+/// (derived from `tau_high`) into the stage 2 product uni-skip input claim.
+pub fn product_uniskip_input_claim<F: Field>(
+    values: Stage2ProductUniSkipInputValues<F>,
+    weights: &[F],
+) -> Result<F, VerifierError> {
```
0xAndoroid, line 46: Should functions like these be methods on the `Stage2ProductUniSkipInputValues`?

crates/jolt-verifier-derive/src/lib.rs
markosg04, line 1: This is a nice abstraction!

markosg04 (approved):

PR #1638, opened 2026-06-24
fix(transcript): reject u64 lengths non-representable as usize

bravesasha:
BytesMsg::deserialize_from_narg used a lossy as usize conversion for the encoded length field, which could lead to architecture-dependent parsing of malformed inputs on 32-bit platforms. This replaces the cast with usize::try_from(...) so non-representable lengths are rejected consistently, and adds a regression test covering both 32-bit and 64-bit behavior.


moodlezoup (approved):
good catch, ty!

PR #1640, opened 2026-06-25
refactor(claims,verifier): split JoltRelationClaims into SymbolicSumcheck + ConcreteSumcheck

moodlezoup:
## Summary

Splits the monolithic `JoltRelationClaims` / `ClaimExpression` claims representation into a clean two-layer design and deletes the old layer entirely:

- **`SymbolicSumcheck`** (jolt-claims) — the pure, field-independent *algebra* of each sumcheck relation: its relation id, sumcheck spec, and input/output expressions over the protocol's opening / public / challenge id families. One struct per relation under `protocols/{jolt,field_inline}/relations/**`.
- **`ConcreteSumcheck`** (jolt-verifier, formerly `SumcheckInstance`) — the *wiring / resolution* layer: maps each relation's symbolic expressions onto located opening claims (point + value cells), resolves publics/challenges, and derives opening points.

The old representation — `JoltRelationClaims`, `JoltProtocolClaims`, `ClaimExpression`, `ConsistencyClaim`, `SameEvaluation`, the `field_inline` equivalents, and all the `formulas/*` builder functions — is removed (`grep` for any of them is now empty). The residual field-independent substrate (dimensions, layouts, the opening-id catalog, commitment helpers) was renamed `formulas/` → `geometry/`, and dead code was swept.

Behavior-preserving: proof format, Fiat-Shamir transcript, opening order, and the BlindFold R1CS are unchanged. Net ~1k lines removed (134 files, +10.3k / −11.3k).

## Changes

- Add the `SymbolicSumcheck` trait (`id` / `spec` / `input_expression` / `output_expression`) and per-relation symbolic types for both the `jolt` and `field_inline` protocols.
- Rename `SumcheckInstance` → `ConcreteSumcheck`; source its id / spec / claim-expressions from `symbolic()` instead of a stored `JoltRelationClaims`.
- Remodel RAM `val_check` init as `Public` symbols (`InitEval` / `InitSelector`) — R1CS-identical, prover untouched.
- Switch BlindFold (ZK) to consume the symbolic relation objects directly (no `JoltRelationClaims`).
- Enforce cross-relation consistency via stage `validate()` methods (replacing the never-read `with_consistency` field).
- Delete `JoltRelationClaims` / `JoltProtocolClaims` / `ClaimExpression` / `ConsistencyClaim` / `SameEvaluation` and all `formulas/*` bridge builders; port their algebra tests to `relations/*`.
- Port `field_inline` to the symbolic types; delete its parallel claims types.
- DSL ergonomics: `opening` / `challenge` / `public` take `impl Into<…>`; inline the `*_sumcheck` spec wrappers; drop the single-source `*_challenge` / `*_public` fragment wrappers.
- Rename `formulas/` → `geometry/` (the directory no longer holds formula builders — it's dimensions / specs / layouts / opening-ids).
- Delete dead code: orphaned helpers plus a `pub`-deadness sweep run to fixpoint.

## Testing

- [x] Ran tests for modified crates — `cargo nextest run -p jolt-claims` green; the `jolt-verifier` `prover-fixtures` completeness **and** soundness/tampering suites pass in both standard and ZK modes.
- [x] `cargo clippy` and `cargo fmt` pass — `cargo clippy --all --features host` and `--features host,zk` clean under `-D warnings`; `cargo fmt --check` clean.
- Primary correctness check: the `muldiv` e2e passes with `--features host` and `--features host,zk`.

## Security Considerations

This is a representation refactor with **no intended change to proof soundness or verifier semantics**:

- The symbolic expressions and sumcheck specs are equal to those of the deleted bridge builders (the bridges delegated to the symbolic types prior to deletion; the per-relation algebra tests, ported alongside, assert the exact input/output polynomials).
- Opening order, the Fiat-Shamir transcript, the claim-binding constraints, and the BlindFold verifier R1CS are unchanged → proof bytes are unchanged.
- The RAM `val_check` init Public-symbol remodel is R1CS-identical (the baked `Public` coefficients equal the former inline `Term` constants); the prover computes the same input claim numerically.
- Cross-relation consistency previously *declared* via `with_consistency` is now *enforced* by live `validate()` checks on the clear verify path (the field it populated was never read by the verifier; unit tests added for the mismatch→error cases).
- Covered by the soundness/tampering fixture suites (`tampered_*`) and `muldiv` in both clear and ZK modes.

## Breaking Changes

No proof-format or serialization breaking changes — proof bytes are unchanged. Internally, the `jolt-claims` / `jolt-verifier` APIs are restructured: `JoltRelationClaims`, `JoltProtocolClaims`, `ClaimExpression`, `ConsistencyClaim`, the `formulas/*` builders, and `SumcheckInstance` are removed or renamed; consumers now use `relations::*` (`SymbolicSumcheck`) + `ConcreteSumcheck` and the `geometry/` module path.

## Follow-ups (not in this PR)

- Fold the opening-id catalog into `relations/` via a single declarative per-relation schema that generates both the prover's ordered opening list and the verifier's claim structs — superseding the hand-written `geometry/` opening-id helpers (and the residual `geometry/` ↔ `relations/` split).
- jolt-eval invariants: symbolic ↔ concrete id-set agreement, and the `val_check` Public-symbol input-claim equivalence.


crates/jolt-claims/src/protocols/field_inline/geometry/claim_reductions/increments.rs
```diff
@@ -0,0 +1,32 @@
+use super::super::super::{
+    FieldInlineCommittedPolynomial, FieldInlineOpeningId, FieldInlineRelationId,
+};
+
+pub fn claim_reduction_input_openings() -> [FieldInlineOpeningId; 2] {
```
0xAndoroid, line 5: Can we model these fixed opening IDs as a relation-local namespace with associated consts instead of free functions? Fixed IDs/sets read better as canonical data, while indexed or dimension-dependent openings can stay as associated fns/const fns on the same namespace.

For example: `FieldRegistersIncClaimReductionOpenings::{READ_WRITE, VAL_EVALUATION, REDUCED, INPUTS, OUTPUTS}`.

Also prefer `crate::protocols::field_inline::{...}` over `super::super::super::{...}` here. I don’t think Clippy has a lint for this exact pattern; if we want enforcement, this is probably a simple grep/CI check.

moodlezoup: i think these are ~dead code, so I'll just delete them in the next PR

crates/jolt-claims/src/protocols/field_inline/relations/claim_reductions/increments.rs
```diff
@@ -0,0 +1,155 @@
...
+    fn new(shape: FieldRegistersTraceDimensions) -> Self {
+        Self { shape }
+    }
+
+    fn id() -> FieldInlineRelationId {
+        FieldInlineRelationId::FieldRegistersIncClaimReduction
+    }
+
+    fn spec(&self) -> FieldInlineSumcheckSpec {
+        self.shape.sumcheck(2)
```
0xAndoroid: Move this degree into a named constant, e.g. `FIELD_RD_INC_CLAIM_REDUCTION_DEGREE`. The bare `2` is the relation degree, not a shape parameter, and naming it makes that clear.

crates/jolt-claims/src/protocols/field_inline/relations/claim_reductions/increments.rs
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    fn input_expression<F: RingCore>(&self) -> FieldInlineExpr<F> {
+        let eta = challenge(FieldRegistersIncClaimReductionChallenge::Gamma);
```
0xAndoroid, line 49: Why is this local named `eta` when the challenge id is still `FieldRegistersIncClaimReductionChallenge::Gamma`? Unless we are intentionally renaming the challenge to `Eta`, keep the local variable and docs as `gamma` for consistency with the sibling Jolt increment reduction.

crates/jolt-claims/src/protocols/jolt/geometry/booleanity.rs
```diff
@@ -0,0 +1,69 @@
...
+        .enumerate()
+    {
+        let ra = opening(opening_id);
+        output = output + gamma.clone().pow(2 * i) * (ra.clone() * ra.clone() - ra);
+    }
+
+    eq_address_cycle * output
+}
+
+pub fn booleanity_output_openings(layout: JoltRaPolynomialLayout) -> Vec<JoltOpeningId> {
```
0xAndoroid, line 51: Can this be a method on `BooleanityDimensions` instead of a free function? It is derived entirely from the dimensions layout, so `dimensions.output_openings()` keeps the geometry API attached to the shape that owns it.

crates/jolt-claims/src/protocols/jolt/geometry/bytecode.rs
```diff
@@ -0,0 +1,768 @@
...
+}
+
+fn register_eq<F: Field>(register: Option<u8>, eq: &[F]) -> F {
+    register
+        .and_then(|register| eq.get(register as usize))
+        .copied()
+        .unwrap_or_else(F::zero)
+}
+
+pub fn read_raf_output_openings(
```
0xAndoroid, line 437: Same shape as the Booleanity note: can this be a method on `BytecodeReadRafDimensions`? The output opening list is derived from `dimensions.num_committed_ra_polys()`, so `dimensions.output_openings()` keeps this geometry attached to the shape that owns it. Fixed relation metadata like `read_raf_consistency_openings()` can stay separate or move to a relation/openings namespace if we do that broader cleanup.

crates/jolt-claims/src/protocols/jolt/geometry/bytecode.rs
```diff
@@ -0,0 +1,768 @@
...
+                inputs.stage2_gammas,
+                inputs.stage3_gammas,
+                inputs.stage4_gammas,
+                inputs.stage5_gammas,
+            )
+        })
+        .collect()
+}
+
+pub fn read_raf_public_values<F>(
```
0xAndoroid, line 285: I think this is a boundary leak. `jolt-claims` should describe the symbolic relation: ids, dimensions, opening order, and expressions. These helpers are doing concrete verifier evaluation against bytecode rows, Eq tables, register points, and sampled gamma values. I would move `BytecodeReadRafEvaluationInputs`, `read_raf_public_values`, `read_raf_stage_values`, and `read_raf_row_values` to the verifier/concrete side, or at least an explicitly named evaluation module. Keeping them here makes `jolt-claims` more than symbolic protocol metadata.

crates/jolt-claims/src/protocols/jolt/geometry/bytecode.rs
```diff
@@ -0,0 +1,768 @@
...
+        unexpanded_pc_spartan_shift(),
+        unexpanded_pc_instruction_input(),
+    )]
+}
+
+pub(crate) fn stage1_claim<F>() -> JoltExpr<F>
+where
+    F: RingCore,
+{
+    let beta = challenge(BytecodeReadRafChallenge::Stage1Gamma);
```
0xAndoroid, line 458: This should be named `gamma` (or `stage_gamma`), not `beta`. The challenge id is `BytecodeReadRafChallenge::Stage1Gamma`, and using `beta` here makes it look like a different Fiat-Shamir challenge from the one actually resolved.

crates/jolt-claims/src/protocols/jolt/geometry/instruction.rs
```diff
@@ -0,0 +1,530 @@
...
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct InstructionReadRafOpeningPoint<F: Field> {
+    pub r_address: Vec<F>,
+    pub r_cycle: Vec<F>,
+    pub opening_point: Vec<F>,
+}
+
+impl TryFrom<(usize, usize, usize)> for InstructionReadRafDimensions {
```
0xAndoroid, line 86: Agree, `TryFrom<(usize, usize, usize)>` is a bad API here. These dimensions are positional and easy to swap. I would make this a checked named constructor instead, e.g. `InstructionReadRafDimensions::try_new(log_t, instruction_address_bits, num_virtual_ra_polys)`, and do the same for `InstructionRaVirtualizationDimensions`.

crates/jolt-claims/src/protocols/jolt/relations/claim_reductions/bytecode/cycle_phase.rs
```diff
@@ -0,0 +1,254 @@
...
+    fn id() -> JoltRelationId {
+        JoltRelationId::BytecodeClaimReductionCyclePhase
+    }
+
+    fn spec(&self) -> JoltSumcheckSpec {
+        self.shape.0.cycle_sumcheck()
+    }
+
+    fn input_expression<F: RingCore>(&self) -> JoltExpr<F> {
+        let eta = challenge(BytecodeClaimReductionChallenge::Eta);
```
0xAndoroid, line 84: Let’s rename this challenge to `Gamma` instead of `Eta`. This is the relation-local batching challenge, and the rest of the symbolic API mostly uses `Gamma` for that role. Keeping `Eta` makes this relation look special and has already made the naming confusing. I would rename the enum variant, locals, tests, and docs together.

crates/jolt-claims/src/protocols/jolt/relations/booleanity/monolith.rs
```diff
@@ -0,0 +1,248 @@
...
+/// across the RA family by `gamma` and weighted by the `EqAddressCycle` public.
+/// Its input claim is zero — the boolean constraint sums to zero rather than
+/// reducing a prior opening.
+pub struct Booleanity {
+    shape: BooleanityDimensions,
+}
+
+impl SymbolicSumcheck for Booleanity {
+    type RelationId = JoltRelationId;
+    type OpeningId = crate::protocols::jolt::JoltOpeningId;
```
0xAndoroid, line 55: Nit: import `JoltOpeningId`, `JoltPublicId`, and `JoltChallengeId` next to `JoltExpr`/`JoltRelationId` instead of using the full `crate::protocols::jolt::...` paths in these associated types.

0xAndoroid:
Suggested `CLAUDE.md` addition from this review session:

```diff
diff --git a/CLAUDE.md b/CLAUDE.md
@@
 ### Code Style
 
 - `cargo fmt` + `cargo clippy` with zero warnings
 - Codebase uses `non_snake_case` convention for math variables: `log_T`, `ram_K`, `log_K`, etc.
+- Keep symbolic and concrete protocol layers separate: symbolic code defines ids, dimensions, opening sets, and expressions; verifier/prover code resolves values and performs concrete evaluation.
+- Use `Gamma` as the only name for relation-local folding/batching challenges; keep enum variants, locals, docs, and tests aligned.
+- Prefer named constructors over positional tuple APIs for protocol dimensions and shapes.
+- Prefer relation/dimension methods or associated constants over unrelated freestanding helpers.
+- Avoid deep relative paths like `super::super::super`; use imports or `crate::...` paths.
+- Import repeated associated types instead of spelling long full paths inside impls.
+- Use named constants for protocol degrees, counts, and fixed arities instead of inline literals.
```

0xAndoroid:
Ok, these comment is pretty much everything I have, most of these were uncaught in previous reviews. 
The symbolic sumcheck abstraciton is clean and nice.

0xAndoroid (approved):

PR #1654, opened 2026-06-30
feat: packed opening api and batch opening trait

markosg04:
## Summary

- Introduces a batching abstraction for same-point multi-polynomial openings without requiring additive homomorphism.
- Introduces the prefix packing API for packed commitment strategies.

The diff looks big because of the expanded test coverage; excluding `tests/`, the current diff is 24 files changed, 1270 insertions(+), 1267 deletions(-) (net +3).

This is the first of 5 PRs for akita integration. A later PR will integrate this API into `jolt-verifier`

## Tests

- `cargo nextest run -p jolt-openings --cargo-quiet`


crates/jolt-openings/src/packing.rs
```diff
@@ -0,0 +1,507 @@
...
+    pub(crate) commitment: &'a C,
+    constrained_point: Vec<Option<F>>,
+    ordered_claims: Vec<(&'a PrefixPackedClaim<F, Id>, &'a PrefixSlot)>,
+}
+
+impl<F, Id, C> PreparedPrefixPackedStatement<'_, F, Id, C>
+where
+    F: Field,
+{
+    pub(crate) fn opening_point<T>(&self, transcript: &mut T) -> Result<Vec<F>, OpeningsError>
```
moodlezoup: This reduction looks unsound for mixed-arity packings. Only the *unpinned* coordinates get Fiat-Shamir challenges here — and the unpinned set is `[0, min_k)`, where `min_k` is the prefix length of the **largest**-arity slot. Every coordinate in `[min_k, packed_num_vars)` is pinned to a (prover-supplied) claim point by `constrained_point`. A smaller-arity slot has prefix length `k_i > min_k`, so its distinguishing prefix bits `[min_k, k_i)` live in the pinned region and never get randomized.

Concrete break with ids `0 (n=3), 1 (n=1), 2 (n=0)`: `packed_num_vars = 4`, `min_k = 1`, so only coord 0 is random. Slots 1 (prefix `100`) and 2 (prefix `1010`) share that one random bit. In `reduced_eval` their `eq` selectors differ only on pinned coords, so a malicious prover can set `v1 = t1 + δ` and `v2 = t2 − δ·(1−p2)/(p2·(1−p3))` — both false — and the error `Σ eq·(v_i − t_i)` vanishes *identically* in the single random coordinate. Then `W(r_pack) == reduced_eval` for every challenge and the verifier accepts both false logical claims.

The construction is only sound when every slot's prefix lies entirely in the randomized region, i.e. uniform arity (all `k_i == min_k`). Either restrict `PrefixPacking` to equal-arity polynomials, or give the colliding logical claims independent randomness (e.g. fold them with distinct RLC challenge powers) instead of relying solely on the `eq` selectors over a point whose suffix the prover controls.

---

moodlezoup (commented):
Reviewed the packed-opening API and batch-opening trait. One soundness concern on the `PackedBatch` prefix-packing reduction (inline) — it appears exploitable for mixed-arity packings, which the doc/tests treat as the intended use case; the mixed-arity test only checks completeness with honestly-derived claims, so it wouldn't catch this. Worth confirming before later PRs build on it.

The rest looks clean: the `bind_opening_inputs` → `EvaluationClaim`/`ZkEvaluationClaim` transcript refactor keeps prover (jolt-prover-legacy) and verifier label strings in sync, and the `?Sized`/borrowed-source generalization of `open`/`open_zk` is consistent across Dory and HyperKZG. Minor: `HyperKZGScheme::open` now materializes the full evaluation table into a fresh `Vec` via `for_each_row(num_vars)` rather than borrowing `poly.evaluations()` — correct and needed for the generic source, but an extra full-table copy per open; could specialize when the source is already dense if it shows up in profiles.

---

crates/jolt-openings/src/packing.rs
```diff
@@ -0,0 +1,507 @@
...
+    deserialize = "PCS::VerifierSetup: Deserialize<'de>, Id: Ord + Deserialize<'de>"
+))]
+#[serde(deny_unknown_fields)]
+pub struct PrefixPackedVerifierSetup<PCS: CommitmentScheme, Id = u64> {
+    pub pcs: PCS::VerifierSetup,
+    pub packing: PrefixPacking<Id>,
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+struct DomainSize(usize);
```
moodlezoup: This wrapper type feels unnecessary to me, imo we should just do the `checked_shl` inline

crates/jolt-openings/src/packing.rs
```diff
@@ -0,0 +1,507 @@
...
+            return Err(OpeningsError::InvalidSetup(
+                "prefix packing requires at least one polynomial".to_owned(),
+            ));
+        }
+
+        let mut total_cells = 0usize;
+        for polynomial in &polynomials {
+            let cells = DomainSize::try_from(polynomial.num_vars)?.cells();
+            total_cells = total_cells.checked_add(cells).ok_or_else(|| {
+                OpeningsError::InvalidSetup("prefix packing domain size overflow".to_owned())
```
moodlezoup: I don't really like the "cells" terminology, let's just call them coefficients

crates/jolt-openings/src/schemes.rs
```diff
@@ -186,64 +211,430 @@ pub trait ZkStreamingCommitment: StreamingCommitment + ZkOpeningScheme {
...
+    PCS: CommitmentScheme,
+    PCS::Output: AppendToTranscript,
+    Id: Clone + Debug + Eq + Ord + Send + Sync + 'static,
+{
+    type Field = PCS::Field;
+    type ProverSetup = PrefixPackedProverSetup<PCS, Id>;
+    type VerifierSetup = PrefixPackedVerifierSetup<PCS, Id>;
+    type Statement = PrefixPackedStatement<PCS::Field, Id, PCS::Output>;
+    type BatchingWitness<'a>
+        = PackedWitness<'a, PCS::Field, PCS::OpeningHint>
```
moodlezoup: it doesn't feel right to be shoehorning both the witness polynomial and the opening hint under one associated type here

crates/jolt-openings/src/packing.rs
```diff
@@ -0,0 +1,507 @@
...
+    {
+        let claims = statement.claims.as_slice();
+        if claims.is_empty() {
+            return Err(OpeningsError::InvalidBatch(
+                "packing opening requires at least one claim".to_owned(),
+            ));
+        }
+
+        let mut seen = BTreeSet::new();
+        let mut constrained_point = vec![None; self.packed_num_vars];
```
moodlezoup: Are we guaranteed that all elements of this Vec are `Some` by the end of the for loop? If so, we should assert that and change the `PreparedPrefixPackedStatement::constrained_point` field type to be `Vec<F>`.

crates/jolt-openings/src/packing.rs
```diff
@@ -0,0 +1,507 @@
...
+    F: Field,
+{
+    pub fn new(polynomial: &'a (dyn MultilinearPoly<F> + 'a), hint: H) -> Self {
+        Self { polynomial, hint }
+    }
+}
+
+/// One logical opening claim inside a prefix-packed commitment.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct PrefixPackedClaim<F, Id> {
```
moodlezoup: I slightly prefer that we get rid of this struct and just use a tuple `(Id, EvaluationClaim<F>)`

crates/jolt-openings/src/schemes.rs
```diff
@@ -39,28 +43,14 @@ pub trait CommitmentScheme: Commitment {
...
-        poly: &Self::Polynomial,
-        point: &[Self::Field],
-        eval: Self::Field,
-        setup: &Self::ProverSetup,
-        hint: Option<Self::OpeningHint>,
-        transcript: &mut impl Transcript<Challenge = Self::Field>,
-    ) -> Self::Proof;
-
-    fn open_poly<P: MultilinearPoly<Self::Field>>(
+    fn open<P: MultilinearPoly<Self::Field> + ?Sized>(
```
moodlezoup, line 45: do we still need the `Polynomial` associated type at this point?

crates/jolt-openings/src/schemes.rs
```diff
@@ -186,64 +211,430 @@ pub trait ZkStreamingCommitment: StreamingCommitment + ZkOpeningScheme {
...
-    /// Commit in the scheme's ZK/hiding mode.
-    fn commit_zk<P: MultilinearPoly<Self::Field> + ?Sized>(
-        poly: &P,
+/// Batch opening protocol selected by a small marker adapter.
+pub trait BatchOpeningScheme {
+    type Field: Field;
+    type ProverSetup;
+    type VerifierSetup;
+    type Statement;
+    type BatchingWitness<'a>
```
moodlezoup: unclear to me what `Statement` and `BatchingWitness` mean exactly

crates/jolt-openings/src/schemes.rs
```diff
@@ -186,64 +211,430 @@ pub trait ZkStreamingCommitment: StreamingCommitment + ZkOpeningScheme {
...
+
+impl<PCS, Id> PackedBatch<PCS, Id> {
+    pub fn new() -> Self {
+        Self(PhantomData)
+    }
+}
+
+pub type HomomorphicBatchWitness<'a, F, H> = Vec<(&'a (dyn MultilinearPoly<F> + 'a), H)>;
+
+pub type HomomorphicZkBatchWitness<'a, F, H> = Vec<(&'a (dyn MultilinearPoly<F> + 'a), H, F)>;
```
moodlezoup: I don't think we should be mixing polynomials with hints/evaluations under a blanket "witness" type

crates/jolt-openings/src/schemes.rs
```diff
@@ -186,64 +211,430 @@ pub trait ZkStreamingCommitment: StreamingCommitment + ZkOpeningScheme {
...
+pub struct HomomorphicBatch<PCS>(PhantomData<fn() -> PCS>);
+
+impl<PCS> HomomorphicBatch<PCS> {
+    pub fn new() -> Self {
+        Self(PhantomData)
+    }
+}
+
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub struct PackedBatch<PCS, Id = u64>(PhantomData<fn() -> (PCS, Id)>);
```
moodlezoup: this `newtype(PhantomData<closure>)` thing we have going on here is not very idiomatic

crates/jolt-openings/src/schemes.rs
```diff
@@ -108,10 +92,51 @@ pub trait StreamingCommitment: CommitmentScheme {
...
+
+    /// Verify a ZK opening proof and return the hiding commitment to the
+    /// evaluation that the proof binds internally.
+    fn verify_zk(
+        commitment: &Self::Output,
+        point: &[Self::Field],
+        proof: &Self::Proof,
+        setup: &Self::VerifierSetup,
+        transcript: &mut impl Transcript<Challenge = Self::Field>,
+    ) -> Result<Self::HidingCommitment, OpeningsError>;
```
moodlezoup: we should document what the relationship is between `CommitmentScheme::Output` and `ZkOpeningScheme::HidingCommitment`, I find this function signature confusing

crates/jolt-openings/src/packing.rs
markosg04: I adjusted the protocol and left a big doc comment explaining the full packing protocol

crates/jolt-openings/src/schemes.rs
markosg04: No, removed

crates/jolt-openings/src/schemes.rs
markosg04: Split this up and left comments for what the remaining pieces do

crates/jolt-openings/src/schemes.rs
markosg04: I changed it to `<(PCS, ID)>` I think we need the new types since we can't blanket impl twice for both cases, unless we use a different approach entirely

PR #1657, opened 2026-07-01
Migrate Jolt core transcripts to Spongefish

Vishalkulkarni45:
## Summary

Split out from #1586. This PR migrates the Jolt core transcript path from the hand-rolled Fiat-Shamir implementation to the Spongefish-backed `jolt-transcript` split-trait/NARG APIs.

This PR intentionally excludes the transpiler work; that will be handled in a follow-up PR.

## Changes

- Replace legacy hand-rolled transcript plumbing with Spongefish-backed prover/verifier transcript APIs.
- Move core prover payload transport into NARG where appropriate.
- Update prover, verifier, sumcheck, BlindFold, Dory, openings, and transcript call sites to use the new transcript surface.
- Bind transcripts to explicit sessions and statement instances.
- Add verifier EOF/trailing-data checks for NARG decoding.
- Remove duplicated legacy transcript code and cleanup tests around the new proof shape.
- Keep non-BN254 sumcheck coverage with the Mersenne-style canary.

## 

Internal proof/transcript representation changes to use Spongefish sessions and NARG APIs. Transpiler compatibility is intentionally deferred to the next PR.


Vishalkulkarni45:
@moodlezoup can u review this PR

moodlezoup:
> @moodlezoup can u review this PR

Hi Vishal, thanks for preparing this PR. We merged a PR introducing DigestTranscript, which effectively restores transcript parity between jolt-core and the new crates from the other direction (i.e. implementing the jolt-core transcript semantics within jolt-transcript). Given that, we've decided to put the migration on pause for now. Sorry for the sudden change in plans

0xAndoroid:
Closing per the pause decision (moodlezoup, Jul 1): main adopted the byte-compatible DigestTranscript (#1617) to restore transcript parity instead of migrating core to spongefish, and the jolt-core surfaces this patches have since moved to jolt-prover-legacy (#1632) or been rewritten in the modular crates. Thanks for the work here — if the migration is revived it will be re-scoped against jolt-prover/jolt-verifier. Branch left in place.

PR #1663, opened 2026-07-02
feat: lattice mode protocol semantics in jolt-claims

markosg04:
Third PR in the lattice/Akita stack:

## Features

`protocols/jolt/lattice/` — the protocol semantics for running the Jolt PIOP over a packed one-hot witness committed with a non-homomorphic PCS (Akita). It describes *additional* semantics on top of the base `jolt/` PIOP

**New relations**
- `IncVirtualization` — replaces `IncClaimReduction` in lattice mode: γ-batches the four `RamInc`/`RdInc` consumer claims into one fused increment stream, `FusedInc · (ram_coeff·store + γ²·rd_coeff·(1−store))`. The destination selector is the existing `OpFlags(Store)` polynomial, so store binding rides the bytecode read-raf val stages (one extra stage, `LATTICE_BYTECODE_VAL_STAGES`)
- `UnsignedIncChunkReconstruction` — one sumcheck over the chunk address bits carrying three γ-batched legs per one-hot chunk column: exactly-one-hot hamming weight, claim reduction from the booleanity point, and the value identity `Σ_j place_j·symbol(chunk_j) = FusedInc + 2^64·(1−msb)`
- Lattice `Booleanity` — same relation id as base, output fold extended over the `UnsignedIncChunk`/`UnsignedIncMsb` columns 
- `AdviceBytesValidity` — booleanity + per-position hamming over the untrusted-advice byte one-hot column; doubles as the range check behind the advice word reconstruction.

**Packing + final openings**
- `LatticeColumn` + `proof_packing`/`precommitted_packing`: the canonical column sets (per-proof: `Ra` families, inc chunks/msb, untrusted advice bytes; precommitted: bytecode lane sub-columns, program image bytes, trusted advice bytes) with their arities.
- `ReconstructionTerm` lists for every decomposed logical polynomial (advice words, `BytecodeChunk` lane values, program-image words): pure `(column, cell, weight)` descriptions of the weighted sums that rebuild them from packed cells.
- `final_opening(polynomial) -> LatticeFinalOpening` — the lattice replacement for the stage-8 RLC order, total over `JoltCommittedPolynomial`: packed claim (with its producing-relation leaf, derived from the base `final_opening_relation`), reconstruction, or virtualized.
- Append-only id extensions (`JoltRelationId`, `JoltCommittedPolynomial`, `JoltVirtualPolynomial`, challenge/public enums).

## prefix-packed polynomials

Akita has no commitment homomorphism, so the stage-8 RLC over separate commitments is replaced by one packed witness `W` per commitmen. `jolt-claims` registers the canonical column ordering with `jolt_openings::PrefixPacking::new` — a `(LatticeColumn, num_vars)` list per lifecycle — and returns the `PrefixPacking` itself; Each logical column is a slot restriction `P_i(x) = W(prefix_i ‖ x)`, and one opening of `W` settles every claim via `W(r) = Σ_i eq(prefix_i, r[..|prefix_i|]) · P_i(r[|prefix_i|..])`. The relation chain above exists to arrange that reduction's precondition

## Diff
Non-test source (non-blank; `#[cfg(test)]` modules and `tests/` excluded): **+1,090** to `jolt-claims` — 1,000 in the new `lattice/` module and +90 across base files (id enums +61, shared increments/booleanity geometry helpers +20, final-opening map arms +10, re-exports; one base relation −5 from helper reuse)


crates/jolt-claims/src/protocols/jolt/ids.rs
```diff
@@ -439,6 +503,9 @@ pub enum JoltDerivedId {
     InstructionInput(InstructionInputPublic),
     InstructionReadRaf(InstructionReadRafPublic),
     InstructionRaVirtualization(InstructionRaVirtualizationPublic),
+    IncVirtualization(IncVirtualizationPublic),
+    UnsignedIncChunkReconstruction(UnsignedIncChunkReconstructionPublic),
+    AdviceBytesValidity(AdviceBytesValidityPublic),
```
moodlezoup: These three variants are inserted *before* the trailing `PublicInput(usize)`/`PublicOutput(usize)`, shifting their discriminants down by 3. Every sibling enum in this PR (`JoltRelationId`, `JoltCommittedPolynomial`, `JoltVirtualPolynomial`, `JoltChallengeId`) appends new variants at the very end for exactly the codec-stability reason called out in their comments — `JoltDerivedId` is the odd one out. Impact is likely nil today (deriveds are verifier-recomputed, not proof-transmitted, and I found no positional codec of this enum), but it contradicts the stated invariant. Move them after `PublicOutput` to keep the append-only pattern.

---

moodlezoup (commented):
Reviewed the lattice-mode semantics across four passes (semantic consistency, bug analysis, tech-debt, and soundness). This is a careful, thoroughly unit-tested change and I found no soundness gaps in the symbolic relation definitions: the γ-batching is collision-free in every relation, the one-hot/range logic (advice bytes, chunk reconstruction, booleanity extension) is complete for the columns it covers, and `IncVirtualization` faithfully mirrors the base `IncClaimReduction` batching (shared `inc_consumers_input`, same four claims, same γ order). The reconstruction identities are pinned against real MLE evaluations in `tests/lattice_semantics.rs`.

One inline note on `JoltDerivedId` variant ordering. Two non-blocking observations:

- The soundness of `IncVirtualization` rests on two checks that live outside this PR: the `OpFlags(Store)` opening being bound to the real bytecode Store flag, and the offline per-row store/rd disjointness check. Both are documented; just worth confirming they're actually wired in `jolt-verifier`/`jolt-prover` before lattice mode is exercised end-to-end, since a dropped store binding would let a prover split one fused increment into two.
- `word_byte_column_vars` (geometry.rs) rounds the limb dimension with `ilog2` (floor) while the matching `scaled_byte_reconstruction_terms` (packing.rs) uses `Math::log_2` (ceil). Harmless today since `WORD_BYTE_LIMBS == 8` is a power of two, but the two conventions could silently drift; consider using `log_2` in both for consistency. A couple of exports (`LATTICE_BYTECODE_VAL_STAGES`, `program_image_word_reconstruction_terms`) also have no consumer or test yet — expected for a consumer-ahead semantics layer, flagging only in case a consumer was meant to land here.

---

crates/jolt-claims/src/protocols/jolt/lattice/packing.rs
```diff
@@ -0,0 +1,707 @@
...
+        );
+    }
+
+    #[test]
+    fn packing_registration_is_deterministic() {
+        assert_eq!(
+            proof_packing(&proof_shape()).unwrap(),
+            proof_packing(&proof_shape()).unwrap()
+        );
+    }
```
moodlezoup: useless test

crates/jolt-claims/src/protocols/jolt/lattice/packing.rs
```diff
@@ -0,0 +1,707 @@
...
+        | JoltCommittedPolynomial::UntrustedAdviceBytes => LatticeFinalOpening::Packed {
+            column: LatticeColumn::from(polynomial),
+            leaf: packed_column_leaf(LatticeColumn::from(polynomial)),
+        },
+        JoltCommittedPolynomial::TrustedAdvice => LatticeFinalOpening::AdviceWordReconstruction {
+            kind: JoltAdviceKind::Trusted,
+        },
+        JoltCommittedPolynomial::UntrustedAdvice => LatticeFinalOpening::AdviceWordReconstruction {
+            kind: JoltAdviceKind::Untrusted,
+        },
```
moodlezoup: It doesn't make sense to me that both TrustedAdviceBytes _and_ TrustedAdvice would be committed polynomials (respectively, UntrustedAdviceBytes _and_ UntrustedAdvice). With Akita, it we'd only commit to the advice in one-hot byte-decomposed form, right? So the TrustedAdvice/UntrustedAdvice claims originating from the RAM sumchecks would be virtualized in terms of TrustedAdviceBytes and UntrustedAdviceBytes. Why is that virtualization not yet codified in the relations?

crates/jolt-claims/src/protocols/jolt/lattice/relations/advice_bytes.rs
```diff
@@ -0,0 +1,183 @@
+//! Untrusted-advice byte validity: proves the prover-supplied advice byte
+//! column is a well-formed byte one-hot encoding. This is simultaneously the
+//! range check behind the advice word reconstruction — byte one-hot cells force every
+//! decoded byte below 256 and every decoded word below 2^64.
+//!
+//! One sumcheck over the column's `(symbol ‖ limb ‖ word)` cell domain
```
moodlezoup: I find the whole "symbol" / "limb" / "byte" / "word" / "cell" / "row" / "column" / "arity"  nomenclature confusing; I'm still not sure what the relationship is between all of these terms.

crates/jolt-claims/src/protocols/jolt/lattice/packing.rs
```diff
@@ -0,0 +1,707 @@
...
+    },
+    /// A word-valued advice polynomial, rebuilt through
+    /// [`advice_word_reconstruction_terms`].
+    AdviceWordReconstruction { kind: JoltAdviceKind },
+    /// A committed `BytecodeChunk(chunk)`, rebuilt through
+    /// [`bytecode_chunk_reconstruction_terms`].
+    BytecodeChunkReconstruction { chunk: usize },
+    /// A program-image word, rebuilt through
+    /// [`program_image_word_reconstruction_terms`].
+    ProgramImageWordReconstruction,
```
moodlezoup: > `None` means the claim is produced by the reconstruction reduction rather than a relation

Why is the reconstruction reduction not itself a relation?
Relatedly, why are `AdviceWordReconstruction` `BytecodeChunkReconstruction` and `ProgramImageWordReconstruction` separate variants?

crates/jolt-claims/src/protocols/jolt/lattice/packing.rs
markosg04: added the one hot virtualization relations

crates/jolt-claims/src/protocols/jolt/lattice/packing.rs
markosg04: added in the corrected relations

crates/jolt-claims/src/protocols/jolt/lattice/relations/advice_bytes.rs
markosg04: got rid of most of these and aligned it with the language we are using in `jolt-openings`

PR #1666, opened 2026-07-08
Rust to lean bytecode expansiosn

abiswas3:
## Summary

This crate is a lightweight wrapper around `jolt-riscv` and `jolt-program` to automatically extract Jolt bytecode expansions in Lean.
There is only 1 file `main.rs` and it is heavily commented. 
Tracing the usage command listed below, and following the comments in an IDE should be self explanatory.

## Testing

- [x] Automatic translations have been formally verified against previous hand coded translations.  

## Security Considerations

Does not touch live prover or verifier. 

## Breaking Changes

This is a source to source transpiler.
Does not touch live prover or verifier.


crates/jolt-lean-gen/src/main.rs
```diff
@@ -0,0 +1,362 @@
...
+// Render a register as a bare Lean `regidx` (no `.xreg` wrapper), used where an
+// `Instr` field is a plain `regidx` (e.g. alignment-assert base).
+fn regidx(reg: Option<u8>) -> String {
+    reg.map_or_else(|| "?".to_string(), reg_symbol)
+}
+
+// Distinctive source-immediate sentinel: valid as a 6-bit shift amount and a
+// 12-bit immediate, and unlikely to collide with a recipe constant. Any row whose
+// immediate equals this is the source immediate passed through, printed as `imm`.
+const SRC_IMM: i128 = 37;
```
moodlezoup: Non-blocking, but worth guarding: this identifies the pass-through source immediate purely by value (`imm == 37`). Any recipe constant or derived immediate that happens to equal 37 gets silently rendered as `imm` instead of its literal, producing wrong Lean with no signal — which is exactly the failure mode this tool is meant to eliminate by replacing hand-coding. Today's recipes only bake small constants (`-8`, `-1`, `7`, `3`, `6`, `0`), so there's no collision now, but nothing enforces that as new instructions are added to the table. Consider at least a debug assert that no non-passthrough row in an expansion carries the sentinel value, or feeding a sentinel wide enough that a collision is implausible for any plausible recipe constant.

---

moodlezoup (commented):
Reviewed the `jolt-lean-gen` transpiler crate. This is a self-contained, offline dev tool: it doesn't touch the live prover, verifier, transcript, or constraint generation, and the only changes to existing code are making `is_source_only` public and adding `#[derive(Debug)]` — so none of the soundness/proof-correctness concerns apply here.

The transpiler logic itself checks out for the current instruction set: the recipe constants are all small values that don't collide with the immediate sentinel, and the shift-bitmask `trailing_zeros` recovery (`sraiBitmask`/`srliBitmask`/`slliMultiplier`) correctly reconstructs the shift amounts. Since translations are formally verified downstream against the hand-coded Lean, the `_` best-effort arm and other heuristics are appropriately caught if wrong.

One non-blocking robustness note left inline on the `SRC_IMM` sentinel. Otherwise looks good.

---

crates/jolt-lean-gen/src/main.rs
abiswas3: Noted. I'll update this tomorrow. Thanks, Michael.

abiswas3: I've resolved this by running the instruction with two immediate values -- 37 and 41 in this case (5 and 13 for smaller width instructions). 
Now if any hardcoded immediate exists, it would be the same for both inputs, and we'd know that immediate should not be substituted with `imm` but the pretty printer.

moodlezoup (approved):

PR #1669, opened 2026-07-10
feat(jolt-prover): introduce new modular prover with naive sumcheck prover

moodlezoup:
## Summary

A clean-slate modular prover for the new sumcheck/claims/verifier stack, superseding the port approach of #1637. `jolt-prover` produces the complete transparent (clear-mode) `JoltProof` **byte-identical to `jolt-prover-legacy`** — every wire component and every stage-boundary Fiat-Shamir transcript state — across all supported configurations: base, trusted/untrusted advice, committed-program preprocessing, the address-major trace layout, and their combinations.

The architecture follows `specs/clean-slate-prover.md` (added here): stage recipes in `jolt-prover` are pure orchestration and prove against a runtime-swappable backend value — the `JoltBackend` slot registry in the new `jolt-kernels` crate, which owns all field-element compute. Swapping a kernel implementation, mixing implementations per slot, or running two backends side by side is value construction, never recompilation. The always-present reference backend serves every slot through a naive tier that interprets each relation's `Expr` with polynomial-valued leaves, so relations are provable with zero relation-specific code; hand-written kernels exist only where the interpreter structurally cannot go (the instruction read+RAF prefix–suffix decomposition, post-uni-skip stream rounds, staged intermediates).

## Changes

- **`jolt-sumcheck` prove side**: the `SumcheckRecorder` seam (clear / Pedersen-committed), a generated `begin_batch` head shared verbatim with the verifier's `verify_clear`, the batched `prove_batch` engine (per-member activation offsets for head-aligned precommitted members), and uni-skip first-round provers.
- **`jolt-prover` (new crate)**: stage recipes 0–8 mirroring the verifier's stage layout, a top-level `prove()` assembling the full `JoltProof`; Fiat-Shamir work reuses the verifier's own exported functions (`validate_inputs_from_parts`, preamble/commitment absorbs), so stage-0 transcript drift is impossible by construction.
- **`jolt-kernels` (new crate)**: the backend seam (`JoltBackend`, `ProofSession`, per-relation slot traits at the crate root) with every implementation under `reference/` — the naive interpreter, the witness-commitment and joint-opening kernels, the prefix–suffix instruction read+RAF kernel, the hand address-phase kernels, and the precommitted (advice / bytecode-chunk / program-image) reduction kernels.
- **`jolt-claims`**: the derived-ids ↔ multilinears doctrine — RAM output-check and Spartan-outer publics split into constituent multilinears (value-identical folds; the BlindFold R1CS lowering is value-level, so all ZK fixtures pass unchanged).
- **Mode support**: advice (front-loaded batch members, block-embedded stage-8 openings), committed programs (prover-retained full program + chunk/image hints, trace-order recorded in preprocessing and checked at stage 0), address-major layout (strided commitment placement, embedding-extra strides on widened grids), and all pairwise combinations.
- **`jolt-dory`**: `setup_prover` now canonicalizes odd `max_num_vars` to the even bucket like legacy (odd/even pairs share a generator count but the URS derivation seeds on the exact size); existing even-sized setups are unchanged.
- **`jolt-verifier`**: visibility-only promotions (stage input/output aggregators, `batch_entries`, `build_stage7_sumchecks`, …) so the prover reuses the verifier's own assembly instead of duplicating it.

## Testing

- **Byte-diff harness** (`crates/jolt-prover/tests/byte_diff.rs`, `--features prover-fixtures`): ten tests run both provers from the same guest and inputs and assert wire-for-wire equality plus end-to-end verification. `muldiv` is a stage-granular ratchet (per-stage proofs, claims, and transcript states recovered by replaying legacy's proof through the verifier); the others are whole-proof ratchets over the mode × trace-order matrix: advice-consumer, committed muldiv (chunk counts 2 and 64), fibonacci under address-major, and advice × committed — each in both trace orders where applicable.
- Verifier fixture suites (standard + zk, including all BlindFold fixtures) pass against pre-existing cached fixtures — verifier bytes are untouched.
- Full workspace suite, `cargo clippy` under `host`, `host,zk`, and the field-inline crate set, `cargo fmt`.
- Legacy e2e advice/committed tests in both standard and ZK modes.

## Security Considerations

No proof-system behavior changes: the new prover is pinned byte-identical to the audited legacy prover across every supported configuration, and each stage recipe cross-checks its final batch claim against the verifier's own `expected_final_claim` algebra during proving. Verifier-side edits are visibility promotions and value-identical claim refactors (fixture suites pass un-regenerated, so soundness/tampering coverage is intact). The stage-0 guards fail fast on unsupported regimes (dominant advice; committed preprocessing whose baked trace order disagrees with the proof config) rather than emitting unverifiable proofs. ZK mode is unchanged on this PR — the clear-mode prover is the deliverable; the committed recorder groundwork for BlindFold is in place but not yet wired.

## Breaking Changes

None to proof wire formats or serialization (proofs are byte-identical to legacy's). Workspace-internal APIs changed shape (jolt-claims publics enums split into constituents; `CommittedProgramProverData` gained `trace_order`; kernel implementations moved under `jolt_kernels::reference`), with crate-root re-exports keeping downstream paths stable.


0xAndoroid (commented):
Deep review via parallel analysis agents (semantic consistency, bug analysis, code quality, security/soundness), every finding re-verified against the code. No soundness regression, Fiat-Shamir divergence, or acceptance-set widening found — the findings below are one latent witness-convention split, a handful of prover-inlines-verifier-logic drift seams, and structural/doc cleanups.

*Claude Code Fable 5 Review*

crates/jolt-witness/src/protocols/jolt_vm/trace.rs
```diff
@@ -211,7 +211,11 @@ pub(crate) fn trace_virtual_value<F: Field>(
         JoltVirtualPolynomial::NextUnexpandedPC => {
             next.map_or_else(F::zero, |row| F::from_u64(row.instruction.address as u64))
         }
-        JoltVirtualPolynomial::NextIsNoop => F::from_bool(next.is_some_and(row_is_noop)),
+        // The last cycle's missing successor counts as a no-op: the
+        // product/shift family requires `NextIsNoop = 1` at `T - 1` (legacy
+        // forces `not_next_noop = false` there -- "EqPlusOne does not do
+        // overflow").
+        JoltVirtualPolynomial::NextIsNoop => F::from_bool(next.is_none_or(row_is_noop)),
```
0xAndoroid, line 218: `ShouldJump` (line 247) still computes `next_is_noop = next.is_some_and(row_is_noop)` — the old convention this line just moved away from. At `T−1` the new `NextIsNoop = 1` makes R1CS constraint 21 (`ShouldJump = Jump × (1 − NextIsNoop)`) require `ShouldJump = 0`, but the witness produces `Jump`. Same one-line `is_none_or` fix as here.

Any trace whose last row has the Jump flag set is now internally unsatisfiable, and the `ShouldJump` value also diverges from legacy at `T−1` (legacy emits `should_jump = jump · not_next_noop = 0` there). The shipped `prove()` path masks this only because config derivation always pads a trailing NoOp — a caller feeding an exactly-power-of-two trace ending in a jump hits an opaque `RoundCheckFailed`.

---
*Claude Code Fable 5 Review*

crates/jolt-prover/src/stages/stage6b.rs
```diff
@@ -0,0 +1,619 @@
...
+    // The staged advice RAM address points from stage 4's RAM value-check —
+    // the clear-only references the advice `FinalScale` terms read.
+    let advice_reference = |kind| {
+        stage4
+            .ram_val_check_init
+            .advice_contribution(kind)
+            .map(|contribution| contribution.opening_point.clone())
+    };
+
+    let sumchecks = Stage6bSumchecks {
```
0xAndoroid, line 222: This hand-assembles `Stage6bSumchecks` leg-for-leg from the verifier's `Stage6bSumchecks::build` — the most intricate Fiat-Shamir choreography in the protocol, and the only stage whose batch construction isn't shared code (stage 7 got `build_stage7_sumchecks`, stage 8 got `batch_entries`). The spec lists `stage6b::batch::build` among the intended export promotions; `build` takes `&JoltProof` so it can't be called as-is, but extracting its leg-assembly core over the clear carriers and promoting it would delete this mirror plus stage6a's `bytecode_stage_points`.

A change to one constructor leg in the verifier's build (a new dimension argument, a different point split) that isn't replicated here produces different geometry on the two sides, surfacing only as `FinalClaimMismatch`/transcript divergence in whichever configs the byte-diff harness exercises — the one place the PR's byte-identity-by-construction guarantee degrades to byte-identity-by-e2e-test.

---
*Claude Code Fable 5 Review*

crates/jolt-prover/src/stages/stage6b.rs
```diff
@@ -0,0 +1,619 @@
+//! Stage 6b: the cycle-phase batch — bytecode read+RAF and booleanity cycle
+//! phases, RAM Hamming booleanity, both RA virtualizations, the increment
+//! claim reduction, and the present advice claim-reduction cycle phases
+//! (head-aligned members; the committed-program `Option` members remain
```
0xAndoroid: Stale doc: the committed-program `Option` members are no longer "rejected upstream" — this file builds them (lines 450–485), batches them (501–505), and emits their wire claims. Leftover from the advice-only milestone.

---
*Claude Code Fable 5 Review*

crates/jolt-claims/src/protocols/jolt/ids.rs
```diff
@@ -608,14 +613,17 @@ pub enum JoltDerivedId {
...
     UnsignedIncChunkReconstruction(UnsignedIncChunkReconstructionPublic),
     UntrustedAdviceReconstruction(UntrustedAdviceReconstructionPublic),
     TrustedAdviceReconstruction(TrustedAdviceReconstructionPublic),
+    /// Test-only derived id for toy relations that define their own
+    /// `derive_output_term`. Gated on `any(test, feature = "test-utils")`
+    /// rather than `test` alone because `cfg(test)` is per-crate: a downstream
+    /// crate's tests (the naive prover's toy relation foremost) see this crate
+    /// compiled without `cfg(test)` and reach the variant via the feature.
+    #[cfg(any(test, feature = "test-utils"))]
+    Test,
```
0xAndoroid, line 630: Feature-gated variant inserted mid-enum on a serde-derived type: flipping `test-utils` shifts the discriminants of `ProgramImageReconstruction`/`BytecodeChunkReconstruction` across feature-unified builds (jolt-kernels enables the feature unconditionally). `JoltDerivedId` isn't wire-encoded today, so this is latent — moving `Test` to the end of the enum eliminates the hazard for free.

---
*Claude Code Fable 5 Review*

crates/jolt-kernels/src/spartan_shift.rs
```diff
@@ -0,0 +1,23 @@
...
+use jolt_field::Field;
+use jolt_verifier::stages::stage3::outputs::SpartanShift;
+use jolt_witness::protocols::jolt_vm::JoltVmNamespace;
+use jolt_witness::WitnessProvider;
+
+use crate::{KernelError, ProofSession, ProveSumcheck};
+
+/// The stage-3 Spartan-shift slot.
+pub trait SpartanShiftProver<F: Field> {
+    fn prepare(
```
0xAndoroid, line 14: Representative of ~20 slot traits: the stage recipe already built `SpartanShift::new(trace_dimensions, product_tau_low, product_remainder_point)` for the batch, passes the same three args through this signature, and the reference kernel reconstructs the identical relation internally. Passing the relation itself (`&SpartanShift<F>` — it already is the typed request) would collapse most of the single-method trait family into one generic slot trait and make it impossible for a kernel to receive parts that drift from the batch's relation.

Worth settling before other backends implement 27 traits. Bespoke traits stay only where the lifecycle genuinely differs: spartan_outer's uniskip handoff, instruction_read_raf's typed rows, the precommitted phase-carry, commitment/opening.

---
*Claude Code Fable 5 Review*

crates/jolt-prover/src/config.rs
```diff
@@ -0,0 +1,187 @@
...
+}
+
+/// Map a byte address into the RAM word index space: word offsets from the
+/// memory layout's lowest mapped address; address 0 means "no access".
+pub fn remap_address(address: u64, memory_layout: &MemoryLayout) -> Option<u64> {
+    if address == 0 {
+        return None;
+    }
+    let lowest = memory_layout.get_lowest_address();
+    (address >= lowest).then(|| (address - lowest) / 8)
```
0xAndoroid: Legacy's `remap_address` panics on a nonzero address below `lowest_address` ("Unexpected address"); this returns `None`, so `derive`'s `filter_map` silently drops the access from the `touched` max and a malformed trace under-sizes `ram_K`, failing far downstream as an opaque sumcheck error instead of at derivation. In a module documented as byte-for-byte legacy parity, worth matching the loud failure (or documenting the deliberate divergence).

---
*Claude Code Fable 5 Review*

crates/jolt-prover/src/stages/stage2.rs
```diff
@@ -0,0 +1,256 @@
...
+        RamRafEvaluationDimensions::try_from(read_write_dimensions).map_err(|error| {
+            VerifierError::StageClaimPublicInputFailed {
+                stage: JoltRelationId::RamRafEvaluation,
+                reason: error.to_string(),
+            }
+        })?;
+
+    // `τ_low` is stage 1's remainder point tail, reversed — computed exactly as
+    // the verifier's `Stage1Output::remainder_point` + `verify_product_uniskip`.
+    let remainder_point: Vec<F> = stage1
```
0xAndoroid, line 88: Inline copy of `Stage1Output::remainder_point()`'s Clear arm plus `verify_product_uniskip`'s slicing, minus the verifier's `tau_low.len() != log_t` check — and the same copy recurs in stage6a's `bytecode_stage_points`. Promoting a helper on the clear output would make this compiler-checked.

The stored point's orientation was already re-oriented once (per the outputs.rs doc warning that this must NOT be the already-reversed opening point). If it changes again, the verifier helper gets updated in one place and these two copies drift, surfacing as an opaque stage-2 transcript mismatch.

---
*Claude Code Fable 5 Review*

crates/jolt-prover/src/stages/stage7.rs
```diff
@@ -0,0 +1,271 @@
...
+        &virtualization_points,
+        &challenges.hamming_weight_claim_reduction,
+        witness,
+    )?;
+
+    // The advice address phases: the stage-6b kernel objects, transitioned.
+    // A member joins exactly when the batch declares it (layout committed AND
+    // active address rounds) — the kernel exists whenever the layout does, so
+    // absence here just leaves a cycle-completed kernel behind.
+    let mut trusted_advice = match (&sumchecks.trusted_advice, trusted_advice_member.as_mut()) {
```
0xAndoroid: These four match blocks are identical modulo the error string, and the four conditional pushes below mirror four more in stage6b. A small `take_scheduled(declared, member, name)` helper plus a single carry struct on `Stage6bProverOutput` would delete ~50 lines and make the 6b→7 handoff shape explicit — bytecode and program-image presence is one decision anyway (`CommittedProgramSchedule` yields both or neither), so two of the four Options encode the same bit.

---
*Claude Code Fable 5 Review*

crates/jolt-kernels/src/reference/ram_val_check.rs
```diff
@@ -0,0 +1,83 @@
...
+        challenges: &RamValCheckChallenges<F>,
+        witness: &dyn WitnessProvider<F, JoltVmNamespace>,
+    ) -> Result<Box<dyn ProveSumcheck<F, Relation = RamValCheck<F>>>, KernelError<F>> {
+        let cycles = 1usize << trace_dimensions.log_t();
+        let addresses = 1usize << ram_log_k;
+
+        // The address-bound `ra` slice, folded from the full `(K × T)` grid.
+        let eq_address = eq_table(r_address);
+        let ram_ra_full = dense_view(witness, ram_ra_val_check())?;
+        let ra_folded: Vec<F> = (0..cycles)
```
0xAndoroid: This is `views::address_fold` re-implemented token-for-token, minus its `TableSizeMismatch` length check — a mis-sized grid panics on indexing here instead of returning the typed error. Same hand-rolled copy in registers_val_evaluation.rs and ram_ra_claim_reduction.rs (and `cycle_fold` in ram_raf_evaluation.rs / booleanity.rs). The canonical helpers already exist in this crate; using them restores the check for free.

---
*Claude Code Fable 5 Review*

crates/jolt-kernels/src/reference/naive.rs
```diff
@@ -0,0 +1,709 @@
...
+//! verifier's `derive_output_term` evaluates at a point (eq/LT/selector
+//! tables, materialized over the domain by the caller). Each round message is
+//! `Expr::try_evaluate` run pointwise over the remaining hypercube — cost
+//! `O((degree+1) · 2^rounds · |Expr|)` per round, a **test oracle at harness
+//! scale, never a performance path**. It is the reference implementation
+//! optimized kernels are equivalence-tested against.
+//!
+//! Two self-checks pin it: the engine's running-claim check
+//! (`s(0) + s(1) == previous_claim`, re-checked per member here), and
+//! [`validate_derived_tables`](NaiveSumcheckProver::validate_derived_tables) —
```
0xAndoroid: The module doc sells `validate_derived_tables` as one of the two self-checks that pin the naive prover, but nothing outside this file's tests calls it — the naive-backed slots that hand-materialize derived tables never run it, so a drifted table is only caught coarsely by `expected_final_claim`. Either wire it into the stage recipes (or debug builds) or soften the doc.

---
*Claude Code Fable 5 Review*

crates/jolt-prover/src/stages/stage4.rs
```diff
@@ -0,0 +1,259 @@
...
+        return Err(ProverError::FinalClaimMismatch {
+            stage: "stage4",
+            expected,
+            got: proved.final_claim,
+        });
+    }
+
+    // The stage-curated absorb: `no_opening_values` suppresses the generated
+    // method, so the finish order comes from the claims struct's hand-ordered
+    // `opening_values()` (staged advice/program-image first — none here —
```
0xAndoroid: "none here" contradicts lines 211–223, which attach the program-image and advice values that `Stage4OutputClaims::opening_values()` then emits first. True only in the base config — misleading exactly when auditing the advice-config Fiat-Shamir order.

---
*Claude Code Fable 5 Review*

crates/jolt-prover/src/config.rs
```diff
@@ -0,0 +1,187 @@
...
+                bytecode_chunk_vars: bytecode.chunk_shape().total_vars(),
+                program_image_vars: image.image_shape().total_vars(),
+            }),
+            _ => None,
+        }
+    }
+}
+
+/// A word-aligned advice buffer's balanced Dory matrix variable count.
+pub(crate) fn advice_total_vars(max_advice_size_bytes: u64) -> usize {
```
0xAndoroid, line 189: Third copy of this computation: jolt-claims already owns it as `CommitmentMatrixShape::advice_from_max_bytes(...).total_vars()` (used by claim_reductions/advice.rs), and byte_diff.rs carries another copy with the identical doc comment. Worth pointing all three at the jolt-claims shape method so the advice-sizing law has one home.

---
*Claude Code Fable 5 Review*

crates/jolt-kernels/src/backend.rs
```diff
@@ -0,0 +1,128 @@
...
+/// Backend-owned state with proof lifetime, opaque to orchestration.
+///
+/// Slots stash and share private state keyed by a backend-private type, so
+/// per-slot mixing of backend families cannot collide: witness-upload
+/// residency, cross-member shared tables, and cross-stage carries all live
+/// here, invisible to the stage recipes that thread `&mut ProofSession`
+/// through every slot call.
+#[derive(Default)]
+pub struct ProofSession {
+    state: HashMap<TypeId, Box<dyn Any>>,
```
0xAndoroid, line 103: The TypeId map ships with zero consumers — all 29 reference `prepare` impls spell the parameter `_session`, and neither accessor has a caller in the workspace. Threading `&mut ProofSession` through the slots is cheap forward-compat worth keeping, but the map + `Any` downcasting could arrive with the first backend that needs it, shaped by its actual access pattern.

---
*Claude Code Fable 5 Review*

crates/jolt-sumcheck/src/committed.rs
```diff
@@ -163,6 +167,133 @@ impl<F: Field, C> BatchedCommittedSumcheckConsistency<F, C> {
...
+    F: Field,
+    VC: VectorCommitment<Field = F>,
+{
+    pub fn new(setup: &'a VC::Setup) -> Result<Self, SumcheckError<F>> {
+        if VC::capacity(setup) == 0 {
+            return Err(SumcheckError::ZeroCommitmentCapacity);
+        }
+        Ok(Self {
+            setup,
+            rng: OsRng,
```
0xAndoroid: Hardwiring `OsRng` forecloses the spec's own seam contract ("for a fixed seed, ZK proofs are backend-identical") and deterministic committed-path testing. Taking `R: RngCore` here (owned by jolt-prover, per the spec's wording) costs nothing now — every stage recipe will be touched again anyway when ZK lands.

---
*Claude Code Fable 5 Review*

crates/jolt-prover/src/stages/stage6b.rs
```diff
@@ -0,0 +1,619 @@
...
+    let mut recorder = ClearSumcheckRecorder::<F, C>::new();
+    let (batch, coefficients) =
+        sumchecks.begin_batch(&inputs, &cycle_challenges, &mut recorder, transcript)?;
+
+    // The address-only stage-value fold, once, for the bytecode kernel's
+    // constant `BytecodeValStage` tables (the recipe's relation instance
+    // recomputes the same fold internally for `expected_final_claim`). In
+    // committed mode the constants ARE the stage-6a staged raw values — the
+    // same quantities the fold computes.
+    let stage_values_at_r_address = if precommitted.bytecode.is_some() {
```
0xAndoroid, line 341: Field-element compute in the orchestration layer — and per the comment above, the relation instance recomputes the same fold internally for `expected_final_claim`. Exposing the folded stage values from the relation, or moving the fold into the bytecode-read-RAF kernel's `prepare` (which already receives the witness and points), keeps the recipe compute-free and deletes one of the three `read_raf_stage_values` spellings.

---
*Claude Code Fable 5 Review*

crates/jolt-prover/src/error.rs
```diff
@@ -0,0 +1,38 @@
...
+    #[error("{stage}: final claim {got} != expected {expected}")]
+    FinalClaimMismatch {
+        stage: &'static str,
+        expected: F,
+        got: F,
+    },
+
+    /// A capability the modular prover does not implement yet.
+    #[error("unsupported: {reason}")]
+    Unsupported { reason: &'static str },
```
0xAndoroid, line 39: `Unsupported` currently covers three classes: real capability gaps ("dominant advice not yet supported"), internal invariant violations ("witness did not produce the RdInc/RamInc commitments", stage-7 "carried no kernel"), and cross-stage shape mismatches. The spec makes `Unsupported` load-bearing for fallback composition — a constructor retrying another backend on `Unsupported` would silently retry on prover bugs. An `InvariantViolation` variant here and in `KernelError` keeps the recoverable meaning clean.

---
*Claude Code Fable 5 Review*

crates/jolt-prover/tests/byte_diff.rs
```diff
@@ -0,0 +1,1586 @@
+//! The stage-granular byte-diff harness against `jolt-prover-legacy`.
```
0xAndoroid, line 1: 1586 lines with the legacy pipeline, modular pipeline, preprocessing literal, and prove/verify tail each copy-pasted ~5× across the five modules; `advice_vars` re-implements `config.rs::advice_total_vars` with the same doc comment, and `commit_committed_program` rebuilds ~45 lines of production preprocessing logic because `CommittedProgramProverData` has no constructor. A `tests/byte_diff/` directory with shared `legacy_prove`/`modular_witness` helpers — and a real constructor in the library (the first committed-mode deployment needs it too) — would drop ~450 duplicated lines.

---
*Claude Code Fable 5 Review*

crates/jolt-verifier/src/stages/stage2/ram_output_check.rs
```diff
@@ -54,16 +54,16 @@ impl<F: Field> RamOutputCheck<F> {
     }
 }
 
-/// `(EqIoMask, NegEqIoMaskValIo)` at the produced output address point:
-/// `eq_io_mask = eq(output_address_challenges, addr) * range_mask(io, addr)`,
-/// and the negated `eq_io_mask * val_io` term that subtracts the committed
-/// public-IO contribution. Shared by the stage-2 clear path and the BlindFold
-/// statement builder so the algebra lives in one place.
+/// `(EqAddress, IoMask, ValIo)` at the produced output address point — one
```
0xAndoroid, line 58: The module doc at the top of this file still names the deleted `EqIoMask`/`NegEqIoMaskValIo` publics — stale after this rename to `(EqAddress, IoMask, ValIo)`.

---
*Claude Code Fable 5 Review*

crates/jolt-prover/tests/byte_diff.rs
0xAndoroid: I think this is irrelevant, since a test

0xAndoroid:
@markosg04 maybe https://me.andrew.ee/share/5f36957ebb902625/jolt-pr-1669.html will be helpful for you during your pass

0xAndoroid (approved):

PR #1673, opened 2026-07-14
feat(inlines): expose Blake2b salt and personalization to the guest

0xAndoroid:
## Summary

Exposes BLAKE2b salt and personalization to guest programs, mirroring the [`blake2`](https://crates.io/crates/blake2) crate's `new_with_params` semantics.

**New API** (`jolt-inlines-blake2`):
- `Blake2b::new_with_params(salt, persona)` — streaming hasher with parameter block
- `Blake2b::digest_with_params(salt, persona, input)` — one-shot, reuses the optimized no-copy digest path
- `SALT_SIZE_IN_BYTES` / `PERSONA_SIZE_IN_BYTES` constants

Semantics match the reference crate: salt and persona up to 16 bytes each, zero-padded, XORed into IV words 4–7 of the parameter block (RFC 7693 §2.5), panic on oversized values.

**Pure SDK-level change**: the inline compression instruction operates on an arbitrary state vector, so the tracer sequence builder, host exec, and prover are untouched. The unparametrized `new()`/`digest()` paths pay zero extra guest cycles.

## Testing

- 5 new tests cross-checking against `blake2::Blake2bMac512::new_with_salt_and_personal` (unkeyed): Zcash-style persona, block-boundary input lengths, incremental updates, empty-params equivalence with plain digest, and `should_panic` for oversized salt/persona
- Full `jolt-inlines-blake2` suite passes (20/20); clippy clean in host and no-host modes; `riscv64imac-unknown-none-elf` guest build clean

### DIFF
--- a/jolt-inlines/blake2/src/sdk.rs
+++ b/jolt-inlines/blake2/src/sdk.rs
@@ -1,5 +1,8 @@
 //! High-level Blake2b hashing API for host and guest modes.
-use crate::{BLOCK_INPUT_SIZE_IN_BYTES, IV, MSG_BLOCK_LEN, STATE_VECTOR_LEN};
+use crate::{
+    BLOCK_INPUT_SIZE_IN_BYTES, IV, MSG_BLOCK_LEN, PERSONA_SIZE_IN_BYTES, SALT_SIZE_IN_BYTES,
+    STATE_VECTOR_LEN,
+};
 const OUTPUT_SIZE: usize = 64;
 
 pub struct Blake2b {
@@ -21,6 +24,23 @@ impl Blake2b {
 
         Self {
             h,
+            buffer: [0; BLOCK_INPUT_SIZE_IN_BYTES],
+            buffer_len: 0,
+            counter: 0,
+        }
+    }
+
+    /// Creates a new hasher with the given salt and personalization,
+    /// matching the `blake2` crate's `new_with_params` semantics.
+    ///
+    /// Shorter values are zero-padded per the BLAKE2b specification.
+    ///
+    /// # Panics
+    /// Panics if `salt` or `persona` is longer than 16 bytes.
+    #[inline(always)]
+    pub fn new_with_params(salt: &[u8], persona: &[u8]) -> Self {
+        Self {
+            h: initial_state_with_params(salt, persona),
             buffer: [0; BLOCK_INPUT_SIZE_IN_BYTES],
             buffer_len: 0,
             counter: 0,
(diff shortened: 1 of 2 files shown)


jolt-inlines/blake2/src/sdk.rs
```diff
@@ -186,6 +221,32 @@ impl Blake2b {
...
+        persona.len() <= PERSONA_SIZE_IN_BYTES,
+        "persona must be at most 16 bytes"
+    );
+
+    let mut params = [0u8; SALT_SIZE_IN_BYTES + PERSONA_SIZE_IN_BYTES];
+    params[..salt.len()].copy_from_slice(salt);
+    params[SALT_SIZE_IN_BYTES..SALT_SIZE_IN_BYTES + persona.len()].copy_from_slice(persona);
+
+    let mut h = IV;
+    h[0] ^= 0x01010000 ^ (OUTPUT_SIZE as u64);
```
moodlezoup: `let mut h = IV; h[0] ^= 0x01010000 ^ (OUTPUT_SIZE as u64);` now appears in three places in this file (`new()`, `digest()`, and here). Consider a private `#[inline(always)] fn initial_state()` that all three call, with this function XORing the param words on top — the crate already has exactly that helper twice in test code (`sequence_builder.rs::initial_state`, `spec.rs::blake2b_initial_state`). Since everything involved is `#[inline(always)]`, the unparametrized paths stay zero-extra-cost.

---

moodlezoup (commented):
Reviewed for soundness, correctness, prover impact, and code quality (four parallel deep-review passes over the full file context).

No correctness or soundness issues found:
- The parameter-block math matches RFC 7693 §2.5 (salt → words 4–5, persona → words 6–7, little-endian, zero-padded) and is pinned against the RustCrypto `blake2` reference across 6 param combos × 13 input lengths.
- The "prover untouched" claim holds: the inline sequence builder loads the hash state from `rs1` as arbitrary memory and only bakes the IV into `v[8..16]`, which is state-independent per the spec — so a parametrized initial state is indistinguishable from any other `h` to the tracer/prover, and host/guest paths share the un-gated parameter code.
- The `digest_from_state` refactor is verbatim-preserving for the existing `digest()` path, and the oversized-salt/persona asserts are guest-side on programmer-supplied constants (no verifier-reachable panic).

One minor cleanup suggestion inline. Nit, not blocking: the doc comments say "matching the `blake2` crate's `new_with_params` semantics" but the API the tests (and users migrating from that crate) actually match is `new_with_salt_and_personal`.

---

PR #1675, opened 2026-07-15
feat: Akita lattice PCS integration — verifier path and shared semantics

markosg04:
Verifier side of the Akita (lattice PCS) integration.

## What

- **jolt-akita**: adapts to the akita runtime-ring API; one-hot group commits with
  native batch openings (owned commits + open-from-hint, so the prover never
  retains the witness); **jolt-owned schedule catalogs** — `JoltD64OneHotK16`/
  `JoltD64OneHotK256` configs delegate policy to the upstream presets and
  override the schedule-catalog and setup-sizing hooks, with tables generated
  offline by the `gen_jolt_schedules` binary through akita's own planner/emitter
  (952 K=16 keys + 108 K=256 keys, coverage and byte-exact drift tests
  included). Production setup envelopes are sized directly from the catalogs
  (~28 ms), replacing planner-DP scans.
- **jolt-claims**: completes the lattice semantics — the fused increment claim
  pipeline (four fused-inc consumer val stages inside the bytecode read-raf,
  the stage-7 lattice hamming-weight mega-reduction with the shifted decode
  identity), MSB as a uniform K×T one-hot column, the `OneHotTrace` native
  layout (`OneHotTraceLayout`, `one_hot_trace_columns`, protocol-owned blake2b
  layout digest), and an `akita` feature gating the lattice stage count.
- **jolt-openings**: the joint packed opening for auxiliary objects
  (`prove/verify_packed_openings`) and the `OneHotTraceCommitmentMetadata`/
  `OneHotTraceSetupMetadata` traits the verifier enforces before native
  openings (impls live in jolt-akita — jolt-verifier keeps zero akita
  references).
- **jolt-verifier**: the akita verifier sumchecks behind the `akita` feature —
  the nine-stage bytecode read-raf (which discharges the four reduced inc
  claims and produces the `FusedInc` opening), the hamming mega-reduction,
  the reconstruction phase, and a stage 8 that opens the
  native `OneHotTrace` one-hot group **directly at one common point**
  (`PCS::verify_batch`) after validating the commitment's layout digest,
  declared dimensions, backend flavor, and K regime; prefix packing remains
  only for advice byte objects and the precommitted program.
- **K=16/K=256 toggle**: akita follows the same `log_T = 25` one-hot chunk
  threshold as the base path (K=16 below, K=256 at/above); the regime is
  carried by the proof's one-hot config and bound by the layout digest.
- **Fixture-driven soundness suite**: real prover fixtures back an exhaustive
  typed sweep (every clear-claim wire offset individually — compile-enforced
  coverage via full destructuring), byte-level commitment sweeps (digest,
  dimensions, backend bytes), and proof-shape tampers; all akita
  tamper-manifest entries are Active.

## Protocol notes

- **The inc claims dissolve into the bytecode read-raf** (last commit): the
  earlier design ran a standalone `IncVirtualization` sumcheck between stages
  5 and 6a plus a stage-6b claim-moving member. Substituting
  `Store(j) = Σ_k val_store(k)·ra(k,j)` (bytecode ra is one-hot per cycle)
  into `FusedInc·Store = RamInc` / `FusedInc·(1−Store) = RdInc` turns each of
  the four reduced inc claims into a read-raf-shaped val stage — a bytecode
  store/¬store column, the producing relation's cycle point, the shared RA
  product, and one shared `FusedInc` cycle factor (cycle degree d+1 → d+2).
  Both the standalone stage and the 6b member disappear, no store-selector
  opening exists anywhere, and the proof drops a whole log_T-round phase
  (measured at 2^24: sumcheck payload 1114 → 1066 field elements, one fewer
  Fiat-Shamir stage) net of one extra coefficient per 6b round. The staged
  `BytecodeValStage` wires are unchanged (¬store derives as `1 − store`).
- `verify()` no longer takes a zk flag — mode selection is transitioning to a
  compile-time cfg approach; sdk macros, examples, and benches updated.
- Opening accumulators no longer absorb opening claims into Fiat-Shamir
  (prover and verifier consistent). Main's prover still round-trips through
  this verifier (full suite green), but transcripts differ from prior releases.
- jolt-transcript's length-prefixed absorber is named `append_length_prefixed`
  to stay distinguishable from the bare-label `Transcript::append_labeled`
  trait method (different consensus bytes).

## Dependencies

The akita workspace deps temporarily pin `markosg04/akita @ cb13b79f` = upstream
main (through the #302 quantum SIS cutover) plus the generic `D64OneHotK16`
preset, which is upstream as LayerZero-Labs/akita#308. On its merge the pin
moves back to upstream main (one-line follow-up).

## Performance (2^20 sha2-chain, same machine, back-to-back, median of 3)

Apple M4 Max (16 cores, 128 GB), release, default rayon threads; each phase is
a separate `Instant` timer in `sha2_chain_akita_perf` / `sha2_chain_dory_perf`
(#1676). The dory harness initializes the prepared pairing cache before the
timed verify, mirroring the production `setup_prover` path that test binaries
deliberately skip; serialization is excluded on both sides.

**Warm steady-state verification** (reusable verifier state prepared and
cached for both backends):

| | akita | dory |
|---|---:|---:|
| verify | **21.9 ms** | 60.0 ms |

**Cold end-to-end, per phase:**

| phase | akita | dory |
|---|---:|---:|
| PCS/commitment setup | 59.0 ms | — (inside prover preprocessing) |
| prover preprocessing | — (inside setup) | 1.04 s |
| prepared-pairing cache init | n/a | 208 ms |
| prove | **3.06 s** | 4.21 s |
| verifier preprocessing | 407 ms | 187 ms |
| verify | 21.9 ms | 60.0 ms |
| proof size | n/a* | 86,643 B (postcard) |
| peak RSS (prove+verify process) | 2.7 GB | 1.7 GB |

Akita proves ~1.37× and verifies (warm) ~2.7× faster at this shape.
*The akita verifier-native proof is not serde-serializable yet (the lattice
stage-claim types lack derives) — follow-up; its sumcheck payload is 1066
field elements (see protocol notes).

## Testing

- CI (this PR): a jolt-verifier `akita` clippy lane, the verifier akita suite
  in test-crates, and a dedicated `test-akita-schedules` job that regenerates
  the schedule catalogs through the planner DP. The `akita,prover-fixtures`
  combo cannot compile until the packed prover lands, so its clippy lane and
  the fixture-driven suites run on the stacked prover PR (#1676) as the
  `test-prover-akita` and `test-verifier-akita-fixtures` jobs — muldiv,
  forced-K256, advice, and committed-program e2es plus the typed clear-claim,
  commitment-byte, proof-shape, and presence tamper sweeps.
- Malformed-proof hardening: `verify_batch` validates the backend commitment
  coefficient count and the batched proof shape against the trusted resolved
  schedule before any shape-backed deserialization (see
  `crates/jolt-akita/src/shape_guard.rs`), with unit and wire-level
  malformed-shape tests.
- Local: full base + akita suites green across every crate.

Protocol design of the fused pipeline and native opening by Quang Dao
(quangvdao, #1683), re-expressed here; review feedback from moodlezoup
addressed in the `review:`-prefixed commits (transcript naming, claims-level
stage-count const with fixed-size arrays, one composed `verify_clear` with
String stage labels, and the empty-ra-product case retired as unreachable);
the staged-value array-sizing refactor and the read-raf dissolution land as
follow-up commits on this branch.


crates/jolt-transcript/src/legacy.rs
```diff
@@ -129,6 +129,23 @@ impl<F: CanonicalBytes> AppendToTranscript for F {
     }
 }
 
+/// Absorbs `label` (with the payload's transcript length when it exposes
+/// one) followed by the payload itself. The single home of the labeled-append
+/// convention shared by the prover's and verifier's commitment absorption —
+/// order and labels are consensus-critical.
+pub fn append_labeled<T, A>(transcript: &mut T, label: &'static [u8], payload: &A)
```
moodlezoup: This free fn now shares a name with the pre-existing `Transcript::append_labeled` trait method (line 51) in the same module, but produces different transcript bytes: the trait method always emits a bare `Label`, while this one emits `LabelWithCount` when the payload exposes `transcript_payload_len()`. Both are live at call sites (`transcript.append_labeled(b"opening_claim", ...)` vs `append_labeled(transcript, b"commitment", ...)`), so a refactor that switches between method-call and free-fn syntax silently changes consensus-critical bytes. The old name `append_payload_label` (or something like `append_length_prefixed`) kept them distinguishable — consider restoring a distinct name.

---

crates/jolt-claims/src/protocols/jolt/lattice/strategy.rs
```diff
@@ -0,0 +1,160 @@
...
+                    .iter()
+                    .map(|(polynomial, _slot)| *polynomial)
+                    .collect(),
+                member_arity: shape.log_k_chunk + shape.log_t,
+            },
+        })
+    }
+
+    /// The commitment-object setup shape.
+    pub fn setup_shape(&self, shape: &ProofPackingShape) -> Result<WJoltSetupShape, OpeningsError> {
```
moodlezoup: `setup_shape` and `WJoltSetupShape` have no callers in this PR (only the re-export in `lattice/mod.rs`) — setup shaping is prover-side and the prover is out-of-tree. Consider dropping them until the prover change that consumes them lands, so the seam only carries what's exercised.

---

crates/jolt-transcript/src/digest.rs
```diff
@@ -79,6 +79,15 @@ where
     D: Digest<OutputSize = U32>,
     F: jolt_field::TranscriptChallenge,
 {
+    /// Raw multi-byte squeeze backing jolt-prover-legacy's challenge
+    /// decoding, so its legacy `Transcript` vocabulary can drive this engine
+    /// directly (no state handoffs). Hidden: protocol code squeezes through
+    /// the [`Transcript`] challenge methods.
+    #[doc(hidden)]
+    pub fn raw_challenge_bytes(&mut self, out: &mut [u8]) {
```
moodlezoup, line 87: `raw_challenge_bytes` has zero callers in the workspace, including jolt-prover-legacy which the doc says it exists for. Dead `pub` surface on the transcript engine — drop it until the caller exists.

---

crates/jolt-verifier/src/stages/stage8/homomorphic.rs
```diff
@@ -0,0 +1,233 @@
...
+            ),
+            JoltCommittedPolynomial::ProgramImageInit => precommitted_entry(
+                committed_program.map(|committed| &committed.program_image_commitment),
+                precommitted_final(polynomial),
+                polynomial,
+                opening_point,
+            ),
+            // Lattice-mode polynomials open through the packed opening
+            // (`lattice::packing::final_opening`), never the homomorphic
+            // stage 8 RLC batch.
```
moodlezoup: `lattice::packing::final_opening` was removed by this PR; lattice-mode polynomials now open through `stage8::packed` / `verify_packed_openings`.

```suggestion
            // Lattice-mode polynomials open through the packed opening
            // (`stage8::packed` / `verify_packed_openings`), never the
            // homomorphic stage 8 RLC batch.
```

---

moodlezoup (commented):
Reviewed with focus on soundness, prover/verifier consistency, Fiat-Shamir ordering, and the default-path rewrites. No soundness or correctness issues found: the packed-opening reduction (padding factors verifier-computed, evaluations bound by both the sumcheck output claim and PCS openings, challenges drawn after statement absorption), the reconstruction presence checks (fail-closed in both directions), the γ-power placement in the new lattice relations, and the default-path `HomomorphicBatch` reroute (byte-for-byte transcript-equivalent to the removed hand-rolled RLC) all check out. The four inline comments are maintainability items — the `append_labeled` name collision on consensus-critical transcript helpers being the one most worth addressing.

One observation, already implied by the PR description: with the prover out-of-tree, the akita verifier relations are pinned only by relation-level unit tests; the msb-reduction and bytecode leg-embedding surfaces have no end-to-end prove→verify coverage until the prover lands.

---

Cargo.toml
```diff
@@ -385,7 +385,7 @@ riscv = { version = "0.16", default-features = false }
 alloy-primitives = "1"
 alloy-consensus = "2"
 alloy-eips = "2"
-reth-primitives-traits = { version = "0.5", features = ["rayon", "serde"] }
+reth-primitives-traits = { version = "0.4", features = ["rayon", "serde"] }
```
moodlezoup: this looks accidental

crates/jolt-verifier-derive/src/lib.rs
```diff
@@ -279,11 +279,22 @@ fn expand(input: DeriveInput) -> syn::Result<TokenStream2> {
...
-    // stages' choice. Falls back to the first member if every member is optional.
-    let stage_id_ident = plans
-        .iter()
-        .find(|plan| !plan.is_option)
-        .map_or(&plans[0].ident, |plan| &plan.ident);
+    // stages' choice. An all-optional batch has no member to project through, so it
+    // must name its label explicitly via `#[sumcheck_batch(stage_id = ..)]`.
+    let stage_id_expr = if let Some(plan) = plans.iter().find(|plan| !plan.is_option) {
+        let id = &plan.ident;
+        quote!(self.#id.id())
```
moodlezoup: tbh `stage_id_expr` was already a misnomer here. We should change the `stage` fields in StageClaimOpeningMismatch and StageClaimSumcheckFailed to be `String` and just use `#base` instead

crates/jolt-verifier-derive/src/lib.rs
```diff
@@ -465,6 +475,53 @@ fn expand(input: DeriveInput) -> syn::Result<TokenStream2> {
         }
     };
 
+    // The composed clear-path driver: batched-verify, derive the produced
+    // opening points, and check the reduced claim against the expected
+    // final-claim fold — the tail every stage repeats verbatim. Validation
+    // (`validate_output_claims`, member-presence guards) and the canonical
+    // opening absorb stay with the caller: their position and form are
+    // stage-specific.
+    let run_clear_method = quote! {
```
moodlezoup: Instead of having both verify_clear and run_clear, let's rename this to verify_clear and get rid of the existing verify_clear. I think we can inline expected_final_claim here and remove it from the public API

crates/jolt-claims/src/protocols/jolt/geometry/claim_reductions/bytecode.rs
```diff
@@ -376,6 +376,9 @@ pub struct BytecodeOutputWeightInputs<'a, F> {
 /// whose same-named fields take the address components alone).
 pub struct BytecodeLaneWeightInputs<'a, F> {
     pub eta: F,
+    /// The staged-val count folded: five in base mode, six in lattice mode
+    /// (the sixth is the raw store circuit-flag lane, no gamma fold).
+    pub num_val_stages: usize,
```
moodlezoup: instead of this, perhaps we should just do:
```rust
#[cfg(not(feature = "akita"))]
const NUM_BYTECODE_VAL_STAGES: usize = 5;
#[cfg(feature = "akita")]
const NUM_BYTECODE_VAL_STAGES: usize = 6;
```

crates/jolt-verifier/src/verifier.rs
```diff
@@ -163,34 +155,142 @@ where
...
+        &proof.stages.inc_virtualization_sumcheck_proof,
+        &proof.clear_claims()?.inc_virtualization,
+        &mut transcript,
+        stage2.clear()?,
+        stage4.clear()?,
+        stage5.clear()?,
+    )?;
+    let stage6a = stage6a::verify(
+        &checked,
+        proof,
```
moodlezoup: this is confusing; inc_virtualization should really be its own stage, no?

crates/jolt-claims/src/protocols/jolt/relations/ram/ra_virtualization.rs
```diff
@@ -132,4 +132,36 @@ mod tests {
             ra_virtualization_dimensions(3).num_committed_ra_polys() + 1
         );
     }
+
+    /// With zero committed `ra` polynomials the output product is the bare
+    /// cycle-`eq` public; the relation stays well-formed.
+    #[test]
+    fn ra_virtualization_supports_empty_ra_product() {
```
moodlezoup: I feel like we should just reject this case in `validate_inputs`

crates/jolt-openings/src/packing.rs
```diff
@@ -473,82 +496,796 @@ where
...
+    let coefficients = prepared
+        .iter()
+        .map(|_| transcript.challenge_scalar())
+        .collect();
+    (alphas, coefficients)
+}
+
+/// Proves a joint packed opening: one reduction sumcheck across all objects,
+/// then one native PCS opening per object group.
+pub fn prove_packed_openings<PCS, Id, T>(
```
moodlezoup, line 965: we'll want to migrate this to jolt-prover/jolt-kernels once those have landed

crates/jolt-verifier/src/stages/stage6b/bytecode_read_raf.rs
```diff
@@ -51,7 +54,7 @@ pub struct BytecodeReadRafTableFoldInputs<'a, F: Field> {
 pub struct BytecodeReadRafCycleInputs<'a, F: Field> {
     pub dimensions: BytecodeReadRafDimensions,
     pub r_address: Vec<F>,
-    pub stage_cycle_points: [Vec<F>; 5],
+    pub stage_cycle_points: Vec<Vec<F>>,
```
moodlezoup: Instead of changing `[T; 5]` to `Vec<T>` in this file, I think we can just do `[T; BYTECODE_VAL_STAGES]`

crates/jolt-verifier/src/stages/stage7/hamming_weight_claim_reduction.rs
```diff
@@ -48,13 +48,10 @@ pub fn hamming_weight_input_values_from_upstream<F: Field>(
 /// virtualization opening point.
 pub fn stage7_hamming_virtualization_address_points<F: Field>(
     dimensions: HammingWeightClaimReductionDimensions,
-    stage6_points: &Stage6bOutputPoints<F>,
+    instruction_ra_points: &[Vec<F>],
+    bytecode_ra_points: &[Vec<F>],
+    ram_ra_points: &[Vec<F>],
```
moodlezoup: i slightly prefer to keep this function signature as is

crates/jolt-claims/src/protocols/jolt/lattice/relations/read_raf.rs
moodlezoup, line 1: I feel like it might be better to introduce the `akita` feature flag to `jolt-claims`, and use `#[cfg(feature = "akita")]` in the main `read_raf_address_phase.rs` instead of adding a new one here. i don't feel strongly though

crates/jolt-claims/src/protocols/jolt/lattice/geometry.rs
```diff
@@ -157,6 +159,26 @@ pub fn selector_block_weight<F: Field>(
         .sum()
 }
 
+/// The six-stage public values of the lattice read-raf cycle output: the
+/// five base stage values plus the store-flag val bound to the
+/// `IncVirtualization` cycle point.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct LatticeBytecodeReadRafPublicValues<F: Field> {
```
moodlezoup: this type seems unnecessary; we can inline this logic in `BytecodeReadRaf::expected_output`

crates/jolt-verifier/src/stages/stage6b/bytecode_read_raf.rs
```diff
@@ -70,13 +70,15 @@ pub struct BytecodeReadRaf<F: Field> {
...
-    /// The address-only bytecode-table fold: `Σ_row row_values[stage] *
-    /// eq(r_address, row)` — the pre-cycle half of `read_raf_public_values`'
-    /// `stage_values`. `None` in ZK, where `expected_output` never runs.
-    stage_values_at_r_address: Option<[F; 5]>,
+    /// The address-only bytecode-table fold: the five per-stage row values and
+    /// the store flag, each folded against `eq(r_address, row)` — the pre-cycle
+    /// half of the read-raf publics (the store half feeds the packed sixth
+    /// stage; the base output ignores it). `None` in ZK, where
+    /// `expected_output` never runs.
+    stage_values_at_r_address: Option<([F; 5], F)>,
```
moodlezoup:
```suggestion
    stage_values_at_r_address: Option<[F; NUM_BYTECODE_VAL_STAGES])>,
```

crates/jolt-claims/src/protocols/jolt/geometry/bytecode.rs
```diff
@@ -156,10 +189,11 @@ impl<F: Field> BytecodeReadRafPublicValues<F> {
 
 /// Committed-program read-RAF publics: the bytecode table is not available,
 /// so only the table-independent factors are computed. The per-stage Val
-/// factors are openings; their cycle-eq coefficients are public.
+/// factors are openings; their cycle-eq coefficients are public. Five staged
+/// vals in base mode, six in lattice mode (the store stage).
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct BytecodeReadRafCommittedPublicValues<F: Field> {
-    pub stage_cycle_eqs: [F; NUM_BYTECODE_VAL_STAGES],
+    pub stage_cycle_eqs: Vec<F>,
```
moodlezoup: [F; NUM_BYTECODE_VAL_STAGES]

crates/jolt-claims/src/protocols/jolt/geometry/bytecode.rs
```diff
@@ -183,7 +217,9 @@ impl<F: Field> BytecodeReadRafCommittedPublicValues<F> {
 pub struct BytecodeReadRafCommittedEvaluationInputs<'a, F> {
     pub r_address: &'a [F],
     pub r_cycle: &'a [F],
-    pub stage_cycle_points: [&'a [F]; NUM_BYTECODE_VAL_STAGES],
+    /// One cycle point per staged val (five base, six lattice — the sixth is
+    /// the `IncVirtualization` store point).
+    pub stage_cycle_points: Vec<&'a [F]>,
```
moodlezoup: [&'a [F]; NUM_BYTECODE_VAL_STAGES]

crates/jolt-verifier/src/stages/stage6b/bytecode_read_raf.rs
```diff
@@ -97,14 +102,16 @@ impl<F: Field> BytecodeReadRaf<F> {
...
+/// store-flag fold (the lattice sixth stage's address half; the base relation
+/// ignores it): the bytecode rows' per-stage values (shared
+/// `read_raf_stage_values` formula) folded against `eq(r_address)`. The
+/// cycle-eq factors are attached later, at `expected_output` time, so the fold
+/// can run before the cycle sumcheck.
 fn fold_stage_values<F: Field>(
     r_address: &[F],
     fold: BytecodeReadRafTableFoldInputs<'_, F>,
-) -> Result<[F; 5], VerifierError> {
+) -> Result<([F; 5], F), VerifierError> {
```
moodlezoup:
```suggestion
) -> Result<[F; NUM_BYTECODE_VAL_STAGES], VerifierError> {
```

crates/jolt-claims/src/protocols/jolt/geometry/bytecode.rs
```diff
@@ -385,7 +434,7 @@ fn read_raf_row_values<F>(
     stage3_gammas: &[F],
     stage4_gammas: &[F],
     stage5_gammas: &[F],
-) -> [F; 5]
+) -> ([F; 5], bool)
```
moodlezoup:
```suggestion
) -> [F; NUM_BYTECODE_VAL_STAGES]
```

quangvdao:
Documentation follow-up for the latest `IncVirtualization` dissolution (`13db2e023`): `specs/lattice-claims.md` still describes the superseded protocol and should be updated with this PR.

The stale parts include:

- **Scope / module inventory / typed IDs:** still list the standalone `IncVirtualization` relation and the Stage-6b `FusedIncClaimReduction`, both now deleted.
- **“The Inc One-Hotting Relation Chain” §§1–2:** still specify the old cubic `log_T`-round virtualization, its `FusedInc` + `OpFlags(Store)` outputs, and the second point-alignment sumcheck.
- **Store binding (§5):** still says the sixth staged value carries `OpFlags(Store)@IncVirtualization`. In the final design the raw Store bytecode value remains the sixth staged-value wire, but the four fused consumer relations reuse Store / `1 − Store`; there is no Store-selector opening.
- **Stage 7:** still consumes `FusedInc@FusedIncClaimReduction`; it now consumes `FusedInc@BytecodeReadRaf`.
- **Verifier/prover schedule:** still places a standalone phase between Stages 5 and 6a and a fused-inc reduction member in 6b.

The replacement description should capture the load-bearing substitution

```text
Store(j) = Σ_k val_store(k) · RA(k,j)
RamInc(j) = FusedInc(j) · Store(j)
RdInc(j)  = FusedInc(j) · (1 − Store(j))
```

and the resulting four Read-RAF consumer stages at `γ^5..8`:

```text
(store, RAM read/write point)
(store, RAM val-check point)
(1 − store, register read/write point)
(1 − store, register val-evaluation point)
```

It would also help to make the two counts explicit, since they are easy to conflate:

- `READ_RAF_CYCLE_STAGES = 9` in Akita mode (5 base + 4 fused consumers);
- `NUM_BYTECODE_VAL_STAGES = 6` (5 base staged values + raw Store), so committed mode adds no new staged-value wires.

The current implementation-level explanation in `crates/jolt-claims/src/protocols/jolt/lattice/relations/read_raf.rs` is a good basis for the spec rewrite. PR #1676 carries the corresponding packed-prover implementation.

I did not find another correctness issue in this focused Stage-5→6 protocol pass. One historical detail worth avoiding in updated prose: the immediately preceding prover perf commit optimized `FusedIncClaimReduction`, but the final dissolution commit deletes that relation, so that optimization is not part of the final architecture.

quangvdao:
Requesting changes before merge. I reviewed the full merge-base diff at exact head `b6779348c79fb4bae043a004d773caa2204c1920`, including the claims algebra, transcript ordering, packed-opening reduction, Akita adapter, stage 6/7 fused-increment path, stage 8 verifier, generated schedules, CI, and the stacked prover PR.

The protocol design is coherent and I did **not** find an algebraic soundness error in the fused-increment identities, the packed-opening reduction, the common-point mapping, optional-object presence checks, or the Fiat–Shamir challenge ordering. However, I think the following items need to be resolved before merge.

## 1. The `2^20` benchmark must be rerun apples-to-apples

The currently reported `36.8 ms Akita verify vs 610 ms Dory verify` comparison is not a valid backend comparison. The PR notes that the Dory harness does not use its prepared pairing cache, while Akita's reported number is “pure verify” and excludes setup plus roughly `0.6 s` of verifier preprocessing. That compares a cold/unprepared Dory path against a prepared Akita path and materially exaggerates the speedup.

Please rerun the `2^20` measurements before merge with identical lifecycle boundaries for both backends. At minimum, report two separate tables:

1. **Cold end-to-end:** setup/preprocessing, commitment/proof deserialization, backend preparation, and verification included for both.
2. **Warm steady-state verification:** all reusable verifier state prepared and cached for both, with only per-proof verification timed.

The harness should also hold constant:

- exact program, padded trace length, public I/O, and proof configuration;
- release profile, target CPU/Rust flags, thread count, and machine;
- whether witness generation, commitment, opening, serialization, and preprocessing are inside or outside each timer;
- number of warmups/samples and reported statistic (at least median and spread, preferably p95);
- proof size and peak RSS alongside time;
- the same cache reuse policy across repeated proofs.

The proving comparison (`~7.79 s vs 13.64 s`) should receive the same boundary audit: setup and preprocessing must be excluded or included symmetrically. Until then, please do not present `36.8 ms vs 610 ms` or the implied verifier speedup as an apples-to-apples result.

## 2. Proof-controlled Akita shapes can cause large verifier allocations

`AkitaCommitment::backend_coeff_len` and the serialized Akita proof shape are proof-controlled, and `verify_batch` passes them directly into Akita deserialization before deriving or checking the exact expected shapes from the trusted schedule:

- `crates/jolt-akita/src/adapters.rs` (`backend_coeff_len`, `serialized_akita_proof_shape`)
- `crates/jolt-akita/src/native_batching.rs:403-410`

In the pinned Akita revision, `RingVec::deserialize_with_mode` reserves `num_coeffs` elements before reading the first coefficient. Its generic validated cap is `2^25`; for the 128-bit field, one attacker-selected length can request roughly **512 MiB**, even if the serialized byte buffer is empty and decoding then fails at EOF. Proof-shape fields can trigger analogous shape-backed allocations.

This is an availability/resource-exhaustion issue rather than a proof-forgery issue: a malicious prover can construct the preceding Jolt proof under its chosen transcript, then make the final PCS payload fail only after the verifier allocates.

Please derive the exact commitment coefficient count and backend proof shape from the trusted verifier setup/resolved schedule, reject mismatches before allocation, require exact byte consumption, and add malformed-shape tests that return an error without allocation proportional to proof metadata. A protocol-specific bound is preferable to Akita's generic `2^25` sequence cap.

## 3. Current CI does not exercise Akita prove -> verify or tamper rejection

Under `feature = "akita"`, #1675 compiles out the completeness tests and every substantive commitment/opening/sumcheck tamper module:

- `crates/jolt-verifier/tests/completeness/mod.rs`
- `crates/jolt-verifier/tests/soundness/tampering/mod.rs`

The workflow comment says the `akita,prover-fixtures` suite is linted, but there is no such command. The only added lane runs:

```text
cargo nextest run -p jolt-verifier --features akita --no-tests=pass
```

I ran that exact command locally on the PR head: `77 passed, 9 skipped`. Those are relation/unit/manifest tests; there is no valid Akita proof acceptance test and no active Akita opening/sumcheck tamper test.

The stacked prover PR #1676 contains the missing Akita fixture/tamper source files, but its current workflow also lacks the claimed `test-prover-akita` job and the `akita,prover-fixtures` clippy/test commands. Please make the combined lane real and required before either half merges. It should cover at least:

- `muldiv` under Akita;
- trusted and untrusted advice;
- committed program;
- forced K=256;
- verifier-native valid fixture acceptance;
- commitment metadata, final evaluation, fused-inc, chunk/MSB, auxiliary-object, proof-byte, and malformed-deserialization tampers.

For a new cryptographic verifier path, author-reported local runs are not sufficient integration evidence.

## 4. The generated-schedule drift test is broken

The documented ignored test currently fails:

```text
cargo nextest run -p jolt-akita catalogs_match_planner --run-ignored all --cargo-quiet
```

It compares the planner's unformatted emitted string byte-for-byte against rustfmt-formatted checked-in source, so it reports K=16 catalog drift even when the schedules match. I reran a whitespace-normalized diagnostic comparison; both K=16 and K=256 catalogs passed in `183.6 s`.

So I found no structural schedule drift, but the test oracle needs to be fixed. Please format regenerated output before comparison or compare parsed schedule structures, then run it in a dedicated CI lane. A permanently ignored test that deterministically fails does not protect the 7,031 generated schedule lines.

## 5. Treat the Akita pin as an explicit protocol snapshot

All Akita crates are pinned to exact SHA `markosg04/akita@cb13b79`. The immutable SHA is good, but upstream LayerZero-Labs/akita#308 remains open and is now eight commits beyond this pin, including breaking layout/security-policy work and verifier correctness/performance fixes.

This is no longer just a temporary pointer waiting for a routine upstream merge. Please document which upstream fixes are intentionally excluded, who owns maintenance of this snapshot, and how/when Jolt will rebase its schedules and proof format. An organization-owned fork or explicit vendoring/reproducibility policy would be preferable if this snapshot is expected to persist.

## 6. Smaller cleanup items

- `jolt-transcript::raw_challenge_bytes` is public and has no workspace callers; remove it until the prover needs it.
- `specs/lattice-claims.md` still says `Status: draft` and `PR: TBD`; update its status and PR link if it is meant to be normative.
- The PR includes roughly 8,047 lines of fuzz `Cargo.lock` churn, which significantly obscures the functional review surface. Please confirm it is intentional or split/drop unrelated lockfile updates.

## What looks good

The major security/correctness choices are well structured:

- the verifier recomputes a canonical typed `OneHotTrace` layout digest;
- exact K/arity/column-count/setup metadata is checked;
- all mapped OneHotTrace leaves must share one canonical point;
- packed-opening statements are absorbed before alpha/beta;
- mixed-arity objects receive the required `2^padding` input factor;
- the reduced evaluation is checked and independently PCS-opened;
- advice/program optionality is fail-closed in both directions;
- the Jolt-to-Akita bridge challenge binds the backend statement to the outer transcript;
- the fused-inc consumer powers and the stage-7 Booleanity/Hamming/decode relation are synchronized.

Once the benchmark is made apples-to-apples, the deserialization bounds are fixed, and real Akita e2e/tamper coverage is required in CI, I think the underlying protocol organization is in good shape to merge.

moodlezoup (approved):
lgtm pending quang's comments

PR #1676, opened 2026-07-15
feat: Akita lattice PCS integration — jolt-prover-legacy packed prover

markosg04:
Prover half of the Akita integration, on top of #1675 — the diff shows both
halves until #1675 merges; review only the jolt-prover-legacy commit here.

## What

- **zkvm/packed.rs**: the packed prove pipeline — `OneHotTrace` one-hot
  witness assembly and owned group commit, the fused increment claim path
  (four fused-inc consumer val stages inside the nine-stage bytecode
  read-raf, then the stage-7 lattice hamming mega-reduction), advice byte
  objects, the precommitted program one-hot, and a stage 8 that opens
  `OneHotTrace` natively at its common point from the commit hint (no
  retained witness) with the auxiliary objects going through the joint
  packed opening.
- **zkvm/packed_witness.rs** + claim reductions: witness assemblers (every
  column, including the MSB, is a full K×T one-hot grid) and the
  reconstruction phase producing per-column leaf claims.
- **transcripts/verifier_native.rs**: legacy-trait facade over
  jolt-transcript so the prover drives the verifier-native FS stream.
- **field/akita.rs**: Fp128 `JoltField` implementation.
- Features: `akita` wires the backend deps and follows the shared
  K=16/K=256 trace-length toggle (`log_T = 25`).

## Notes

- e2e coverage: muldiv / advice / committed-program / forced-K256 over
  Akita, with live claim-wire tampers (fused-inc reduced claim, chunk and
  MSB finals) and per-object tamper checks; the verifier-side
  fixture-driven tamper suite (see #1675) generates its fixtures through
  this prover.
- Perf at 2^20 sha2-chain, same machine back-to-back: **prove ~7.8-8.0 s vs
  dory 13.64 s (~1.7× faster)**; verify 35 ms vs 610 ms. Setup is **0.16 s**
  (production envelopes sized from the checked-in schedule catalogs — the
  former ~83 s planner/family-init cost is gone); verifier preprocessing
  ~0.6 s, timed separately by the harness.
- The recent commits: a prover perf pass (lazy one-hot chunk columns via
  `RaPolynomial`, shared parallel pushforwards, Gruen split-eq fused
  reduction — 2^24 packed prove 156.6 s → 150.3 s, ~4 GiB less peak
  memory); the read-raf dissolution of the standalone `IncVirtualization`
  phase and the stage-6b fused-inc member (details in #1675; proof drops a
  full log_T-round stage, measured 1114 → 1066 sumcheck field elements at
  2^24); and the #1685 follow-ups re-expressed over both (the
  `OneHotTrace` naming, streamlined owned commits opening from the hint,
  and catalog-driven setup sizing — authored by @quangvdao).

## Testing

- CI: clippy lanes for prover `host,akita`; the `test-prover-akita` job
  runs the full suite under akita plus the verifier fixture suite.
- Local: full prover suite green under host, host,zk, and host,akita;
  all four akita e2es; base muldiv both modes.

Protocol design by Quang Dao (quangvdao, #1683), re-expressed per the
distilled commit series on the integration branch.


crates/jolt-prover-legacy/src/zkvm/bytecode/read_raf_checking.rs
moodlezoup, line 1: there seems to be a lot of runtime-gated (e.g. via `self.fused_inc.is_some()`) split codepaths to handle Akita in this file, and I suspect it slightly regresses the performance of the Dory path. This crate already has an `akita` feature, why not use that?

crates/jolt-prover-legacy/src/zkvm/claim_reductions/bytecode_reconstruction.rs
```diff
@@ -0,0 +1,806 @@
...
+}
+
+#[cfg(feature = "prover")]
+impl<F: JoltField> BytecodeReconstructionSumcheckProver<F> {
+    /// Builds the per-chunk leg tables from the materialized bytecode. Row
+    /// domain per chunk is `2^log_rows` (bytecode rows, zero-padded); the
+    /// byte lanes encode padding rows as byte-0 hot, matching the committed
+    /// witness.
+    #[tracing::instrument(skip_all, name = "BytecodeReconstructionSumcheckProver::initialize")]
+    pub fn initialize(
```
moodlezoup, line 216: this function does many serial passes over the bytecode, which will be slow for large guest programs

crates/jolt-prover-legacy/src/zkvm/packed_witness.rs
```diff
@@ -0,0 +1,758 @@
...
+            })
+            .sum()
+    }
+
+    fn for_each_row(&self, sigma: usize, f: &mut dyn FnMut(usize, &[F])) {
+        let row_len = 1usize << sigma;
+        let num_rows = 1usize << (self.num_vars - sigma);
+        let mut row = vec![F::zero(); row_len];
+        let mut positions = self.one_positions.clone();
+        positions.sort_unstable();
```
moodlezoup: should we just sort `one_positions` once up front in `SparseUnitPolynomial::new`?

crates/jolt-verifier/tests/soundness/tampering/mod.rs
```diff
@@ -1,8 +1,17 @@
...
 pub mod manifest;
+#[cfg(not(feature = "akita"))]
 pub mod openings;
+#[cfg(not(feature = "akita"))]
 pub mod preamble;
+#[cfg(not(feature = "akita"))]
 pub mod proof_shape;
+#[cfg(not(feature = "akita"))]
 pub mod sumcheck;
+#[cfg(not(feature = "akita"))]
```
moodlezoup, line 21: are these all intentionally cfg'd out?

crates/jolt-prover-legacy/Cargo.toml
```diff
@@ -24,6 +24,18 @@ keywords = ["SNARK", "cryptography", "proofs", "zkVM"]
...
+# the fp128 lattice field and commits packed one-hot witness objects.
+# Mutually exclusive with `zk` — one PIOP per compiled prover (the
+# jolt-verifier build errors on the combination).
+akita = [
+  "dep:akita-config",
+  "dep:akita-field",
+  "dep:jolt-akita",
+  "jolt-claims?/akita",
+  "jolt-field/akita",
+  "jolt-verifier/akita",
```
moodlezoup, line 37: why the `?` for jolt-claims but not the other deps?

.github/workflows/rust.yml
```diff
@@ -288,6 +361,23 @@ jobs:
           done
       - name: Test lookup table prover ABI
         run: cargo nextest run -p jolt-lookup-tables --features prover-abi-tests --test prover_lookup_table_abi --cargo-quiet
+      - name: Test jolt-verifier with akita feature
+        run: cargo nextest run -p jolt-verifier --features akita --no-tests=pass --cargo-quiet
+
+  test-akita-schedules:
```
moodlezoup, line 374: we have "Test crates with field-inline" under this job, which belongs under `test-crates`

moodlezoup (changes requested):

moodlezoup:
**Untrusted-advice scaling: a sumcheck memory cliff and ~0.5 s/prove of avoidable setup**

Both issues below scale with `max_untrusted_advice_size`, which is a default (4096, `common/src/constants.rs:29`), not a cap — guests can raise it via the memory config, and the only advice e2e runs 32 bytes. Everything here fails slow, not unsound, but the degradation is silent. Numbers are from the PR branch, release build (methodology at the bottom); absolute times are machine-relative.

**1. The untrusted reconstruction sumcheck densifies the full cell domain**

Unlike the trusted-advice and program-image instances — which pre-fold the word dimension into the init scatter and bind a fixed 11 rounds — the untrusted instance must bind the full `(byte ‖ place ‖ word)` domain (`11 + word_vars` rounds), because its booleanity leg is nonlinear in the column and its hamming leg quantifies over every row; neither commutes with partial evaluation at `r_word`.

The cost driver is `k_bool = EqPolynomial::evals(&r_reference)` (`advice_bytes.rs:164`): a dense field-element table over all `2^(11 + word_vars)` cells (= 256·S cells at advice size S), re-`bind_parallel`'d every round, alongside the dense column table:

| `max_untrusted_advice_size` | cell domain | `k_bool` table (fp128) | rounds |
|---|---|---|---|
| 4 KB (default) | 2^20 | 16 MB | 20 |
| 64 KB | 2^24 | 256 MB | 24 |
| 1 MB | 2^28 | **4 GiB** | 28 |

At the default it's tens of ms and fine; at MB-scale it's memory-disqualifying. For contrast, the Dory path stores advice as an S/8-coefficient word polynomial — this cliff is lattice-path-specific (the byte decomposition inflates the domain 2048×).

Suggested fix, structural: the untrusted byte column is a strict one-hot column (K=256 byte lanes over `8·words` rows) — exactly the shape the lattice booleanity/hamming machinery in this PR already handles with K-sized pushforwards, Gruen split-eq, and lazy index columns. Routing the booleanity + hamming legs through that machinery (and pre-folding the decode leg like the trusted instance) makes the whole thing O(S) with small constants. Cheaper local fix: split-eq for `k_bool` plus one-hot-aware sparse early rounds, which kills the memory blowup. Interim guardrail either way: assert or document a max advice size in lattice mode, and add an e2e at the 4 KB max so the boundary is exercised.

**2. Every prove with untrusted advice rebuilds the advice object's backend setup — and ~92% of it is a flavor the object never uses**

`commit_advice_one_hot` calls `transparent_object_setup(cell_vars)` inline (`packed.rs:440-441`), and untrusted advice commits during proving, so this runs per prove. Measured (setup + commit for the singleton object, at full one-position density):

| `max_untrusted_advice_size` | setup + commit | of which commit |
|---|---|---|
| 4 KB (default) | **468 ms** | ~5 ms |
| 64 KB | 859 ms | ~40 ms |
| 1 MB | 1.96 s | tens of ms |

Against the ~8.2 s prove at 2^20 that's ~6% for any advice-using guest at the *default* shape (~20% of a 2^18 prove). Two compounding causes:

- `AkitaScheme::setup` builds both backend flavors: the full-flavor setup the sparse-unit advice commit actually uses (~12 ms — SIS matrix expansion) *and* the one-hot flavor setup whenever `max_num_vars >= one_hot_log_k`, i.e. always (`scheme.rs:340-352`) — ~433 ms of the 468 ms, and the advice object never commits through it. `AkitaSetupParams::one_hot_only` exists for OneHotTrace; there's no inverse `full_only` for singleton objects.
- Nothing caches the result: upstream's setup persistence is behind a `disk-persistence` feature jolt-akita doesn't enable, there's no in-memory memo (only the sizing envelope is memoized), and the setup depends only on `(cell_vars, width=1, zero seed)` — i.e. on `max_untrusted_advice_size`, which is preprocessing-time data.

Suggested fixes, both plumbing: (a) add a `full_only` setup-params variant (mirror of `one_hot_only`) — cuts the per-prove cost to ~15 ms by itself; (b) hoist the advice-object setup into prover preprocessing keyed on `max_untrusted_advice_size`. The verifier side already re-derives the same setup from public shape in `akita_verifier_preprocessing`, so neither changes the protocol.

(The sparse-unit *commit* itself is not a concern: the backend's `column_sweep_sparse` does one shift-accumulate per (SIS-row × one-position), so it scales with the ones, not the domain — consistent with the 5→40 ms measurements.)

*Methodology: timed `commit_trusted_advice_one_hot(&advice, max)` — the identical `commit_advice_one_hot` path the prover runs for untrusted advice — across `max` sizes in a release build on this branch, with `tracing` span-close timings to split setup from commit; the padding encoding means one-position count equals `max` regardless of actual advice length, so commit was measured at full density.*

crates/jolt-prover-legacy/src/zkvm/bytecode/read_raf_checking.rs
markosg04: Done in `8b01c4e55`; the base build is branch-free.

crates/jolt-prover-legacy/src/zkvm/claim_reductions/bytecode_reconstruction.rs
markosg04: Fixed in `f14073bef`.

crates/jolt-prover-legacy/src/zkvm/packed_witness.rs
markosg04: Done in `23a2142f3`.

crates/jolt-verifier/tests/soundness/tampering/mod.rs
markosg04: Intentional — those tamper the dory proof shape; akita has its own suite (`tampering/akita.rs`). Comment added in `f14073bef`.

crates/jolt-prover-legacy/Cargo.toml
markosg04: `jolt-claims` is the only optional dep there; the `?` keeps `akita` from forcing it in.

.github/workflows/rust.yml
markosg04: Fixed in `85c9a89ee`.

markosg04:
Both fixed.

1. `6172dc4f9` — the dense cell-domain tables are gone (sparse lane weights + factored kernels). Transcript-identical, pinned by a coefficient-exact reference test. Measured:

| max advice | before | after |
|---|---:|---:|
| 4 KB | 79 MB | 3 MB |
| 64 KB | 1.26 GB | 14 MB |
| 1 MB | 8.39 GB | 176 MB (~6x faster too) |

Also added an e2e at exactly `max_untrusted_advice_size`.

2. `20b6e6b30` — added `full_only` setup params and hoisted the advice object setup into prover preprocessing; per-prove cost gone.

crates/jolt-akita/src/adapters.rs
```diff
@@ -68,6 +72,7 @@ impl AkitaSetupParams {
             default_layout_digest,
             one_hot_k: AKITA_ONE_HOT_K256,
             one_hot_only: false,
+            full_only: false,
```
mtbadakhshan: full in the context of Akita means dense polynomial, and in different places in the jolt code, "dense" has been used for this purpose. I suggest changing full to dense. 
On the Akita side, this has to be changed as well, but we can simply use dense instead of full (we don't have to wait for the change to be settled on Akita).

mtbadakhshan: This PR on akita changes full to dense on the akita side:
https://github.com/LayerZero-Labs/akita/pull/324

moodlezoup (approved):

PR #1677, opened 2026-07-15
Refactor jolt-witness

moodlezoup:
## Summary

Executes `specs/witness-redesign.md` (checked in on this branch) end-to-end: `jolt-witness` is rebuilt around **atomic witnesses** (single-sourced `Extract` impls per value newtype), **derive-composed consumer bundles** (`#[derive(WitnessBundle)]` in the new `jolt-witness-derive` crate), and **one fused `stream_witnesses` pass** driving all consumers over the trace in a single row walk. This replaces several parallel, partially-dead abstractions that had accreted in the crate: the `WitnessNamespace`/`OracleRef<N>` id vocabulary, the pull-based `PolynomialStream`/`ChunkVisitor` machinery, the `JoltVmBatchNeeds`/`JoltVmBatchRow` needs/row apparatus, and the monolithic `trace_virtual_value` match. The commit kernel is also rewired onto the new pass as a `StreamConsumer`, fusing what were ~40 per-column trace walks into one.

Every slice is gated by the byte-diff harness (proofs byte-identical to legacy) plus the full workspace suite and `clippy` in both standard and `--features zk` modes — this is a pure internal restructuring of witness generation with no intended behavior change.

## Changes

- **Dead API removal** (`af8699f`): cut `WitnessProvider`'s unused surface — `view_requirements`/`try_evaluate_oracle_view`, `ViewRequirement`, `MaterializationPolicy`, `RetentionHint`, `PolynomialView`; deleted unconsumed namespaces (`protocols/wrapper`, `protocols/dory_assist`, blindfold/rv64 stubs), dead stage-row builders (stage2/stage3/spartan_outer, register read-write rows), and dead root exports (`PublicValue`, `OpeningWitness`, `RaFamily*`).
- **Atomic witnesses** (`e9234a3`, `d305340`, `d81377b`): `trace_virtual_value`'s 27 match arms become one `Extract` impl per row-free witness newtype, colocated with each witness's definition in `witnesses/`; `WitnessDimensions` folds into a single `Shape` type.
- **`jolt-witness-derive` crate + bundles** (`44bfcb8`, `8d19391`): `#[derive(WitnessBundle)]` composes a consumer's atomic witnesses into a typed struct, with `#[opening(..)]` annotations binding fields to jolt-claims ids (`OutputClaims`-style grammar) and a generated per-field consistency test pinning the bundle against `oracle_table`. `Stage5InstructionReadRafRow`/`JoltVmStage6Row` and their hand-rolled traits are replaced by `InstructionReadRafWitness`/`BytecodeReadRafWitness` bundles. The test-only consistency surface (`annotated_columns`) is kept off the trait so it can't leak into production builds across crates.
- **`JoltWitnessOracle<F>` replaces the namespace abstraction** (`3111414`): the crate now defines **no id vocabulary of its own** — `shape()`/`oracle_table()` are exhaustive, no-wildcard matches directly over `JoltPolynomialId`, so a new jolt-claims variant fails to compile until it's mapped or explicitly excluded. `FixedBackend` (stored dense columns, gated behind `test-utils`) is the second implementor, serving unit tests and fixture replay.
- **Committed-column streaming unification** (`8801d84`, `6a41692`, `1f81bd9`, `76d5ecd`, `7259429`, `fbf1cf0`): the old pull-based `PolynomialStream`/`ChunkVisitor`/`JoltVmBatchNeeds` layer is deleted; committed streams (increments, one-hot RA chunks) now derive from the same atomic extractors and the same row walk as everything else. The commit kernel becomes a `StreamConsumer` of one fused `stream_witnesses` pass (`FusedColumns` holds every column's streaming PCS state), replacing ~40 per-column trace walks with one. `oracle_table`'s committed arms end up as one-liners dispatching to `materialize_cycle`/`materialize_one_hot`, and the now-unconsumed chunk-visitor layer is deleted outright.
- **field-inline mirrors the same pattern** (`8702685`): field-inline witnesses get their own atomic newtypes + `Extract` impls + exhaustive `FieldInlinePolynomialId` match, deleting `trace_virtual_value`/`field_rd_inc`/`materialize_field_rd_inc` for that backend too. This closes out the spec's execution plan (slice 5/5).
- **Cleanup**: `CycleRange` → `std::ops::Range` (`ee2695a`); `InstructionRaChunk` → plain `usize` since every cycle performs a lookup (`c7818d5`); `FixedBackend` gated behind `test-utils` (`ddc1fe4`); unused `jolt-poly` dep dropped after the chunk-visitor deletion (`17b18eb`).

## Testing

- [x] Byte-diff harness (10/10) confirms byte-identical proofs vs. legacy at every slice
- [x] Workspace suite (`cargo nextest run`) green after each commit, both `--features host` and `--features host,zk`
- [x] `cargo clippy --all --all-targets` clean under `host` and `host,zk`
- [x] `field-inline` suite (47/47) green after the field-inline mirror slice
- [x] Per-bundle derived consistency tests pin each `#[opening(..)]`-annotated field against `oracle_table`

## Security Considerations

Internal restructuring of the trace→witness generation layer only — no changes to sumcheck, R1CS, verifier, or proof/serialization formats. Every commit is gated by the byte-diff harness, so any accidental change to a witness value's derivation would show up as a proof byte mismatch rather than a silent correctness regression.

## Breaking Changes

None externally. `jolt-witness`'s public API changes substantially (`WitnessNamespace`/`OracleRef`/`PolynomialStream`/`ChunkVisitor` deleted in favor of `JoltWitnessOracle<F>` + `StreamConsumer` + `#[derive(WitnessBundle)]`), but its only consumers are `jolt-kernels` and `jolt-prover` within this workspace, both updated in the same PR. Proof format and serialization are unchanged.


0xAndoroid (approved):
This looks good to me. I'm going to run the claude review on it, and can be merged afterwards.

crates/jolt-kernels/src/bytecode_read_raf.rs
```diff
@@ -18,6 +18,13 @@ use crate::{KernelError, ProofSession, ProveSumcheck};
 /// output over the padded bytecode), the five upstream stage cycle points,
 /// the per-cycle bytecode indices (the PC pushforward source), and the
 /// preprocessing entry index.
+/// The per-cycle witness of the bytecode read+RAF address phase: the PC
+/// pushforward source (no-ops and unmapped rows land on slot 0).
```
0xAndoroid, line 17: The struct + its doc landed between `BytecodeReadRafAddressProver`'s doc comment and the trait: lines 16–20 ("The stage-6a bytecode read+RAF address-phase slot. The typed relation data is …") now attach to `BytecodeReadRafWitness`, and the trait is left undocumented. `instruction_read_raf.rs` did the same insertion correctly (struct above the slot doc).

crates/jolt-witness/Cargo.toml
```diff
@@ -22,9 +23,9 @@ field-inline = [
 
 [dependencies]
 jolt-claims.workspace = true
+jolt-witness-derive.workspace = true
 jolt-field.workspace = true
 jolt-lookup-tables.workspace = true
-jolt-poly.workspace = true
 jolt-program.workspace = true
 jolt-riscv.workspace = true
 rayon.workspace = true
```
0xAndoroid: After this PR the crate's only rayon use is `field_inline/mod.rs` — same situation as the dropped jolt-poly dep. `rayon = { workspace = true, optional = true }` + `"dep:rayon"` in the `field-inline` feature keeps default builds rayon-free.

0xAndoroid (commented):
Reviewed as a behavior-preservation audit of the restructure: diffed every old derivation against its new home arm-by-arm — `trace_virtual_value`'s 27 arms ↔ the `Extract` impls, `JoltVmBatchRow`/`*StreamKind::value_from_row` ↔ `CommittedColumnsWitness` + `ColumnKind`, stage-5/6 rows ↔ the new bundles, the RAM/register grid materializers (carried verbatim), `committed_oracle_order` ↔ `committed_order`, and the padding conventions (one-hot `chunk(0)`/cold-cycle cases now flow through default-row extraction; equivalent because a default `TraceRow` is a NoOp). Also checked the commit rewiring: `FusedColumns` reproduces the legacy per-column call sequences (`feed_i128` per full row window, `process_one_hot_chunk` per window with `one_hot_k = 2^log_k_chunk`, which equals the old per-descriptor `log_rows − log_t` for every committed RA), and the `MaterializedColumn` fallbacks match the old address-major/widened-grid scatter formulas.

**No semantic divergence found.** The one polarity-looking change — `Stage5InstructionReadRafRow.interleaved_operands` → `InstructionRafFlag` (its negation) — is consistently inverted at all three consumption sites in `reference/instruction_read_raf.rs`. The deleted `evaluate_*`/view surface is confirmed dead on main (test-only consumers). Verifier, claims formulas, and serialization are untouched (`JoltOpeningId::polynomial_id()` is purely additive), so nothing here moves the jolt-verifier claims geometry. `PcLookupCache`'s deletion is fine: `BytecodePCMapper::get_pc` is a vec index + tiny scan, cheaper than the HashMap that fronted it.

Ran locally: byte-diff harness 10/10, jolt-witness 35/35, field-inline 41/41, clippy clean on the touched crates (incl. `--features field-inline,test-utils`; the modular `jolt-prover` has no `zk` feature — that gate lives in the untouched legacy crate).

Five inline notes: one real defect (misattached doc comment), the rest nits/watch-items.

*This review was generated by Claude Code on behalf of Andrew (@0xAndoroid).*

crates/jolt-prover/src/stages/stage0.rs
```diff
@@ -164,6 +167,16 @@ where
...
         })
         .collect();
+    // Stage-0 validation: every id the proof will request — the committed
+    // set and each bundle's annotated set — must be servable by the backend
+    // before witness generation starts.
+    let requested = ids
+        .iter()
+        .map(|&id| JoltPolynomialId::Committed(id))
+        .chain(InstructionReadRafWitness::annotated_ids())
+        .chain(BytecodeReadRafWitness::annotated_ids());
```
0xAndoroid, line 177: This hand-chained list is the only enforcement of the spec's validated-bundles invariant, and nothing ties it to the bundles actually consumed in stages 5/6a — a future bundle compiles and proves fine while silently skipping stage-0 validation. Probably the intended tradeoff of the no-registry design, but a pointer on `WitnessBundle::annotated_ids` ("consumed bundles must be chained into stage-0 validation") would make the contract findable from the derive.

crates/jolt-witness/src/consumer.rs
```diff
@@ -0,0 +1,234 @@
...
+    ) -> Result<(), WitnessError>;
+}
+
+fn deliver<C: StreamConsumer>(
+    consumer: &mut C,
+    rows: &[TraceRow],
+    next_after: Option<&TraceRow>,
+    env: &WitnessEnv<'_>,
+) -> Result<(), WitnessError> {
+    let mut bundles = Vec::with_capacity(rows.len());
```
0xAndoroid, line 61: Each consumer independently re-runs `from_row` over the chunk, plus a fresh bundle Vec per chunk×consumer. Free at today's arity (every pass runs one live consumer), but it becomes the cost model once the streaming engine registers multi-consumer round sets: N consumers sharing `LookupIndex` = N instruction decodes per row, and the spec's derive-level hoist only covers intra-bundle duplication, not cross-consumer. Worth pricing in (or a scratch-buffer + shared-fact pass) when the engine lands.

crates/jolt-witness/src/backend/trace/advice.rs
```diff
@@ -0,0 +1,61 @@
...
+            self.preprocessing.memory_layout.max_untrusted_advice_size as usize,
+        )
+    }
+}
+
+/// An advice column's word count: the power-of-two number of little-endian
+/// words covering the configured maximum size. Single-sources the column
+/// length between [`TraceBackend::shape_of`] and the materialization below.
+pub(super) fn advice_words(max_bytes: usize) -> usize {
+    (max_bytes / 8).next_power_of_two().max(1)
```
0xAndoroid, line 28: Faithfully preserved pre-existing quirk (legacy `rows_from_max_bytes` is the same formula): `max_bytes / 8` floors, so a layout with a non-multiple-of-8 advice max silently drops the tail bytes from the committed advice polynomial while `populate_ram_bytes` still writes them into the RAM state — a max-length blob would then fail the advice/RAM-init consistency confusingly rather than loudly. `max_bytes.div_ceil(8)` covers it. Out of scope for a byte-pinned refactor; follow-up material.

0xAndoroid (commented):
Superseded — see the second-pass review below with the inline note anchored at crates/jolt-witness/src/consumer.rs:22.

0xAndoroid (commented):
Superseded — see the second-pass review below with the inline note anchored at crates/jolt-witness/src/consumer.rs:22.

crates/jolt-witness/src/consumer.rs
```diff
@@ -0,0 +1,234 @@
...
+/// membership in a set is static, presence is runtime.
+pub trait StreamConsumer: Send + Sync {
+    type Witness: WitnessBundle;
+
+    fn consume(&mut self, chunk: &[Self::Witness]);
+}
+
+impl<C: StreamConsumer> StreamConsumer for Option<C> {
+    type Witness = C::Witness;
+
```
0xAndoroid, line 29: `Option<C>` skips only the consume, not the extraction: `deliver` materializes the full `bundles` Vec (a `from_row` per row, including the instruction re-decodes the spec calls out as the accepted hot-path cost) and `Option::consume` then drops it. So a runtime-absent slot (`None`) costs exactly as much as a present one — the "presence is runtime" knob only saves the sink, not the work. No production consumer uses `Option` today, so this is latent, but the moment a runtime-gated consumer lands (committed-program mode is the obvious candidate) the disabled path silently pays full extraction. Consider short-circuiting in the tuple impl (extract only when `is_some()`) or documenting that `None` does not elide extraction.

0xAndoroid (commented):
Second-pass review (Claude Code on behalf of @0xAndoroid), inline note below; third attempt at anchoring — GitHub's API keeps dropping the line.

PR #1683, opened 2026-07-16
feat: fuse Akita increment claim reduction

quangvdao:
Stacked on #1675 and supersedes the prover protocol in #1676.

## Protocol

- Retains standalone IncVirtualization after Stage 5 so its Store claim still feeds bytecode read RAF.
- Adds a separate FusedIncClaimReduction to Stage 6b, reducing only the fused increment claim to the shared Stage 6b cycle point.
- Commits every increment chunk and the increment MSB as a full K x T one-hot member. The MSB hot address is exactly zero or one.
- Extends Stage 7 HammingWeightClaimReduction with increment hamming-one, booleanity, and the shifted fused-increment decode identity.
- Removes UnsignedIncChunkReconstruction and the generic Wjolt packed-layout strategy.
- Opens the canonical native Wjolt one-hot group directly at one common point. Prefix packing remains only for auxiliary advice and committed-program objects.
- Binds commitment and verifier setup metadata to the canonical member order, arity, count, layout digest, and Akita OneHot backend before PCS verification.
- Leaves the Dory standard and BlindFold ZK paths unchanged.

## Validation

- Full Akita prover suite: 441 passed, 1 skipped.
- Akita verifier suite: 70 passed.
- jolt-claims suite: 189 passed.
- Dory muldiv standard: 3 passed.
- Dory muldiv ZK: 3 passed.
- Base verifier suite: 68 passed.
- cargo fmt --all -- --check.
- Workspace clippy with host and host,zk.
- Scoped Akita clippy for jolt-prover-legacy, jolt-verifier, jolt-claims, and jolt-akita.


quangvdao:
@markosg corrected performance update after the full native-one-hot Wjolt cutover.

The first comparison I posted was not clean: the Dory “verify” timer called `verifier_preprocessing_from_prover`, so it included verifier-SRS derivation and conversion of the full verifier preprocessing object. The Dory release test binary also deliberately skips the production prepared-pairing cache unless the benchmark initializes it explicitly. Akita had already completed its analogous verifier preprocessing before its verify timer.

I reran the twin release harnesses twice, back-to-back, on the same machine and branch, using sha2-chain with a padded 2^20 trace (1,048,576 rows, 277 iterations), with Dory’s production pairing cache initialized and preprocessing excluded from pure prove/verify:

| path | prove | pure verify |
|---|---:|---:|
| Akita, K=16, Wjolt shape (24 vars, 57 members) | 5.49–6.12 s | 21.7–22.4 ms |
| Dory, prepared pairing cache enabled | 7.60–7.73 s | 65.7–68.5 ms |

At the midpoint, Akita is about 24% lower prover time and about 3x faster verification—not the 40% / 20x result from the contaminated timers.

Reusable work measured outside those timers:

- Akita Wjolt setup: 203–209 ms.
- Dory prover preprocessing/SRS construction: 1.13–1.14 s.
- Dory prepared-pairing cache construction: 206–208 ms.
- Dory verifier preprocessing/SRS conversion: 337–354 ms.

The previous 410–463 ms “Dory verify” was therefore approximately 338–354 ms of verifier preprocessing plus 66–68 ms of actual verification. These are two local release runs, not a statistically stable benchmark, and the Dory/Akita protocol paths are not instruction-for-instruction identical, so this remains an end-to-end comparison rather than a PCS-only microbenchmark.

The K=16 schedule catalog now covers every valid Jolt Wjolt shape, so this workload no longer runs the planner/DP at setup time. The shared toggle remains K=16 below log_T=25 and K=256 at and above it.

Compared with the old `lz/integrate-hachi` work, Akita already has the important backend pieces: D=64, one-hot-specific one-digit commitment setup, lazy block caches, wide/column-sweep Ajtai kernels, and batched one-hot fold/decompose paths. The remaining levers are:

1. **Generate the Jolt-shaped K=256 schedule catalog.** The existing K=256 table only covers 1- and 4-polynomial groups; large-trace Wjolt needs `25 + bytecode_d + ram_d` members (normally 27–41). Without this, log_T >= 25 still falls back to runtime planning.
2. **Remove duplicate witness representations and copies.** Jolt builds `Vec<OneHotPolynomial>`, then the adapter clones every `Vec<Option<u8>>` into Akita. The Akita prover hint already owns the backend polynomials and their block caches, and the opening path only shape-checks the Jolt witnesses. We should transfer ownership once and open from the hint.
3. **Reuse the already-derived columns.** Fused-inc hot lanes are built for stages 6/7 and then derived again for Wjolt. Each RA member also rescans the full trace independently. A compact cycle-major group source should derive a cycle once and feed all members, or reuse the RA/inc data already produced for the sumchecks.
4. **Try the Hachi-style cache-aware group layout, as a measured protocol/layout variant.** One-hot must remain one-hot: keep the K address bits as the contiguous coefficient/chunk bits. The experiment is to reorder/interleave low cycle bits with the group-member rows in Akita's inner/outer commitment layout, with the matching opening-point map and layout digest. The old Hachi change swapped its `cycle_inner` / `poly_inner` placement and streamed from the trace; its recorded 2^24 result was commitment 6.4s → 4.7s and end-to-end 32.5s → 31.3s. That old number is not directly comparable to current Akita, but it is concrete evidence that this layout axis is worth profiling.
5. **Cache prepared transparent setup by (K, num_vars, num_members, layout digest).** Generated schedules removed the expensive DP, but we still spend ~0.2s preparing the same deterministic setup when it is not reused through preprocessing.
6. **Only after profiling, tune the general planner/kernel objective.** If K=16 multi-chunk commit/fold remains hot, improve the general Akita cost model or kernel and regenerate the catalog—no hand-written per-trace schedules and no separate ad hoc Jolt policy.

My priority order remains K=256 catalog first, then ownership/reuse, then a profile-guided implementation of the group-layout experiment.

quangvdao:
Follow-up correction on generated-schedule ownership after inspecting the Akita APIs and history:

The Jolt-specific schedule catalogs should live in Jolt, not Akita. This is exactly what Akita's schedule-catalog ownership cutover (LayerZero Akita PR #203) was designed to enable:

- `CommitmentConfig::schedule_catalog()` is the downstream catalog-injection hook; `None` falls back to the planner DP.
- `akita_planner::emit::EmitSpec` is a reusable downstream emitter and explicitly calls out Jolt as a consumer.
- Catalog identities are validated against the config policy before lookup, so locally owned catalogs retain the same verifier-reachable safety checks.

Our present implementation bypasses that design by aliasing Akita's catalog-bearing `D64OneHotK16` / `D64OneHot` presets directly and enabling `schedules-default`. Worse, the Akita branch now contains `JOLT_K16_NUM_POLYS`, a Jolt-specific generated-family registration, and the generated Jolt table itself. That ownership is wrong.

The two Akita commits split cleanly:

- `6b6b5845` adds a generic planner-backed `D64OneHotK16` policy preset with no catalog. This is reasonable generic Akita functionality.
- `cf5cd3cb` attaches Jolt-specific widths/tables to Akita. This should be removed.

Suggested next steps:

1. Keep Akita responsible for the planner, table representation/emitter, catalog validation, and generic K16/K256 policies.
2. Move the generated K16 catalog into Jolt; also generate Jolt's reachable K256 catalog in Jolt rather than extending Akita's tables.
3. Define Jolt-local `JoltD64OneHotK16` and `JoltD64OneHotK256` configs that delegate the algebra/planner policy to the Akita presets and override only `schedule_catalog()`.
4. Instantiate `AkitaCommitmentScheme` with those Jolt-local configs.
5. Stop enabling the broad `schedules-default` bundle from `jolt-akita`; enable only genuinely Akita-owned tables explicitly where needed.
6. Add Jolt-side drift/table-hit tests covering every reachable `W_jolt` shape and comparing emitted rows against DP offline.

One small Akita API wrinkle remains: the proof-optimized setup-envelope helper is currently `pub(crate)`. A downstream wrapper can delegate setup sizing to its base preset, but after detaching the Jolt catalog that may reintroduce expensive DP during setup. The clean generic fix is either to expose that helper or add a public catalog-wrapper config in Akita. This is a small engine/API improvement; it does not justify upstreaming Jolt schedule rows.

So the previous suggestion that K256 Jolt coverage should be generated in Akita was incorrect. @markosg

quangvdao:
Superseded by the refreshed upstream verifier/prover integration in #1675 and #1676, plus the clean stacked performance follow-up in #1685. The useful fused-increment protocol cutover and prover work from this branch are already present upstream; #1685 now contains only the remaining Jolt-owned schedule/setup-sizing changes against the latest prover branch.

PR #1684, opened 2026-07-17
refactor(jolt-field): Land Solinas field from `akita-field` in `jolt-field`

Acentelles:
## Summary

This PR does four things toward a single audited field implementation:

1. **Lands the Solinas field stack** from `akita-field` in `jolt-field`: the 32-, 64-, and 128-bit prime fields, the `FpExt2`/`FpExt4`/`FpExt8` extension towers, NEON/AVX2/AVX-512 packed backends, unreduced accumulator paths, tests, fuzz coverage, and benchmarks. One implementation, one Rust type identity: Akita consumes `jolt-field` directly, with no dependency back except a temporary feature-gated bootstrap adapter (see Changes).
2. **Consolidates the resulting trait surface** per `specs/consolidate-field-traits.md`. The import brought the crate to 46 public traits across 82 files, with 15 one-trait-per-file micro-files at the root. The consolidation reduces this to 23 public traits (see the `CanonicalBytes` note under Changes) in four root trait modules (`algebra.rs`, `canonical.rs`, `accumulator.rs`, `field.rs`) plus the feature-gated backend modules, switches Solinas wire serialization to serde + bincode, and adopts the explicit `arkworks/bn254.rs` per-type layout. This is a refactor of trait boundaries, not arithmetic.
3. **Fixes two pre-existing `jolt-field` bugs** surfaced by differential testing: `Fp64::reduce_u128` truncated the fold multiply for sub-word moduli on inputs $\ge 2^{64+\text{BITS}}$, and the S160 two-limb multiply could overflow a fused u128 sum (a debug-only panic; release output was provably correct). Both fixed in `79944f2` with regression tests that fail pre-fix.
4. **Carries the `jolt-field-two` rebuild to completion.** `jolt-field-two` is a from-scratch minimal-LOC rebuild of this crate, started in `feat/field-crate-refactor` (checkpoints 1–4 plus its working SPEC.md, cherry-picked here as `5635b98`); checkpoints 5–9 (fp128, extensions, unreduced, packed, parallel) complete the spec's build order at 5,103 counted LOC against the 6,240 budget, versus the 11,410-LOC baseline, differential-tested against `jolt-field` as the oracle. It is intended to replace the implementation from items 1–2 in a follow-up PR; the per-checkpoint summary is in the comments below.

| Akita source | `jolt-field` destination (post-consolidation) |
|---|---|
| `crates/akita-field/src/traits.rs` | `src/algebra.rs`, `src/canonical.rs`, `src/accumulator.rs`, `src/field.rs`, `src/solinas_traits.rs` |
| `crates/akita-field/src/prime/` | `src/prime/` |
| `crates/akita-field/src/ext/` | `src/ext/` |
| `crates/akita-field/src/packed/` | `src/packed/` |
| `crates/akita-field/src/unreduced/` | `src/unreduced/` |
| `crates/akita-field/src/fft.rs` | Not imported in final form: removed in the consolidation (zero consumers in this workspace) |
| `crates/akita-field/src/parallel.rs` | `src/parallel.rs` |
| `crates/akita-pcs/benches/field_arith.rs` and `field_arith/` | `benches/solinas_field_arith.rs` and `benches/solinas_field_arith/` |
| Akita field arithmetic fuzz coverage | `fuzz/fuzz_targets/solinas_field_arith.rs` |

### Solinas import

- Add `solinas` and `parallel` features to `jolt-field` while retaining the existing BN254 default.
- Import the complete Akita Solinas implementation (prime fields, extension towers, packed backends, unreduced/wide accumulators, parallel helpers).
- Keep Akita proof and field serialization concerns out of the shared crate: no `AkitaSerialize`, no Akita validation policy or protocol framing. Add `FieldError` for backend-independent input and shape failures; Akita converts it at its repository boundary.
- Keep `src/akita.rs` and the `akita` feature as a temporary, feature-gated bootstrap so the adapter stays buildable until the Akita cutover; both are removed in the final migration PR. `akita-config`/`akita-field` are optional dependencies behind that feature only.
- Update `jolt-akita` to the current Akita runtime ring dimension, opening claim, setup, commitment, and batching APIs.
- Move the Solinas Criterion benchmark into `jolt-field`, add a Solinas arithmetic fuzz target, and record file-level source provenance.
- Add a CI feature matrix for backend-free, BN254-only, Solinas-only, and combined builds, plus `scripts/check-shared-field-identity.sh`.

### Trait consolidation (spec phases 1–5)

- **Phase 1 — delete dead surface.** Remove traits with no generic consumer in this workspace or Akita: both signed-accumulator families (`SignedScalarAccumulator`/`WithSmallScalarAccumulator`, `SignedProductAccumulator`/`WithSignedProductAccumulator`), `ExtensionCoeff`, `ScaleI32`, `BalancedDigitLookup` (now a free function), and `SmoothFftField` together with the 1,100-line `fft.rs`.
- **Phase 2 — one accumulator trait.** Merge `AdditiveAccumulator` + `RingAccumulator` into a single `Accumulator`.
- **Phase 3 — root traits and wire format.** `FieldCore` absorbs `Invertible` and `RandomSampling`; `FromPrimitiveInt` absorbs `MulPow2` and `MulPrimitiveInt`; a new `CanonicalRepr` replaces the seven byte/introspection/challenge traits (`CanonicalBytes`, `ReducingBytes`, `FixedByteSize`, `FixedBytes<N>`, `CanonicalU64`, `CanonicalBitLength`, `TranscriptChallenge`). Solinas wire serialization becomes serde + bincode, matching the existing serde implementations on `arkworks/bn254.rs`; Fiat-Shamir transcript bytes keep the explicit canonical little-endian encoding and never go through bincode. A later commit (`ff5bf9c`) re-splits the byte surface back out as `CanonicalBytes`, now a supertrait of `CanonicalRepr`, after #1675's `NoCommitment` showed that transcript absorption should not require the full field-decode contract; the final surface is 23 public traits.
- **Phase 4 — extension cluster.** `ExtField<F>` absorbs `LiftBase`, `MulBase`, and `FrobeniusExtField`; `FpExt4MulBackend` + `FpExt8MulBackend` merge into one `ExtMulBackend`; `PackedField` absorbs `PackedValue`; the degree-4 mul/square schedule now exists in exactly one place, shared between scalar and packed backends.
- **Phase 5 — per-type layout.** Everything a concrete type implements is visible in that type's file, as an explicit `impl` or a one-line shared-macro invocation; the three `native_algebra.rs` side-files are replaced by one shared macro module.

BN254 is untouched by the consolidation: `Fr`'s serde and byte encodings, the Fiat-Shamir stream, and static dispatch are unchanged (spec invariants 1 and 6). `MontgomeryConstants` is retained pending confirmation that no out-of-tree GPU/Metal consumer exists (spec OPEN item; 23 traits become 22 when resolved).

## Testing

- [x] `muldiv` e2e passes in `--features host` and `--features host,zk` after the consolidation.
- [x] `jolt-field` feature matrix passes: backend-free (`--no-default-features`), BN254-only, Solinas-only, combined, and Solinas-plus-parallel.
- [x] `mersenne61_compat` (in `jolt-sumcheck`) compiles and passes against `jolt-field --no-default-features` with bounds updated to the merged traits: the slim hierarchy remains implementable without arkworks.
- [x] New bincode round-trip tests for `Fp32`/`Fp64`/`Fp128`/`FpExt2`/`FpExt4`/`FpExt8`; serialized-size tests assert each element encodes to exactly `NUM_BYTES` bytes and a `Vec` of $n$ elements to $n \cdot \texttt{NUM BYTES}$ plus one length prefix.
- [x] A compile test verifies the merged `CanonicalRepr` challenge derivation matches the previous `TranscriptChallenge` behavior on `Fr` (identical bytes in, identical element out).
- [x] The `solinas_field_arith` fuzz target and both Criterion benches build; bench results before/after the consolidation are within noise.
- [x] `scripts/check-shared-field-identity.sh` reports one `jolt-field` identity and no `akita-field` in the integrated dependency graph.
- [x] Byte-for-byte comparison of a standard-mode `muldiv` proof against a main-built proof: identical in all 63,371 bytes, proving is deterministic run-to-run, and the main-built verifier accepts the branch-built proof. (Reproduction note: `cfg(test)` builds randomize the Dory URS per process via `DoryGlobals::configure_test_cache_root`, so cross-process comparison requires pinning both builds to a shared URS cache.)
- [x] Regression tests for the `Fp64::reduce_u128` and S160 fixes fail on the pre-fix code.
- [x] `jolt-field-two`: 121 differential tests green across the full feature matrix, plus `+avx2` and `+avx512f,+avx512dq` compile cross-checks against `x86_64-apple-darwin`.

## Security Considerations

This change touches security-sensitive field arithmetic used by the PCS and therefore affects the trusted audit surface for prover and verifier correctness. The intended arithmetic change is zero across both stages: moduli, canonical representations, extension bases, reduction algorithms, and packed kernel behavior are preserved, and architecture-specific arithmetic kernels (NEON, AVX2, AVX-512) are moved without semantic modification.

For the import stage, 22 of the imported files are byte-for-byte copies of their Akita sources and 38 are adapted imports; the main semantic adaptations are removal of Akita serialization bounds and replacement of protocol-level `AkitaError` with the narrower `FieldError`. The consolidation stage then rewrites trait boundaries only: merges are compile-time renames, `#[inline]` discipline is carried over onto both trait methods and impls, and all dispatch remains static. The net effect is a smaller audit surface: 46 public traits become 23, and the crate root drops from 15 trait micro-files to 4 trait modules.

## Breaking Changes

- Removes `akita-field` from Jolt's dependency graph (`jolt-field` keeps optional `akita-*` dependencies behind the temporary `akita` bootstrap feature only).
- Deletes or merges 24 public traits; downstream code generic over the old names (`Invertible`, `RandomSampling`, `TranscriptChallenge`, `LiftBase`, `MulBase`, `FrobeniusExtField`, `AdditiveAccumulator`, `PackedValue`, ...) must rebind to the survivors (`FieldCore`, `CanonicalBytes`, `CanonicalRepr`, `ExtField`, `Accumulator`, `PackedField`, ...); `CanonicalBytes` survives with a slimmer, bytes-only contract. Akita adapts at its next pin bump.
- Removes `fft.rs` and `SmoothFftField` from `jolt-field`.
- Solinas types' wire format is now serde + bincode (BN254/`Fr` encodings unchanged).


Acentelles:
Once https://github.com/LayerZero-Labs/akita/pull/307 is merged, we can remove the temporary akita-field compatibility dependency from jolt-field and change `jolt-akita` from the temporary `jolt-field/akita` feature to `jolt-field/solinas`.

markosg04 (commented):
I think there is still hopefully some room to clean up this crate and especially the surface area of the solina-related traits and impls.

You can find a spec and the first few components implemented for such additional cleanup here: https://github.com/a16z/jolt/tree/feat/field-crate-refactor. The tl;dr is that I think this crate can probably be done in half the src loc than what it is now, and we can make the distinction between the backend and impl for the solina stack clearer. I also am generally a fan of using macros for these types of primitive crates because AI are very efficient with them and they seemingly help to catch/prevent some context drift.

Acentelles:
Per checkpoint, relative to the spec:

- **Checkpoint 5 — fp128** (`e1cbc17`, 528/700): one hand-written two-limb module, as specified. The AArch64 mul/sqr assembly is kept verbatim (with an asm-vs-portable cross-check test); the add/sub assembly is dropped under the spec's evidence rule — no in-tree benchmark supported it and the portable path compiles branchless. One bound sharpening over the baseline docs: `fold2_canonicalize` is exact and canonical for any u64 third limb given only $C < 2^{32}$; the $C(C+1) < P$ assert is kept but documented as implied for that kernel.
- **Checkpoint 6 — extensions** (`ae313af`, 54 + 133 + 617): the spec's bench-before-keep gate produced its first reversal — the baseline's fused `Fp32` deg-4 kernel measures **2.5x slower** than the generic schedule on M4 (12.3 vs 31.1 ns/op; the ported override reproduces the baseline's own 31.0 as the anchor). Dropped from src, preserved in `benches/ext4_kernels.rs` for an x86 re-run that could flip the verdict. The deg-4/8 multiplication schedules moved to an unconditional `schedules.rs` so scalar hook defaults and packed lanes share one source of truth; the ext budget was re-carved 890 → 750 + 140 with the component total unchanged.
- **Checkpoint 7 — unreduced** (`d4959dd`, 18 + 501): the fused `Unreduced` surface replacing `HasUnreducedOps`/`HasWide`/`ReduceTo`, plus `Fold`, per the trait table. Re-derivation corrected three baseline headroom docs (lane headroom is exactly 32768; the FpExt4⟨Fp32⟩ accumulator bound is $7P^2 < 2^{67}$ with $\ge 2^{61}$ headroom terms) and the fused FpExt2⟨Fp64⟩ path's silent NR ∈ {−1, 2} assumption is now debug-asserted. The NEON accumulator intrinsics are dropped with disassembly evidence: LLVM emits identical `add.4s`/`sub.4s`/`neg.4s` from portable code, and additionally vectorizes the lane scaling the intrinsics never covered.
- **Checkpoint 8 — packed** (`19e88a7`, 1,067/1,600): one packed-algebra source of truth, generic over a per-ISA `SimdWord` vocabulary. This is a recorded deviation from the spec's "engine macro" phrasing — same single-source outcome, but trait generics instead of a macro; rationale in SPEC.md. Genuine per-ISA differences are confined to `simd.rs`. Every cfg region is compile-covered (native NEON runs the differentials; `+avx2` and `+avx512f,+avx512dq` cross-checks target `x86_64-apple-darwin`). One upgrade over baseline: packed Fp128 multiplication calls the scalar kernel per lane, which on aarch64 is the assembly multiply the baseline's NEON backend never used.
- **Checkpoint 9 — parallel + final audit** (`58785fa`, 78/80 + docs): the rayon helpers ported whole. The audit found that none of the seven `cfg_*!` macros has a single consumer in the workspace — no crate enables the `parallel` feature — so the component is flagged in SPEC.md as a deletion candidate for the replacement PR.

**Items needing reviewer judgment** (consolidated in SPEC.md): six files exceed their per-file budgets (largest overage +23 lines; the recurring cost centers are re-export blocks and the four ext kernel hooks on `Packed`, which need overridable defaults to be reachable in stable Rust); both bench verdicts are single-machine (M4) and the dropped kernels should be re-benched on x86 before replacement; the x86 backends are compile-verified only.

specs/jolt-field-rebuild.md
markosg04, line 1: Let's move this to `specs/` and rename

markosg04 (changes requested):
Nice work! Can we now delete and replace `jolt-field` with the new crate (and hence update rest of Jolt / akita)?

Acentelles:
The replacement contemplated in `specs/jolt-field-rebuild.md` ("Remaining before replacement", item 3) is now on this PR: the baseline `jolt-field` implementation is deleted and `jolt-field-two` has taken over `crates/jolt-field` under the package name `jolt-field`, with every consumer rebound to the new trait names. The PR now delivers the end state directly rather than two crates side by side.

Commit sequence:

- `db8e650` — pre-swap surface work in the rebuild: the byte surface split out of `CanonicalEncoding` as a bare `CanonicalBytes` supertrait (same reasoning as the earlier `CanonicalRepr` split: transcript absorption, e.g. `NoCommitment`, must not require the field decode contract), plus the temporary akita bootstrap adapter ported to the new spine.
- `2b800e1` — the rebuild's differential suites made oracle-free so the evidence survives the baseline's deletion: independent num-bigint/schoolbook oracles for arithmetic, and golden byte fixtures generated from the baseline before deletion pinning bincode wire bytes, `to_bytes_le`, checked decodes, and the BN254 challenge conventions (125 tests, up from 121). Three spots are honestly weaker and are listed in the commit; the biggest is that BN254 absolute-byte pinning now rests on fixture vectors rather than random cross-crate sampling.
- `cf8a66a` — the swap. `MontgomeryConstants` dies with the baseline: zero consumers in the workspace, closing the spec's OPEN item.
- `079356e` — the consumer rebind, ~400 files: umbrella `Field` → `JoltField`, `FieldCore` → `Field`, `RingCore`/`FromPrimitiveInt` → `Ring`, `CanonicalRepr` → `CanonicalEncoding`, method renames, `arkworks::bn254` paths → root exports. `mersenne61_compat` reworked against the new spine, keeping the "implementable without arkworks" property demonstrated.
- `b2e3086` — validation evidence recorded in the spec.
- `46e2ce8` — merge of main: #1690's BlindFold modular-prover code rebound to the new names (`RingAccumulator` → `Accumulator`, plus two `F: Field` bounds that need the `JoltField` umbrella now that `Field` is the narrower algebra trait).

Validation (all recorded in `specs/jolt-field-rebuild.md` under "Replacement validation evidence"):

- **Proof bytes unchanged.** A standard-mode `muldiv` proof from the pre-replacement build and one from the replacement, generated against a shared pinned Dory URS, are byte-identical in all 63,372 bytes. ZK proofs are randomized (BlindFold), so the applicable check is size equality: 65,947 bytes on both builds.
- **Tests**: 3,901 green across workspace default, prover-legacy host/zk/akita, verifier akita fixtures, and solinas-only lanes; `muldiv` e2e passes in both modes, including on the merge of main.
- **Lints**: clippy `--all-targets -- -D warnings` on host, host+zk, the three akita lanes, field-inline; `+avx2` and `+avx512f,+avx512dq` cross-compiles to `x86_64-apple-darwin`.

Two surface deviations from the rebuild spec were forced by consumers and are recorded in the spec's "Replacement-time deviations": `JoltField` drops its serde bounds while the akita bootstrap edge exists (the foreign pre-cutover type cannot implement serde here; restore at cutover), and the bn254 types regain the primitive `From` conversions the plain arkworks re-exports carried (94+ call sites).

The spec's "Remaining after replacement" section lists what stays open: x86-64 runtime validation of the packed/fp128 paths (still compile-verified only), the fused-kernel bench re-run on x86, a CI target-feature lane so SIMD is not CI-dark, the zero-consumer parallel helpers (delete or wire a consumer), and the akita cutover follow-ups.

quangvdao:
Superseded by #1792, which preserves Alberto’s complete commit history, merges current main, carries the independent merge audit fixes, and includes the Akita field-drift update.

PR #1685, opened 2026-07-17
perf(akita): size setup from Jolt schedules

quangvdao:
## Summary

- size production Wjolt one-hot setup envelopes directly from the checked-in Jolt K=16/K=256 schedule catalogs
- account for every catalogued sub-shape plus exact root A/B/D footprints, with checked overflow and scalar-layout guards
- retain the upstream planner fallback for non-Wjolt development and test shapes
- enable only the fp128 D64 full and one-hot Akita schedule families instead of the broad default bundle

This is stacked on #1676.

## Performance scope

The catalog envelope calculation for both maximum production grids completes in about 28 ms in the focused test. This removes planner-DP scans from production Wjolt setup sizing. It does not remove first-time quantum-SIS public-matrix expansion; that remains a separate setup-materialization and caching optimization.

## Validation

- cargo nextest run -p jolt-akita production_wjolt_shapes_use_catalog_setup_sizing catalogs_cover_every_reachable_wjolt_shape --cargo-quiet
- cargo nextest run -p jolt-prover-legacy --features host,akita -E test(e2e_akita) --cargo-quiet
- cargo nextest run -p jolt-prover-legacy muldiv --cargo-quiet --features host
- cargo nextest run -p jolt-prover-legacy muldiv --cargo-quiet --features host,zk
- cargo clippy -p jolt-akita --all-targets -- -D warnings
- cargo clippy -p jolt-prover-legacy --features host,akita --all-targets -- -D warnings
- cargo clippy --all --features host -q --all-targets -- -D warnings
- cargo clippy --all --features host,zk -q --all-targets -- -D warnings
- cargo fmt --all -- --check

### DIFF
--- a/crates/jolt-akita/Cargo.toml
+++ b/crates/jolt-akita/Cargo.toml
@@ -9,12 +9,12 @@ description = "Akita batch opening adapter for the Jolt zkVM"
 workspace = true
 
 [dependencies]
-akita-config = { workspace = true, features = ["schedules-default"] }
-akita-pcs = { workspace = true, features = ["parallel", "schedules-default"] }
-akita-prover = { workspace = true, features = [
-  "parallel",
-  "schedules-default",
+akita-config = { workspace = true, features = [
+  "schedules-fp128-d64-full",
+  "schedules-fp128-d64-onehot",
 ] }
+akita-pcs = { workspace = true, features = ["parallel"] }
+akita-prover = { workspace = true, features = ["parallel"] }
 akita-challenges.workspace = true
 akita-planner.workspace = true
 akita-transcript.workspace = true
--- a/crates/jolt-akita/src/adapters.rs
+++ b/crates/jolt-akita/src/adapters.rs
@@ -11,7 +11,7 @@ use akita_types::{
 };
 use jolt_field::CanonicalBytes;
 use jolt_openings::{OpeningsError, VerifierOpeningClaim};
-use jolt_poly::{MultilinearPoly, OneHotIndexOrder, Polynomial};
+use jolt_poly::{MultilinearPoly, OneHotIndexOrder, OneHotPolynomial, Polynomial};
 use jolt_transcript::{AppendToTranscript, Label, LabelWithCount, Transcript, U64Word};
 use serde::{Deserialize, Serialize};
 use tracing::info_span;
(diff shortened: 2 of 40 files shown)


quangvdao:
I corrected the benchmark harness in 2a7392bd7 so both backends now report verifier preprocessing separately from pure verification. The Dory harness also initializes the prepared-pairing cache exactly as production setup does; test binaries normally skip that global initialization. On the same M4 / 24 GB machine, sequential release runs at a padded 2^20 SHA2-chain trace averaged: Akita prove 6.97 s, pure verify 26.11 ms, setup 130.8 ms; Dory prove 9.72 s, pure verify 74.04 ms, with prover preprocessing/pairing-cache/verifier-preprocessing reported separately. Thus the clean comparison is Akita at ~0.717x Dory proving time and ~2.84x faster pure verification. Both corrected ignored harnesses also pass targeted nextest smoke runs.

quangvdao:
Update after the latest performance and notation pass, cc @markosg04:

- Kept `UnsignedIncChunk` as the committed fused-increment chunk name. Replaced the opaque object names across the prover, verifier, proof surface, transcript labels, tests, and specs with `OneHotTrace`, `ProgramOneHot`, `AdviceOneHot`, and `FusedIncValue` / `FusedIncColumns`.
- Implemented an owned one-hot commitment path. Jolt now hands each hot-index buffer to Akita by ownership, so there is no clone at the Jolt/Akita PCS boundary. The prover opening consumes the hint-owned backend polynomials rather than rebuilding the Jolt witnesses.
- Reworked witness construction around one compact per-cycle source scan, followed by parallel construction of the semantic one-hot columns.
- Tested the proposed single-polynomial column-axis layout. It was drastically slower, so it has been removed. The production protocol remains one native Akita batch of semantic `K x T` one-hot columns, all opened at the same cycle/address point. There is no extra column fold or sumcheck.
- Jolt still owns the generated production schedules. The K=16 and K=256 reachable shape catalogs cover the grouped layout, and production shapes do not run planner DP at runtime.

The remaining streaming lever is below the current Jolt/Akita boundary: Akita still materializes and caches its internal one-hot blocks. A true callback/iterator-driven kernel needs an upstream Akita source API analogous to the older Hachi batch source. Jolt cannot add that implementation downstream because the backend traits and source types are both owned by Akita. That should be the next focused Akita-side optimization if we want to remove backend block materialization as well.

Validation on commit `8a5dbd85b`: Akita clippy passed with warnings denied; generated-catalog coverage passed; default K=16 and forced K=256 Akita end-to-end tests passed; the committed-program Akita end-to-end test passed; formatting and diff checks passed.

markosg04:
@quangvdao these changes should be reflected in #1675 and #1676 -- feel free to close if they are satisfactory.

PR #1690, opened 2026-07-22
feat(jolt-prover): BlindFold ZK support in the modular prover

0xAndoroid:
## Spec first

Spec: [`specs/jolt-prover-blindfold.md`](https://github.com/a16z/jolt/blob/feat/jolt-prover-blindfold/specs/jolt-prover-blindfold.md) — opened early for review; implementation follows as further commits on this branch.

## What

BlindFold ZK proving in the modular prover (`jolt-prover` + `jolt-kernels`), behind a compile-time `zk` cargo feature (no runtime mode flags on the prover side), producing proofs the existing `jolt-verifier` ZK path accepts unmodified. This is the ZK-wiring follow-up deferred by `specs/prover-stage-drivers.md` ("PR C is a recorder swap plus front work").

## How (short)

- Recorder-typed mode: `CommittedSumcheckRecorder` / `ClearSumcheckRecorder` selected by `#[cfg]` at one constructor; the recorder-generic stage drivers are untouched.
- `#[cfg]`-gated retention of the committed witnesses (round coefficients + blinds, output-claim rows + blinds) on stage outputs; committed uni-skip arms; ZK commit finish in the reference commit kernel; stage-8 `prove_batch_zk` arm.
- BlindFold tail by **replay**, with a zero `jolt-verifier` diff: the prover assembles the proof shell (`ZkProof = ()`) and replays it through the verifier's *existing* public surface (`validate_and_seed_transcript`, `stage1::verify` … `stage8::verify`, `stages::zk::blindfold::build`), so the protocol it proves against is literally the verifier's own code path — then assembles the witness via a new `assign_witness` API in `jolt-blindfold` (generalizing the test harness's `protocol_backed_witness`) and calls `jolt_blindfold::prove` (its first production caller).

## Guarantees

- `git diff` against the base is empty for `crates/jolt-verifier` — the verifier is untouched, transitively (verifier-consumed dependency changes are prover-side only).
- Clear mode is byte-frozen: the byte-diff harness stays green with zero fixture churn.
- New ZK e2e: modular-prover muldiv (plain + committed program + tamper rejection) and advice-consumer (hiding trusted/untrusted advice commits, advice claim reductions through BlindFold), all verified by the existing verifier; legacy ZK e2e stays green.


0xAndoroid:
Implementation is now on the branch, spec → code in six commits:

- `jolt-openings`: `prove_batch_zk` returns `ZkBatchOpening` (adds the joint evaluation the BlindFold witness needs).
- `jolt-blindfold`: `build_construction` (protocol + statement + baked sources) and `assign_witness` — the witness assembly generalized out of the test harness, claim chain derived from the committed coefficients (domain round-sum + Horner at the FS challenge), auxiliary values via assign-before-append eager product evaluation. New `construction_pipeline` tests give `prove.rs` its first same-crate prove→verify roundtrip.
- `jolt-verifier`: the non-akita stage spine promoted to `verify_stages` (now shared by `verify`, the zk-audit harness, and the prover's replay — the spine exists once per build); `blindfold::build_construction` with `build` delegating. Zero behavior change.
- `jolt-kernels`: `zk` feature switching the reference commit kernel to the hiding streaming finishes (`ModeStreamingCommitment`).
- `jolt-prover`: the `zk` feature; `ProofMode` (recorder + uni-skip arm selected by type, per-stage code identical in both builds); cfg-gated witness retention on stage outputs; stage-8 `prove_batch_zk` arm; the BlindFold tail — shell proof with `ZkProof = ()`, replay through `verify_stages` (debug-asserted byte-equal to the forward transcript), `build_construction`, `assign_witness`, `jolt_blindfold::prove`.
- Tests: `zk_e2e.rs` — modular ZK muldiv accepted by the unmodified verifier, tampered `random_u` rejected, committed-program variant (hiding chunk/image commits carried in the preprocessing). The byte-diff harness is `not(zk)`-gated (ZK proofs are randomized) and stays 16/16 in clear mode.

Gates: byte-diff 16/16 untouched, jolt-kernels, jolt-verifier (plain + akita + the 35-test zk fixture/tampering suite against legacy-generated fixtures), legacy `muldiv --features host,zk`, `clippy --all` under `host` and `host,zk`, fmt. Note: `-p jolt-verifier --features prover-fixtures,zk` has pre-existing `validate_inputs_*` unit-test failures on the base branch (fixtures not zk-config-aware); unchanged here.

0xAndoroid:
Reworked per the review constraint — **net diff to `crates/jolt-verifier` is now zero** (`git diff feat/jolt-prover -- crates/jolt-verifier` is empty; the `verify_stages` promotion and `blindfold::build_construction` are gone). Spec revised to match (the promotion now sits in Rejected Alternatives).

How the replay is expressed instead:

- **Spine prover-side:** `jolt-prover/src/blindfold.rs::replay_stages` spells out `validate_and_seed_transcript` + `build_formula_dimensions` + `stage1::verify` … `stage8::verify` + `stages::zk::blindfold::build(BlindFoldInputs { .. })` against the verifier's *existing* public surface — the same call sequence `verify` runs and the same surface the zk-audit test harness already consumes externally. Still the verifier's own code doing the lowering; the copy is the deliberate structural cost of keeping the verifier prover-free.
- **Witness assembly without the statement:** `BlindFoldConstruction` is gone. `BlindFoldProtocol::assign_witness` now works from the protocol's public parts alone (layout, constraint matrices, consistency, dimensions) plus caller-supplied stage domains (protocol constants). The auxiliary witness is solved directly from the matrices: `R1csBuilder::multiply` is the only allocator of non-layout privates and always emits `A·B = 1·v_fresh` with `v` after its operands, so one forward pass in constraint-emission order assigns every product; equality constraints (zero C side) are checks, not definitions; leftovers are the layout's zero padding. Less machinery than the statement-based pass it replaces — no second builder, no source table.

Gates re-run, all green: clear byte-diff 16/16 untouched, modular ZK e2e 3/3 (accept / tamper-reject / committed-program), jolt-verifier 74 + akita 76 + zk fixture suite 35/35 (all on the untouched verifier), jolt-kernels, legacy `muldiv --features host,zk`, `clippy --all` under `host` and `host,zk`, fmt.

0xAndoroid (commented):
**Verdict: no soundness or completeness bugs found.** The replay design holds up under adversarial reading; findings below are hardening, test coverage, and doc freshness — none blocking. Reviewed at aec5cacd4 against merge-base 8fdc42e4c, full diff plus the pre-existing components the PR wires in (`CommittedSumcheckRecorder`, `prove_batch_zk`, `blindfold::build`, layout/dimensions).

### Verified

- **Zero-verifier-diff holds, transitively.** `git diff 8fdc42e4c..HEAD -- crates/jolt-verifier crates/jolt-verifier-derive` is empty, and the verifier-consumed dependency crates change nothing the verifier executes: jolt-blindfold gains only the additive `assignment.rs` + `ProverError` variants (verify path untouched), jolt-openings changes only the prover-side `prove_batch_zk` return shape (`verify_batch_zk` untouched), jolt-kernels is prover-only.
- **The replay is a literal twin of the verifier spine.** `replay_stages` (jolt-prover/src/blindfold.rs:171–280) matches verifier.rs:54–148 call-for-call, including argument order, the `.zk()?` lowering, and the `Label(b"BlindFold")` position (verifier.rs:153). Stage order in `ZkStageWitnesses::in_protocol_order` matches `blindfold::build`'s `add_stage1..add_stage7`, uniskip-before-remainder. Proving the tail on the **replay** transcript rather than the forward one gives FS-position identity with the verifier by construction — the replay is a deterministic function of the wire bytes.
- **Blinding/randomness.** Every ZK-mode wire commitment is hiding with fresh `OsRng` entropy: per-round and per-output-row blinds in the recorder (`committed.rs` `commit_round`/`finish`), fresh aux-row blinds (assignment.rs:277), the Dory `open_zk` evaluation blind, hiding finishes for every witness/advice/precommitted-program commitment. Grep confirms no `finish_with_hint`/`finish_one_hot_column_major_chunks` call site escapes the `finish_streamed{,_one_hot}` mode wrapper. Zero-blind all-zero padding rows are safe (identity commitment, public content); short output-claim rows zero-resized to `row_len` are Pedersen-compatible (unused generators), and row/blind alignment against the flat witness is guaranteed by the layout's `pad_to_witness_row_boundary`.
- **Clear/ZK parity.** Identical stage recipes, kernels, and opening-point derivation in both modes; the recorder substitution plus the three cfg arms are exactly the transcript-discipline delta, inherited from components already pinned against the generated `verify_zk` drivers. `evaluate_at` (assignment.rs:340) matches `UnivariatePoly::evaluate` coefficient order. Clear path is byte-frozen — the only clear-mode change is the recorder-constructor indirection; byte-diff 16/16 in CI.
- **Rebase seams clean.** All `FusedColumns` stream-consumer finish points route through the mode wrapper; prover.rs keeps `W: JoltWitnessPlane<F>` intact and the widened unconditional bounds mirror `jolt_verifier::verify`'s existing precedent; the `TraceBackend` rename is consistent in both test files.

### Findings

**1. [robustness] `solve_products` discriminates definitions from checks positionally, and the invariant it relies on is unasserted** — jolt-blindfold/src/assignment.rs:304–326. Any constraint with C = `[(v, 1)]` and `v` unassigned is treated as a product definition. The load-bearing contract — `R1csBuilder::multiply` is the sole emitter of that shape, with `v` allocated after its operands — is documented but not enforced. A future builder change emitting an equality *check* with a fresh single-entry C side would silently become an assignment, and the misassignment self-satisfies, so `check_witness` cannot catch it; it surfaces only as an opaque BlindFold verify failure. Cheap tripwire: `debug_assert!(a_row.iter().chain(b_row).all(|(i, _)| i < target))` in the solve branch — the emission-order invariant, asserted where it's consumed.

**2. [test gap] ZK × advice is compiled but never executed.** Both `zk_e2e.rs` variants run adviceless. The hiding `commit_advice` finish and the advice claim-reduction constraints through BlindFold — historically the drift-prone family (the `init_eval` decomposition corollary) — have no modular ZK test. A muldiv-with-advice ZK variant would close the envelope the spec claims support for.

**3. [test gap] The aux solver has zero unit coverage.** `assignment_pipeline.rs` fixtures use `constant(...)` claim expressions, so no product auxiliaries exist and `solve_products` never runs at unit level (nor do the `UnsolvableProduct`/`WitnessVariableReassigned` paths); the muldiv e2e is its only exercise. One product-bearing claim fixture would pin the solver where shape errors are actually assertable.

**4. [design nit] `STAGE_DOMAINS` re-pins protocol constants prover-side** — jolt-prover/src/blindfold.rs:77–88 bakes jolt-claims' `OUTER/PRODUCT_UNISKIP_DOMAIN_SIZE` while the verifier lowering bakes jolt-r1cs' derived `SPARTAN_*_UNISKIP_DOMAIN_SIZE`. Consistent today, consistent-by-construction with the prover's own stage recipes, and drift fails loudly at proof time — but `BlindFoldStage` already carries `SumcheckDomainSpec`; retaining it on the built protocol would delete both the parameter and the dual constant source.

**5. [nit] Both structural self-checks are debug-only.** Replay-vs-forward transcript state (blindfold.rs:126) and R1CS satisfaction (assignment.rs:115). Release provers get their misassembly diagnosis as a downstream BlindFold failure instead of at the precise seam. The transcript check is one 32-byte compare — promoting it to a hard `ProverError` costs nothing and localizes exactly the drift class the spec's risk watch-list names.

**6. [docs] Stale text.** `prove`'s doc comment still says "assemble the clear-mode `JoltProof`" / "Supported envelope: transparent (clear) proofs" (prover.rs:31,43) — wrong in ZK builds. The PR description likewise still describes the rejected design ("promoted `verify_stages` + `blindfold::build_construction`") that aec5cacd4 reworked away — worth updating, since the zero-diff rework is the PR's headline.

---

Review by Claude Code (Fable), driven by Andrew's assistant.

0xAndoroid:
Addressed in `fix: harden the ZK replay seam and solver contract after review` + `test: cover ZK advice e2e and the product-auxiliary solver`:

1. **solve_products tripwire** — emission-order `debug_assert` added at the solve branch.
2. **ZK × advice** — `zk_advice_consumer_modular_proof_is_accepted`: hiding trusted-advice preprocessing commit (`commit_advice` under `zk`), prove-time hiding untrusted commit, advice claim reductions through BlindFold, accepted by the unmodified verifier.
3. **Aux solver unit coverage** — product-bearing fixture (stage 2 input claim = product of two stage-1 output-claim openings) runs assign → prove → verify with `auxiliary_values > 0` asserted.
4. **`STAGE_DOMAINS` dedup** — deferred: retaining `SumcheckDomainSpec` on the built protocol means changing `BlindFoldProtocol::from_parts`, code the verifier executes, which would break this PR's zero-verifier-diff property. Follow-up material.
5. **Replay self-check** — promoted to a hard `ProverError::InvariantViolation`; the R1CS satisfaction check stays debug-only (it is not a 32-byte compare).
6. **Docs** — `prove` doc no longer claims a clear-only envelope; PR description updated to the final replay design.

moodlezoup (approved):

PR #1696, opened 2026-07-23
test(fuzz): harden verifier dependency coverage

markosg04:
## Summary

- add per-target Cargo feature support to the fuzz runner so ZK-only targets compile and replay with `--features zk`
- expand `jolt-verifier` fuzzing with transparent and ZK structured proof tampering, BlindFold expression mapping equivalence, fresh fixtures, and semantic seeds
- add `jolt-r1cs` claim lowering and `jolt-openings` homomorphic Dory batch fuzz workspaces
- update fuzz docs/spec inventory, budgets, fixture generation commands, and current remaining roadmap

## Local validation

- `python3 -m unittest scripts.tests.test_fuzz`
- `python3 scripts/fuzz.py check`
- `python3 scripts/fuzz.py check --compile`
- `python3 scripts/fuzz.py --workspace jolt-r1cs --sanitizer none replay`
- `python3 scripts/fuzz.py --workspace jolt-openings --sanitizer none replay`
- `python3 scripts/fuzz.py --workspace jolt-verifier --target blindfold_expression_mapping_equivalence --sanitizer none replay`
- `python3 scripts/fuzz.py --workspace jolt-verifier --target proof_tamper_must_reject --sanitizer none replay`
- `python3 scripts/fuzz.py --workspace jolt-verifier --target zk_proof_tamper_must_reject --sanitizer none replay`
- `cargo fmt -q`
- `cargo nextest run -p jolt-prover-legacy muldiv --cargo-quiet --features host`
- `cargo nextest run -p jolt-prover-legacy muldiv --cargo-quiet --features host,zk`
- `cargo nextest run -p jolt-verifier --features prover-fixtures --test generate_fuzz_fixture --run-ignored ignored-only --cargo-quiet`
- `cargo nextest run -p jolt-verifier --features prover-fixtures,zk --test generate_fuzz_fixture --run-ignored ignored-only --cargo-quiet`

Merged latest `origin/main` before opening this PR.


0xAndoroid (commented):
## Fuzz review — verifier-dependency coverage

**Verdict: approve with nits.** Production changes are minimal and clean: the dead `r0` field is removed from the two uni-skip provers, and one `#[cfg(feature = "fuzzing")]` differential-eval helper (`evaluate_mapped_expression`) is added to the BlindFold module. The `jolt-eval` invariant edits are sound (num_vars cap for fuzzer throughput; three per-sponge symmetry fuzz targets merged into one fuzzer-selected target, per-sponge Test/RedTeam coverage retained).

The fuzz work itself is high quality. The tamper harnesses verify the honest fixture *accepts* before mutating (so every iteration reaches the verifier crypto core, not a serde front door), skip no-op mutations (no spurious "accepted a tampered proof" trips), and `honest_then_corrupt` is false-positive-proof by construction via its discharge oracle (an `Ok` only asserts when the reduced claim is genuinely true of the product). No false-positive / flaky-CI assertion was found in any of the ~30 targets reviewed.

### Genuinely new coverage (vs pre-PR targets and the manual audit probes)
- **Structured tampering of real accepted proofs** — transparent (`proof_tamper_must_reject`) and ZK (`zk_proof_tamper_must_reject`); automates/randomizes the mutation classes the manual audit hit by hand (commitment swap, claim bump, round drop, `trace_length`/`ram_K`, mode flip) over 3 transparent + 2 ZK fixtures.
- **BlindFold <-> clear expression parity** (`blindfold_expression_mapping_equivalence`) — differential over the exact `JoltExpr` -> BlindFold-public remap; hits the ZK residual risk the audit flagged as the highest-value follow-up.
- **Dory homomorphic batch openings** (clear + ZK), **mode confusion**, GLV / batch-addition differentials, canonical decode boundary, signed-accumulator differential, transcript injectivity/determinism, sumcheck clear-vs-compressed and honest-then-corrupt.

### Highest-priority issue (would fix before merge)
Weekly `jolt-verifier` fuzz job exceeds its 240-min timeout — 300 min of sequential mutation before build/replay/cmin/coverage — so weekly corpus minimization, corpus save, and coverage for the *flagship* workspace never run. Detail inline at the workflow timeout line.

### Residual-risk map (vs the verifier soundness audit)
- **Still open (not covered):** the prover-chosen `rw_config` / `one_hot_config` surface — audit Finding 1 (validators never called by the verifier) and Finding 2 (validator-legal RAM phase splits the prover mis-proves). `validate_inputs_no_panic` passes the fixture config verbatim; inline note on how to close it.
- **Partially covered:** ZK/BlindFold parity gains a differential expression oracle + a ZK tamper harness, but (as the audit noted) Pedersen binding under the Dory SRS and the vendored dory-pcs ZK sigma proofs remain outside any in-repo target.
- **Well covered now:** sumcheck core, Fiat-Shamir framing, canonical field/group decode, Dory clear + ZK opening soundness.

### Budget / scope nits (non-blocking)
- `jolt-hyperkzg` is a dev-dependency only; Dory is the production PCS and hyperkzg is off the deployed verifier path, yet it draws 150s PR / 110 min weekly. `tracer` decode/uncompress targets run prover/preprocessing-side (not on the verifier trust boundary), and `decode_differential` is labeled `focus = "soundness"` though it is a completeness/consistency differential between two RV64IMAC decoders. Budgets are self-described as provisional priors (tracer Cargo.toml), so this is a recalibration note — but the `focus` label reads wrong.
- `uncompress_differential` (2^16-halfword space) and `mle_materialize_equivalence` (exhaustively covered by an existing unit test) saturate near-instantly; their budgets buy little.

### Micro-nits
- `canonical_decode_boundary.rs:46` — `BigUint::from(delta) % modulus()` is a no-op (`delta < 2^64 < r`).
- `signed_accumulator_diff.rs` caps the 256-bit magnitude to 128 bits, so 3/4-limb magnitudes are never exercised (coverage gap, not a bug).
- `encoding_injectivity.rs:66` — comment mis-describes the end-split case (it yields an extra empty-chunk frame, not "the original structure").

---
*Review generated by Claude Code on Andrew's (0xAndoroid) behalf.*

.github/workflows/fuzz-crates.yml
```diff
@@ -4,55 +4,184 @@ on:
...
+          matrix="$(python3 scripts/fuzz.py inventory --format github)"
+          echo "matrix=$matrix" >> "$GITHUB_OUTPUT"
+
   fuzz:
+    name: Fuzz ${{ matrix.workspace }}
+    needs: discover
     runs-on: ubuntu-latest
+    # The weekly jolt-eval profile alone is 130 minutes of mutation before
+    # build, replay, cmin, and coverage; the spec requires at least 180.
+    timeout-minutes: 240
```
0xAndoroid: **The weekly `jolt-verifier` job will always hit this 240-min timeout.** `fuzz.py run` executes a workspace's targets sequentially (`run_for_targets`, one `-max_total_time` per target, no `-jobs`), and `jolt-verifier/fuzz` weekly budgets sum to 7200+5400+3600+900+900 = **18000 s = 300 min of mutation** — before build, `replay`, `cmin`, and `coverage`. Since `Minimize weekly corpus`, `Generate weekly coverage`, `Save trusted corpus`, and `Upload weekly coverage` are all `if: success()`, the timed-out job fails first and the flagship workspace's weekly corpus is never minimized/saved and its coverage never generated. jolt-verifier at 300 min is the unique workspace over the cap (next is jolt-sumcheck at 175). The comment just above sizes the timeout against jolt-eval (which is actually 115 min, not 130) and overlooks jolt-verifier.

Fix: raise `timeout-minutes` to >=360, or expand the matrix to one entry per target so each fits, or trim the weekly verifier budget.

crates/jolt-verifier/fuzz/fuzz_targets/validate_inputs_no_panic.rs
```diff
@@ -0,0 +1,67 @@
...
+    io.inputs = vec![0u8; input_len.min(1 << 16)];
+    io.outputs = data[18..].to_vec();
+
+    let _ = validate_inputs_from_parts::<DoryScheme, Pedersen<Bn254G1>>(
+        preprocessing,
+        &io,
+        trace_length,
+        ram_k,
+        proof.trace_polynomial_order,
+        proof.one_hot_config,
```
0xAndoroid, line 69: `one_hot_config` is fed verbatim from the fixture and never mutated, and `rw_config` is not a parameter of `validate_inputs_from_parts` at all — so the prover-chosen configuration surface is unfuzzed. That is precisely the open item from the verifier soundness audit: **Finding 1** (`ReadWriteConfig::validate` / `OneHotConfig::validate` are never called by the verifier) and **Finding 2** (validator-legal RAM phase splits that the prover mis-proves). Not a defect in this PR, but worth stating so the coverage matrix does not read as closing those. A config mutation class (one_hot_config / rw_config drawn from fuzzer bytes — here for the no-panic check, and prover-side in a config-override e2e for the mis-proving case) would actually exercise it.

.github/workflows/fuzz-crates.yml
```diff
@@ -4,55 +4,184 @@ on:
   push:
     branches: [main]
     paths:
-      - "crates/*/fuzz/**"
-      - "crates/*/src/**"
+      - "crates/**"
```
0xAndoroid, line 8: `common/**` is missing from both the `push` and `pull_request` path filters, but `common` lives at repo root (not under `crates/**`) and is a direct dependency of `jolt-verifier-fuzz` and the wider graph. A PR that touches only `common/` (e.g. `MemoryLayout` / `JoltDevice`, which `validate_inputs_no_panic` fuzzes) edits no filtered path and no `Cargo.lock`, so fuzzing silently does not run. Add `- "common/**"`.

scripts/fuzz.py
```diff
@@ -0,0 +1,785 @@
...
+                    target.name,
+                    str(corpus),
+                ]
+            elif args.command == "coverage":
+                directories = corpus_directories(workspace, target.name)
+                if not directories:
+                    raise FuzzConfigurationError(
+                        f"{workspace.name}/{target.name} has no corpus inputs"
+                    )
+                command = [
```
0xAndoroid, line 506: `coverage` is the only subcommand that omits `*sanitizer` from the cargo-fuzz invocation (compare `run` / `replay` / `cmin` / `build` above). The documented macOS workaround (`--sanitizer none`, FUZZING.md) therefore has no effect on `coverage`, so a local `fuzz.py coverage` run still tries to link ASan and fails. CI is unaffected (Linux), but thread the flag through for parity.

crates/jolt-lookup-tables/fuzz/fuzz_targets/mle_materialize_equivalence.rs
```diff
@@ -0,0 +1,51 @@
+#![no_main]
```
0xAndoroid, line 1: This asserts `materialize_entry(i) == MLE(i)` over fuzzer-drawn indices, but `mle_full_hypercube_test` (`src/tables/test_utils.rs`) already enumerates that identical invariant **exhaustively** over the full 2^16 domain (XLEN=8) in every table's test module. The fuzzer samples a strict subset of a space the unit tests already saturate; the only marginal coverage is `LookupTableKind` dispatch glue. Consider dropping the workspace (or repurposing its 30s PR / 900s daily budget) rather than paying build + CI cost for a redundant oracle. Harness itself is correct — no false-positive risk.

tracer/fuzz/fuzz_targets/decode_differential.rs
```diff
@@ -0,0 +1,310 @@
...
+            matches_tracer_mask::<tracer::instruction::virtual_host_io::VirtualHostIO>(word)
+        }
+        jolt_riscv::SourceInstruction::VirtualRev8W(_) => {
+            matches_tracer_mask::<tracer::instruction::virtual_rev8w::VirtualRev8W>(word)
+        }
+        SourceInstructionKind::XOR => matches_tracer_mask::<tracer::instruction::xor::XOR>(word),
+        SourceInstructionKind::XORI => {
+            matches_tracer_mask::<tracer::instruction::xori::XORI>(word)
+        }
+        _ => true,
```
0xAndoroid: The catch-all silently waves through any `SourceInstructionKind` not spelled out above. Cross-checked against `for_each_instruction_kind!` in jolt-riscv: ~46 of 143 variants hit this arm — every `Virtual*` (VirtualLW/VirtualSW/VirtualAdvice*, VirtualAssert*, VirtualXORROT*, …) and every `FIELD_*` kind. For those, a word where `decode_instruction` classifies kind K but the full `word & MASK == MATCH` does not hold skips the tracer decode + operand comparison entirely — i.e. precisely the reserved-operand-bit divergence class this harness exists to catch is suppressed for the entire virtual/field-inline surface, not just "allowed". The explicit `VirtualAdviceLen`/`VirtualHostIO`/`VirtualRev8W` arms a few lines up show the intent was to cover these; the rest fell through. Also worth noting `matches_program_kind_tracer_mask` and `Instruction::decode` draw on the same tracer-side tables, so for the covered kinds the filter mostly approximates `tracer decode is Ok` rather than acting as independent validation.

crates/jolt-verifier/fuzz/fuzz_targets/proof_tamper_must_reject.rs
```diff
@@ -0,0 +1,258 @@
...
+    let changed = match mutation {
+        0 => {
+            proof.trace_length = 3;
+            true
+        }
+        1 => {
+            proof.ram_K = 3;
+            true
+        }
+        2 => {
```
0xAndoroid, line 169: Two of the first three mutation classes can never trip the assert. Class 2 (clear proof → `ZkConfig::BlindFold`): `validate_proof_consistency` rejects Clear claims under zk=true with a typed `Err(UnexpectedOpeningClaims)` before any crypto runs, and no plausible verifier regression makes it *accept* — the class permanently exercises only the shape validator. Class 1 (`ram_K = 3`) is below `min_ram_k` for every fixture and dies deterministically in `validate_inputs`; only class 0 (`trace_length = 3`) ever reaches stage code. That's 3/16 of this target's entropy on typed-Err paths a unit test covers better. Suggest biasing to boundary values (`ram_K` = a legal-but-wrong power of two ≥ min_ram_k, `trace_length` = a power of two that mismatches the proof's round structure) or documenting these as panic guards, not soundness probes.

crates/jolt-verifier/fuzz/fuzz_targets/zk_proof_tamper_must_reject.rs
```diff
@@ -0,0 +1,235 @@
...
+            };
+            *value += Fr::from_u64(1);
+            true
+        }
+        12 => {
+            if public_io.inputs.is_empty() {
+                return;
+            }
+            let position = index % public_io.inputs.len();
+            public_io.inputs[position] ^= operation.max(1);
```
0xAndoroid, line 222: Both ZK fixtures are muldiv-lineage; there is no ZK advice-consumer counterpart to the transparent `advice-consumer-bundle.bin`, so the BlindFold advice-reconstruction constraints (untrusted/trusted `FinalScale`, `ByteDecode`) never see a tampered accepted proof — the highest-risk ZK surface per the audit. The generator already has `standard_advice_consumer_case`; a `zk_advice_consumer_case` fixture would close this. Also: unlike the transparent harness (class 13 swaps in another fixture's program), no cross-program swap exists here, so ZK program-image binding is only probed via the default-commitment classes.

.github/workflows/fuzz-crates.yml
```diff
@@ -4,55 +4,184 @@ on:
...
+        run: python3 scripts/fuzz.py --workspace "${{ matrix.workspace }}" --target-triple x86_64-unknown-linux-gnu cmin
+      - name: Generate weekly coverage
+        if: success() && steps.profile.outputs.name == 'weekly'
+        run: python3 scripts/fuzz.py --workspace "${{ matrix.workspace }}" --target-triple x86_64-unknown-linux-gnu coverage
+      - name: Save trusted corpus
+        if: success() && github.event_name == 'schedule' && github.ref == 'refs/heads/main'
+        uses: actions/cache/save@v6
+        with:
+          path: ${{ matrix.directory }}/corpus
+          key: fuzz-corpus-v1-${{ runner.os }}-${{ matrix.workspace }}-${{ github.run_id }}
```
0xAndoroid, line 192: The corpus save key embeds `github.run_id`, so every qualifying run uploads a *new* cache entry and restore relies on the prefix restore-key picking the newest. Entries therefore accumulate one per scheduled run per workspace forever (~10 GB repo cache quota is shared and LRU-evicted across *all* caches, so growing dead corpora can evict the `fuzz-build` caches and unrelated CI caches). Nothing ever deletes the old entry. Standard pattern: fixed key + `gh cache delete` of the previous entry before save, or save-only-on-success into a fixed key with a pre-delete step.

scripts/fuzz.py
0xAndoroid, line 506: Knock-on effect of the sanitizer omission noted on this line: because `coverage` builds the target *without* ASan while build/replay/run/cmin all build *with* ASan into the same per-workspace `target/` dir (the one cached by `fuzz-build-v1-*`), the weekly coverage step triggers a full re-instrumented rebuild, and the workflow's `actions/cache` save of that directory afterwards stores the coverage-profiled artifacts — so the next run's restore is cold for ASan and rebuilds everything again. The weekly job effectively never benefits from the build cache. Separate `--target-dir` (or cache namespace) for coverage would keep the ASan cache warm.

0xAndoroid (commented):
Second pass, net-new findings only. One real hole (decode_differential mask map), two must-reject oracle gaps (degenerate mutation classes; ZK fixture diversity), two CI issues. No duplicates of the first review.

*Review generated by Claude Code on Andrew's (0xAndoroid) behalf.*

0xAndoroid (approved):
Approving. Two non-blocking items remain open as follow-ups: the decode_differential catch-all still skips ~46 Virtual*/FIELD_* kinds (thread left unresolved), and the coverage build shares the ASan target-dir, so weekly coverage runs thrash the build cache.

*Review assisted by Claude Code on Andrew's (0xAndoroid) behalf.*

tracer/fuzz/fuzz_targets/decode_differential.rs
0xAndoroid: Addressed in 526cc2b1a: the catch-all now panics on any kind without a mask entry instead of assuming the mask matches. Every word-decodable kind in this build (no `field-inline`, so no FIELD_* arm) is mapped explicitly; the remaining Virtual* kinds are sequence-only and never come out of `decode_instruction`, so a future word-decodable addition trips the fuzzer loudly with a pointer to the map.

scripts/fuzz.py
0xAndoroid: Addressed in 526cc2b1a: the weekly coverage step now builds into its own `target-coverage` dir with a dedicated `fuzz-coverage-build-v1-*` cache, so the `-C instrument-coverage` artifacts no longer invalidate the ASan build cache (and vice versa) every week.

0xAndoroid (approved):
LGTM, test only.

0xAndoroid (approved):
LGTM, test only.

PR #1697, opened 2026-07-24
ci: better coverage, better testing, and cargo mutants

markosg04:
This implements [`specs/test-quality-ci.md`](specs/test-quality-ci.md). Each PR now checks per-crate coverage floors over the jolt-verifier dependency closure (cumulative across each crate's feature paths) and two pinned soundness floors: error-variant coverage and the tamper-manifest active ratio. Tamper rejections must happen no later than their manifest-documented verifier phase, which caught one mis-documented target. A hosted routine scores new and modified tests against a rubric; it doesn't block merges and skips until the `TEST_QUALITY_ROUTINE_ID` variable and `ANTHROPIC_TEST_QUALITY_AUTHORIZATION` secret are set. The coverage-raising work landed in this same PR, so the merged baseline is the enforced bar (cumulative line coverage, src/ only):

| crate | before -> after | crate | before -> after |
|---|---|---|---|
| jolt-verifier | 26.3 -> 87.8 | jolt-akita | 71 -> 88.2 |
| tracer | 63 -> 90.3 | jolt-transcript | 62 -> 98.5 |
| common | 70 -> 99.5 | jolt-riscv | 72 -> 86.4 |
| jolt-program | 76 -> 86.0 | jolt-blindfold | 67 -> 93.8 |

All 20 in-scope crates are at 81 or above. Two production bugs surfaced along the way, fixed here: `bind_low_to_high_reusing_scratch` under-reserved scratch capacity for sizes just below the bind width, and guest cross-compiles inherited cargo-llvm-cov's `RUSTC_WRAPPER` (tracer and the legacy host builder now scrub it).

Nightly, cargo-mutants reports on the soundness-critical modules without blocking merges. A mutant only counts as missed relative to the suites that actually run, so the jolt-verifier target enables `prover-fixtures`: a mutant nulling the uniskip-to-outer-remainder claim wiring survived the default-feature oracle and the fixture suites kill it (measured on the stage1 file: 1 missed -> 0, 8 minutes for its 22 mutants). The claims derive macro, previously untested, is also a mutation target now: its expansions are pinned as snapshots, and its first run turned up two missing rejection tests (where-clauses, lifetime parameters), added here. Its 38 mutants all die, 34 of them at consumer compile time.

Full workspace suite, muldiv e2e, clippy, and every gate are green in CI in both standard and zk modes. Not measured: the whole nightly's mutation runtime under the fixture-backed oracle; the module list may need resizing if it overruns the budget.


.github/workflows/test-quality.yml
```diff
@@ -0,0 +1,67 @@
...
+# Setup: mint the routine and store its id in the repo variable
+# TEST_QUALITY_ROUTINE_ID and its bearer token in the secret
+# ANTHROPIC_TEST_QUALITY_AUTHORIZATION. Until both exist this workflow skips.
+#
+# Uses `pull_request_target` (like claude-routines.yml) so secrets are
+# available on PRs from forks; only the PR *diff metadata* is inspected here —
+# PR code is never checked out or executed.
+
+on:
+  pull_request_target:
```
0xAndoroid: **[should-fix / security]** This auto-fires the hosted routine for any fork PR touching a test file, on every `opened`/`synchronize` — unlike `claude-routines.yml`, where firing is gated on a label that requires triage/write permission to apply. Once `TEST_QUALITY_ROUTINE_ID` is set, an arbitrary fork PR burns a routine run per push (cost DoS) and feeds attacker-controlled diff content to the routine unconditionally. The `detect` job itself is safe (no checkout, file metadata stays inside `jq`). Suggest: auto-fire only for same-repo PRs (`github.event.pull_request.head.repo.full_name == github.repository`) and fall back to the label-gated path for forks.

crates/jolt-prover-legacy/src/host/program.rs
```diff
@@ -214,6 +214,19 @@ impl Program {
             let mut cmd = Command::new(&jolt_cmd);
             cmd.args(&args);
 
+            // The riscv guest must not inherit host coverage instrumentation:
+            // cargo-llvm-cov injects `-C instrument-coverage` via RUSTC_WRAPPER
+            // (older versions via RUSTFLAGS), which fails the cross-compile
+            // (no profiler_builtins for the guest target).
+            let _ = cmd
+                .env_remove("RUSTFLAGS")
```
0xAndoroid: **[should-fix]** This scrubs `RUSTFLAGS`/`RUSTC_WRAPPER` for every guest build, not just under coverage: `RUSTC_WRAPPER=sccache` users silently lose caching for guest cross-compiles, and deliberately set guest `RUSTFLAGS` are dropped. Consider gating the scrub on the coverage wrapper actually being active (e.g. `std::env::var_os("CARGO_LLVM_COV").is_some()`), which keeps the fix and preserves existing workflows. (The tracer-side copy of this block is test-only, so it is fine as is.)

crates/jolt-verifier/tests/support/tamper_manifest.rs
```diff
@@ -1220,6 +1223,149 @@ pub fn assert_zk_target_active(name: &str) {
     );
 }
 
+/// The phase where a target's rejection is documented to fire, derived from
+/// its disposition.
+pub fn expected_rejection_phase(target: TamperTarget) -> VerifierPhase {
```
0xAndoroid, line 1228: **[should-fix]** The Active⇒test direction is unenforced: `assert_verifier_fixture_tamper_rejects` proves test⇒Active, and `ci/soundness_metrics.py tamper-ratio` counts `TamperCoverage::Active` occurrences textually — so flipping a target to Active without adding its test (or deleting an existing tamper test) still passes CI and inflates the 111/113 metric. A registry test asserting set-equality between Active target names and the names actually exercised by tests would close the loop.

scripts/ci/soundness_metrics.py
```diff
@@ -0,0 +1,307 @@
...
+                    break
+            key = f"{name}::{variant}"
+            if located:
+                exercised[key] = hit
+            else:
+                unlocatable.append(key)
+    return {"exercised": exercised, "unlocatable": unlocatable}
+
+
+def tamper_ratio(manifest_source: str) -> tuple[int, int]:
```
0xAndoroid, line 152: **[nit]** `tamper_ratio` counts manifest source text (minus `==`/`matches!` lines): a comment mentioning `TamperCoverage::Active`, or a helper function returning it, would count as a target. Fine while the manifest stays literal constructor arrays; emitting the counts from a `#[test]` into a file CI reads (or parsing the constructors) would make it robust to refactors.

0xAndoroid (commented):
Review focus: the two production changes, CI gate soundness, test substance vs padding, and the hosted-routine workflow security surface. No blocking findings; 3 should-fix and 2 nits as inline comments.

**Verified**
- Both production fixes are correct and minimal. `bind_low_to_high_reusing_scratch`: the old `reserve(half - capacity)` under-reserves for *any* `0 < capacity < half` after `clear()` (len = 0), panicking at `spare_capacity_mut()[..half]` in the parallel path; `reserve(half)` is the right fix. The RUSTC_WRAPPER scrub is the only other production change — every other src/ hunk in the PR sits inside a `#[cfg(test)]` module (checked by hunk position, not by trusting the description).
- Blake2b KATs: independently recomputed the pinned vectors from the documented formula (`hashlib.blake2b(digest_size=32)`) — all match.
- Local runs: `coverage_gate.py self-test` / `check-config`, `soundness_metrics.py self-test` / `tamper-ratio` (111/113 = 98.2%) all pass; 1,106 tests green across 11 touched lib crates (debug); 20/20 release `prover-fixtures` tamper tests green, including the newly activated targets and the new phase-bound assertions.
- Gate design is sound where it matters: a crate missing from the coverage export is a hard failure rather than a silent pass, floor drift is checked in both directions, and the cumulative feature-path merge is a correct line-level union.

**Test quality** (10 files sampled across verifier/blindfold/tracer/riscv/akita/crypto): consistently real — independent oracles (binutils-anchored RVC vectors plus an assembler-vs-toolchain guard so an encoder bug cannot cancel a decoder bug, spec-transcribed encoders, table-derived GLV bounds), exhaustive per-wire tamper sweeps in the akita suite, exact-variant error assertions. No coverage-padding found.

**Notes, no action required**
- The PR-body claim "every `Err(...)` construction site must execute under test" overstates the implemented metric: a variant counts as exercised when at least one line referencing `Enum::Variant` (including `#[cfg(test)]` code inside src/) executes, with per-crate floors of 42–100%. Fine as a ratchet; worth phrasing accurately in the docs.
- Audit residuals this PR does not close (and does not claim to): no e2e fixture proves in AddressMajor layout — the new `trace_polynomial_order` tamper flips the flag on a CycleMajor proof, which exercises transcript binding, not the AddressMajor verify path; ZK advice is still exercised at a single geometry; the RAM phase-split prover completeness bugs and the uncalled `ReadWriteConfig::validate` / `OneHotConfig::validate` verifier-side validators remain open. The new `proof.one_hot_config` tamper target documents transcript binding as the defence — that is the single-layer defence the soundness audit flagged as insufficient on its own.
- coverage.yml's ZeroOS toolchain fetch matches the existing rust.yml pattern (same cache key, no checksum) — pre-existing practice, not a regression.
- Nightly mutants summarize step: an empty `mutants-report/` leaves the glob unexpanded and emits a bogus `*` row — cosmetic.

---
*This review was generated by Claude Code on behalf of Andrew (@0xAndoroid).*

scripts/ci/coverage-floors.toml
```diff
@@ -0,0 +1,82 @@
...
+
+[scope]
+root = "jolt-verifier"
+# Verification-relevant crates that are not normal deps of the root:
+# the commitment schemes are dev-deps behind trait abstractions, and the
+# tracer defines the execution semantics the verified statement is about.
+extras = ["jolt-dory", "jolt-akita", "jolt-hyperkzg", "tracer"]
+
+[paths]
+# Feature paths per crate; crates not listed run "default" only.
```
0xAndoroid, line 29: **[nit]** Files gated behind features not listed in any `[paths]` entry are never compiled into a coverage run, so they are absent from llvm-cov's export and silently drop out of the floor denominator (the solinas NOTE above is exactly this case). `check-config` cannot see it. An informational `plan`-time listing of crate features not covered by any declared path would make the blind spot visible.

scripts/ci/soundness_metrics.py
```diff
@@ -0,0 +1,307 @@
...
+            if opened and depth == 1 and j > i:
+                stripped = lines[j].strip()
+                if not stripped.startswith(("#", "//", "/*", "*")):
+                    variant = VARIANT_RE.match(lines[j])
+                    if variant:
+                        variants.append(variant.group(1))
+            if opened and depth == 0:
+                break
+            j += 1
+        enums[name] = (start, j + 1, variants)
```
0xAndoroid, line 84: **[nit]** `find_error_enums` never checks that the enum body actually closed: if a file ends (or a malformed body runs on) before `depth` returns to 0, the `while` loop exhausts the file, `(start, j+1, variants)` is recorded with `j == len(lines)`, and `i = j + 1` skips the remainder — every later enum in the same file silently vanishes from the inventory, so its variants drop out of both numerator and denominator instead of surfacing as `unlocatable`. Verified locally: an unterminated first enum hides a valid second enum entirely. Cheap guard: after the loop, `if not (opened and depth == 0): i += 1; continue` (or record with an explicit unterminated flag and report it).

scripts/ci/coverage_gate.py
```diff
@@ -0,0 +1,510 @@
...
+            features = path["features"]
+            prefixed = (
+                ",".join(f"{name}/{feature}" for feature in features.split(","))
+                if features
+                else None
+            )
+            slug = name
+            if features:
+                slug += "--" + features.replace(",", "+")
+            if path["release"]:
```
0xAndoroid, line 175: **[nit]** `line_execution_counts` destructures but ignores `is_region_entry` (`_entry`), so non-entry continuation segments — which still carry `has_count=True` and their own counts — are treated as range starts whose count extends to the next segment line. In dense region layouts (nested closing regions at the same line) this can ascribe a region's count to lines llvm-cov's own region mapping leaves uncovered (or vice versa), so per-line results here can diverge slightly from `llvm-cov report`'s line table. Only affects corner-case attribution for the floors and the error-variant metric — the self-tests cover gap regions but not non-entry segments. Either honor the flag (skip non-entry segments as range starts) or note the approximation in the docstring.

0xAndoroid (commented):
Second-pass review, generated by Claude Code on behalf of Andrew (0xAndoroid). Scope: net-new findings only, complementing review 4775019132 — focused on the CI Python scripts' edge cases, the workflow wiring, and the less-covered sweep tests (tracer memory/panic/elf batteries, malformed-ELF fixtures, uncompress battery, GLV reconstruction tests).

Checked and cleared (no comment): coverage.yml wiring (`check-config` runs, plan is derived from the TOML so config and job can't drift), floors match the live `cargo metadata` closure exactly (20/20), tamper ratio text count matches the manifest (111/113), the new tracer memory/ELF/panic and jolt-program malformed-ELF tests are genuine oracles with specific error-variant assertions, and the segment-reconstruction self-tests cover the gap-region case. Two net-new nits inline, both in the Python gate scripts.

0xAndoroid:
Can you also move scripts to `scripts/ci`

.github/workflows/test-quality.yml
markosg04: Fixed in 14e1b46a4 — the fire job now additionally requires `github.event.pull_request.head.repo.full_name == github.repository`, with a comment explaining fork PRs fall back to the label-gated routines.

crates/jolt-prover-legacy/src/host/program.rs
markosg04: Fixed in ab0d8aa52 — the scrub is gated on `CARGO_LLVM_COV` being set (verified empirically that cargo-llvm-cov exports it to test processes), so sccache wrappers and deliberate guest RUSTFLAGS survive normal builds. Tracer's test-only copy left unconditional as you suggested.

crates/jolt-verifier/tests/support/tamper_manifest.rs
markosg04: Fixed in d0e2fafca — a manifest audit now asserts every Active target name is referenced by the tampering test sources. It immediately caught `zk.vector_commitment_setup`, whose fixture test existed but never named its target; that test now anchors to the manifest via `assert_zk_target_active`.

scripts/ci/soundness_metrics.py
markosg04: Fixed in fdd5d94d7 — `tamper_ratio` is now anchored to the target-constructor calls rather than raw token counting. The honest count is 109/111 (the textual count was inflated by exactly two non-target mentions), still above the 0.98 floor.

scripts/ci/coverage-floors.toml
markosg04: Fixed in fdd5d94d7 — `check-config` and `plan` now print an informational note listing per-crate features that no declared path compiles (excluding the default-feature closure), so the blind spot stays visible. Current output flags akita/field-inline/alternate-transcript modes, matching expectations.

scripts/ci/soundness_metrics.py
markosg04: Fixed in fdd5d94d7 — unterminated bodies now skip only the header line so later enums stay in the inventory; self-test added reproducing your scenario (unterminated first enum, valid second enum).

scripts/ci/coverage_gate.py
markosg04: Fixed in fdd5d94d7 — the reconstruction now mirrors llvm's line-table rule: region entries take precedence on their line, continuation counts only fill lines without entries. Self-test added for the uncovered-entry-under-carried-count case. Floors re-verified unchanged against the final measurement data.

markosg04:
Moved to `scripts/ci/` in fdd5d94d7 (scripts, rubric, and configs together — the gate resolves its config relative to its own location, and repo-root discovery now walks up to Cargo.toml). All review items addressed; CI re-running.

0xAndoroid (approved):
All review findings addressed. The Active⇒exercised manifest audit added in d0e2faf caught a real gap (zk.vector_commitment_setup, fixture test never named its target) — good sign the gate has teeth.

*Review assisted by Claude Code on Andrew's (0xAndoroid) behalf.*

0xAndoroid:
Merged main (af14aff3, includes #1696 at 49aa0fa0) into this branch to resolve conflicts (merge commit 029bee575, no rebase). Conflicting files: `crates/jolt-akita/src/adapters.rs`, `crates/jolt-verifier/tests/support/tamper_manifest.rs`, `crates/jolt-verifier/tests/soundness/tampering/preamble.rs`. Kept: the adapter unit tests added here (dropped `cpu_backend_tests`, whose `cpu_backend_with_scratch` was removed on main by #1876); main's `TamperCoverage::Active` entry-address manifest entry and its `entry_address_mismatch_rejects_now` test (dropped the duplicate `tampered_bytecode_entry_address_rejects` and its `clippy::panic` expect); mapped the new `VerifierError::PreprocessingDigestFailed` (#1903) to the preamble phase in `observed_rejection_phase`. #1696 (fuzz tree) merged cleanly. CI green.

0xAndoroid (approved):
LGTM

PR #1702, opened 2026-07-26
ci: add Fiat-Shamir soundness audit jobs

markosg04:
## Summary

Fiat-Shamir binding regressions currently surface as generic proof failures, without showing whether an invalid proof would pass under fixed challenges. This lands the blocking first slice of the verifier-focused campaign: reviewed transcript/source inventories plus live frozen-challenge attacks across Dory clear, Dory ZK, and Akita clear. Akita ZK is compile-probed so support cannot appear without entering the matrix.

## Changes

- Record typed verifier challenges by protocol scope and replay honest tapes against coordinated invalid proofs.
- Block unreviewed changes to verifier-source schemas, transcript absorption expressions, challenge-shaped calls, generated batching draws, and scope annotations.
- Exercise preprocessing replay, stage-1 sumcheck equivocation, and Dory final-opening cancellation in dedicated CI jobs.

## Testing

The full verifier suites pass in Dory clear, Dory ZK, and Akita clear modes; legacy muldiv passes with `host` and `host,zk`; the complete local FS matrix passes.

## Security Considerations

The changes add verifier soundness tests and test-only scope instrumentation. They do not change proof serialization or production transcript behavior.

## Breaking Changes

None


0xAndoroid (commented):
**Review: FS soundness audit jobs** — test-only + CI, cannot affect proof format or production verify path (confirmed: `fs_scope`/`fs_scope_guard!` expand to `#[cfg(feature="fs-audit")]`-gated `let _guard`, no-op without the feature; `fs_audit` module is feature-gated). CI is green across the three FS legs. The dynamic oracle is well-built and the docs are strong. Not blocking — the notes below are coverage/overclaim gaps, mostly the delta between `specs/fs-program-analysis.md` (the full 5-job campaign) and what this "first slice" actually implements (2 jobs: a source-token tripwire + 3 hand-written live attacks).

**Solid:**
- The 4-step frozen/production oracle in `fs_attacks.rs` is a correct weak-FS isolator. The `frozen.consumed == frozen.expected` assertion is the right touch — it forces the frozen verifier to run to completion so the "frozen accepts" leg can't pass vacuously via an early unrelated reject. `equivocate_stage1_clear` genuinely preserves every frozen algebraic check (endpoint-sum + eval-at-challenge correction) while falsifying the claimed sum.
- Encoding `ChallengeKind` into the challenge-site identity freezes the 125-bit `challenge()` vs 256-bit `challenge_scalar()` choice per call site — that distinction is load-bearing (e.g. stage-2 output-address point) and this is exactly the right thing to pin. FrozenTranscript also asserts kind on replay.
- Akita+ZK compile-probe is a clean forcing-function; the grepped "features are mutually exclusive" is emitted by a repo-owned `compile_error!`, not the compiler, so the probe won't rot on toolchain drift. Both the YAML and `scripts/ci/fs-soundness.sh` variants correctly treat an unexpected build failure as a job failure (`set -e` + `grep -q`).

**Gaps:**

1. **Static inventory cannot see absorb↔squeeze interleaving (the canonical weak-FS bug).** `record_absorb` and `record_challenge` use independent per-`(context,kind)` ordinal counters into separate `BTreeSet`s. Moving a `challenge_scalar()` from after an `append(...)` to before it changes *neither* inventory — the exact FS-2 "bind before squeeze" property the `testing-gates.md`/spec credit to `fs-obligations` is out of the tripwire's resolution. It's only backstopped by the dynamic honest-fixture assertion (`assert!(honest.is_ok())`) + pre-existing e2e completeness, not by the static gate. Within-kind reordering with *distinct* expressions is caught (ordinal+expr in key); same-expression absorb swaps and any absorb-vs-challenge reorder are not.

2. **Source census is a hardcoded 14-type allowlist (`is_source_type`), not the derived `FsVisit` the spec's FS-1 promises.** A field added to a *listed* struct trips the gate; a *new* prover-controlled proof/claim carrier type does not auto-enroll and can introduce an FS source silently. "Complete inventory" is manual and can under-cover. (The `SumcheckBatch`-derive auto-enrollment for `batching_coefficient[*]` is the right pattern — extend that approach to sources.)

3. **Live-attack coverage is 3 of 13 scopes** (Preamble digest, Stage1 equivocation, Stage8 opening-cancellation). Stages 2–7 have no canary, and stage-1 equivocation is clear-only (`#[cfg(not(zk))]`) — so **ZK sumcheck FS binding has no live attack at all**; dory-zk only runs preprocessing-digest + final-opening-cancellation. Acknowledged as first-slice (synthesis/nightly sweep/mutation testing from the spec are not in this PR), but worth stating plainly so the gate isn't over-trusted.

4. **False-positive absorb enrollment.** The `receiver == "self"` / `starts_with("append")` heuristic captures non-transcript methods, e.g. `BlindFoldStatement::build::append#0::self.append(builder,&layout,claim_sources)` (R1CS builder, not a transcript). Harmless to soundness (a tripwire over-including is safe) but it dilutes signal and makes unrelated refactors of those methods trip the FS review gate.

5. **Domain separation is frozen-by-snapshot, not positively checked.** The inventory pins current `Label(b"...")` bytes (dropping/renaming one requires review), but nothing asserts the label set is injective / collision-free, and the label namespace still carries no type tag. A snapshot regression is caught; a pre-existing collision would be blessed.

**Cost / flake:**
- `fs-attacks-smoke` triples fixture generation (`cargo install --profile ci` + real muldiv proofs per leg, 3–7 min each). Two fragility vectors: the external ZeroOS musl download from a GitHub release (network/asset-rename/rate-limit), and `~/.cache/dory` restored under `shared-key: release-tests` with `save-if: false` — a cross-workflow cache read that regenerates the SRS (slow) if the producing workflow's cache is absent.
- `fs-obligations` shells `cargo metadata --locked` and `syn::parse_file`s the whole in-repo closure inside a `#[test]`; it hard-panics on a stale lockfile or any closure file syn can't parse, coupling the gate to lock/toolchain freshness. `rust_sources` skips files by stem (`prove`/`prover`/`tests`) and fns by name prefix (`prove*`/`commit_round*`) — heuristic; a differently-named prover-only fn in-closure would be scanned, and a verifier fn named `prove*` would be skipped.

**Nits:** `FsScope` enum and the `validate_fs_scope` allowlist in the derive macro are two sources of truth for the same 13-name set (self-correcting at compile time, but a manual double-update on every new stage). Thread-local `CURRENT_SCOPE` won't propagate to any off-main-thread challenge draw — none exist today (FS is serial), but a future parallel draw would mis-attribute scope silently. To actually be "blocking" as the PR states, these jobs must be added to branch-protection required checks.

_Review written with Claude Code assistance._

0xAndoroid (commented):
**Second-pass review — net-new findings only** (first pass from 0xAndoroid already covered the interleaving-blindness, 14-type allowlist, 3/13 scope coverage, false-positive absorb enrollment, domain-separation-by-snapshot, cost/flake, and the dual-source-of-truth nits). These three are additive, all in the audit harness itself.

**1. `AuditTranscript` collapses `challenge_vector` / `challenge_scalar_powers` into single atomic draws, hiding per-element `#[must_use]` drops.** `AuditTranscript::challenge_vector`/`challenge_scalar_powers` each record one `draw()` against one call to the inner `challenge_vector`/`challenge_scalar_powers`. But the *production* trait defaults (`crates/jolt-transcript/src/legacy.rs:76-90`) implement those as `(0..len).map(|_| self.challenge())` and a `challenge_scalar()` seed + local power loop — i.e. `len` independent scalar squeezes. In production each element challenge is its own `#[must_use]` `challenge()` return; in the audit it's one opaque `Vec` record. Consequence: if a verifier refactor drops one element of a returned `Vec` (uses `vec[..k-1]`, or lets a `challenge()` inside a custom loop be discarded), the gate can't distinguish it — the tape records the whole `Vec` as consumed and `consumed == expected` still passes. The strongest guarantee in the first-pass review (kind-encoding per call site) holds at *call* granularity but not at *element* granularity for the multi-draw APIs. Consider recording vector/powers element-wise so an unused or short-consumed element registers as a divergence.

**2. `first_value_divergence` is positional — it only proves a *same-index value* changed, not that the mutation altered the schedule.** `fs_attacks.rs` asserts `tape.first_value_divergence(&mutated_tape).is_some()` as the "mutation altered a production challenge" leg. But the impl is `self.records.iter().zip(&other.records).find_map(...)` — index-wise. It returns `Some` only if the mutation changed a challenge *value at the same ordinal*. The two canonical weak-FS structural changes are invisible to it: (a) an absorb↔squeeze reorder that leaves the recorded challenge *values* at their indices (values identical, schedule different), and (b) a *dropped* absorb that shifts subsequent squeezes — the tapes then have different lengths, and `zip` truncates to the shorter, silently passing if the shared prefix happens to match. So the "did alter a challenge" assertion can certify a mutation as soundness-relevant when the transcript schedule actually changed structurally and the test is no longer isolating the intended defense. A length/`id`-sequence equality check before the value comparison would close it.

**3. Two compile-couplings make the inventory gate non-idempotent and over-strict.** (a) `validate_fs_scope` in the derive parses the attribute as a bare `Ident`, so `#[fs_scope(Stage1)]` compiles but `#[fs_scope(scope = Stage1)]` (or any `name = value` form) is a hard compile error — the scope language silently forks between the attribute macro and anything hand-writing the guard. (b) `SCOPE_INVENTORY`/`ABSORB_INVENTORY` entries store the raw attribute/call token stream (`attribute.meta.to_token_stream()` / full `expression`) as the identity. Any formatting-only or token-level change to a scope attribute or absorb call regenerates a *different* inventory line and fails the gate even when semantics are unchanged — the census can't be re-run idempotently across a rustfmt/`cargo fmt` bump or macro-authoring style change without a bless. Both push the gate toward "diffs the source text" rather than "diffs the FS-relevant facts."

_Review written with Claude Code assistance._

0xAndoroid (approved):

PR #1674, opened 2026-07-15
docs: specify modular verifier panic-freedom remediation

quangvdao:
## Summary

- add a pinned remediation spec for making the new modular `jolt-verifier` panic-free
- classify wire, typed API, dependency, resource-exhaustion, and latent invariant surfaces by reachability and repository ownership
- document exact overlap with open PRs, especially preserving and extending #1609 rather than duplicating its serialization work
- propose a dependency-aware Jolt/Dory PR sequence, acceptance criteria, adversarial tests, fuzzing, and deployment guidance

## Scope

The primary target is the new modular `crates/jolt-verifier` and its SDK byte entry points. Verifier-relevant Dory/PCS/transcript dependencies are included. `jolt-prover-legacy` is explicitly excluded from the panic-freedom contract.

The audit is pinned to Jolt `78ae842c2cbc016d3b8f0e8dfeb30880aebaba8d`, Dory `56a7243aceb4d74e19a576a0245028d8d117507e` (`dory-pcs` 0.4.0), and the Arkworks patch at `76bb3a4518928f1ff7f15875f940d614bb9845e6`.

## Why

Malformed proof bytes are only one part of the verifier boundary. Serialized preprocessing and public I/O, nested Dory canonical data, directly constructed typed values, unchecked domain arithmetic, and allocation amplification also need explicit ownership and fail-closed behavior. This spec records those surfaces and separates urgent canonical fixes from API-disruptive or defense-in-depth follow-ups.

## Impact

Documentation only. This PR does not change verifier behavior. It provides the implementation and review plan for subsequent Jolt and Dory PRs.

## Validation

- verified the pinned Jolt and inspected PR-tip commits locally
- reviewed the complete spec in multiple independent audit and reconciliation passes
- `git diff --cached --check`
- Rust tests not run because this change adds documentation only

PR #1703, opened 2026-07-26
chore(lints): panic, unsafe, and numeric discipline across the jolt-verifier and deps

markosg04:
The workspace lints deny `panic!`/`unwrap`/`expect`, but a verifier consuming untrusted proof bytes could still panic through slice indexing and `unreachable!`, silently wrap release-mode arithmetic, and use `unsafe` without restriction.

This implements `specs/verifier-closure-lints.md` (spec in the diff): crate-root lint attributes across the 19 verifier-runtime crates, every existing violation fixed or carrying an expect-with-reason at the narrowest scope. Four tiers: panic-macro and wildcard-match denies everywhere; `forbid(unsafe_code)` on the 15 crates with no unsafe, with jolt-crypto's ten inline slice casts collapsed into two macro-generated helpers and jolt-dory's transmutes backed by compile-time layout asserts; `indexing_slicing` denied in the 15 control-plane crates (~290 fixes, mostly restructures to iterators/split_at and `.get` with existing typed errors, zero new error variants); and numeric discipline in jolt-verifier itself (~160 fixes: casts became `From`/`TryFrom`, proof-influenced arithmetic became checked ops, plus a `clippy.toml` disallowing wrapping/unchecked integer methods and HashMap/clock types, since verification must stay a deterministic pure function).

Hot paths kept their shape: prover-shared loops carry narrow expects stating the structural bound instead of new bounds checks, and hot-path indexing in jolt-poly/jolt-lookup-tables/jolt-claims/jolt-field stays warn-level for benchmark-gated follow-up PRs (a spec non-goal).

Two finds along the way: the stage4 public-RAM segment bound computed `start_index + len` with wrapping u128 arithmetic, so a wrapped sum would pass the domain check. Not reachable today (segment indices derive from u64-scale layout addresses, so the sum cannot approach 2^128), but the check trusted a distant constructor invariant and is now locally sound via checked_add. And jolt-blindfold's verify path panicked on malformed folded-eval arrays (now `LengthMismatch`). Also, `common` was outside the workspace lint regime entirely (no `[lints]` inheritance) and now inherits it.

Clippy is clean workspace-wide in `host` and `host,zk` and per-crate across the feature combos (zk, akita, field-inline, prover-fixtures). Full workspace suite, `muldiv` e2e in both modes, and the verifier fixture suites (honest-proof replays with identical transcripts) all pass. Not verified: criterion benchmarks (no hot-loop code changed shape, so no movement is expected) and 32-bit targets, where u64-to-usize layout conversions now error instead of truncating (reasoned, not executed).


0xAndoroid (commented):
Reviewed the full diff plus the post-merge tree (lint attributes, `clippy.toml`, all verifier-stage rewrites, the jolt-crypto/jolt-dory unsafe containment, and the closure claim against `cargo metadata`). This is enforcement, not theater: all 19 `indexing_slicing` expects in the control-plane crates sit on `#[cfg(test)]` modules — production lib code carries zero indexing escape hatches — and the ~14 production `arithmetic_side_effects` expects in jolt-verifier each state a type-level bound (ilog2 < 64, u8-derived counts) that holds regardless of proof input. `num::ilog2` panics on zero but every caller is behind an `is_power_of_two` validation or a `next_power_of_two()` input. The closure boundary checks out: the workspace normal-dep closure of jolt-verifier is exactly the 16 named crates, +3 PCS crates = 19. Both headline fixes are real: the stage4 segment bound (`start_index + words.len()` u128 wrap → `checked_add().is_some_and(end <= ram_K)`) and the blindfold `folded_eval_outputs[index]` OOB panic → `LengthMismatch`. FS-sensitive rewrites (transcript codec `split_first_chunk`, digest squeeze, compressed round-poly append) are byte-equivalent by inspection; the empty-coeff early return in `append_to_transcript` is unreachable in accepting runs because `check_round_well_formed` rejects `len < 2` in the same round.

Findings, roughly in priority order:

1. **The two bug fixes ship without regression tests.** The PR adds exactly one test (`prove_batch_rejects_zero_max_degree`, covering the one new error variant). The spec's testing letter is satisfied vacuously — both fixes reuse pre-existing variants — but a demonstrated domain-check bypass (segment with `start_index` near `u128::MAX`) and a demonstrated verify-path panic (truncated `folded_eval_outputs`) are precisely the inputs that should be pinned. Both tests are a few lines against the existing fixture proofs.

2. **The one known proof-influenced wrap stays outside the numeric tier.** `jolt-claims` `geometry/dimensions.rs` — `phase3_cycle_rounds` / `raf_evaluation_rounds` / `output_check_rounds` compute `log_t (+ log_k) - phase1_num_rounds`, which underflows when the prover-chosen `rw_config` phase split exceeds `log_t`. The RAM path is eagerly guarded (`RamRafEvaluationDimensions::try_from` at stage2 entry), but the registers `ReadWriteDimensions` built at stage4 (`proof.rw_config.register_dimensions(...)`) has no eager phase-split validation — only the lazy check inside point derivation. jolt-claims is carved out of `arithmetic_side_effects` for prover-perf reasons, but these are const-fn geometry accessors, not hot loops. Either extend the numeric tier to the geometry module, hoist an eager `validate_phase_split` at dimension construction, or name this site in the spec's residual-risk list with the same specificity as the shift-overflow note.

3. **Spec/impl drift on tier 3:** the spec says all 15 control-plane crates get `#![deny(clippy::indexing_slicing)]`, but the two proc-macro crates don't carry it — 13 do. Defensible (proc-macro panics are compile-time), but the spec carves the derive crates out only for `wildcard_enum_match_arm`; tier 3 should say the same or the attribute should be added as the regression guard the acceptance criteria imply.

4. **Boundary residuals worth naming in the spec:** (a) the external crates (`dory` / dory-pcs, the arkworks fork) are proof-byte-reachable and outside crate-root attribute reach — this is exactly #1674's dependency-surface tier; (b) `jolt-sdk` (`deserialize_verifier_object`, the WASM byte entry) is outside the 19 — a thin fallible bincode wrapper today, but it's the byte boundary #1674 names as its primary target. Neither this spec nor #1674's references the other; a cross-reference would keep the two panic-freedom efforts from drifting (resource-exhaustion/allocation amplification also remains lint-invisible and belongs to #1674's plan).

5. **Fail-open vs fail-closed inconsistency (nit):** some "unreachable by invariant" conversions clamp instead of erroring — `saturating_sub` in stage5 `reconstruct_r_address` and stage6b `committed_output_claims`, and the `selector_table` rewrite in jolt-openings (`table.iter_mut().skip(offset).zip(evals)`) now silently truncates a write where the old code panicked on an oversized `evals`. Elsewhere the identical shape got `checked_sub` → typed error (stage4 advice block, stage8 `missing_vars`). When the invariant is internal both are unreachable, but the checked-error form is uniformly fail-closed and self-documenting; the codebase now carries both idioms for the same situation. (The stage6a `saturating_sub`s are under exact branch guards — fine.)

6. **Vec transmute residual:** jolt-dory's `transmute::<Vec<Bn254G1>, Vec<ArkG1>>` (pre-existing; this PR adds the size/align const asserts). The asserts don't cover what the std contract actually leaves open — `Vec` layout across distinct generic instantiations is unspecified — and the documented-sound form (`ManuallyDrop` + `Vec::from_raw_parts(ptr.cast(), len, cap)`) is the same cost. Worth doing while the file is hot, given the PR's own "unsafe behind checked layout invariants" bar. The jolt-crypto slice-cast helpers are sound as written (`repr(transparent)` + `from_raw_parts`).

7. **Pre-existing file-level `#![expect(clippy::expect_used)]` blankets remain** in jolt-hyperkzg/scheme.rs and jolt-dory/transcript.rs. The covered `.expect()`s are prover-path today, but a file-wide blanket is wider than the "narrowest scope" rule this PR enforces everywhere else, and nothing keeps future verify-path code in those files from hiding under it.

None of these block the direction — the tier design is sound, the escape-hatch discipline is real, and the diff is behavior-preserving everywhere I could trace (including the 32-bit-relevant `u64`-domain comparisons in `verifier.rs`, which strictly improve on the old truncating casts). 1 and 2 are the ones I'd want addressed before or immediately after merge.

*Review written with Claude Code assistance.*

0xAndoroid (commented):
Second-pass review. Independent of the first pass; net-new items only.

**1. BlindFold: `random_witness` rows at dedicated final-opening coordinates never get masked — ZK hole (prove.rs)**

In `jolt-blindfold/src/prove.rs`, the Nova hiding instance is masked in three steps: (a) padding rows zeroed, (b) dedicated final-opening rows zeroed, (c) eval outputs/blindings scattered into `random_witness_rows[coordinate.row][coordinate.column]`. Step (b) covers only rows that appear in `final_coordinates`, but `witness_coordinate` maps an arbitrary R1CS variable index to a `(row, column)` grid position — a *different* variable can share the row and column of an eval coordinate (the same cell carries at most two dedicated variables: one evaluation + one blinding, which steps b–c do cover). Any *non*-final-opening variable whose index happens to collide with a dedicated row keeps its random value, and more importantly the cells of the folded witness at non-dedicated coordinates are unconstrained random values while `fold` only checks the dedicated coordinates and the R1CS relation. This is exactly the masking-class bug the protocol's `witness_rows.dedicated` bookkeeping exists for; recommend a test that the random instance's folding leaves zero residual witness at every committed coordinate (current tests exercise valid-layout paths only). ~60% this is exploitable rather than defense-in-depth, because the R1CS constraint set does not pin every witness cell.

**2. `kzg_verify_batch` scalar/base length mismatch from non-canonical `com` length (jolt-hyperkzg/src/kzg.rs)**

`bases` gets `com.len() + 4` entries; `scalars` gets `q_powers_multiplied.len() + 4` where `q_powers = challenge_powers(q, k)` and `k = com.len()` — consistent. But the earlier change `bases.extend_from_slice(&com[..k])` → `bases.extend_from_slice(com)` silently dropped the slice cap, and `scalars.extend_from_slice(&q_powers_multiplied[..k])` → `extend_from_slice(&q_powers_multiplied)` likewise. Both are len `k` today so this is behavior-preserving, but the truncation was the only guard if `challenge_powers` ever returns more than `k` entries; the MSM then panics on length mismatch deep inside `P::G1::msm` instead of returning `false`. Minor; a `debug_assert_eq!(q_powers_multiplied.len(), k)` restores the invariant.

**3. HyperKZG `DeriveSetup`: `assert!` on untrusted setup length, panic instead of error (jolt-hyperkzg/src/scheme.rs)**

`split_at(capacity)` + `rest.first().expect("length checked by the assert above")` relies on a pre-existing `assert!(source.g1_powers.len() >= capacity + 1)`. In a lint-hardening PR whose goal is "no panics on untrusted input," a derive-setup path that panics on a short SRS (attacker-influenced in any deployment where setup is loaded, not compiled in) is an odd survivor — same shape got converted to `HyperKZGError::SrsTooSmall` in `kzg_commit`. Convert to `Result` or document why this path can never see untrusted input.

**4. `BytecodePCMapper::try_new` double allocation + first-bucket semantics change (jolt-program/src/preprocess/bytecode.rs)**

```rust
let mut indices = vec![vec![(0, last_pc)]];   // one bucket with the sentinel
indices.resize_with(Self::index_count(bytecode)?, Vec::new);  // grows to index_count
```

If `index_count` returns 0 (empty bytecode), `indices` stays at length 1 — pre-change it would be empty; later `get_mut(bytecode_index)` silently accepts index 0. Also `vec![...]` + `resize_with` allocates twice vs the old single `vec![Vec::new(); n]`. Trivial: `let mut indices = vec![Vec::new(); Self::index_count(bytecode)?]; indices.first_mut()…push((0, 0))` keeps both properties.

**5. RAM preprocess `#[expect]` reason states the invariant it fails to establish (jolt-program/src/preprocess/ram.rs)**

```rust
let remapped_index = (chunk_address / 8 - min_bytecode_address / 8) as usize;  // ← underflow if a later chunk's address < min? no — min over all; OK
bytecode_words[remapped_index] = word;  // expect reason: "num_words covers every word between min and max"
```

The `#[expect]` reason covers the *bound* but not the actual invariant: `memory_init` sortedness. `chunk_by` groups *consecutive* equal `address / 8`; unsorted input produces multiple chunks for the same word, and later chunks overwrite earlier ones (last-write-wins) — same as before the PR, but now a byte-level silent-corruption path carries an annotation claiming it's structurally safe. If `memory_init` is not guaranteed sorted by the caller, this needs `sort` or a merge; if it is, the reason should say so. Also `num_words` (u64→usize `as` cast, pre-existing) still truncates on 32-bit for >4GiB images, contradicting the PR's own 32-bit rationale.

**6. `spartan_outer_row_weights`: zip truncation if lagrange weight count ≠ row group size (jolt-r1cs/src/constraints/jolt.rs)**

`centered_lagrange_evals(SPARTAN_OUTER_UNISKIP_DOMAIN_SIZE, uniskip)` returns a per-`uniskip` vector; the two row groups are zipped against it. Old code indexed `lagrange_weights[index]` with an expect-able bound; new zip silently truncates the *shorter* side — if `SPARTAN_OUTER_FIRST_GROUP_ROWS.len() != lagrange_weights.len()` (a constant drift in a future refactor), rows at missing indices get weight zero instead of a compile-time panic. These are all `const` arrays so the lengths are checkable: `const _: () = assert!(SPARTAN_OUTER_FIRST_GROUP_ROWS.len() == …)` or keep the indexed form with a narrow expect.

**7. `impl From<&HyperKZGProverSetup>` file-level expectation is unverifiable (jolt-hyperkzg/src/types.rs)**

`g1_powers[0]` / `g2_powers[0]` / `g2_powers[1]` indexed under one `#[expect(indexing_slicing, reason = "SRS generation always produces at least one G1 power and exactly two G2 powers")]`. But `HyperKZGProverSetup` is a public struct with public fields — any caller can build one with empty vectors, bypassing `setup_from_secret`. Either make the fields private/crate, or destructure: `let [g1, ..] = &prover.g1_powers[..] else { … }` with a typed error. As-is the reason asserts an invariant the type does not enforce.

**8. Spec-vs-code: `clippy.toml` `arithmetic-side-effects-allowed = ["F", …]` only matches the literal identifier `F`**

Four blindfold `add_stageN` functions were re-signed from `PCS::Field` to a bare `F: Field` generic purely so clippy.toml's identifier match fires, with a comment admitting it. This is a lint-config-driven API change: any future function that keeps `PCS::Field` spelling loses the exemption and re-introduces arithmetic-side-effects noise in field arithmetic. Prefer `allow`ing the concrete type paths (`jolt_field::Fr` is already there; add the trait projections clippy actually sees) or accept function-level expects, rather than contorting signatures.

**9. `prove_batch` inactive-member branch now silently no-ops when `batched_coefficients` is empty (jolt-sumcheck/src/prover.rs)**

`if let Some(constant) = batched_coefficients.first_mut()` replaces an unconditional `batched_coefficients[0] += …`. The empty case is exactly `max_degree == 0 && max_num_vars > 0`, which `ZeroBatchDegree` rejects one block earlier — but only when `max_num_vars > 0`. With `max_num_vars == 0 && max_degree == 0` and a member that reports `active` for round 0 (impossible per `described.rounds`, enforced nowhere), the contribution is silently dropped. Fold the invariant: `debug_assert!(prelude.max_degree >= 1 || max_num_vars == 0)` at the top, or keep the direct index with an expect.

**10. Round-proof `append_to_transcript` silently no-ops on empty coefficients (jolt-sumcheck/src/round_proof.rs)**

`let Some((constant, rest)) = coeffs.split_first() else { return; }` — an empty-coefficient round poly now appends *nothing* to the transcript (not even the label), diverging prover/verifier Fiat-Shamir if a zero-coefficient poly ever reaches this path. `check_round_well_formed` rejects `< 2` on the verify side, but `append_to_transcript` runs before/without that check on the prove side. If empty is impossible, `debug_assert!(!coeffs.is_empty())`; the silent early-return is the one transcript-absorption path in the closure that can desync without an error.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code) (AI-assisted review — 0xAndoroid)

0xAndoroid (commented):
Second-pass review. Independent of the first pass; net-new items only.

**1. BlindFold: `random_witness` rows at dedicated final-opening coordinates never get masked — ZK hole (prove.rs)**

In `jolt-blindfold/src/prove.rs`, the Nova hiding instance is masked in three steps: (a) padding rows zeroed, (b) dedicated final-opening rows zeroed, (c) eval outputs/blindings scattered into `random_witness_rows[coordinate.row][coordinate.column]`. Step (b) covers only rows that appear in `final_coordinates`, but `witness_coordinate` maps an arbitrary R1CS variable index to a `(row, column)` grid position — a *different* variable can share the row and column of an eval coordinate (the same cell carries at most two dedicated variables: one evaluation + one blinding, which steps b–c do cover). Any *non*-final-opening variable whose index happens to collide with a dedicated row keeps its random value, and more importantly the cells of the folded witness at non-dedicated coordinates are unconstrained random values while `fold` only checks the dedicated coordinates and the R1CS relation. This is exactly the masking-class bug the protocol's `witness_rows.dedicated` bookkeeping exists for; recommend a test that the random instance's folding leaves zero residual witness at every committed coordinate (current tests exercise valid-layout paths only). ~60% this is exploitable rather than defense-in-depth, because the R1CS constraint set does not pin every witness cell.

**2. `kzg_verify_batch` scalar/base length mismatch from non-canonical `com` length (jolt-hyperkzg/src/kzg.rs)**

`bases` gets `com.len() + 4` entries; `scalars` gets `q_powers_multiplied.len() + 4` where `q_powers = challenge_powers(q, k)` and `k = com.len()` — consistent. But the earlier change `bases.extend_from_slice(&com[..k])` → `bases.extend_from_slice(com)` silently dropped the slice cap, and `scalars.extend_from_slice(&q_powers_multiplied[..k])` → `extend_from_slice(&q_powers_multiplied)` likewise. Both are len `k` today so this is behavior-preserving, but the truncation was the only guard if `challenge_powers` ever returns more than `k` entries; the MSM then panics on length mismatch deep inside `P::G1::msm` instead of returning `false`. Minor; a `debug_assert_eq!(q_powers_multiplied.len(), k)` restores the invariant.

**3. HyperKZG `DeriveSetup`: `assert!` on untrusted setup length, panic instead of error (jolt-hyperkzg/src/scheme.rs)**

`split_at(capacity)` + `rest.first().expect("length checked by the assert above")` relies on a pre-existing `assert!(source.g1_powers.len() >= capacity + 1)`. In a lint-hardening PR whose goal is "no panics on untrusted input," a derive-setup path that panics on a short SRS (attacker-influenced in any deployment where setup is loaded, not compiled in) is an odd survivor — same shape got converted to `HyperKZGError::SrsTooSmall` in `kzg_commit`. Convert to `Result` or document why this path can never see untrusted input.

**4. `BytecodePCMapper::try_new` double allocation + first-bucket semantics change (jolt-program/src/preprocess/bytecode.rs)**

```rust
let mut indices = vec![vec![(0, last_pc)]];   // one bucket with the sentinel
indices.resize_with(Self::index_count(bytecode)?, Vec::new);  // grows to index_count
```

If `index_count` returns 0 (empty bytecode), `indices` stays at length 1 — pre-change it would be empty; later `get_mut(bytecode_index)` silently accepts index 0. Also `vec![...]` + `resize_with` allocates twice vs the old single `vec![Vec::new(); n]`. Trivial: `let mut indices = vec![Vec::new(); Self::index_count(bytecode)?]; indices.first_mut()…push((0, 0))` keeps both properties.

**5. RAM preprocess `#[expect]` reason states the invariant it fails to establish (jolt-program/src/preprocess/ram.rs)**

```rust
let remapped_index = (chunk_address / 8 - min_bytecode_address / 8) as usize;  // ← underflow if a later chunk's address < min? no — min over all; OK
bytecode_words[remapped_index] = word;  // expect reason: "num_words covers every word between min and max"
```

The `#[expect]` reason covers the *bound* but not the actual invariant: `memory_init` sortedness. `chunk_by` groups *consecutive* equal `address / 8`; unsorted input produces multiple chunks for the same word, and later chunks overwrite earlier ones (last-write-wins) — same as before the PR, but now a byte-level silent-corruption path carries an annotation claiming it's structurally safe. If `memory_init` is not guaranteed sorted by the caller, this needs `sort` or a merge; if it is, the reason should say so. Also `num_words` (u64→usize `as` cast, pre-existing) still truncates on 32-bit for >4GiB images, contradicting the PR's own 32-bit rationale.

**6. `spartan_outer_row_weights`: zip truncation if lagrange weight count ≠ row group size (jolt-r1cs/src/constraints/jolt.rs)**

`centered_lagrange_evals(SPARTAN_OUTER_UNISKIP_DOMAIN_SIZE, uniskip)` returns a per-`uniskip` vector; the two row groups are zipped against it. Old code indexed `lagrange_weights[index]` with an expect-able bound; new zip silently truncates the *shorter* side — if `SPARTAN_OUTER_FIRST_GROUP_ROWS.len() != lagrange_weights.len()` (a constant drift in a future refactor), rows at missing indices get weight zero instead of a compile-time panic. These are all `const` arrays so the lengths are checkable: `const _: () = assert!(SPARTAN_OUTER_FIRST_GROUP_ROWS.len() == …)` or keep the indexed form with a narrow expect.

**7. `impl From<&HyperKZGProverSetup>` file-level expectation is unverifiable (jolt-hyperkzg/src/types.rs)**

`g1_powers[0]` / `g2_powers[0]` / `g2_powers[1]` indexed under one `#[expect(indexing_slicing, reason = "SRS generation always produces at least one G1 power and exactly two G2 powers")]`. But `HyperKZGProverSetup` is a public struct with public fields — any caller can build one with empty vectors, bypassing `setup_from_secret`. Either make the fields private/crate, or destructure: `let [g1, ..] = &prover.g1_powers[..] else { … }` with a typed error. As-is the reason asserts an invariant the type does not enforce.

**8. Spec-vs-code: `clippy.toml` `arithmetic-side-effects-allowed = ["F", …]` only matches the literal identifier `F`**

Four blindfold `add_stageN` functions were re-signed from `PCS::Field` to a bare `F: Field` generic purely so clippy.toml's identifier match fires, with a comment admitting it. This is a lint-config-driven API change: any future function that keeps `PCS::Field` spelling loses the exemption and re-introduces arithmetic-side-effects noise in field arithmetic. Prefer `allow`ing the concrete type paths (`jolt_field::Fr` is already there; add the trait projections clippy actually sees) or accept function-level expects, rather than contorting signatures.

**9. `prove_batch` inactive-member branch now silently no-ops when `batched_coefficients` is empty (jolt-sumcheck/src/prover.rs)**

`if let Some(constant) = batched_coefficients.first_mut()` replaces an unconditional `batched_coefficients[0] += …`. The empty case is exactly `max_degree == 0 && max_num_vars > 0`, which `ZeroBatchDegree` rejects one block earlier — but only when `max_num_vars > 0`. With `max_num_vars == 0 && max_degree == 0` and a member that reports `active` for round 0 (impossible per `described.rounds`, enforced nowhere), the contribution is silently dropped. Fold the invariant: `debug_assert!(prelude.max_degree >= 1 || max_num_vars == 0)` at the top, or keep the direct index with an expect.

**10. Round-proof `append_to_transcript` silently no-ops on empty coefficients (jolt-sumcheck/src/round_proof.rs)**

`let Some((constant, rest)) = coeffs.split_first() else { return; }` — an empty-coefficient round poly now appends *nothing* to the transcript (not even the label), diverging prover/verifier Fiat-Shamir if a zero-coefficient poly ever reaches this path. `check_round_well_formed` rejects `< 2` on the verify side, but `append_to_transcript` runs before/without that check on the prove side. If empty is impossible, `debug_assert!(!coeffs.is_empty())`; the silent early-return is the one transcript-absorption path in the closure that can desync without an error.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code) (AI-assisted review — 0xAndoroid)

moodlezoup (commented):
this looks fine to me. we seem to have a lot of `expect`s for `clippy::indexing_slicing` and `clippy::unreachable`, I wonder if we should just cut them from the `deny` list?

moodlezoup (approved):

PR #1706, opened 2026-07-27
perf(akita): stream one prefix-packed trace one-hot polynomial

quangvdao:
## Metadata and scope

- **Base:** `a16z/jolt:perf/optimize-akita-prover` at `9afe014b78501d067a56be51e2ec3e10e7fa7901`
- **Head:** `ecef63c558da7da3cf527bbd550b3db6e09e806f`
- **Surface:** 6 commits, 24 files, 2,775 additions, 9,571 deletions
- **Supporting Akita revision:** [`quangvdao/akita@a96b26fa`](https://github.com/quangvdao/akita/commit/a96b26fa04fb2e08d14e46d2f128a1168885beca)

This replaces the per-semantic-polynomial Akita boundary for trace-derived one-hot data with one prefix-packed physical polynomial and Jolt-owned streaming commitment/opening kernels.

## Motivation

The old adapter presented instruction, bytecode, RAM, and fused-increment one-hot columns to Akita as separate logical polynomials. That boundary prevented Jolt from exploiting the actual witness source: one trace row determines every semantic hot lane.

It also encouraged materializing column-major one-hot indices and Akita `FlatBlocks` before commitment/opening. At production sizes those structures dominate memory even though the trace already contains all information needed to reconstruct them.

The new invariant is:

```text
one trace row
  -> all semantic hot lanes for that cycle
  -> one linear streaming pass
  -> several live Akita block accumulators
```

## Final protocol contract

All semantic columns are represented by one multilinear polynomial:

```text
P(selector || cycle || address)
```

Each selector slot owns one contiguous `K * T` segment. Unused capacity slots are identically zero.

The selector registry has a fixed prefix and a variable suffix:

| Regime | Fixed instruction | Fixed increment chunks | MSB | Variable bytecode/RAM | Capacity |
|---|---:|---:|---:|---:|---:|
| K=16 | 32 | 16 | 1 | at most 15 | 64 |
| K=256 | 16 | 8 | 1 | at most 7 | 32 |

The layout rejects:

- chunk widths other than 4 or 8;
- noncanonical instruction-column counts;
- semantic-column counts above the fixed capacity;
- non-power-of-two or non-ring-aligned trace geometry;
- unsupported kernel dimensions where neither `K | D` nor `D | K`.

The kernels are const-generic over `D`; tests cover D=64, 128, 256, and 512.

## Transcript and opening reduction

Stage 8 resolves the semantic leaf claims at the common `(cycle || address)` point, binds that complete evaluation vector to the transcript, and only then samples the selector point.

The physical opening claim is:

```text
point = selector || cycle || address
eval  = Σ_i eq(selector, i) * semantic_eval_i
```

Padding slots contribute zero. Binding the semantic vector before sampling the selector is load-bearing: the prover cannot choose semantic evaluations after seeing the random reduction point.

The verifier reconstructs the same singleton point/evaluation and verifies exactly one Akita polynomial.

## Change-surface inventory

| Layer | Before | After |
|---|---|---|
| Claims/layout | Variable-length native batch of semantic columns | Fixed-capacity prefix-packed singleton with digest-bound ordering |
| Trace source | Per-column one-hot objects | Row-major `TraceOneHotRows` provider |
| Commitment | Akita per-polynomial callbacks and cached blocks | Jolt-owned multi-accumulator trace sweep |
| Evaluation fold | Materialized one-hot block traversal | Direct block/position weighting from trace rows |
| Decompose fold | Akita `FlatBlocks` and rotated tables | Parallel position-owned sparse challenge accumulation |
| Stage 8 | Batch opening for all semantic columns | One selector-reduced opening |
| Planner | Minimum estimated payload | Minimum root rank within configurable payload slack |
| Schedules | Hundreds of native batch-width rows | 19 singleton packed-shape rows |
| Dependency wiring | Machine-local Akita checkout | Pinned public fork revision |

```mermaid
flowchart LR
    Trace["Vec<Cycle> + fused increment"] --> Rows["TraceOneHotRows::fill_row"]
    Rows --> Commit["Jolt commit kernel"]
    Rows --> EvalFold["Jolt evaluation fold"]
    Rows --> Decompose["Jolt parallel decompose fold"]
    Commit --> PCS["one Akita commitment"]
    EvalFold --> PCS
    Decompose --> PCS
    Claims["semantic Stage-8 claims"] --> Bind["transcript bind"]
    Bind --> Selector["sample selector"]
    Selector --> Singleton["one packed point/evaluation"]
    Singleton --> PCS
```

## Owned kernel geometry

For `K >= D`, one trace row spans `K / D` ring elements. The commit kernel buckets all semantic lanes by ring offset, loads each active A position once, and updates the corresponding selector-slot accumulators.

For `K < D`, one ring contains `D / K` trace rows. The production K=16 path now consumes the compact row lanes directly in commitment, evaluation fold, and decompose fold. At `T=2^20`, this removes about 60 million `(column, coefficient)` tuples, or roughly 0.96 GB of transient tuple writes per complete traversal.

When one semantic segment spans one or more complete Akita blocks:

- commitment and evaluation fold use a deterministic row-aligned task grid;
- at `T=2^20` on the benchmark host, the natural 8 trace-block tasks become 40 stealable partial tasks;
- once the natural block count is sufficient (`T=2^24` has 64), the kernel does not oversplit;
- partial accumulators merge in fixed order without changing A-matrix positions;
- decompose fold partitions the position domain, so each worker owns disjoint output rows and scans only its matching trace ranges.

The trace source also reuses the fused-increment lane arrays that stages 6/7 already materialize. No production path constructs an Akita `OneHotPoly`, `FlatBlocks`, or per-column hot vector. The D=64 decompose path may build a bounded rotation table up to 64 MiB; every other dimension and larger geometry uses the generic sparse fallback.

## Planner and generated surface

The supporting Akita change introduces:

```text
MinRootRankThenPayloadWithinSlack { slack_permille }
```

Jolt selects a 1% payload envelope, then minimizes the root inner commitment matrix output rank before payload size. This makes the A width versus live-accumulator count an explicit tuning knob.

The policy parameter is included in generated-catalog identity digests. Jolt's one-hot catalogs were regenerated for one physical polynomial:

- K=16: packed arities 22–34, 13 entries;
- K=256: packed arities 38–43, 6 entries.

This removes roughly 9,500 lines of batch-width schedule tables.

## Performance

Measured on a 10-core Apple M4 with 24 GB RAM on 2026-07-28. Each headline run was release-built first, cooled for at least 120 seconds, run with no concurrent Cargo/prover process and one-minute load below 4, and wrapped in `/usr/bin/time -l`.

Exact command:

```bash
PERF_LOG_T=N /usr/bin/time -l cargo nextest run --release \
  -p jolt-prover-legacy --features akita \
  -E 'test(sha2_chain_akita_perf)' \
  --run-ignored all --no-capture --cargo-quiet
```

### Final scaling

| Padded trace | Actual trace | K / selector capacity | Setup | Prove | Verify | Peak RSS |
|---:|---:|---:|---:|---:|---:|---:|
| 2^20 | 1,048,576 | 16 / 64 | 99 ms | **4.09 s** | 17 ms | **1.61 GB** |
| 2^22 | 4,194,304 | 16 / 64 | 175 ms | **15.38 s** | 22 ms | **5.07 GB** |
| 2^24 | 16,777,216 | 16 / 64 | 0.98 s | **69.29 s** | 59 ms | **11.72 GB** |

All three runs completed with zero swaps.

### Optimization progression at 2^20

| Revision | Change | Prove | Peak RSS |
|---|---|---:|---:|
| `20af98f35` | Initial single-poly owned kernels | 4.39 s | 1.63 GB |
| `303bfcb4f` | Row-aligned task grid + fused-inc reuse | 4.18 s | 1.61 GB |
| `1671a391e` | Direct K<D lane traversal | **4.09 s** | **1.61 GB** |

The final result is 300 ms / **6.8% faster** than the initial single-polynomial implementation.

The initial trace attributed 2.17 s to commitment and 0.92 s to the complete Akita opening. The final trace reports:

- `TracePackedOneHot::commit_inner`: 1.92 s;
- complete Akita opening: 0.90 s;
- `TracePackedOneHot::decompose_fold`: 0.30 s.

Commit accumulation remains the dominant optimization target; commitment decomposition is negligible at this shape.

The standalone `trace_one_hot` benchmark is now production-aware: `JOLT_AKITA_TRACE_LOG_T` selects the trace length, K=16/capacity-64 is used below `logT=25`, K=256/capacity-32 thereafter, and explicit K/column overrides are validated. This replaces the previous hard-coded K=256 benchmark, which accidentally measured an 8x larger physical domain for the requested 2^20–2^24 range.

## Correctness and validation

Coverage:

- exact commitment/opening parity against a materialized packed `OneHotPolynomial`;
- selector reduction matches direct evaluation of the physical packed polynomial;
- zero-padding selector slots evaluate to zero;
- capacity and canonical instruction-width rejection;
- cross-D row mapping for K=16 and K=256 at D=64/128/256/512;
- materialized parity for blockwise evaluation and decompose-fold kernels in both `K > D` and `K < D` geometries;
- end-to-end proofs in normal K=16, forced K=256, and committed-program modes;
- Stage-8 commitment metadata rejection before PCS verification.

Final local gates:

- `cargo nextest run -p jolt-akita --cargo-quiet` — 42 passed;
- packed legacy Akita muldiv filter — 3 passed (K=16, forced K=256, committed program);
- `cargo clippy -p jolt-akita --all-targets -- -D warnings` — passed;
- `cargo clippy -p jolt-prover-legacy --features akita --all-targets -- -D warnings` — passed;
- `cargo fmt -q` and `git diff --check` — passed.

The pushed checkpoint at `303bfcb4f` completed the full GitHub matrix with all 27 reported checks green, including packed Akita tests, ZK tests, clippy, formatting, schedule-catalog drift, verifier fixtures, and modular crates. The final two commits trigger the same matrix again.

## Compatibility and deliberate limits

- This changes the OneHotTrace commitment layout, transcript, layout digest, setup shape, and proof wire behavior. Proofs and setup metadata from the old native-batch layout are intentionally incompatible.
- Advice and committed-program objects keep their existing packed flows.
- Tensor root challenges are deliberately unsupported for this trace source; the generated Jolt catalogs use flat root challenges.
- The production config remains D=64, while the source kernels support future D=128/256/512 schedules.

## Reviewer map

- `crates/jolt-akita/src/trace_onehot.rs` — streaming source and owned commit/open/decompose kernels
- `crates/jolt-claims/src/protocols/jolt/lattice/{packing,strategy,geometry}.rs` — selector registry, capacity, digest, reduction
- `crates/jolt-prover-legacy/src/zkvm/packed.rs` — trace-row derivation and prover Stage 8
- `crates/jolt-verifier/src/stages/stage8/packed.rs` — singleton verifier reduction
- `crates/jolt-akita/src/configs.rs` — planner policy and setup sizing
- `crates/jolt-akita/src/schedules/` — singleton generated catalogs
- `crates/jolt-akita/benches/akita_paths.rs` — production K16/K256 benchmark geometry

## 2026-07-28 kernel optimization update

Commit `0aeab5459` adds the next owned-kernel pass. The generic pieces are bulk row filling, active-column-only task buffers, skipped empty wide reductions, and a bounded dense-rotation option with a sparse fallback. The only production specialization is an explicit `D=64 && K=16` dispatch that fuses the four trace-row shifts targeting one ring accumulator. The fused algebra primitive itself is generic over both `D` and the fixed shift count and lives in [`quangvdao/akita@a96b26fa`](https://github.com/quangvdao/akita/commit/a96b26fa04fb2e08d14e46d2f128a1168885beca).

### A/B results at 2^20

The host became heavily contended during the candidate run: the unchanged evaluation-fold control phase slowed from 59.20 ms to 88.78 ms. Therefore the phase speedups below normalize candidate times by that 1.50x control ratio; raw wall times are retained only as diagnostics.

| Metric | `1671a391e` baseline | `0aeab5459` candidate | Improvement |
|---|---:|---:|---:|
| Trace one-hot commitment | 1.925 s | 1.48 s normalized | **1.30x / 23.3% faster** |
| Trace one-hot decompose fold | 302 ms | 105 ms normalized | **2.87x / 65.1% faster** |
| Complete prove | 4.099 s | 3.45 s normalized | **1.19x / 15.9% faster** |
| Same-session process instructions | 100% | 86.4% | **13.6% fewer** |
| Same-session process cycles | 100% | 75.6% | **24.4% fewer** |

The raw candidate prove samples were 4.49 s median and 3.94 s best, versus the earlier clean 4.09 s baseline. They are consistent with the counter/phase result but are not used as the headline speedup because background load was not controlled.

### Larger traces

Both candidate proofs completed and verified with zero swaps. Their wall times were degraded by sustained macOS Storage Management and UI load, so the useful A/B signal is memory:

| Padded trace | Baseline RSS | Candidate RSS | Change | Candidate prove under load |
|---:|---:|---:|---:|---:|
| 2^22 | 5.07 GB | 4.88 GB | **3.8% lower** | 21.39 s |
| 2^24 | 11.72 GB | 10.37 GB | **11.5% lower** | 93.15 s |

### Rejected experiment

A grouped-A traversal was also implemented and measured before being reverted. It reduced trace decoding but multiplied negacyclic shift arithmetic: after normalizing by the same unchanged control phase, commitment remained about **2x slower** and complete proving about **51% slower**. This is why the final design keeps the existing A-position task grid and instead fuses same-destination shifts.

Additional local gates for this update: all 18 `jolt-akita` library tests passed; K=16 and forced-K=256 legacy end-to-end tests passed; `jolt-akita --all-targets` clippy and legacy Akita clippy passed; formatting, lockfile validation, and `git diff --check` passed.

### Perfetto kernel phase tracing

Commit `ecef63c55` adds fixed-cardinality phase spans around the owned OneHotTrace kernels. The spans separate commitment preparation, accumulation, partial merging, and decomposition; evaluation accumulation, merging, and block reduction; and decompose-fold rotation preparation, challenge accumulation, digit expansion, and witness construction. Span arguments record `D`, `K`, active/capacity columns, block/task geometry, fused-shift dispatch, weight kind, and dense-rotation sizing. No span is created inside a trace-row, ring, block, Rayon-task, or accumulator loop.


quangvdao:
@mgeorghiades Draft is open against your perf branch so you can follow the protocol/layout diff early. The correctness cutover is in the initial commit; I am now profiling and iterating on the 2^20–2^24 performance matrix and will update the PR with final commits and measurements.

quangvdao:
Optimization checkpoint pushed as 303bfcb4f.

This keeps the single prefix-packed OneHotTrace protocol and improves the owned kernels:
- splits the under-parallelized K=16 root commit/evaluation geometry into deterministic row-aligned partial tasks (8 -> 40 tasks at T=2^20 on this 10-core M4; no oversplitting once the natural block count is sufficient),
- merges partial accumulators in a fixed order without changing A-matrix positions,
- reuses the fused-increment one-hot lanes already materialized for stages 6/7,
- removes redundant row-buffer clearing under the TraceOneHotRows overwrite contract,
- adds kernel-local tracing spans for the next attribution pass.

Validation at this checkpoint: jolt-akita 42/42 tests pass; all three packed muldiv tests pass, including forced K=256 and committed-program mode. Before this checkpoint, the exact T=2^20 production run was 4.39s prove / 18ms verify / 1.63GB peak RSS; the trace attributed 2.17s to commitment and 0.92s to opening. The optimized release measurement is building now; I will post the before/after and T=2^20..2^24 scaling results next.

quangvdao:
Final perf pass is pushed at `1671a391ed2d2895407da8554ca832a7d4b13cad`, and I updated the PR body with the full architecture/diff and corrected production-geometry results.

On the 10-core M4 / 24 GB machine:
- `2^20`: prove 4.09 s, peak RSS 1.61 GB (down from the first single-poly checkpoint at 4.39 s: -6.8%)
- `2^22`: prove 15.38 s, peak RSS 5.07 GB
- `2^24`: prove 69.29 s, peak RSS 11.72 GB
- zero swaps in all three runs

The hot-kernel trace at `2^20` is 1.92 s custom commit, 0.90 s total opening, and 0.30 s decompose/fold. Final local validation is clean: 42 jolt-akita tests, normal/forced-K256/committed-program muldiv e2e, both relevant clippy invocations, fmt, and diff-check. The final commits keep the single physical prefix-packed polynomial and add deterministic parallel trace tasks, reuse of fused increment lanes, and direct flattened K16-lane consumption.

quangvdao:
Published the next one-hot kernel optimization for review.

Jolt head: a740e209c
Akita dependency: quangvdao/akita@0d38c126

The D=64/K=16 commit scheduler groups semantic columns that share the same four-row shift tuple. It computes each shared shifted source-ring sum once and fans it out directly to the matching commitment accumulators. Akita only exposes generic wide-ring coefficient access; the trace geometry and scheduling remain owned by Jolt.

Validation on the pushed, remote-pinned tree:
- jolt-akita library: 18/18 passed (including K=16, K=256, generic mapping, and D=512 opening parity)
- legacy end-to-end: K=16, forced K=256, and committed-program tests all passed

2^20 trace measurements on this M4, warm real-prover run:
- total prove: 3.16 s (previous clean baseline 4.09 s, 22.7% faster)
- TracePackedOneHot::commit_inner: 1.205 s (previous clean baseline 1.92 s, 37.2% faster)
- Akita batched opening: 0.700 s
- TracePackedOneHot::decompose_fold: 0.109 s

A fixed-four-thread exact-executable Criterion A/B bracketed the commit speedup at 14.9-31.2% under thermal drift (old 2.698 s; candidate 1.857 s on the cleanest adjacent pair). No architecture-specific intrinsics are used.

quangvdao:
Perf follow-up pushed.

Jolt head: `ac23eb7bc7700a4609b087fe402fa845f9643cc9`

Supporting Akita head: `27cdbb01cbb29784d90763d2df2472743e43c93b`

## What changed

- Added a D=64/K=256 commit specialization that stages 8,192 trace rows once, then streams one A rank at a time. Only 29 contiguous wide accumulators are hot for the active rank; the dimension-generic path remains the fallback for other `(D, K)` geometries.
- Reordered dense opening rotations into trace-block-major / active-column-major order and stopped building tables for selector-padding columns.
- Increased decompose position parallelism to four tasks per Rayon worker and writes directly into final `par_chunks_mut`, eliminating the nested-vector flatten copy.
- Fused up to four K<D trace-row rotations per destination pass.
- For K>=D, streams dense rotated rows in fixed eight-vector tiles. This was the dominant K=256 opening win.
- Returns the single-digit compressed buffer directly instead of expanding/copying it.
- Updated the Akita pin to an owned-coefficient witness constructor. `Vec<[F; D]>` and `Vec<[i32; D]>` now flatten into the witness without element-wise copies.
- Added benchmark controls for path selection, shorter Criterion windows, and one-shot trace commitment so T=2^25 can be measured without materializing the 3.5 GiB logical one-hot matrix.

## Isolated commitment A/B

Same optimized binary, 10 Rayon threads, synthetic 29-column K=256 trace:

| Geometry | Generic | Rank-tiled | Change |
|---|---:|---:|---:|
| T=2^22, mean of two interleaved one-shot runs | 10.712 s | 9.099 s | **15.1% faster** |
| T=2^25, first cold comparison of the initial rank-tiled form | 122.722 s | 92.656 s | **24.5% faster** |

The shorter T=2^22 pair is the more stable claim. Long T=2^25 wall times vary substantially with package temperature. In the first cold comparison, reported max RSS also fell from 15.95 GiB to 11.18 GiB. The final full T=2^25 proof reported 13.10 GiB max RSS.

Tile sweep:

- 4,096 rows: rejected (`133.39 s` at T=2^25).
- 8,192 rows: retained.
- 16,384 rows: slightly worse than 8,192 at T=2^22 (`9.29 s` vs `9.10 s` mean).

## Full-proof Perfetto results

For T=2^20 through T=2^24, the baseline is the exact previous PR head `a740e209c`. These use K=16. The K=16 commit kernel was not changed in this follow-up, so its observed timing movement is machine/thermal variation, not a claimed kernel speedup.

| T | Prove | Commit accumulate | Akita opening | Decompose accumulate | Evaluate/fold |
|---|---:|---:|---:|---:|---:|
| 2^20 baseline | 3.270 s | 1.262 s | 0.701 s | 99.6 ms | 56.7 ms |
| 2^20 new | 3.169 s | 1.165 s | 0.677 s | **78.6 ms** | 63.1 ms |
| 2^22 baseline | 11.865 s | 4.877 s | 1.742 s | 481.2 ms | 270.9 ms |
| 2^22 new | 10.734 s | 4.666 s | **1.495 s** | **341.3 ms** | 251.9 ms |
| 2^24 baseline | 62.568 s | 27.511 s | 6.326 s | 2.048 s | 1.140 s |
| 2^24 new | 50.118 s | 20.567 s | **5.350 s** | **1.511 s** | 1.108 s |

The repeatable targeted K=16 result is decompose/fold: **21% / 29% / 26% faster** at T=2^20 / 2^22 / 2^24. Whole Akita opening improved 14–15% at T=2^22 and T=2^24.

The T=2^24 total-proof reduction is an observed result, not fully attributable to this patch: unrelated sum-check spans were also faster in the new run. For example, stage 3 and stage 6a were flat, while stages 4 and 5 varied by roughly 28% and 33% despite no changes there.

At T=2^25 (K=256), comparing the same final source before and after dense-row tiling:

| Variant | Prove | Commit accumulate | Akita opening | Decompose accumulate | Evaluate/fold |
|---|---:|---:|---:|---:|---:|
| Source-strided dense rotations | 206.582 s | 72.353 s | 60.578 s | 13.265 s | 2.925 s |
| Four-vector tiles | 229.029 s | 90.723 s | 54.031 s | 3.754 s | 2.964 s |
| Eight-vector tiles (retained) | 202.161 s | 79.171 s | **49.916 s** | **3.507 s** | 2.743 s |

The isolated K=256 decompose span improved **73.6%** (`13.265 s -> 3.507 s`). Eight-wide was a further 6.6% over four-wide. Total opening improved 17.6% in the final run. Commit and total-proof movement across these three long runs is thermally noisy; the interleaved one-shot A/B above is the commitment result to use.

The owned Akita handoff makes witness construction negligible: at T=2^25 it is 34.8 ms after a 3.51 s accumulation. At T=2^24 it fell from 14.6 ms to 5.9 ms, while also removing the transient coefficient copies.

## Rejected hypotheses

- K=16 partial-tuple shift fusion plus compact group lookup: no statistical improvement.
- Compact lookup alone: no statistical improvement.
- K=16 coefficient-plane batching (B=4): decisive 56.9% regression; scalar/scattered updates lost the vectorized full-ring shift.
- K=256 4,096-row and 16,384-row tiles: both worse than the retained 8,192-row tile.

## Validation

- `cargo nextest run -p jolt-akita --lib --cargo-quiet` — 18 passed.
- `cargo nextest run -p jolt-prover-legacy --lib muldiv_e2e_akita --cargo-quiet --features host,akita` — ordinary K=16, forced K=256, and committed-program tests all passed.
- `cargo clippy -p jolt-akita --all-targets -- -D warnings` — clean.
- `cargo clippy -p jolt-prover-legacy --lib --features host,akita -- -D warnings` — clean.
- `cargo fmt --check` and `git diff --check` — clean.
- Supporting Akita: 232 `akita-prover` library tests passed; `akita-prover` + `akita-types` Clippy passed under `-D warnings`.

quangvdao:
K=256 commitment update pushed in fe4a318a4.

The real T=2^25 SHA2-chain Perfetto run now shows:

| Span | ac23eb7bc baseline | fe4a318a4 | Change |
|---|---:|---:|---:|
| Wall span | 217.10 s | 196.00 s | 9.7% faster |
| prove_packed | 202.16 s | 181.55 s | 10.2% faster |
| Akita commitment | 90.48 s | 79.41 s | 12.2% faster |
| TracePackedOneHot::commit_inner | 79.30 s | 63.28 s | 20.2% faster |
| trace_onehot_commit_accumulate | 79.17 s | 63.20 s | 20.2% faster |
| Akita batched opening | 49.92 s | 53.18 s | 6.5% slower in this run (opening code is unchanged) |

What changed:

- Jolt now exposes an always-present semantic-column mask to the owned trace kernel.
- In the current K=256 layout, 26 of 29 columns are always present: 16 instruction chunks, the bytecode chunk, eight unsigned-increment chunks, and the increment MSB. Only the three RAM columns are sparse.
- The D=64/K=256 specialization stages four trace rows and their 16 A rings, then combines the four dense contributions before touching each destination accumulator. The hot source+destination set is about 90 KiB per task (32 KiB source tile + 58 KiB destinations).
- Sparse RAM columns retain direct one-source shift accumulation and are routed by one presence mask per row.
- The always-present declaration is checked fail-closed: a missing lane in a declared-dense column returns InvalidInput.
- Sources that do not provide the mask, other D/K geometries, and the existing generic/rank-tiled path are unchanged.

Validation:

- The full release T=2^25 proof completed successfully and produced the final Perfetto trace.
- The focused forced-K256 end-to-end proof passed during development.
- jolt-akita and jolt-prover-legacy host+akita compile checks passed.
- cargo fmt and git diff --check pass.

Rejected experiments / useful warnings:

- Treating every real column as dynamically fuseable regressed accumulation to roughly 111 s.
- Fusing only instruction+bytecode (17 dense / 12 sparse) regressed accumulation to 107.94 s. The nine increment columns being structurally always present is what makes this design win.
- Column tiling repeatedly rescanned A and was materially slower.
- Larger synthetic block batches and column batches were misleading because the synthetic source did not reproduce Jolt's dense/sparse structure.
- A raw machine-specific NEON linear-add variant did not beat the compiler-generated wide-ring operations and was removed.

Remaining K=256 levers for the next handoff:

1. Share each A ring across two trace blocks. This halves A loading/widening, but grows live destinations from 58 KiB to 116 KiB. The most plausible sweep is block-batch=2 with row-fusion=2 versus row-fusion=4. I stopped this run to push the measured result; no temporary environment knobs or unmeasured defaults were committed.
2. Improve the four-distinct-source negacyclic primitive. Its coefficient loop still has runtime wrap/sign branches. A generic branch-segmented implementation could preserve portability while improving vectorization.
3. Add a small synthetic dense/sparse equivalence test that explicitly supplies the always-present mask. The full Jolt proof exercises it today, while the existing jolt-akita synthetic sources intentionally fall back via mask=0.

Perfetto artifacts retained locally:

- Baseline: sha2-chain-2exp25-akita-onehot-batch8-final-20260729.perfetto.json
- Final: sha2-2exp25-akita-k256-dense26-final-20260729.perfetto.json
- Rejected dense-17 run: sha2-2exp25-akita-k256-dense17-20260729.perfetto.json

quangvdao:
Superseded by #1715, which rebuilds the packed-onehot optimization series on current main and carries the evolved protocol/performance work. The one remaining unique candidate from this branch, fe4a318a4 (dense K256 trace-row fusion), is being evaluated for porting to #1715 before validation.

PR #1708, opened 2026-07-27
fix(akita): fp128 lookup address alias

moodlezoup:
## Summary

Closes a soundness break in the instruction-lookup argument under `--features akita`, where a prover can commit a lookup address that satisfies every existing constraint yet makes the lookup table return a different value.

The Akita modulus `p = 2^128 - 2^32 + 22537` is **below** `2^128`. The identity-RAF leg is the only thing tying the committed 128-bit address `k` to `RightLookupOperand`, and it does so via `Identity(k) = k mod p` — so it pins `k` only up to its residue class. The committed address domain (`2^128` cells) is larger than the field, so every honest index `r < 2^32 - 22537` has a second committable preimage `r + p`, which the table (reading the low `XLEN` bits) evaluates differently. Booleanity, Hamming weight, the R1CS operand rows and commitment binding are all satisfied *exactly* — no probabilistic step is involved.

This is a property of the **field**, not the PCS. BN254 is unaffected: an aliased address there is `≈2^254` and would need 254 address variables, of which only 128 exist.

### The fix

Fold `γ³·Ũ(r_address)` into read-RAF's existing `EqRafFlag` derived public, where `Ũ` is the multilinear indicator that the address's upper half is all ones:

```
EqRafFlag:  γ²·Ĩ − γ·L̃ − γ²·R̃   →   γ²·Ĩ − γ·L̃ − γ²·R̃ + γ³·Ũ(r_a)
```

The input claim gains no `γ³` term, and both sides are polynomials in `γ` whose coefficients are transcript-fixed before `γ` is drawn — so Schwartz–Zippel in `γ` forces `Σ_j eq(r_red,j)·RafFlag_j·U(k_j) = 0`, hence `U(k_j) = 0` pointwise for every identity-RAF cycle. Every alias `r + p < 2^128` has an all-ones upper limb, so all are excluded.

The converse is what keeps it cheap: `U(k) = 0 ⟹ k < 2^128 − 2^64 < p` (asserted at compile time), so a surviving address is already the canonical representative and the identity leg pins it exactly — **with no range bound on `RightLookupOperand` needed**. That matters, because advice rows set `RafFlag` but row 10 excludes them, leaving their right operand genuinely unconstrained.

### Why not the witness column in the design docs

`akita_fix.md` / `akita_fix_followup.md` specify a committed `B` column plus an R1CS row `(f_add+f_sub+f_mul)·B = 0`. This PR deliberately does not do that — reviewers should read the diff against that expectation. Since `B = RafFlag · U` is already masked by `RafFlag`, and `RafFlag` already multiplies `EqRafFlag`, the column's only job would be to be proven zero, which read-RAF does directly. The column version would touch ~50 files across **two** independent R1CS encodings (`jolt-prover-legacy/src/zkvm/r1cs/` is what `prove_packed` uses for akita; `crates/jolt-r1cs/` is what the verifier checks), three copies of the Spartan row-group tables, both BlindFold mirrors, and the BN254 proof shape. The cost of the simplification is `(3 + log_T)/|F| ≈ 2^-121`; the column version isn't exact either, since it proves its row through Spartan and still binds `B` through this same read-RAF term.

One behavioural difference: masking by `InstructionRafFlag` (`!is_interleaved_operands()`) also covers **advice**, not just add/sub/mul. Strictly stronger, and free. The docs should be reconciled with what shipped.

## Changes

**Alias fix** (commit 2, gated on `akita`)

- `jolt-claims`: `CANONICAL_INSTRUCTION_ADDRESS` const + `upper_half_all_ones()`. The gate is read from `jolt-claims` rather than re-derived downstream — `jolt-kernels` has no `akita` feature, so a local `cfg!` there would silently be `false` and desync prover from verifier.
- `jolt-verifier` stage 5: the `γ³` term, plus a length check on the reconstructed address point (a wrong length would mis-slice silently rather than fail).
- `jolt-prover-legacy`: `UpperHalfAllOnesPolynomial` + `Prefix::UpperHalfAllOnes`, a fourth accumulator in the fused `init_Q_raf` scan (identity rows only — that is what supplies the `RafFlag` mask), wired through the address message, the address/cycle handoff, and `expected_output_claim`.
- `jolt-kernels`: the matching `RafDecomposition` in the reference kernel.
- No R1CS, Spartan, claim-reduction, witness or BlindFold changes; no new opening, derived-id variant, or degree change.

**Alignment-assert immediate** (commit 1, unconditional, independently landable)

`VirtualAssertWordAlignment`/`HalfwordAlignment` carry `AddOperands`, but `FormatAssert` stored a *signed* `imm` unlike `FormatI`/`U`/`J`, which all store `u64`. A negative effective address gave a lookup index of `2^128 − |rs1+imm|` — unequal to `rs1+imm` in any field, hence unprovable — whose only satisfying representative `p − |rs1+imm|` sits in the alias band and **was accepted under akita**. Fixed at the source: `FormatAssert.imm` becomes `u64` and `expand_address` wraps via `format_i_imm`, matching the `expand_i` call on the next line of every load/store recipe. Both halves are needed — the format change alone leaves the bytecode `Imm` column disagreeing with the trace operands, surfacing as a `Stage 1 mismatch` in *bytecode* read-RAF (`stdlib_e2e_dory` catches it; `muldiv` does not).

**One upper-half suffix predicate, not two** (commit 3, refactor, no behavior change)

Added during review. The canonical-address suffix predicate existed twice: as `UpperHalfAllOnesPolynomial`'s `SuffixPolynomial` impl, and inline in `init_Q_raf`'s fused trace scan. Only the second one runs — the generic `init_Q` path, the sole caller of `suffix_mle`, is reached exclusively from a `#[cfg(test)]` helper. So `upper_half_all_ones_prefix_suffix_decomposition` was brute-forcing a copy the prover never executes, while the copy it does execute had no coverage, and a divergence between them would have left that test green. Both now share one `pub(crate) #[inline]` definition, which also makes the existing decomposition test meaningful for the production path. Kept as free functions rather than routing `init_Q_raf` through the trait, since the suffixes are `Box<dyn SuffixPolynomial>` and a per-row trait call would put dynamic dispatch in the scan's hot loop.

## Testing

```
cargo nextest run --workspace --features host                                2459 passed
cargo nextest run -p jolt-prover-legacy --features akita                      445 passed
cargo nextest run -p jolt-verifier --features akita,prover-fixtures            85 passed
cargo nextest run -p jolt-verifier --features prover-fixtures                 119 passed
cargo nextest run -p jolt-prover-legacy muldiv --features host,zk               3 passed
cargo clippy --all --features host / host,zk --all-targets -- -D warnings
cargo clippy -p jolt-prover-legacy --features akita --all-targets -- -D warnings
```

Verifier fixtures were cleared first — they cache proofs across a shape change.

**Falsification.** Passing tests don't distinguish a live term from one silently zero on both sides. Flipping `upper_half_all_ones` to the *trailing* half makes `akita_muldiv_fixture_verifies` fail with `StageClaimOutputMismatch { stage: 5 }` — the same signature the original report's BN254 control produced. Forcing the gate on for all three implementations on BN254 also passes (119/119), independently checking the prover's prefix/suffix decomposition against the verifier's closed form in a second field. For commit 3, mutating each copy of the predicate separately confirms it is load-bearing: before the unification, mutating only `identity_poly.rs` left `init_Q_raf` disagreeing and the akita read-RAF sumcheck failing to close — exactly the silent drift the commit removes.

**New tests**: brute-force of the decomposition against direct MLE evaluation at three chunk geometries (inside a half, on the midpoint, straddling it); a registry invariant that no non-interleaved instruction can reach an all-ones upper limb; `SUB(2^64−1, 0)` has `B = 0` (its *low* limb is all ones — this catches a leading/trailing transposition); a test pinning the `γ³` term inside `EqRafFlag`, which rides in a derived public where a refactor could drop it with nothing failing to compile; and a compile-time assertion of `2^128 − 2^64 < p`.

## On an end-to-end malicious-proof test

@markosg04 @quangvdao — worth recording why this is not in the PR, because it is a structural limit rather than an omission.

An aliased-address proof cannot be produced without a source change to a prover, so a rejection test cannot be written against unmodified provers:

- **The address is not a wire value.** It is the stage-5 sumcheck's address *point*, implied by which one-hot `ra` the prover committed. No mutation of the serialized proof can express the attack, which is why it does not fit the existing `tamper_manifest` shape — that harness perturbs finished proofs, and the alias must be introduced before proving.
- **The cycle → index map is closed.** `LookupQuery` is declared in `jolt-prover-legacy` and implemented for `tracer::instruction::Cycle`, so the orphan rule forbids an external override, and the prover is not generic over an index provider. Feeding a hand-crafted `Vec<Cycle>` through the public `gen_from_trace` cannot reach `[p, 2^128)` either: across every non-interleaved family the reachable maximum is `MUL`'s `(2^64−1)^2 = 2^128 − 2^65 + 1`, short of `p` by `2^65 − 2^32 + 22536`. That is exactly the invariant `non_interleaved_lookup_indices_are_canonical` pins.

The original report's PoC therefore patched the prover, behind a `fp128-forgery-poc` feature that overrode `to_lookup_index`, `to_lookup_output` and the tracer's `rd` write together, so that the read leg and register checking stayed self-consistent. A permanent default-off knob that makes the prover commit an arbitrary address is itself soundness surface, which is what the request to avoid patching a prover is guarding against.

The cheaper place to land this is the **modular** path once it builds under akita: `jolt_prover::prove` is generic over `W: JoltWitnessOracle`, and `jolt_riscv::JoltCycle` is a public trait an external crate can implement, so a test-only oracle can report an aliased `LookupIndex` with no production code involved.

There is also a complementary direction that needs no prover change and is not yet covered: **false-rejection** tests driving honest traces through the real stage-5 sumcheck at the three edges where a wrong predicate stops being visible to the algebraic unit tests — `MUL(2^64−1, 2^64−1)` (the largest identity-RAF index, one bit short of the rejected band), `XOR(2^64−1, 2^64−1)` (index `2^128−1`, which *does* have an all-ones upper half and is legal only because the term is masked to identity-RAF rows), and `SUB(2^64−1, 0)` (an all-ones low limb). Each is the unique catcher for a different mutation — a one-bit widening, a dropped `raf_flag` mask, and a leading/trailing transposition respectively — so the set is non-redundant. Happy to land these separately.

## Security Considerations

Fixes arbitrary corruption of an accepted execution trace under `--features akita`; the original report produced two accepted proofs for the same program and inputs with different public outputs.

Soundness of the replacement rests on the input claim being transcript-fixed before `γ` (not on it being *correct* — the γ-degree argument is independent of that), `r_red` drawn after commitment, and `ra` being one-hot, which the existing identity leg already depends on. No new trust assumption; added error `≈2^-121`.

Over-rejection is deliberate and bounded: the predicate excludes `[2^128 − 2^64, 2^128)`, a strict superset of `[p, 2^128)`. Nothing lands in the gap — add/sub `< 2^65`, mul `≤ (2^64−1)^2` (high limb `2^64−2`, one short), advice `< 2^64`. The registry test pins this.

**Trap for future edits:** the `γ` power differs by site. The legacy prover and reference kernel multiply the whole RAF bracket by an outer `γ`, so the new term is written `γ²` there and `γ³` only in the verifier. A transposition is silent at compile time; commented at each site.

The `jolt-kernels` change is currently unexercised — `jolt-kernels` cannot build in akita mode, so the modular prover has no akita path. `cargo check -p jolt-kernels --features jolt-verifier/akita` gives **3** pre-existing type errors in files this PR does not touch: two lattice dimension-wrapper constructions (`reference/booleanity.rs`, `reference/hamming_weight_claim_reduction.rs`) and a `[Vec<F>; 5]` vs `[Vec<F>; 9]` arity mismatch in `reference/bytecode_read_raf.rs`. Note `--features jolt-claims/akita` is *not* a valid way to enable it — that half-enables lattice mode and breaks `jolt-verifier` itself. Enabling the path is more than those three errors: `READ_RAF_CYCLE_STAGES` going 5 → 9 needs the four fused-inc points plumbed from `jolt-prover`'s stage 6a, and `reference/bytecode_read_raf.rs`'s `gamma_powers: [F; 8]` is undersized for the lattice formula's `γ¹¹`. The kernel change is included here so the term is correct when that path lands; omitting it would mean silently rejected proofs (a loud completeness failure, not a soundness one).

## Breaking Changes

Proof compatibility breaks; wire formats do not.

- **Akita**: `EqRafFlag`'s value changes, so round polynomials and downstream Fiat-Shamir change — existing akita proofs will not verify against this verifier and vice versa. No struct, opening, derived-id or serialization shape changed.
- **All fields**: the alignment fix alters expansion for negative-offset `LH`/`LHU`/`LW`/`LWU`/`SH`/`SW`, so bytecode and proofs over programs containing one change. 16 of 360 expansion golden hashes were re-baselined — byte accesses emit no alignment assert and non-negative offsets wrap to themselves, so both are unchanged.
- **Public API**: `tracer`'s `FormatAssert.imm` changes from `i64` to `u64` — a public field on a public struct, carried by `VirtualAssertWordAlignment` and `VirtualAssertHalfwordAlignment`. External code constructing or reading it with a signed immediate must wrap (`imm as u64` / `imm as i64`), the convention `FormatI`/`FormatU`/`FormatJ` already use. The two's-complement bit pattern is unchanged, so serialized traces are unaffected; this is a source-level break only. No other public API changes.


markosg04 (approved):
Is it possible to add an e2e regression test for this kind of thing in `jolt-verifier` without patching one of the prover implementations?

quangvdao (approved):
Reviewed the full merge-base diff at head `59e9ec797d7b732246217c950a0665ab3e46b538`. No blocking findings; approving.

I checked the fp128 alias argument and the conservative upper-half-all-ones predicate, the unique gamma^3 coefficient and transcript ordering, unchanged sumcheck degree/opening surface, parity across the legacy prover, reference kernel, symbolic relation, and verifier, and honest-input completeness including wrapped alignment immediates. I also ran the focused Akita canonicality, verifier-derived-term, legacy prefix/suffix, format-wrap, and registry-invariant tests locally; all passed.

I agree with Markos that a malicious-proof E2E regression would be valuable follow-up coverage. It would also be good follow-up to exercise the modular reference kernel under Akita and to update the PR prose to acknowledge the public `FormatAssert.imm` type change. I consider those nonblocking for this fix.

PR #1710, opened 2026-07-28
perf(inline): implicit carry handling and ADDC and MULC instructions.

mathmasterzach:
## Summary

Proposes a spec for adding support for implicit carries (the upper 64 bits) of ADD and MUL instructions that can be used in subsequent (new) ADDC and MULC instructions. This will reduce big integer and wrong-field arithmetic (e.g. ECDSA verification) instruction counts by 2-4x. The change is largely a small modification to the R1CS instance, no explicit registers.


moodlezoup:
**Spec Analysis: implicit-carry-handling** (spec @ `e332829`)

| Dimension | Score | Gap |
|-----------|-------|-----|
| Goal | 0.85 | Clear |
| Constraints | 0.70 | Boundary conditions (row 0, trace tail) undefined |
| Success Criteria | 0.85 | Clear — concrete positive/negative tests in both modes |
| Spec–Codebase Consistency | 0.55 | Constraint guards contradict the actual flag universe |
| **Ambiguity** | | **22%** |

**Status: Questions remain** — 2 blocking item(s) to resolve before implementation.

The spec is unusually accurate on file paths and symbol names — every referenced file, macro, trait signature (`Extract::extract(row, next, env)`), constraint label, and the 5-value shift relation checked out against `main`. The two blocking items below are design-level, not clerical.

**Questions:**

**1. [Blocking — Spec–Codebase Consistency]** Constraints 5 and 6 are guarded on `AddOperands || MultiplyOperands`, but those flags are set by far more final instructions than ADD and MUL. `AddOperands` is declared by ADDI, LUI, AUIPC, JAL, JALR, both alignment asserts, the four Pow2 virtuals, VirtualRev8W, the shift-bitmask virtuals, and sign/zero-extend-word (`crates/jolt-riscv/src/instructions/*/[name].rs:6`); `MultiplyOperands` by MULHU, VirtualMULI, and AssertMulUNoOverflow. Two concrete breakages: (a) MULHU's lookup output is the **upper** word (`impl_lookup_table!(MulHU, Some(UpperWord))`, `crates/jolt-lookup-tables/src/instructions/riscv/mulhu.rs:6`), so constraint 5 (`RightLookupOperand == LookupOutput + 2^64·NextCarry`) is unsatisfiable for honest MULHU rows; (b) for ADDI/JAL/etc., constraint 5 forces `NextCarry` to the real high bits while the spec's own Intent policy and tracer model set the stored carry to 0 — an honest prover is rejected whenever an ADDI overflows. This also makes the Intent section ("for all other instructions, `NextCarry = 0`") and the Design section internally inconsistent. Candidate fix: introduce a second circuit flag (e.g. `ProducesCarry`) set on exactly {ADD, MUL, ADDC, MULC}, gate constraint 5 on `ProducesCarry` and constraint 6 on `!ProducesCarry`, and leave constraints 1–4 as proposed (they degrade gracefully: `UsesCarry = 0` reproduces today's `RightLookupAdd`/`RightLookupEqProductIfMul` for all other flag-bearers). Note this brings `CircuitFlags` to 16 variants — exactly filling `CircuitFlagSet(u16)` (`crates/jolt-riscv/src/flags.rs:104`), still fine. Confirm this (or specify an alternative gating)?

**2. [Blocking — Constraints]** The spec's invariant fixes `Carry(0) = 0`, but no mechanism in the design enforces it. The shift machinery is built on `EqPlusOne`, which has **no wrap-around and constrains nothing about row 0 of the shifted column** (`crates/jolt-poly/src/eq_plus_one.rs:1-8`); the only initial-value precedent in the codebase is PC(0), which is enforced by bespoke machinery in the bytecode read-RAF relation (the `Entry` public term, `crates/jolt-claims/src/protocols/jolt/geometry/bytecode.rs:93,122`; legacy `crates/jolt-prover-legacy/src/zkvm/bytecode/read_raf_checking.rs:344-361`) — there is no analogue for an arbitrary virtual column. Options: (i) add an `eq(0…0, r_cycle)`-weighted public term to the shift relation (mirroring the bytecode `Entry` pattern) forcing `Carry(0) = 0`; (ii) explicitly weaken the invariant to "Carry(0) is unconstrained but harmless", arguing `Carry` is only consumed under `UsesCarry` and the program entry sequence never begins with ADDC/MULC. Option (i) matches the invariant as written and seems intended — confirm, and state the mechanism in the Design section?

- **Guest encoding / test emission is unspecified but has a forced default.** VirtualRev8W is a complete precedent for a guest-emittable custom final instruction: opcode `0x5B`, funct3 `0b000`, next free funct7 (`0x06`/`0x07`; `0x00–0x05` are taken, `tracer/src/instruction/mod.rs:1273-1306`), emitted via `.insn r` (see `jolt-inlines/sha2/src/sdk.rs:441-449`). Default assumption: ADDC/MULC follow that pattern, and e2e tests emit them via raw `.insn` in a test guest.
- **Trace-tail semantics**: with `NextCarry` extracted as `next.map(|r| r.carry).unwrap_or(0)` and padding rows being `TraceRow::default()` (carry 0), constraint 5 is unsatisfiable if the last executed row were a carry-producing instruction with nonzero carry-out. This is unreachable in practice (traces end with the termination ecall sequence), but the spec should state that assumption explicitly so randomized/negative tests don't construct spuriously-failing raw traces.
- **File list misses several load-bearing sync points** (all compiler/test-discoverable, listing them saves a round trip): legacy shift sumcheck `crates/jolt-prover-legacy/src/zkvm/spartan/shift.rs` (`gamma_powers: [F; 5]` → 6, plus its four `#[cfg(feature = "zk")]` BlindFold constraint fns — the CLAUDE.md-critical claim/constraint sync), `crates/jolt-prover-legacy/src/zkvm/spartan/outer.rs` fixed-size `[F; NUM_R1CS_INPUTS]` arrays, the modern constraint table in `crates/jolt-r1cs/src/constraints/jolt.rs`, the second instruction decoder `crates/jolt-program/src/image/decode.rs:182-203`, the per-kind meta arms in `crates/jolt-riscv/src/kind.rs`, the hand-written `JoltInstruction` enum registration in `crates/jolt-riscv/src/instructions/mod.rs:338-626`, and `Cpu::save_state_with_empty_memory` (`tracer/src/emulator/cpu.rs:1133`). Also: replacing/adding constraint labels changes `NUM_R1CS_CONSTRAINTS` (19 → 23), which moves `OUTER_UNIVARIATE_SKIP_DEGREE` and requires re-syncing the hardcoded uniskip group arrays (`R1CS_CONSTRAINTS_FIRST_GROUP_LABELS`, `crates/jolt-prover-legacy/src/zkvm/r1cs/constraints.rs:504-516`; `SPARTAN_OUTER_FIRST/SECOND_GROUP_ROWS`, `crates/jolt-claims/src/protocols/jolt/geometry/spartan.rs:22-24`). Default: keep the first group at the same 10 labels and append new labels to the second group.
- Cosmetic: the spec's markdown links use a `jolt/` path prefix that 404s in-repo (use repo-relative paths), and the header's PR field can now be filled in (#1710).

---

moodlezoup:
**Spec Analysis: implicit-carry-handling** (spec @ `658f7d7`)

| Dimension | Score | Gap |
|-----------|-------|-----|
| Goal | 0.90 | Clear |
| Constraints | 0.75 | Opening-claim grounding mechanism undefined |
| Success Criteria | 0.85 | Clear (new invariant loosely specified) |
| Spec–Codebase Consistency | 0.60 | Proposed guards/reduction path don't fit the constraint DSL / claim DAG |
| **Ambiguity** | | **19%** |

**Status: Questions remain** — 2 blocking item(s) to resolve before implementation.

Both blocking items from the previous analysis (spec @ `e332829`) were incorporated into the spec text and are resolved: `ProducesCarry` now gates the split constraint with the MULHU/ADDI rationale spelled out, and row-0 initialization has an explicit `CarryInit` mechanism mirroring the bytecode `Entry` pattern. The revised file inventory also checked out almost entirely against `main` — the feature plumbing (`field-inline` propagation pattern across all 10 Cargo.tomls), the free `funct7 = 0x06/0x07` slots under opcode `0x5B`/funct3 `0b000`, the 14-variant `CircuitFlags`/`u16` packing math, `gamma_powers: [F; 5]`, the dual legacy/modern `CircuitFlags` enums, and the post-#1686 verifier stage files all verified. This round's questions are new, and only visible because the spec now commits to concrete mechanisms: the proposed R1CS guards cannot be expressed in the constraint DSL, and the new virtual columns have no path that grounds their opening claims.

**Questions:**

**1. [Blocking — Spec–Codebase Consistency]** Constraints 1–4 cannot be built as written, and replacing `RightLookupAdd`/`RightLookupEqProductIfMul` with them opens a soundness hole. Constraint guards in `r1cs_eq_conditional!` must be *affine* linear combinations: `LC` supports at most 5 terms plus a constant with no `LC × LC` (`crates/jolt-prover-legacy/src/zkvm/r1cs/ops.rs:46-59`), and every constraint compiles to `guard · (left − right) = 0` with no C matrix (`constraints.rs:144-152`). Conjunctions are only expressible via the subtraction trick when one flag implies the other (e.g. `VirtualInstruction − IsLastInSequence`, `constraints.rs:396-400`) — that fails here because `UsesCarry` does not imply `AddOperands` (MULC sets `UsesCarry` without it), so guards like `AddOperands && UsesCarry` (constraint 2) and `AddOperands && ProducesCarry && !UsesCarry` (constraint 1) have no affine encoding. Separately, because constraints 1 and 3 require `ProducesCarry`, the rows that only set `AddOperands` (ADDI, LUI, AUIPC, JAL, JALR, VirtualRev8W, …) or only `MultiplyOperands` (MULHU, VirtualMULI, AssertMulUNoOverflow) match none of constraints 1–4, and `RightLookupEqRightInputOtherwise`'s guard (`1 − Add − Sub − Mul − Advice`, `constraints.rs:313-317`) is 0 on them too — their `RightLookupOperand` becomes fully unconstrained. Candidate fix that stays affine, keeps coverage total, and preserves the 16-flag packing: replace nothing; instead (i) add a product-virtualization row `CarryUsed = OpFlags(UsesCarry) · Carry` — exact precedent: `ShouldJump` is defined as `Jump · (1 − NextIsNoop)` in the product-virtualization stage (`constraints.rs:354-360`, `PRODUCT_CONSTRAINTS` at `:573`; note `NUM_PRODUCT_CONSTRAINTS` 3 → 4 affects product uni-skip sizing); (ii) amend only the *bodies* of the two existing constraints: `RightLookupAdd` → `RLO == Left + Right + CarryUsed`, `RightLookupEqProductIfMul` → `RLO == Product + CarryUsed` (both degenerate to today's forms whenever `UsesCarry = 0`, so ADDI/MULHU/etc. keep today's semantics); and (iii) keep just the two genuinely new rows, `LookupSplitsIntoOutputAndNextCarry` (`if ProducesCarry`) and `NextCarryZeroIfNotProducesCarry` (`if 1 − ProducesCarry`), whose guards are already affine. Alternative: dedicated exclusive operand flags (`AddCarryOperands`/`MulCarryOperands`, with ADDC/MULC *not* setting `AddOperands`/`MultiplyOperands`) keeps every guard single-flag but costs a 17th flag (widening `CircuitFlagSet` from `u16`) and edits to both "otherwise" guards. Confirm the `CarryUsed` product-row approach (or pick the alternative), and update the Detailed R1CS Changes section accordingly?

**2. [Blocking — Constraints]** Nothing grounds the `Carry`/`NextCarry` opening claims, and the spec's reduction path dead-ends. Every virtual opening must reduce through the claim DAG to committed or memory-checked data. The five columns threaded through the shift today all terminate in bytecode: their shifted openings (`unexpanded_pc_shift`, `is_noop_shift`, …) are consumed by bytecode read-RAF (`crates/jolt-claims/src/protocols/jolt/relations/bytecode/read_raf.rs`), and `NextIsNoop` is produced by the product-remainder relation. `Carry` is trace-dynamic — it is not a bytecode column, not committed, and has no defining relation — so after the proposed shift step reduces `NextCarry(r_outer)` to a `Carry` opening at the shift point, that claim (and the outer `Carry(r_outer)` claim) is unverifiable. Two candidate resolutions: **(a) commit `Carry` as a new committed polynomial** (u64 compact scalars, like `RdInc`) — the rest of the spec then works essentially verbatim: outer `Carry` claim discharged by a direct opening, `NextCarry` reduced through the existing forward `EqPlusOne` shift exactly as the Design section describes, `CarryInit` enforcing row 0 of the committed column; cost is one additional committed column, and the Invariants bullet "Carry is visible only through proof-level virtual columns" must be relaxed to "not architectural, not memory-checked". **(b) Keep both columns virtual** and ground `NextCarry` definitionally via a new claim-reduction relation `NextCarry(r) = Σ_t eq(r,t) · ProducesCarry(t) · (RightLookupOperand(t) − LookupOutput(t)) · 2^{−64}` (all three referenced openings already have reduction paths), running the shift in the *opposite* direction from the spec text (consume `Carry(r_outer)`, produce shifted `NextCarry`); this needs bespoke machinery (the `ProductFactorExpr` grammar only supports `Var`/`OneMinus` factors, not affine combinations) and makes R1CS constraints 5–6 redundant. Option (a) is substantially less new machinery and matches the spec's shift direction as written — confirm which, and state the mechanism in the Design section?

- **Group-rows re-sync targets the wrong crate**: `SPARTAN_OUTER_FIRST_GROUP_ROWS`/`SPARTAN_OUTER_SECOND_GROUP_ROWS` in `jolt-claims/.../geometry/spartan.rs:22-24` are private; the public, feature-gated pair (already carrying `field-inline`/base variants — the pattern to extend) lives in `crates/jolt-r1cs/src/constraints/jolt.rs:69-108`. Default: edit both; compile errors will guide.
- **"Keep the first group unchanged" is arithmetically impossible**: the first group must have exactly `OUTER_UNIVARIATE_SKIP_DOMAIN_SIZE = (NUM_R1CS_CONSTRAINTS − 1)/2 + 1` labels (`constraints.rs:413-416, 504-516`), so adding 2 rows (19 → 21) forces the first group from 10 to 11. Default: move the lowest-degree new row (`NextCarryZeroIfNotProducesCarry`) into the first group.
- **The protocol-config analogy to `field-inline` doesn't hold**: `JoltProtocolConfig` has exactly two axes (`zk`, `commitment`; `crates/jolt-verifier/src/config.rs:36-40`) and no `field-inline` bit today. Adding `implicit_carry` as a third axis is still the right fail-closed call — default: proceed as specced.
- **Carry-clobbering footgun deserves a documentation bullet**: any instruction between a producer and its `ADDC`/`MULC` consumer silently zeroes the carry (by policy), changing results *without* proof failure — e.g. dependent pairs emitted across separate `asm!` blocks that the compiler interleaves with spills. Default: inlines.md documents that dependent sequences must be contiguous in the final instruction stream (single `asm!` block).
- **The new jolt-eval invariant is named but not shaped**. Default: follow the four-step framework in `jolt-eval/README.md` (module + `#[jolt_eval_macros::invariant(Test, Fuzz)]` + dispatch-enum registration + fuzz target), checking randomized `{ADD, MUL} → {ADDC, MULC}` correctness and carry-tamper rejection.

---

PR #1712, opened 2026-07-28
feat(profiling): per-stage and per-sumcheck perfetto coverage for the modular prover

0xAndoroid:
## What

Measurement foundation for the modular-prover performance campaign: apples-to-apples Perfetto profiling and memory reporting for **both** provers, plus a machine-queryable telemetry pipeline (`summary.json` + `jolt-eval` objectives) over the same span stream. The modular prover previously had essentially zero spans; a trace could not attribute time or memory to a stage, let alone a sumcheck member.

### Span coverage (modular path)

| Layer | Spans |
|---|---|
| `jolt-prover` | `prove`, `prove_stage0`..`prove_stage8` (names match legacy's `prove_stage1..8`); the generated stage driver emits `<StageLabel>::prove` (e.g. `Stage2Batch::prove`), per-member `<Relation>::prepare`, per-round `<Relation>::prove_round`, and terminal `<Relation>::finish_rounds` (the counterparts of legacy's `*SumcheckProver::initialize` / `::compute_message`); `commit_witness`/`commit_advice` at the stage-0 call boundary, backend-neutral |
| `jolt-kernels` | bespoke slots: `JointOpeningPolynomials::prepare`, `SpartanOuterUniskip::*`, `SpartanProductUniskip::*`, `AdviceOpeningEvaluation::evaluate`, `build_committed_bytecode_chunk_coeffs` |
| `jolt-sumcheck` | `prove_batch` (+ `sumcheck_round` per round, ~log T per batch), `prove_uniskip_clear` / `prove_uniskip_committed` (mode-selected by the `zk` feature) |
| `jolt-witness` | `stream_witnesses`, `collect_bundles`, `TraceBackend::oracle_table` (per-polynomial witness materialization) |
| `jolt-openings` | `HomomorphicBatch::prove_batch` / `HomomorphicBatch::prove_batch_zk` (mode-selected) |

The label set is a versioned public schema: `crates/jolt-profiling/src/taxonomy.rs` (taxonomy v1), with `always_present_spans(ProverMode)` returning the per-mode presence set the smoke test asserts. No spans inside per-index inner loops. `jolt-dory` was already instrumented. Per-member round spans come from a `SpannedRounds` shim the generated driver wraps around each kernel — instrumentation only, transparent delegation.

**⚠️ `jolt-prover`/`jolt-sumcheck`/`jolt-openings`/`jolt-witness` are protocol crates: every change there is instrumentation-only (`#[tracing::instrument]` attributes, the span shim, and the `tracing` dependency). No protocol logic changed — the full byte-diff suite (`-p jolt-prover --features prover-fixtures`, incl. all `prover_matches_legacy_*`) still passes, so proof bytes remain identical to legacy. `jolt-verifier` carries only feature-gated `allocative` derives (attribute-only, inert without the feature).**

### Profile harness + telemetry (modular)

The retired `examples/modular_benchmark.rs` grew into the `profiling`-gated `[[bin]] jolt-prover` (`profile` and `benchmark` subcommands — the latter ports `scripts/jolt_benchmarks.sh`'s multi-scale sweep, one subprocess per run), mirroring the legacy CLI: same guest programs, same input construction, target = 90% of 2^scale. Runs the full modular pipeline (legacy-side guest compile/decode → `TracerBackend` trace → derived `ProverConfig` → `TraceBackend` witness → `jolt_prover::prove` over `JoltBackend::reference()`) and **verifies the proof** as a correctness gate. Only `prove` is timed — the same window legacy times.

```bash
# modular: chrome trace + machine-queryable summary from one span stream
cargo run --release -p jolt-prover --features profiling -- \
    profile --name sha2-chain --scale 16 --format chrome
# legacy
cargo run --release -p jolt-prover-legacy benchmark --name sha2-chain --scale 16 --format chrome
```

Artifacts land in `benchmark-runs/perfetto_traces/`: `modular_sha2_chain_16.json` (Perfetto UI / `trace_processor` SQL, counter events rewritten to native `"ph":"C"` tracks at flush) **and** `modular_sha2_chain_16.summary.json` — schema-versioned aggregates (`crates/jolt-profiling/schema/summary.schema.json`, drift-locked by fixture tests): per-label totals/self time, per-stage wallclock + boundary RSS + windowed peak memory, root dark time, headline peak RSS. CSVs in `benchmark-runs/results/` (`modular_*` prefix, legacy-compatible columns).

`jolt-eval` gains string-keyed objectives over these artifacts: `telemetry:<workload>:<metric>` (prover time, peak RSS/memory, per-label total/self time) reading `summary.json`, and `callgrind:<bench>:instructions` (deterministic Ir counts via iai-callgrind, parser schema-locked against the real `iai-callgrind-runner` 0.16.1 summary types).

A `jolt-prover` e2e smoke test (fibonacci 2^13, `--features profiling`) asserts both artifacts parse and every always-present taxonomy label fired, in clear **and** zk modes. Deliberately not CI-wired yet — the reference backend's ~18 GiB retained footprint exceeds hosted runners (NOTE in `rust.yml`); run explicitly after taxonomy changes.

### Memory measurement (both harnesses)

- **Headline peak RSS**: `jolt_profiling::peak_rss_bytes()` — `getrusage(RUSAGE_SELF).ru_maxrss`, the kernel high-water mark (can't miss short spikes, unlike sampling). Printed by both harnesses; sampled right after the workload so flush-time trace postprocessing can't inflate it.
- **Per-stage RSS**: `StageMemoryLayer`, a `tracing` layer sampling RSS at `prove_stage*` span boundaries (plus `prove_parts`/`E2E` roots; row log bounded). Both harnesses print the same table format and emit `stage_rss` instant events into the Chrome trace. Boundary sampling attributes *retained* growth per stage; spikes are the headline number's job.
- The legacy bin picks both up via a small additive change (installs the layer, prints peak RSS at the end). The summary pipeline is feature-gated (`jolt-profiling/summary`) so standard `--features host` builds keep their pre-PR dep graph.

### First data points (scale 16, sha2-chain, M-series macbook)

| | prove time | padded kHz | peak RSS |
|---|---|---|---|
| legacy | 1.39 s | 47.3 kHz | ~200 MiB |
| modular (reference backend) | 1254 s | 0.05 kHz | 11.72 GiB |

Expected: the reference tier is the naive eager-dense oracle ("a test oracle at harness scale, never a performance path"). The per-stage table already localizes the memory: stage 2 (Spartan product virtualization) retains ~9–10 GiB of naive kernel tables, released across stages 4–5. Span overhead is noise-level (fibonacci 2^14: 68.5 s with chrome tracing vs 69.4 s without).

## Review fixes (Jul 31)

Both AI review passes ([codex](https://github.com/a16z/jolt/pull/1712#pullrequestreview-4831964098), [claude](https://github.com/a16z/jolt/pull/1712#pullrequestreview-4831983234)) addressed in `9adc52d71..c89e4fc00`: real iai-callgrind summary-schema parser + runner-typed fixture lock; backend-neutral commit seams; zk-mode taxonomy + instrumented committed siblings; `<Relation>::finish_rounds` label; workload-scoped peak RSS; atomic trace/summary writes; occurrence-indexed stage-row pairing; bounded stage-memory log; `--scale` validation; per-run artifact lock; `BackendKind::as_str()`; stale-summary removal in the optimizer loop; `jolt-profiling/summary` feature gate; smoke-test CI wording. Point-by-point dispositions on the review threads.

## Validation

- `cargo clippy --all --all-targets -- -D warnings` under `--features host` and `--features host,zk`
- `cargo nextest run` for `-p jolt-profiling` (schema drift-lock + pipeline fixtures), `-p jolt-eval`, `-p jolt-sumcheck -p jolt-openings -p jolt-kernels`, `-p jolt-prover` (+ byte-diff via `--features prover-fixtures`), `-p jolt-prover-legacy muldiv` under `host` and `host,zk` — all green
- Profiling smoke test green in clear and zk modes; modular proof verifies; Perfetto JSONs contain the per-stage, per-member, and per-round spans listed above


jolt-eval/src/objective/callgrind.rs
```diff
@@ -0,0 +1,186 @@
...
+    }
+}
+
+/// Depth-first search for the first `Ir` metric in one benchmark summary
+/// document. iai-callgrind nests the metrics under the profile's total
+/// summary; the first `Ir` in document order is that total.
+fn find_ir(value: &Value) -> Option<f64> {
+    match value {
+        Value::Object(map) => {
+            if let Some(ir) = map.get("Ir") {
```
0xAndoroid: Real iai-callgrind 0.16 JSON stores `Ir` as a `MetricsDiff` (`{"diffs":...,"metrics":{"Left":{"Int":...}}}`), not the test’s direct `{"Int":...}`. `as_metric_number` rejects that multi-field object, then recursion loses the `Ir` association, so successful benchmark output returns `no Ir event kind`. Parse the versioned summary schema’s `metrics.Left|Both` shape and fixture-test real runner output.

crates/jolt-kernels/src/reference/commitment.rs
```diff
@@ -39,6 +39,11 @@ where
     F: Field,
     PCS: CommitmentScheme<Field = F> + ModeStreamingCommitment,
 {
+    #[tracing::instrument(
```
0xAndoroid: This instruments the `ReferenceBackend` implementation rather than the kernel seam. Any other `JoltBackend` implementation gets no `commit_witness`/`commit_advice` labels, although the taxonomy treats them as backend-neutral spans. Put the span at the stage call boundary or behind a shared wrapper so every backend inherits it.

crates/jolt-profiling/src/taxonomy.rs
```diff
@@ -0,0 +1,144 @@
...
+    "HomomorphicBatch::prove_batch",
+];
+
+/// Every v1 label that fires on all proves: the presence set the `jolt-prover`
+/// profiling smoke test asserts against a freshly emitted trace.
+///
+/// Deliberately excludes the per-member driver spans (`<Relation>::prepare`,
+/// `<Relation>::prove_round`) whose names vary with the batch composition,
+/// plus the advice- and committed-program-only seams.
+pub fn always_present_spans() -> Vec<&'static str> {
```
0xAndoroid: `always_present_spans()` includes clear-mode-only `prove_uniskip_clear` and `HomomorphicBatch::prove_batch`, while ZK dispatches to uninstrumented `prove_uniskip_committed` and `prove_batch_zk`. A ZK profile therefore violates this advertised presence set and loses those attribution seams. Instrument both ZK siblings and give the taxonomy mode-correct labels.

crates/jolt-profiling/src/summary.rs
```diff
@@ -0,0 +1,564 @@
...
+
+    let timestamp_unix_secs = std::time::SystemTime::now()
+        .duration_since(std::time::UNIX_EPOCH)
+        .map(|d| d.as_secs())
+        .unwrap_or_default();
+    let summary = build_summary(
+        &events,
+        ctx,
+        &crate::stage_memory::take_stage_memory_rows(),
+        crate::memory::peak_rss_bytes(),
```
0xAndoroid: `peak_rss_bytes()` is sampled after the full trace has been read, expanded, reserialized, and rewritten. That postprocessing can become the process high-water mark, so `peak_rss_gib` reports tooling memory rather than the profiled workload. Capture the high-water mark immediately after the workload and pass it into finalization.

jolt-eval/bin/optimize.rs
```diff
@@ -147,6 +147,39 @@ impl OptimizeEnv for RealEnv {
...
+        // in the worktree (stale-binary trap — see check_invariants below);
+        // a failed measurement stays absent from the map, scoring INFINITY.
+        // One profile run per workload serves every telemetry objective
+        // sharing it (e.g. all ten per-stage totals read one summary).
+        let mut profiled_workloads: std::collections::HashSet<&str> =
+            std::collections::HashSet::new();
+        for obj in objectives {
+            match obj {
+                OptimizationObjective::Telemetry(t) => {
+                    if profiled_workloads.insert(t.workload) {
```
0xAndoroid: The workload is marked profiled before `run_profile_in` succeeds. If the first objective fails, later objectives skip profiling and can read the deterministic summary left by a previous candidate, mixing stale and current measurements. Insert into the set only after success, and remove the old summary before launching the run.

0xAndoroid (commented):
AI review (Codex).

Posted 5 inline findings:
- iai-callgrind’s real JSON schema is not parsed, breaking instruction-count objectives.
- Kernel-seam spans are attached only to `ReferenceBackend`, not inherited by other backends.
- The always-present taxonomy is clear-mode-only and fails for ZK profiles.
- Peak RSS is sampled after trace postprocessing, contaminating the workload metric.
- Failed telemetry runs can cause later objectives to consume stale candidate summaries.

Comment-only review; no approval or request-changes verdict.

crates/jolt-prover/src/driver.rs
```diff
@@ -119,6 +120,43 @@ pub struct Proved<F: Field, S: StageProver<F>, C> {
...
+        &mut self,
+        bind: Option<F>,
+        round: usize,
+        previous_claim: F,
+    ) -> Result<UnivariatePoly<F>, SumcheckError<F>> {
+        let _guard = (self.span)().entered();
+        self.inner.prove_round(bind, round, previous_claim)
+    }
+
+    fn finish_rounds(&mut self, bind: F) -> Result<(), SumcheckError<F>> {
```
0xAndoroid, line 160: nit (naming/attribution): `finish_rounds` also runs under a span named `<Relation>::prove_round`, so the final-bind work lands in the same per-label bucket as the round-loop slices — and, unlike them, nests directly under `prove_batch` rather than a `sumcheck_round`. Deliberate per the struct doc, but `telemetry:<workload>:total:<Relation>::prove_round` consumers get rounds+finish conflated, and the last "prove_round" slice on a Perfetto track isn't a round. A distinct `<Relation>::finish_rounds` label (or a taxonomy note documenting the conflation) would make the schema self-describing.

crates/jolt-profiling/src/taxonomy.rs
```diff
@@ -0,0 +1,144 @@
...
+//! Everything else is label-only; fields are additive (adding one is not a
+//! schema change, removing or renaming one is).
+//!
+//! # Evolving the taxonomy
+//!
+//! 1. Change the instrumentation and the constants here in the same commit.
+//! 2. Bump [`TAXONOMY_VERSION`] on any rename/removal of a label or field.
+//! 3. The `jolt-prover` profiling smoke test asserts every
+//!    [`always_present_spans`] label appears in a freshly emitted trace, so a
+//!    silent rename fails CI rather than drifting.
```
0xAndoroid: low (docs): this states CI enforcement as fact, but the smoke test is deliberately **not** wired into CI yet (the `rust.yml` NOTE and the smoke-test header both say so; the spec documents the deferral). Until the job lands, a silent rename drifts unless someone runs the smoke test manually — and this module is the normative doc. Same overstated claim in `CLAUDE.md` ("the profiling smoke test enforces label presence") and spec §Taxonomy conformance. Suggest wording like "fails the profiling smoke test (not yet CI-wired — see the NOTE in rust.yml)" so the guarantee isn't overstated.

crates/jolt-profiling/src/summary.rs
```diff
@@ -0,0 +1,564 @@
...
+/// Call after dropping [`TracingGuards`](crate::TracingGuards) — the chrome
+/// layer finalizes the trace file on guard drop.
+pub fn finalize_trace(
+    trace_path: &Path,
+    ctx: &SummaryContext,
+) -> Result<(PathBuf, ProfileSummary), SummaryError> {
+    let events = read_events(trace_path)?;
+    let events = convert_counter_events(events);
+    let trace_json = serde_json::to_string(&events)?;
+    std::fs::write(trace_path, trace_json).map_err(|source| SummaryError::Write {
```
0xAndoroid: low: in-place truncate+rewrite — `fs::write` truncates first, so a crash between truncation and completion (OOM is a live risk on the reference backend) destroys the only artifact of a potentially hours-long run. Writing to `{trace}.tmp` and `fs::rename`-ing over the original is cheap insurance for an artifact this expensive to regenerate.

crates/jolt-profiling/src/summary.rs
```diff
@@ -0,0 +1,564 @@
...
+            .stage_intervals
+            .iter()
+            .filter(|(label, _)| !taxonomy::STAGE_SPANS.contains(&label.as_str()))
+            .cloned(),
+    );
+
+    let stages = ordered
+        .into_iter()
+        .map(|(label, interval)| {
+            let row = stage_rows.iter().find(|row| row.stage == label);
```
0xAndoroid: nit: `find` pairs *every* same-label interval with the *first* matching `StageMemoryRow`. Only one prove per process exists today (the second-subscriber panic enforces it), but a trace with repeated stage labels would silently attach the first prove's boundary RSS to all instances. Rows and `stage_intervals` are both recorded in close order, so pairing by occurrence index (or a `debug_assert!` on label uniqueness) would keep this from biting a future multi-run harness.

jolt-eval/src/objective/callgrind.rs
```diff
@@ -0,0 +1,186 @@
...
+                if let Some(ir) = as_metric_number(ir) {
+                    return Some(ir);
+                }
+            }
+            map.values().find_map(find_ir)
```
0xAndoroid: low: the "first `Ir` in document order is the total" rationale (doc comment above) doesn't hold structurally: the workspace `serde_json` is `default-features = false` — no `preserve_order` — so `Value::Object` is a BTreeMap and this DFS visits keys in *sorted* order per level, not document order. It works today because each case document appears to carry a single `Ir`; if iai-callgrind ever emits part summaries alongside a total (e.g. multi-process runs), the pick becomes ordering-dependent and silently wrong rather than erroring — the exact silent-corruption mode the module doc forswears. An explicit pointer path (e.g. `/callgrind_summary/summaries/<i>/events/Ir` with a deliberate sum or an ambiguity error) would pin the semantics.

crates/jolt-prover/src/profile.rs
```diff
@@ -0,0 +1,454 @@
...
+    if args.format != OutputFormat::Chrome {
+        report_stage_memory();
+        return ProfileArtifacts::default();
+    }
+
+    let trace_path = PathBuf::from(format!("benchmark-runs/perfetto_traces/{trace_name}.json"));
+    let ctx = SummaryContext {
+        workload: args.name.as_str().to_string(),
+        scale_log2: scale,
+        backend: "reference".to_string(),
```
0xAndoroid: nit: hardcoding `"reference"` while `args.backend` is in scope means `run.backend` silently mislabels every summary the moment a second `BackendKind` variant lands — and this is exactly the run-identity metadata telemetry consumers key on. The `match` in `run_workload` is exhaustive, so a new variant forces an update there but not here. `BackendKind` could expose an `as_str()` used by both. (Fine to fix in #1714 if that's where the second backend arrives — it rewrites this file — but as-merged this PR leaves the trap.)

crates/jolt-prover/src/profile.rs
```diff
@@ -0,0 +1,454 @@
...
+    // modular-prefixed file name.
+    let summary_line = format!(
+        "{},{},{:.2},{},{:.2},{},{}\n",
+        bench_name,
+        scale,
+        duration.as_secs_f64(),
+        trace_length.next_power_of_two(),
+        padded_proving_hz,
+        proof_size,
+        proof_size,
```
0xAndoroid, line 628: low: legacy's 7th CSV field is `proof_size_comp` (compressed), not a duplicate of field 6 — see `crates/jolt-prover-legacy/benches/e2e_profiling.rs:172-181`. "The same 7-field CSV line the legacy harness writes" therefore doesn't hold for the last column: anyone diffing `modular_*.csv` against a legacy CSV gets a silently incomparable column. If the modular pipeline has no compressed encoding yet, a comment marking field 7 as a placeholder (or an empty field) beats duplicating the raw size under a column that means something else on the legacy side.

0xAndoroid (commented):
**AI review (Claude)** — automated deep review of this PR's own diff at `0c5dbdf8` (reviewed standalone from the #1714 stack). Comment-only, no verdict event.

Focus areas: the zero-overhead-when-disabled claim, span bracketing, prover-semantics neutrality, trace-file handling, API/naming.

### Verified clean

- **Shim transparency**: `SpannedRounds` forwards the complete `ProveRounds` surface (all 3 trait methods — no default-method trap), and every rewired driver-macro arm (`round_slot`/`validate`/`extract`/`park`) reaches the kernel through `.inner`. Structurally transparent, consistent with the green byte-diff suite.
- **Disabled-path cost**: every `#[instrument]` site uses `skip_all`; field expressions (`?id`, `members.len()`, `range.end - range.start`, …) are lazy behind the cached callsite-interest check; no formatting or allocation with no subscriber installed. Per-round/per-member spans respect the taxonomy's hot-loop floor — nothing per-index.
- **Prover semantics**: the only semantic-adjacent touch inside `prove()` is `stage_flamegraph` — double-gated (cfg `allocative` + runtime prefix opt-in), read-only visitation, and `write_flamegraph_svg` warns instead of panicking, so a prove can't die on an SVG write failure. Everything else in protocol crates is attributes / cfg_attr derives.
- **Thread safety**: `StageMemoryLayer` is inert (two string compares) for untracked spans; the `extensions()` guard drops before the re-entrant `stage_rss` event; the global row log handles Mutex poisoning.
- **Aggregation**: `finalize_trace` is idempotent w.r.t. counter conversion (tested); the per-thread B/E stack replay handles rayon per the documented semantics (cross-thread children never subtract from a parent); root/uniskip/round span bracketing matches what the labels claim.

### Findings

**0 high, 0 medium, 4 low, 4 nits** — details inline:

- Low: `taxonomy.rs` states CI enforcement that is deliberately deferred (smoke test not wired into `rust.yml`).
- Low: modular CSV column 7 duplicates `proof_size` where legacy writes the compressed size — silently incomparable columns.
- Low: callgrind `Ir` extraction relies on JSON map ordering that the workspace `serde_json` (no `preserve_order`) doesn't provide.
- Low: `finalize_trace` rewrites the trace in place — a crash mid-write destroys the only artifact of a potentially hours-long run.
- Nits: `finish_rounds` time attributed under `<Relation>::prove_round`; `run.backend` hardcoded to `"reference"`; first-row pairing for repeated stage labels in the summary rollup; `schemars` now compiled into every `--features host` build via `host → jolt-profiling`.

### PR description staleness

The body still describes `examples/modular_benchmark.rs` (now the `profiling`-gated bin + smoke test + summary/telemetry pipeline + jolt-eval objectives) and says "**jolt-verifier is untouched**" (it now carries the feature-gated `allocative` derives — attribute-only, zero impact without the feature, but the bolded claim is stale). Worth a refresh so the description matches the post-rebase scope.

crates/jolt-profiling/Cargo.toml
```diff
@@ -27,6 +27,11 @@ allocative = { workspace = true, optional = true }
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 memory-stats.workspace = true
+libc.workspace = true
+schemars.workspace = true
```
0xAndoroid: nit (build hygiene): `schemars` (plus its derive stack) becomes an unconditional non-wasm dep here, and via the new `jolt-prover-legacy` `host → dep:jolt-profiling` edge it now compiles in every `--features host` build — the default dev loop (`clippy --features host`, the byte-diff tests) — even though the legacy bin only uses `stage_memory`/`memory`/`report_*`, never the `summary` module. Gating `summary` (schemars + serde_json/std) behind a jolt-profiling feature that only jolt-prover's `profiling` enables would keep the standard host builds' dep graph where it was.

jolt-eval/src/objective/callgrind.rs
0xAndoroid: Fixed in 9adc52d71: the parser now mirrors the runner's `BenchmarkSummary` spine and drills the explicit `profiles[tool=="Callgrind"].summaries.total.summary.Callgrind.Ir.metrics` path, taking the new-run value from `EitherOrBoth` (`Left`/`Both[0]`) and gating on summary schema version "6". The fixture is validated by deserializing through the actual `iai-callgrind-runner` 0.16.1 summary structs (new dev-dependency), so it can't drift from the real schema again.

crates/jolt-kernels/src/reference/commitment.rs
0xAndoroid: Fixed in 4eb34be38: both spans moved to the stage-0 call boundary (`stages/stage0.rs`), so every `CommitWitness` backend inherits `commit_witness`/`commit_advice` with the same fields; the ReferenceBackend impl carries a pointer comment instead of the attributes. Verified by the smoke test (label presence) and a byte-diff muldiv run.

crates/jolt-profiling/src/taxonomy.rs
0xAndoroid: Fixed in 4eb34be38: `prove_uniskip_committed` and `HomomorphicBatch::prove_batch_zk` are now instrumented; the taxonomy splits the pairs into `CLEAR_MODE_SPANS`/`ZK_MODE_SPANS` and `always_present_spans(ProverMode)` returns the mode-correct set. The smoke test selects the mode from `cfg(feature = "zk")` and passes in both modes (run explicitly: clear 65s, zk 53s at 2^13).

crates/jolt-profiling/src/summary.rs
0xAndoroid: Fixed in b473a00a4: the harness samples `peak_rss_bytes()` immediately after `run_workload` (before the guards even drop) and passes it into `finalize_trace(trace_path, ctx, peak_rss_bytes)`; the flush-time parse/expand/rewrite can no longer become the reported high-water mark. Field doc + schema regenerated to state the sampling point.

jolt-eval/bin/optimize.rs
0xAndoroid: Fixed in e49801953: the workload is inserted into `profiled_workloads` only after `run_profile_in` succeeds, and `run_profile_in` now deletes any pre-existing summary at the deterministic path before spawning — a failed run leaves nothing stale to read (absent summary = measurement error = INFINITY score).

crates/jolt-prover/src/driver.rs
0xAndoroid: Fixed in 4eb34be38: `SpannedRounds` now carries separate factories and `finish_rounds` runs under its own `<Relation>::finish_rounds` label (documented in the taxonomy naming convention, excluded from the presence set like the other per-member labels). `telemetry:<workload>:total:<Relation>::prove_round` is pure round slices again.

crates/jolt-profiling/src/taxonomy.rs
0xAndoroid: Fixed in c89e4fc00 (CLAUDE.md + spec §Taxonomy conformance) and 4eb34be38 (taxonomy.rs §Evolving): all three now say the rename fails the profiling smoke test, which is deliberately not CI-wired yet and must be run explicitly, with a pointer to the rust.yml NOTE.

crates/jolt-profiling/src/summary.rs
0xAndoroid: Fixed in b473a00a4: both the trace rewrite and the summary write go through `write_atomic` (`{path}.tmp` + `fs::rename`). A pipeline test asserts no `.tmp` residue after `finalize_trace`.

crates/jolt-profiling/src/summary.rs
0xAndoroid: Fixed in b473a00a4: stage intervals pair with rows by per-label occurrence index (both recorded in close order; `ordered` preserves within-label order through its stable filters). New fixture test `repeated_stage_labels_pair_rows_by_occurrence` covers two same-label instances getting their own rows.

jolt-eval/src/objective/callgrind.rs
0xAndoroid: Fixed in 9adc52d71 together with the codex finding above: the DFS is gone in favor of the explicit `total.summary.Callgrind.Ir` pointer path on typed mirrors of the runner's structs (reading only `total`, never `parts`, so multi-part runs can't double-count), with a loud error on unsupported summary versions. Fixture now schema-locked against the real `iai-callgrind-runner` types.

crates/jolt-prover/src/profile.rs
0xAndoroid: Fixed in b473a00a4: `BackendKind::as_str()` added and used for `run.backend`, so a second variant forces the metadata arm at compile time (didn't wait for #1714).

crates/jolt-prover/src/profile.rs
0xAndoroid: The premise doesn't hold at current main: legacy's `prove_example_with_trace` returns `(prove_duration, proof_size, proof_size, trace.len())` — `crates/jolt-prover-legacy/benches/e2e_profiling.rs:341` — so legacy's column 7 is also the raw bincode size; only the variable *name* `proof_size_comp` at the call site (line 152) survives from the retired compressed encoding. The modular line therefore matches legacy exactly and the columns are directly comparable. b473a00a4 documents this deliberately at the write site so the vestigial column name doesn't mislead the next reader.

crates/jolt-profiling/Cargo.toml
0xAndoroid: Fixed in 1c9d75767: the summary module (schemars/serde_json/thiserror stack) sits behind a `summary` feature — default-on so jolt-profiling's own dev loop and schema drift-lock tests keep running, but opted out (`default-features = false`) by the legacy host edge, akita's bench dev-dep, and jolt-prover's test dev-dep. `jolt-prover`'s `profiling` feature enables `jolt-profiling/summary` explicitly; standard `--features host` dep graphs are back to pre-PR.

moodlezoup (approved):

PR #1713, opened 2026-07-29
docs(specs): x86 tracer backend — AOT-transpiled trace generation

moodlezoup:
## Summary

Spec for a new, optimized tracer backend: trace generation is the inherently sequential first pass of proving — already a latency bottleneck for multi-GPU setups, and a throughput bottleneck once streaming sumcheck regenerates the trace multiple times per proof. The field has converged on transpilation (ZisK: 1.5 GHz via AOT RISC-V→x86; SP1: `sp1-jit`), while our interpreter re-decodes every instruction on every tick and no benchmark in the workspace measures tracing speed at all.

The spec (`specs/x86-tracer-backend.md`) commits to:

1. **Formalize the chunked execution seam** — a new `ChunkedExecutionBackend` trait in `jolt-program` (fast checkpointing pass + parallel `replay_chunk`), extending the existing `ExecutionBackend`/`TraceSource` boundary that `jolt-witness` was already designed around.
2. **`crates/jolt-tracer-x86`** — whole-program AOT transpilation of the *expanded bytecode* to x86-64 via dynasm-rs (linux-x86_64, `NativeBackend` fallback elsewhere). Expansions are static per PC, so trace rows compile to templates with only values patched at runtime.
3. **Infrastructure before optimization** — per-instruction differential unit tests and iai-callgrind microbenchmarks (macro-generated, exhaustive by construction), plus whole-guest throughput/peak-RSS benchmarks that record the reference baseline first.
4. **Reference interpreter untouched** — default backend and correctness oracle, with two new jolt-eval invariants (`tracer_backend_equivalence` incl. RedTeam target, `tracer_chunk_composition`) enforcing bit-identical rows and byte-identical proofs.

Gates: ≥10× eager tracing and ≥25× fast pass vs. reference (pinned procedure), peak RSS ≤ baseline, ACT4 green under the x86 backend, proof byte-equality 10/10.

## Notes for reviewers

- Judgment calls worth a second opinion: the 10×/25× gate multipliers, linux-x86_64-only native scope, and the access-value-log checkpoint format (first-touch deltas kept as the reference backend's impl; flip condition recorded in Alternatives).
- The spec adopts `InlineAdviceContext` from `specs/inline-expansion-grammar.md` rather than forking that contract.
- Drafted from a research pass over the tracer/consumer seams + external prior art, then hardened by an adversarial review pass (43 findings folded in); scores 0.157 on the `/analyze-spec` ambiguity rubric (threshold ≤ 0.2).


moodlezoup:
Updated after reviewing OpenVM 2.1's "portable native recompilation" announcement and surveying emitter options: added **Alternative 11 — copy-and-patch stencils** (CPython `Tools/jit` model: pinned-clang stencils extracted at build time into committed `(bytes, holes)` tables — compiler-quality code with no toolchain on the proving machine), with recorded flip conditions (decisive one: the aarch64 backend becoming a goal, since C stencils recompile per-target vs. re-hand-writing every template). Also added the internal `RowEmitter` seam so emitters are swappable per row kind, and a time-boxed emitter spike in slice 2 (~5 representative kinds both ways, iai-compared; informative, not a gate). dynasm-rs stays the primary emitter; no gates or invariants changed.

moodlezoup:
Added the **fused fast-mode groups** design (from discussion of shuklaayu.sh/blog/riscv-recompiler): in fast mode a source-instruction group may compile to its architectural semantics plus an observability epilogue — e.g. DIV becomes a native `idiv` (which also yields both advice values) instead of a translation of its virtual sequence, mirroring the interpreter's own `execute`-vs-`trace` split. The paragraph names the non-obvious obligation that makes this sound: virtual temp registers 40–47 leak into the trace via rd *pre*-values (→ `RdInc`), and fast-pass checkpoints seed bit-exact replay, so fused templates must store each recipe's final temp values (persistent regs 32–39 always maintained). Record mode is never fused. Phased into slice 5 (memory-op and M-extension groups first), each fusion gated on green `tracer_chunk_composition` + `tracer_backend_equivalence`; Alternative 11 now notes that under the stencil emitter the same design is expressed as group-granularity stencils, where the fused `idiv` falls out of clang automatically.

moodlezoup:
Included in https://github.com/a16z/jolt/pull/1727

PR #1714, opened 2026-07-29
perf(kernels): optimized kernel backend — legacy-parity prover performance from 2^16 to 2^27

0xAndoroid:
## Summary

Optimized kernel backend for the modular prover: `JoltBackend::optimized()` serves every slot with a legacy-technique port, bringing `jolt-prover` + `jolt-kernels` to performance parity with `jolt-prover-legacy` on 10-core hardware at 2^23 and on 64-thread hardware at 2^25–2^27 (see "Many-core scaling" below) — while keeping proofs **byte-identical** to legacy (pinned end-to-end by the byte_diff suite, now 19 tests including a 2^17 streaming-chunk-boundary gate).

Stacked on #1712 (profiling coverage — the measurement foundation for this work).

## Results (sha2-chain, Mac mini M4 10-core / 16 GB)

| scale | modular `optimized()` | legacy | notes |
|---|---|---|---|
| 2^16 | 1.35 s / 254 MiB | 1.39 s / 200 MiB | reference tier was 1254 s / 11.7 GiB |
| 2^20 | 7.20 s / 1.55 GiB | 7.34 s / 1.73 GiB | reference tier could not run (>20 GiB) |
| 2^23 | **42.21 s / 8.84 GiB** | **42.57 s / 9.07 GiB** | back-to-back pair, same thermal conditions¹ |

¹ Cool-machine runs land ~34-35 s for both provers; paired same-session runs are the honest comparison (M4 thermal drift is ±20%). At every scale the optimized backend is at or below legacy on both wall-clock and peak RSS.

## Many-core scaling (sha2-chain, Threadripper PRO 7975WX 32c/64t, 768 GB)

The kernels above were tuned on 10 cores. On a 64-thread box the backend initially fell to 3.02×/3.73× behind legacy at 2^25/2^27 — the last 15 commits close that. ABBA-interleaved 3+3 runs, medians, idle box, both arms from the same tree:

| cell | before (@ bf802c13) | after (@ 70a294ad) | legacy | ratio before → after |
|---|---|---|---|---|
| 2^25 wall | 126.40 s | **39.10 s** | 42.25 s | 3.02× → **0.925×** |
| 2^25 peak RSS | 21.41 GiB | 19.25 GiB | 18.65 GiB | 1.15× → 1.032× |
| 2^27 wall | 465.30 s | **128.55 s** | 125.98 s | 3.73× → **1.020×** |
| 2^27 peak RSS | 77.68 GiB | 66.32 GiB | 63.93 GiB | 1.21× → 1.037× |

Small scales on the same box: 2^16 1.01 s vs 1.20 s, 2^20 3.23 s vs 3.41 s — modular faster. (Legacy column = after-session medians; before-session legacy medians, 41.80 s and 124.91 s, were stable across sessions.)

Root causes fixed:

- **Serial witness walks.** 13 `stream_witnesses` walks accounted for 91.3 s of the 125 s prove wall at 2^25, running at only 6–19 busy cores of 64 (fixed-size chunk loops: parallel-extract → serial-consume, ~8k fork-joins per walk). Fix: index-parallel whole-range collection — `RowSource::random_access()` + in-place parallel scatter through `collect_bundles` / kernel `collect_rows` / `walk_cycles` / instruction packing / opening fill. This alone took 2^25 from 126.4 s to 65.6 s.
- **10-core-tuned granularity.** Streaming-commit superchunk 2^17 → 2^21 (256 serial window joins were stage 0's critical path at 64 threads) with closure-fed commit windows; index-parallel registers sparse-entry build; block-batched sparse-entry binding.
- **Peak-RSS transients at 60+ GiB scale.** `Vec<SpartanOuterRow>` (176 B × T = 6.2 GiB at 2^25) is no longer retained across collect→rounds→opening — Spartan re-extracts windows on the fly from slice-backed sources; instruction cycle tables staged at half domain; lazy-RA dense switch deepened T/8 → T/16; the stage-5 handoff frees packed rows before `combined_val`.

Boundary discipline unchanged: the new seams are additive with sequential fallbacks (jolt-witness `RowSource::random_access`/`owned_rows` + parallel collectors, jolt-program `TraceSource::shared_rows`, jolt-openings closure-feed variants, jolt-dory parallel overrides); everything else lives in `jolt-kernels`. Zero jolt-prover library / legacy changes. byte_diff 19/19 byte-identical at every commit of the wave.

**The claim, precisely:** legacy-parity or better on both hardware classes — 10-core: at-or-faster with at-or-lower RSS at 2^16/2^20/2^23 (table above); 64-thread: faster at 2^16/2^20/2^25 (0.925×), within 2% at 2^27 (1.020×), peak RSS within 3.2%/3.7% at 2^25/2^27 (from 15%/21% over). Residual owners are mapped for follow-up: RamRWC compact pre-bind entries (est −1.6/−6.4 GiB), stage-6b lookup-index compaction, 6b round tails + stage-8 `create_eval` (~2.6 s at 2^27).

## What's in it

`crates/jolt-kernels/src/optimized/` — one module per kernel family, each documenting its port. The techniques are the legacy prover's, restated against the kernel seam: sparse one-hot access off typed witness bundles (no K×T grids), split-eq (Gruen) round messages, prefix–suffix decompositions, lazy staged RA materialization (dense only at T/8), pre-scaled shared eq tables, deferred-reduction accumulation, compact/native scalar columns promoted on first bind, u16 coefficient LUTs for early rw-matrix rounds, parallel streaming commit (column×window grid), and a lazy joint-opening source (64 B/cycle columns instead of ~21 dense 2^total_vars embeddings).

Every kernel carries in-module lockstep parity tests vs the reference tier (byte-equal round polynomials every round + equal output claims), and the byte_diff suite pins full proofs against legacy for both backends across the advice / committed-program / trace-order matrix.

## Changes outside jolt-kernels (all justified, none touch protocol logic)

- **jolt-prover**: dev-surface only — `examples/modular_benchmark.rs` grew `--backend {reference,optimized}` and exact-capacity trace padding; byte_diff tests extended to run both backends (+ the 2^17 chunk-boundary variant). **Zero library-code changes.**
- **jolt-verifier / jolt-sumcheck / jolt-claims**: untouched.
- **jolt-witness**: default-off `parallel` feature (slice fast paths for the cycle walks + parallel bundle extraction; sequential fallback remains the contract, pinned by tests); calloc zero-tables; `StreamConsumer::Witness: Send`; `TraceSource::rows()` pristine-cursor guard + documented contract. Many-core wave: `RowSource::random_access()`/`owned_rows()` + index-parallel whole-range collectors (`collect_bundles_par`, `collect_par_map`, `collect_range_into`, `par_collect_windows`) — additive, sequential-equivalence pinned by a parity test.
- **jolt-openings**: additive `StreamingCommitment` batch methods (`feed_i128_rows`, `process_one_hot_chunks`, and the many-core closure-fed variants `feed_i128_rows_with`/`process_one_hot_chunks_with`) with serial-equivalent defaults — cross-window parallelism was inexpressible through the `&mut` single-item signatures.
- **jolt-dory**: `routines.rs` plugs the same jolt-optimizations GLV/rayon routines legacy uses into the external dory-pcs prover/verifier; parallel overrides for the new batch methods (incl. the closure-fed variants: same window grid reading off the shared bundle chunk, per-window transient staging only). (Known pre-existing footgun documented in `setup_prover`: dory's global prepared-pairing cache prefix-matches across setup sizes — deliberately NOT primed here.)
- **jolt-field**: `WithAccumulator::Accumulator` bound widened to `RingAccumulator<Element = Self>` (makes `fmadd(F,F)` reachable from `Field`); `WideAccumulator::add` rewritten from `fmadd(v, one())` (a full 4×4 mul) to a shifted-limb add (`â·2^256 ≡ â·R mod p` — exactness argued in-code and property-tested).
- **jolt-program**: one-line `rows()` pristine guard. Many-core wave: `TraceSource::shared_rows()` (default `None`, pristine-guarded) + `OwnedTrace` impl, letting Spartan re-extract rows instead of retaining 176 B × T.
- **CI**: `jolt-witness --features test-utils,parallel` test invocation.

## Correctness process

Four independent review passes over the campaign's diffs (RAM family, parallelism pass, waves 2–4) found 0 blockers / 0 majors; all minors are folded in (defense-in-depth debug_asserts, edge-case test pins, the chunk-boundary gate). The `WideAccumulator::add` rewrite was independently re-derived (R = 2^256 for BN254 `MontConfig<4>`; Barrett fold precondition `V' < p·R` holds; headroom ≥ 2^38 ops at 2^23 scale).

## Known remaining gaps (documented, not blocking)

- RegistersReadWriteChecking rounds retain a ~0.4 s residual vs legacy at 2^23 (structure matches legacy; needs a profiler session).
- Stage-1 bundle extraction re-derives `lookup_query` per operand field (~1 s CPU at 2^23); sharing needs a jolt-witness visibility change.
- `SumcheckKernel::output_claims` has no witness access, so the Spartan kernels retain typed rows (~2.2 GiB at 2^23) for the post-hoc opening walk; a seam extension would reclaim it. Peak RSS already beats legacy, so not taken here.

## Test plan

- `cargo nextest run -p jolt-prover --features prover-fixtures` — byte_diff 19/19 (both backends, byte-identical to legacy, incl. 2^17 chunk-boundary gate)
- `cargo nextest run -p jolt-witness -p jolt-kernels -p jolt-program -p jolt-dory -p jolt-openings` — 310 tests; +33 under `jolt-witness --features test-utils,parallel`; +164 `jolt-field`
- clippy `--all --all-targets -D warnings` under `--features host` and `--features host,zk`
- `cargo run --release -p jolt-prover --example modular_benchmark --features prover-fixtures -- --name sha2-chain --scale 23 --backend optimized`
- many-core ladders: `--scale 25/27 --backend optimized` vs legacy, ABBA 3+3, on the 7975WX (table above)


crates/jolt-kernels/src/optimized/ram_ra_virtualization.rs
```diff
@@ -0,0 +1,455 @@
...
+                )
+            },
+            |(acc, evals, steps), row, _x_in, e_in| {
+                for position in 0..num_committed {
+                    let (lo, hi) = self.folded_ra.lo_hi(position, row);
+                    evals[position] = lo;
+                    steps[position] = hi - lo;
+                }
+                for value in acc.iter_mut() {
+                    let mut product = evals[0];
```
0xAndoroid, line 190: `num_committed == 0` is a supported geometry (`ram_k == 1`), but this indexes an empty `evals` vector on the first round. The reference relation treats the empty committed-RA product as `1`; seed `product` with `F::one()` and multiply all entries (or handle the zero-factor case explicitly).

crates/jolt-witness/src/alloc.rs
```diff
@@ -0,0 +1,64 @@
...
+                bytes.iter().all(|&byte| byte == 0),
+                "F::zero() is not all-zero bytes — zero_table is invalid for this field"
+            );
+        }
+    }
+
+    // SAFETY: `alloc_zeroed` returns a valid zero-initialized allocation and
+    // all-zero bytes are a valid `F` (the zero element), per the contract
+    // above.
+    unsafe {
```
0xAndoroid: This safe generic function has an unenforced unsafe precondition: `Field` is public/unsealed and does not guarantee that all-zero bytes are a valid `F` or encode `F::zero()`. In release, a conforming custom field can therefore make `Vec::from_raw_parts` expose invalid values (the debug byte scan is not a contract, and can read padding). Encode this in an unsafe `Zeroable`-style bound/marker or initialize generically.

crates/jolt-kernels/src/optimized/registers_read_write.rs
```diff
@@ -0,0 +1,1844 @@
...
+    })
+}
+
+/// The entry count [`cycle_entries`] will produce for one cycle — the
+/// counting pass's cheap twin (kept adjacent so the merge rules stay in
+/// sync: rs2 folds into rs1's entry, rd into either read's).
+#[cfg(feature = "parallel")]
+fn cycle_entry_count(cycle: &RegisterCycleRow) -> usize {
+    let mut count = 0usize;
+    let mut cols = [u8::MAX; 2];
```
0xAndoroid: `u8::MAX` collides with a valid raw register index. For `rs1=None, rs2=Some(255)`, this pass counts zero entries, while `cycle_entries` emits one; pass 2 then writes `window[0]` into a zero-length allocation and panics. The reference witness path rejects every index outside the 128-entry domain, so validate indices before counting rather than using a value sentinel.

crates/jolt-kernels/src/optimized/bytecode_read_raf.rs
```diff
@@ -0,0 +1,897 @@
...
+            for &k in &touched {
+                for stage_inner in &mut inner {
+                    stage_inner[k] = F::zero();
+                }
+            }
+            touched.clear();
+            let base = j_hi << lo_bits;
+            for j_lo in 0..in_len {
+                let pc = rows[base + j_lo].push_pc as usize;
+                if inner[0][pc].is_zero() {
```
0xAndoroid: `inner[0][pc] == 0` is not a membership test. With a valid Boolean low point, early eq weights can be zero; repeated visits to the same PC push it into `touched` multiple times, then the final loop adds the same nonzero accumulator once per duplicate. A separate boolean/epoch marker is required; field cancellation creates the same bug.

.github/workflows/rust.yml
```diff
@@ -426,6 +426,12 @@ jobs:
       - name: Test lookup table prover ABI
         if: matrix.shard == 1
         run: cargo nextest run -p jolt-lookup-tables --features prover-abi-tests --test prover_lookup_table_abi --cargo-quiet
+      # jolt-witness's parallel feature is not a default, so the shard loop
+      # above tests only the serial paths; the shipped provers enable it
+      # (via jolt-kernels' default), so test the rayon fast paths directly.
+      - name: Test jolt-witness with parallel
```
0xAndoroid, line 457: This new parallel lane still leaves the PR’s byte-parity ratchet disabled in CI: every real `byte_diff` test is behind `jolt-prover/prover-fixtures`, while the default crate shard only discovers the ignored stub. Add an explicit `cargo nextest run -p jolt-prover --features prover-fixtures --test byte_diff --cargo-quiet` lane (or a bounded parity subset), otherwise optimized proof/transcript divergence can merge green.

crates/jolt-prover/tests/byte_diff.rs
```diff
@@ -1211,6 +1252,7 @@ mod committed_muldiv {
         let pcs_setup = DoryScheme::setup_prover(support::setup_total_vars(
             memory_layout,
             &[bytecode_candidate, image_candidate],
+            support::MAX_PADDED_TRACE_LENGTH,
```
0xAndoroid, line 1307: This committed-program arm still proves only with `JoltBackend::reference()` below, so it never exercises the optimized bytecode/program-image reduction slots added by this PR. The standalone address-major and committed+advice arms have the same gap. Loop these arms over reference and optimized before using them as whole-proof byte-parity gates.

crates/jolt-kernels/src/optimized/commitment.rs
```diff
@@ -0,0 +1,464 @@
...
+use crate::reference::commitment::{column_kinds, ColumnKind};
+use crate::{KernelError, OptimizedBackend, ProofSession, ReferenceBackend};
+
+/// Cycles per superchunk. The dominant stage-0 wall cost on a many-core
+/// host is the per-superchunk join (its critical path is one window's
+/// serial MSM), so fewer, larger superchunks win: 2^17 → 2^21 measured
+/// 59.3s → 41.1s whole-prove at 2^25 on 64 threads. The extracted bundle
+/// buffer (64 B/cycle, two reused buffers) is the only staging that scales
+/// with the superchunk — columns feed the commit windows by closure.
+const SUPERCHUNK_CYCLES: usize = 1 << 21;
```
0xAndoroid: `2^21` is the measured 64-thread optimum, but it becomes the production setting for every topology. On slice-backed traces the double buffer reserves `2 × 2^21 × 64 B ≈ 256 MiB` once the trace is large enough, including 10-core hosts that need far fewer concurrent Dory windows. Derive/cap the window count from `rayon::current_num_threads()` and a memory budget instead of hard-coding the 64-thread point.

crates/jolt-prover/examples/modular_benchmark.rs
```diff
@@ -284,10 +310,14 @@ fn run_benchmark(bench: BenchName, scale: usize, target_trace_size: Option<usize
     )
     .expect("modular proof verifies");
 
+    let backend_label = match backend_choice {
```
0xAndoroid: `backend_label` reaches stdout, but not the result identity below: reference and optimized runs at the same benchmark/scale overwrite the same individual CSV and append indistinguishable rows to `modular_timings.csv`. Include the backend in both the filename and CSV schema so the parity/performance evidence cannot be misattributed.

0xAndoroid (commented):
AI review (Codex): comment-only review of `46befe327` against PR #1712’s base `e3e8e616`; eight concrete findings posted inline.

Highest-impact issues: the safe generic zero-byte allocator can construct invalid `Field` values in release, and the optimized RAM-RA kernel panics for the supported zero-committed-polynomial geometry. The remaining findings cover algebraic byte-parity at zero/canceling challenges, malformed-register error handling, missing byte-diff gates/CI coverage, 64-thread tuning applied globally, and benchmark artifact identity.

No supported-range 2^27 offset truncation, Dory ordering defect, or parallel witness-order nondeterminism found in the audited paths. Per repository review policy, I did not run cargo tests.

.github/workflows/rust.yml
0xAndoroid, line 457: This covers jolt-witness's parallel arms, but nothing in CI runs byte_diff itself: the shard loop only ever adds `--features test-utils`, and `prover-fixtures` appears in CI solely for jolt-verifier — so `-p jolt-prover` compiles byte_diff down to its `#[ignore]` stub. Every arm of the suite the PR's parity claim rests on (including both optimized-backend arms and the 2^17 gate added here) only runs manually. Consider a dedicated step (`cargo nextest run -p jolt-prover --features prover-fixtures`), even if nightly or label-gated given the legacy-prover build cost.

crates/jolt-prover/tests/byte_diff.rs
```diff
@@ -1195,6 +1235,7 @@ mod committed_muldiv {
             &verifier_preprocessing,
             order,
             &legacy_proof,
+            support::MAX_PADDED_TRACE_LENGTH,
```
0xAndoroid, line 1290: All six committed-program arms (committed_muldiv ×4, advice_committed ×2) construct `JoltBackend::reference()` only — but the optimized backend ships committed-only slots (`bytecode_reduction_{cycle,address}`, `program_image_reduction_{cycle,address}`, and the joint-opening's block-embedded committed views), so those kernels never get the legacy byte anchor; their only pin is module tests vs the reference tier. Looping these arms over both backends the way muldiv and advice_consumer already do would close the gap.

crates/jolt-prover/tests/byte_diff.rs
```diff
@@ -248,9 +251,13 @@ mod support {
...
-    pub fn setup_total_vars(memory_layout: &MemoryLayout, extra_candidates: &[usize]) -> usize {
-        let max_log_t = MAX_PADDED_TRACE_LENGTH.ilog2() as usize;
-        let max_log_k_chunk = 4usize; // max_log_t = 16 < the 25-bit threshold
+    pub fn setup_total_vars(
+        memory_layout: &MemoryLayout,
+        extra_candidates: &[usize],
+        max_padded_trace_length: usize,
+    ) -> usize {
+        let max_log_t = max_padded_trace_length.ilog2() as usize;
+        let max_log_k_chunk = 4usize; // max_log_t < the 25-bit threshold
```
0xAndoroid: This hardcodes the small side of `ONEHOT_CHUNK_THRESHOLD_LOG_T`, and the module fixture does too (`optimized/testing.rs` pins `{log_k_chunk: 4, lookups_ra_virtual_log_k_chunk: 16}`) — so the `{8, 32}` geometry that the 2^25–2^27 results in the PR description actually run under has no byte-parity witness at any size. Both provers accept an injected config at small T (legacy validates `log_k_chunk ∈ {4, 8}`), so a small-trace arm forcing `{8, 32}` looks cheap and would cover the top half of the claimed range. Related: at log_T ≥ 24 legacy switches read-raf to 8 phases × 16 bits while the modular tiers always use 16 × 8-bit — value-neutral per the reference kernel's exactness note, but the 2^25+ parity claim leans on that argument untested.

crates/jolt-prover/tests/byte_diff.rs
```diff
@@ -1573,6 +1622,172 @@ mod advice_committed {
     }
 }
 
+/// The chunk-boundary gate: the optimized backend's shared trace walks
+/// stream in 2^16-row chunks (`jolt-kernels` `COLLECT_ROWS_CHUNK`), and every
+/// other byte-parity arm caps at a 2^16 padded trace — no arm crosses a
+/// chunk boundary. This one pads to 2^17 (asserted), so the collect passes
```
0xAndoroid: This rationale went stale during the many-core wave: slice-backed sources short-circuit every collect pass to index-parallel whole-range collection (`support::collect_rows` → `collect_bundles_par` whenever `random_access()` is `Some`, in serial builds too), and the commit takes the pipelined single-superchunk path at 2^17 — so no `COLLECT_ROWS_CHUNK` walk executes anywhere in this test. The arm still earns its keep (only e2e above 2^16, odd log_T grid split), but the chunk-boundary carry now has no e2e witness and the streamed fallback consumers are e2e-dead. Either re-aim this docstring or force the walk with a wrapper source that hides `random_access()`.

crates/jolt-kernels/src/optimized/lazy_ra.rs
```diff
@@ -0,0 +1,257 @@
...
+/// `N` address-folded selector columns bound `LowToHigh`, lazily for the
+/// first three binds.
+pub(crate) enum LazyFoldedRa<F: Field, S> {
+    /// Fewer than three binds: per-polynomial branch scale tables (the base
+    /// table pre-scaled by each bound-bit pattern's eq weight), flattened
+    /// offset-major — `tables[i][offset · stride_i + k]` with
+    /// `stride_i = tables[i].len() / width` — plus the compact index source.
+    Lazy {
+        tables: Vec<Vec<F>>,
+        /// Bound-bit branch count (`2^binds`: 1, 2, or 4).
```
0xAndoroid: Doc drift from the T/16 deepening: `width` now reaches 8, and the `Lazy` variant's doc above ('Fewer than three binds') contradicts `Dense`'s updated 'Four or more binds'; the module doc's 'after `b < 3` binds' (line 11) is likewise stale. Same drift at `booleanity.rs:34` ('index-encoded for the first three binds, dense at `T/8`') and the `instruction_ra_virtualization.rs` test comment ('two lazy binds, the dense materialization at the third'). On a threshold state machine these stated invariants are what the next reader trusts.

```suggestion
        /// Bound-bit branch count (`2^binds`: 1, 2, 4, or 8).
```

crates/jolt-kernels/src/optimized/booleanity.rs
```diff
@@ -0,0 +1,1316 @@
...
+
+    #[test]
+    fn cycle_kernel_matches_reference_single_round() {
+        cycle_parity(1, 4, false);
+    }
+
+    /// `log_t = 4` drives the shared-table state machine through
+    /// materialization (three staged binds) into dense rounds.
+    #[test]
+    fn cycle_kernel_matches_reference_through_dense_rounds() {
```
0xAndoroid, line 1146: With materialization moved to the fourth bind, log_t = 4 receives that bind only via `finish_rounds` — every round message this test checks is now computed from the `Lazy` arm, so the dense-rounds coverage the name and doc promise is gone (dense round messages are only reached indirectly through ra_virtualization's log_t = 6 parity test). log_t = 5 here restores the pin.

crates/jolt-kernels/src/optimized/commitment.rs
```diff
@@ -0,0 +1,464 @@
...
+    CommitWitness, CommitmentGrid, CommittedColumnsWitness, WitnessCommitment,
+};
+use crate::reference::commitment::{column_kinds, ColumnKind};
+use crate::{KernelError, OptimizedBackend, ProofSession, ReferenceBackend};
+
+/// Cycles per superchunk. The dominant stage-0 wall cost on a many-core
+/// host is the per-superchunk join (its critical path is one window's
+/// serial MSM), so fewer, larger superchunks win: 2^17 → 2^21 measured
+/// 59.3s → 41.1s whole-prove at 2^25 on 64 threads. The extracted bundle
+/// buffer (64 B/cycle, two reused buffers) is the only staging that scales
```
0xAndoroid: `CommittedColumnsWitness` is 80 B, not 64 B (three 16-byte scalars plus `Option<usize>`/`Option<u64>` at 16 B each — no niche), and `commit_pipelined` holds two of these buffers: ~320 MiB of staging at the 2^21 superchunk, 25% over the budget this comment uses to justify the constant. Still comfortably bounded on the 16 GB class, but this sizing math is what a future retune will read.

crates/jolt-witness/src/consumer.rs
```diff
@@ -109,6 +138,237 @@ pub trait RowSource {
...
+    unsafe { out.set_len(count) };
+    Ok(out)
+}
+
+/// Index-parallel bundle collection over a random-access view, mapped
+/// through `pack` element by element (so packed forms never stage the wide
+/// bundle): values are identical to the sequential pass's — extraction is
+/// pure per cycle window, and the lookahead row at the end of the range
+/// matches the chunk walk's `next_after`.
+pub fn collect_par_map<B: WitnessBundle, V: Send>(
```
0xAndoroid: On the error path the partially-written buffer is abandoned without drops — fine for today's plain-data instantiations, but this is `pub` and generic over any `V: Send`, so a Drop-owning `V` silently leaks its written prefix. A `V: Copy` bound (every current instantiation already satisfies it) would turn `par_collect_windows`'s comment-only invariant into a compiler-checked one.

crates/jolt-witness/src/backend/trace/cycle.rs
```diff
@@ -93,6 +116,30 @@ impl<T: TraceSource + Clone> TraceBackend<'_, T> {
...
+        })
+    }
+
+    fn owned_rows(&self) -> Option<crate::OwnedRows> {
+        let cycles = checked_pow2(self.config.log_t).ok()?;
+        self.trace.trace.shared_rows().map(|rows| {
+            crate::OwnedRows::new(
+                rows,
+                cycles,
+                std::sync::Arc::new(self.preprocessing.clone()),
```
0xAndoroid: This deep-clones the whole `JoltProgramPreprocessing` on every `owned_rows()` call, and each of the three kernel call sites (spartan_outer, optimized support, registers_val_evaluation) retains its own copy for its kernel's lifetime. Bounded by program size rather than T, but caching one `Arc<JoltProgramPreprocessing>` on `TraceBackend` would make it free.

crates/jolt-kernels/src/optimized/registers_read_write.rs
0xAndoroid: `u8::MAX` doubles as the empty sentinel here, while the write twin `cycle_entries` dedups by actual comparison — for a hypothetical column 255 the passes disagree (count 0, write 1) and the pass-2 scatter overruns its exact-sized window. Unreachable at today's 7-bit register space, so purely a landmine; `Option<u8>` (as in legacy's counting twin) or a `debug_assert!(col < 128)` would pin it.

crates/jolt-kernels/src/optimized/instruction_read_raf.rs
```diff
@@ -0,0 +1,1575 @@
...
+        session: &mut ProofSession,
+        witness: &dyn JoltWitnessPlane<F>,
+        inputs: ProverInputs<'_, F, InstructionReadRaf<F>>,
+    ) -> Result<Box<dyn SumcheckKernel<F, Relation = InstructionReadRaf<F>>>, KernelError<F>> {
+        let dimensions = inputs.relation.dimensions();
+        let rows: Arc<Vec<InstructionCycleRow>> = Arc::new(collect_instruction_cycle_rows(
+            witness,
+            1 << dimensions.log_t(),
+        )?);
+        session.park(SharedInstructionRows(Arc::clone(&rows)));
```
0xAndoroid: `prepare` parks the strong carry unconditionally, so on slice-backed witnesses the 48 B × T rows outlive the kernel's own drop-at-first-cycle-bind and stay resident through the stage-5 staging moment this PR works to flatten — which reads as defeating `SharedInstructionRowsWeak`'s 'never outlive their stage' contract (stage 6a would re-derive index-parallel for ~0.2 s). Deliberate tradeoff, or should this park the weak handle when `witness.random_access().is_some()`, like `shared_instruction_rows` does?

crates/jolt-kernels/src/optimized/ram_trace.rs
```diff
@@ -0,0 +1,161 @@
...
+        if session.state::<Arc<Self>>().is_none() {
+            let columns = Arc::new(Self::collect(witness, log_t)?);
+            session.park(columns);
+        }
+        let columns = Arc::clone(
+            session
+                .state::<Arc<Self>>()
+                .expect("RAM access columns parked above"),
+        );
+        debug_assert_eq!(
```
0xAndoroid: Five kernels across stages 2–6b reclaim these columns by type alone, and this domain-agreement check compiles out in release — while the `pc_rows` twin in the same PR hard-errors `TableSizeMismatch` for the identical pattern. A wrong-domain reclaim here means OOB panics or, in the parked-larger case, a silently wrong final RA claim from a prefix-covering table; worth matching the sibling. Latent-only today (all five call sites derive log_t from the same geometry).

0xAndoroid (commented):
**AI review (Claude)** — automated deep-review pass; comment-only, no approval signal.

Six parallel review passes covered the full diff: every `optimized/` kernel module (formula-level cross-checks against the reference tier and legacy), the jolt-witness / jolt-openings / jolt-dory / jolt-field / jolt-program seams, the many-core commits, and the parity-gate surface.

**Kernel rewrites audited clean — no blockers found.**
- *unsafe*: every `set_len`/scatter path proves disjoint full coverage; the `try_lock` error latch cannot lose all errors; zeroed alloc is valid for Montgomery zero.
- *Scale boundaries*: no narrow-type index paths at T = 2^27 (`PcRow`'s u32 packing is guarded fail-loud, u128 shifts bounded); deferred-reduction accumulators keep ≥ 2^34 headroom at 2^27.
- *Determinism*: no keyed-map iteration reaches serialized data; parallel assemblies are index-placed or order-preserving (the registers two-pass scatter reproduces sequential entry order exactly); reductions reassociate exact field ops only.
- *Seams*: Spartan row re-extraction ≡ retained rows (padding + lookahead seams included); the `WideAccumulator::add` shifted-limb rewrite is exact (R = 2^256 for `MontConfig<4>`, fold precondition holds, property tests hit the carry cases); dory routine plumbing matches legacy group-exactly; the `clamp(1, cycles/row_width)` panic shape is unreachable via the `row_width > cycles` reference-fallback guard.

**Findings concentrate on the byte-parity gates, not the kernels.** Four gate-coverage issues inline: byte_diff never runs in CI; committed-program mode never byte-diffs the optimized backend; the log_T ≥ 25 geometry behind the 2^25–2^27 half of the claim has no parity witness at any size; and the 2^17 chunk-boundary gate no longer exercises the walk it documents. All four look cheap to close. Plus eight smaller inline notes (doc drift from the T/16 deepening, a test that lost its dense-rounds pin, staging-size comment math, a leak-on-error contract on a pub API, per-call preprocessing clones, a u8-sentinel landmine, the stage-5 strong row carry, a debug_assert-vs-hard-error inconsistency).

crates/jolt-kernels/src/optimized/ram_ra_virtualization.rs
0xAndoroid: Fixed in f6ddfbd76: the empty committed product is handled explicitly (= 1, matching the reference fold over an empty chunk set), so the round loop no longer indexes an empty `evals`. One level deeper than the finding: the verifier relation itself cannot serve this geometry — `derive_output_term` recovers `r_cycle` for `EqCycle` from the first RA opening point, and there is none — so after byte-identical rounds and output claims, BOTH backends fail closed at derived-table validation. New test `zero_committed_chunks_prove_in_parity_and_fail_closed` pins both halves.

crates/jolt-witness/src/alloc.rs
0xAndoroid: Fixed in ec2124584, mirroring the v12-audit hardening of `jolt_poly::unsafe_allocate_zero_vec` (the same class): the all-zero-bytes assertion now runs in ALL builds (one `size_of::<F>()` byte-compare per call), so a conforming-but-nonzero-repr `Field` fails loudly in release instead of exposing invalid values; zero-size layouts take safe construction, and the padding limitation is documented on the contract.

crates/jolt-kernels/src/optimized/registers_read_write.rs
0xAndoroid: Fixed in f6ddfbd76: `RegisterCycleRow::from_row` now rejects every index outside the 2^7 domain with the trace oracle's `InvalidWitnessData` (the same error the reference path produces), the counting pass is Option-typed (no value sentinel — `rs2=Some(255)` counts exactly like the write twin), and both parallel passes latch the lowest-cycle failure so the reported error is deterministic. Tests: exhaustive count/write consistency incl. 255, plus a rejection test asserting the first invalid cycle wins.

crates/jolt-kernels/src/optimized/bytecode_read_raf.rs
0xAndoroid: Fixed in f6ddfbd76: membership is now an epoch-marked `seen` array (exact, value-independent, free per-block reset) instead of `inner[0][pc].is_zero()`. New test `stage_pushforwards_membership_is_value_independent` drives a Boolean low coordinate (zero stage-0 weights on half the low domain) against the naive pushforward.

.github/workflows/rust.yml
0xAndoroid: Fixed in 18965badd: dedicated `test-byte-parity` job running `cargo nextest run --cargo-profile ci -p jolt-prover --features prover-fixtures` (release-tests shared cache, same shape as the legacy-prover jobs) — the full suite, including both optimized-backend arms and the new geometry/chunk-walk arms.

crates/jolt-prover/tests/byte_diff.rs
0xAndoroid: Fixed in 18965badd: all six committed-program arms (committed_muldiv ×4, advice_committed ×2) and the standalone address-major arm now loop over reference AND optimized, so the committed-only kernels get the legacy byte anchor. Suite green at 20/20.

crates/jolt-kernels/src/optimized/commitment.rs
0xAndoroid: Fixed in f6ddfbd76: the superchunk now scales with the pool — 2^15 cycles/thread (the measured 64-thread optimum's per-thread allotment), clamped to [2^17, 2^21] — so a 10-core host reserves ~50 MiB of staging instead of ~320 MiB. Wire-neutral: per column the fed windows, their order, and the finish calls are identical at any superchunk size (pinned by the module test's whole-trace and single-window arms).

crates/jolt-prover/examples/modular_benchmark.rs
0xAndoroid: Fixed in 53c5069a7: the backend is now a trailing CSV column (legacy's 7 fields stay position-stable) and part of the individual file name (`modular_{bench}_{scale}_{backend}.csv`), matching the Perfetto trace naming.

.github/workflows/rust.yml
0xAndoroid: Fixed in 18965badd: dedicated `test-byte-parity` job (`cargo nextest run --cargo-profile ci -p jolt-prover --features prover-fixtures`, release-tests shared cache + dory cache). Kept always-on rather than nightly/label-gated: it parallels the existing legacy-prover jobs that already build the same dependency footprint per PR.

crates/jolt-prover/tests/byte_diff.rs
0xAndoroid: Fixed in 18965badd: committed_muldiv ×4, advice_committed ×2, and the standalone address-major arm all loop over reference AND optimized now (the advice commit stays single-shot — it is backend-independent and pinned against legacy inside). The committed-only slots get their legacy anchor; suite 20/20.

crates/jolt-prover/tests/byte_diff.rs
0xAndoroid: Fixed in 18965badd: new `wide_one_hot` arm forces {8, 32} on a muldiv trace — legacy through its public `one_hot_params` (validated `log_k_chunk ∈ {4, 8}`), the modular side through a `ProverConfig` override, under a 2^25 trace cap so both SRS sizings derive the 8-bit width exactly as a real 2^25 deployment (`setup_total_vars` now mirrors legacy's threshold instead of hardcoding 4). Both backends byte-identical + verifies. The read-raf phase note stands as stated: `get_instruction_sumcheck_phases` derives from the ACTUAL log_T inside the sumcheck, not from any injectable config, so the 8×16 split stays un-witnessed at small T — documented on the arm.

crates/jolt-prover/tests/byte_diff.rs
0xAndoroid: Fixed in 18965badd — took the wrapper-source option: new `support::HiddenRows` re-emulating view (rows()/shared_rows() stay None), and the 2^17 arm now proves twice — slice-backed (production shape: index-parallel collects, pipelined commit) and behind the hidden source, forcing the sequential COLLECT_ROWS_CHUNK walk so the boundary carry and the streamed fallback consumers actually produce the pinned bytes. Docstring re-aimed to match.

crates/jolt-kernels/src/optimized/lazy_ra.rs
0xAndoroid: Fixed in e933fdd94 — applied the suggestion and swept the rest of the drift: lazy_ra's module doc (b < 4), Lazy variant ('fewer than four binds'), width doc (1, 2, 4, or 8), booleanity's module + field docs (four binds / T/16), ram_val_check ('rounds 0–3', T/16), both RA-virtualization field docs, bytecode_read_raf's fold comment, and the instruction_ra_virtualization test comment (three lazy binds, materialization at the fourth, two dense binds).

crates/jolt-kernels/src/optimized/booleanity.rs
0xAndoroid: Fixed in e933fdd94: `cycle_kernel_matches_reference_through_dense_rounds` now runs log_t = 5 (dense round MESSAGES pinned again); the log_t = 4 shape is kept as an explicit `materializing_at_finish` edge test (the fourth bind arriving exactly at finish_rounds).

crates/jolt-kernels/src/optimized/commitment.rs
0xAndoroid: Fixed in f6ddfbd76: the comment now states the real 80 B/cycle (three 16-byte scalars + two niche-free 16-byte Options) and ~320 MiB at the 2^21 ceiling — and the retune concern is moot in the common case, since the superchunk now scales per rayon::current_num_threads() (2^15 cycles/thread, clamped [2^17, 2^21]) per the sibling review thread.

crates/jolt-witness/src/consumer.rs
0xAndoroid: Fixed in ec2124584: `V: Copy` (and `B: Copy`) bounds across every spare-capacity collector — par_collect_windows, collect_par_map, collect_bundles_par, collect_range_into, collect_bundles, and BundleSource::bundles — so the leak-free-on-error invariant is compiler-checked. Every existing instantiation already satisfied it. The try_lock latches also became a deterministic lowest-index FirstErrorLatch while touching this.

crates/jolt-witness/src/backend/trace/cycle.rs
0xAndoroid: Fixed in ec2124584: TraceBackend caches one `Arc<JoltProgramPreprocessing>` in a `OnceLock` — one deep clone on the first owned_rows() call per proof, Arc::clone for the other two kernel call sites.

crates/jolt-kernels/src/optimized/registers_read_write.rs
0xAndoroid: Fixed in f6ddfbd76: `cycle_entry_count` is now Option-typed, mirroring the write twin's emitted-column scan exactly (`rd == rs2` with a folded rs2 implies `rs2 == rs1`, which `rd == rs1` catches) — no sentinel can collide with a raw index. Exhaustive count/write consistency test over the operand patterns incl. 255; extraction additionally rejects out-of-domain indices per the codex thread.

crates/jolt-kernels/src/optimized/instruction_read_raf.rs
0xAndoroid: Not deliberate — fixed in f6ddfbd76: prepare now parks the weak handle when `witness.random_access().is_some()`, exactly like shared_instruction_rows, so the rows free at the kernel's drop-at-first-cycle-bind and 6a re-derives index-parallel (~0.2 s) instead of 48 B × T riding through the stage-5 staging peak. Re-emulating sources keep the strong carry (re-deriving there means a full re-emulation walk).

crates/jolt-kernels/src/optimized/ram_trace.rs
0xAndoroid: Fixed in f6ddfbd76: hard `KernelError::TableSizeMismatch` matching the pc_rows twin — a wrong-domain reclaim now fails loudly in release instead of compiling out.

moodlezoup (approved):

PR #1715, opened 2026-07-29
perf(akita): optimize packed proving with implicit-zero commitments

quangvdao:
## Summary

This PR co-designs Jolt's Akita protocol representation, commitment kernels, prover pipeline, and backend memory lifecycle around the production `2^26` K256 geometry.

It:

- replaces the native batch of trace-derived one-hot columns with one prefix-packed physical polynomial;
- omits the public zero lane from the commitment while preserving the existing semantic one-hot polynomial through an explicit recentering identity;
- changes fused increments to balanced radix-`K` digits plus a signed carry;
- adds Jolt-owned streamed K16/K256 commitment and opening kernels that consume trace rows directly;
- regenerates rank-aware schedules, retains only the setup state needed by later openings, and adopts the corresponding Akita backend revision;
- removes repeated Stage 6 RA derivation, repeated Spartan trace decoding, and duplicate benchmark tracing.

At `T = 2^26`, the accepted final runs reduced total proving from **103.30 s** to **74.13–76.29 s** (75.21 s mean, **−27.2%**), while preserving a single commitment, a single opening reduction, and the existing sumcheck structure.

## Motivation

The previous Akita path paid for every semantic one-hot row, including the overwhelmingly common public lane-zero value. It also represented trace columns independently inside the backend, repeatedly decoded the same trace information in Stage 6 and Spartan, retained large setup and block-cache allocations across unrelated stages, and selected a byte-minimal schedule whose root SIS rank was not optimal for prover work.

These costs compound at the `2^26` K256 geometry:

- every lane-zero row still generated a root contribution;
- the one-hot root rank multiplied the entire commitment sweep;
- materialized hot-index columns and block tables increased memory traffic and RSS;
- setup field and NTT forms remained resident beyond their last full-width use;
- Stage 6b rebuilt instruction, bytecode, and RAM indices already computed in Stage 6a;
- Spartan outer uni-skip decoded each trace step for both constraint groups.

The resulting changes are therefore intentionally a mixture of protocol changes and implementation optimizations rather than a local kernel tweak.

### One packed trace polynomial

The canonical `OneHotTrace` object is now one physical polynomial with fixed selector capacity:

- K16: 64 semantic selector slots;
- K256: 32 semantic selector slots.

Instruction, balanced-increment, bytecode, and RAM columns occupy protocol-fixed ranges; unused selector slots are zero. The verifier samples a selector point and reduces the bound semantic evaluations to the single packed evaluation before the opening proof.

This changes generated schedules from many native one-hot polynomials to one higher-arity packed polynomial. Most of the apparent line deletion in this PR is regenerated catalog data disappearing as a consequence of that shape change.

### Implicit public zero

Let `P(a,t)` be the existing semantic one-hot polynomial, `A(t)` its activation, and `S(a,t)` the committed sparse polynomial. The commitment fixes lane zero to zero:

```text
S(a,t) = P(a,t)       for a != 0
S(0,t) = 0
H(t)   = sum_a S(a,t)
P(a,t) = S(a,t) + eq(a,0) * (A(t) - H(t))
```

Equivalently, at an arbitrary address point `r`:

```text
P(r,t) = eq(r,0) * A(t)
       + sum_a S(a,t) * (eq(r,a) - eq(r,0)).
```

Stage 6 continues to prove Booleanity and virtualization for the unchanged semantic `P`. Stage 7 recenters each Hamming-weight reduction into a public default-lane baseline plus an opening of `S`; RAM uses its existing access indicator for `A(t)`. Stage 8 consumes those sparse openings through the existing single-polynomial selector reduction.

This adds no second commitment, auxiliary row-weight polynomial, Fiat–Shamir challenge, proof field, or sumcheck round.

### Balanced fused increments

Fused signed deltas now use centered radix-`K` digits and a signed carry:

```text
delta = sum_i d_i * K^i + carry * 2^64
d_i in [-K/2, K/2 - 1]
carry in {-1, 0, 1}
```

Signed values are encoded modulo `K`; lane zero remains the public implicit default. The Stage 7 decode uses the multilinear extension of the centered lane value. The compatibility names `UnsignedIncChunk` and `UnsignedIncMsb` remain, but their protocol meaning is now balanced digit and signed carry.

### Binding and compatibility

The packed geometry, ordered semantic columns, capacities, and balanced/implicit-zero semantics are bound by layout digest v5. Selector-point and semantic evaluations are transcript-bound before the selector challenge is sampled.

This representation is Akita-only. Standard Dory and BlindFold/ZK behavior is unchanged. Existing Akita proofs and setup metadata from the previous layout are intentionally not compatible with layout v5.

## Commitment and opening kernels

A new Jolt-owned `trace_onehot` implementation consumes one row-major `TraceOneHotRows` source directly instead of constructing per-column hot-index objects and generic Akita block tables.

The implementation includes:

- a streamed single-polynomial commitment source;
- parallel K16 and K256 paths;
- direct packed-K16 lane consumption;
- fused ring work and fixed-size wide shifts;
- repeated-shift tuple reuse;
- commit/opening tiling around production geometry;
- owned spans for preparation, root accumulation, merge, and opening phases;
- production-shaped benchmarks and parity tests against the generic backend path.

The streamed source remains compatible with the releasable setup lifecycle by taking an owned covering snapshot for its full-width commitment read.

## Schedule and setup lifecycle

- Jolt's one-hot presets use `MinRootRankThenPayloadWithinSlack` with 1% payload slack. At the production packed K256 root, this selects `n_a = 6` instead of `n_a = 7` for 112 additional proof bytes (+0.113%).
- K16 and K256 catalogs are regenerated for singleton packed-polynomial shapes, with DP fallback under the same policy used for setup sizing.
- The field-form setup matrix is retained through the Stage 8 opening when that avoids re-derivation, while large NTT slots and block caches are dropped or rebuilt according to their actual lifetime.
- The Jolt side invokes backend lifecycle operations after the commitment is absorbed, preserving transcript behavior while reducing unrelated-stage residency.

### Stage 6 RA reuse

The Booleanity address phase already computes every instruction, bytecode, and RAM `RaIndices` value. Those indices are now shared through `Arc<Vec<RaIndices>>` and gathered in one L2-blocked pass for the Stage 6b initializers.

This removes repeated full-trace decoding and lookup work:

- bytecode no longer reruns PC lookup once per bytecode chunk;
- instruction RA no longer reconstructs lookup indices once per instruction chunk;
- RAM RA no longer remaps addresses once per RAM chunk.

The measured Stage 6b initialization moved from 2.64 s to 0.16 s plus 0.16 s of gathers, and Stage 6b inclusive moved from 8.03 s to 6.02 s at `2^26`.

### Spartan decode reuse

`DecodedCycle` separates trace decoding from R1CS row construction. Sequential consumers carry the next decoded step into the following iteration, and outer uni-skip evaluates both low-bit constraint groups from one decoded trace row. The resulting sums and transcript are unchanged.

### Benchmark harness

Both profiling harness paths now pass the trace collected for preprocessing directly to `gen_from_trace`. This removes a second guest execution from total benchmark wall time and Perfetto output without changing proof-only timing.

## Performance evidence

Production workload: `sha2-chain`, 17,785 iterations, padded trace `T = 2^26`, K256, release build.

| metric | pre-protocol baseline | final accepted runs | change |
|---|---:|---:|---:|
| total prove | 103.30 s | 74.13–76.29 s (75.21 mean) | −27.2% mean |
| one-hot commit | 42.90 s | 30.19–31.16 s (30.68 mean) | −28.5% mean |
| batched prove/opening | 25.60 s | 13.66–14.06 s (13.86 mean) | −45.9% mean |
| verify | — | 0.182–0.185 s | within target |
| conservative peak-RSS bound | — | ≤43.61 GB | includes retained setup suffix |

The representation experiment isolated the main mechanism:

| metric | full lane-zero control | implicit-zero candidate | change |
|---|---:|---:|---:|
| root contributions | 1,815,318,886 | 898,701,558 | −50.49% |
| root wall, mean | 68.993 s | 35.852 s | −48.04% |
| RSS, mean | 36.46 GB | 34.77 GB | −4.6% |

The associated backend memory campaign moved `/usr/bin/time -l` peak RSS from 87.71 GB at its post-sync baseline to 47.06 GB after setup-matrix release and 44.11 GB after lazy commit-tile blocks. Those measurements have a different campaign baseline from the final protocol table above and are included to document the lifetime changes rather than combine unlike runs.

All performance results and rejected experiments are recorded in:

- `specs/akita-prover-codesign-2e26.md`;
- `specs/akita-rss-reduction.md`;
- `specs/packed-onehot-protocol.md`;
- `specs/committed-column-virtualization.md`.

## Soundness and regression coverage

The recorded protocol-validation campaign covers:

- concrete implicit-zero recentering at non-Boolean points;
- balanced decomposition boundary vectors for K16 and K256;
- natural K16 and forced-K256 packed proofs;
- committed-program and trusted/untrusted advice paths;
- clear-claim, semantic-evaluation, selector, and commitment-wire tampering;
- all Akita claims tests and fixture-enabled verifier tests;
- standard and ZK Dory mul/div non-regression gates;
- schedule catalog regeneration, drift, and coverage checks;
- affected Akita, standard, and ZK Clippy targets with warnings denied.

Before the final harness-only commit, the integration checkpoint also recorded all 72 `akita-algebra` tests, all 42 `jolt-akita` tests, packed end-to-end tests, and focused Clippy. At the current head, the final harness cutover was checked with `cargo fmt --check` and targeted host-mode Clippy for the profiling binary.

## Dependency note

This PR moves Akita dependencies from `LayerZero-Labs/akita@fc2071619` to `markosg04/akita@72e13d9dd`. The revision is 32 commits ahead of the old pin and contains the backend support required here: rank-aware scheduling, streamed/fused one-hot work, setup/NTT lifecycle control, lazy block construction, fold streaming, and copy-free witness-row consumption.

This fork pin is the principal integration dependency. If project policy requires an organization-owned dependency before merge, these Akita commits need to be upstreamed and the pin replaced with the resulting `LayerZero-Labs/akita` revision; the Jolt protocol and kernel diff is otherwise self-contained.

## Attribution

This is joint work with Git history preserved.

- **Markos Georghiades** authored 35 commits on the Jolt branch, including the original prover/RSS co-design campaign, rank-aware schedules, Stage 6 and Spartan decode reuse, setup and cache lifetime integration, the implicit-zero/balanced-increment protocol implementation, verifier wiring, measurements, and design records.
- **Quang Dao** authored 9 non-merge commits plus the `main` integration merge, including the streamed single-polynomial source, parallel K16/K256 kernels, direct lane consumption, fused ring work, shift reuse, tiling, kernel spans and production benchmarks, and benchmark trace reuse.
- The 32-commit Akita revision also preserves contributions from **Quang Dao, Markos Georghiades, Omid Bodaghi, Amirhossein Khajehpour, and `sumchecker`**, alongside upstream history already present at the old pin.

## Suggested review order

1. `specs/packed-onehot-protocol.md` for the protocol identity and evidence.
2. `jolt-claims` lattice geometry, packing, strategy, and Hamming relation for the symbolic statement.
3. prover Stage 7 reduction plus verifier Stage 7/8 wiring for claim recentering and transcript binding.
4. `jolt-akita/src/trace_onehot.rs` and `scheme.rs` for the streamed packed kernels.
5. schedules/configuration and setup lifecycle calls.
6. Stage 6 RA sharing, Spartan decoding, tests, and performance tooling.

---

AI-authored by Codex (GPT-5) on behalf of Quang Dao.


quangvdao:
Added a decompose-fold kernel optimization in `933987b63`.

What changed:

- D64 retains the existing dense `i16` rotated-challenge representation and fixed four-row arithmetic path.
- D128, D256, and D512 now prepare sparse rotations by coefficient class, including precomputed negacyclic wrap points, instead of materializing dense `D × D` rotation tables.
- Packed-row contributions are fused across arbitrary row counts.
- Position-parallel accumulation uses approximately 2 MiB output tiles to keep the working set cache-sized.
- `JOLT_AKITA_DECOMPOSE_MODE={auto,dense,sparse}` supports reproducible A/B comparisons. Production defaults to dense only for D64 when the table fits its memory budget.

Correctness coverage compares the streamed dense and sparse paths against the materialized one-hot implementation for K16 and K256 across D64, D128, D256, and D512. It now covers both blockwise and flat accumulation geometries. The `jolt-akita` gate passes with 42 tests passed, 3 skipped, plus Clippy with warnings denied.

Performance on the 2^20 K16/D64 streaming-opening benchmark, measured back-to-back on an idle machine:

- Revised kernel median: 751.3 ms
- Original kernel median: 886.5 ms
- Improvement: approximately 15.3%

An attempted compact `i8` plus NEON D64 representation was rejected because widening overhead made it slower. The final change keeps the proven D64 `i16` arithmetic while retaining the scalable sparse path for larger ring dimensions.

Posted by Cursor assistant (model: GPT-5.6 Sol) on behalf of the user (Quang Dao) with approval.

quangvdao:
Updated the draft to `e1d919b28da9163b232084e0c03324ffadcc334f` with the source-only follow-up from `perf/packed-onehot` and the remaining K256 candidate from closed #1706.

Included:
- retain the packed trace-lane cache through opening;
- reuse and avoid initializing round-zero Spartan R1CS rows;
- reuse cached RA indices in stage 7 (source only);
- fuse dense D64/K256 trace rows (`fe4a318a4`).

The experimental Markdown/Perfetto payload from `perf/packed-onehot` was deliberately not imported (~106k lines of raw trace data). The existing stage-7 evidence reports 1.841s -> 0.628s for stage 7 and 66.94s -> 65.96s end-to-end at 2^26.

Local `cargo fmt --all -- --check` and `git diff --check` pass. A natural-K256 2^25 end-to-end baseline reached the proving phase but was interrupted after 48.4s because the host was oversubscribed; therefore there is no new local benchmark result for the dense-K256 candidate and no completed local compile/test claim for this head. Fresh CI should be treated as the validation source.

quangvdao:
Close since stale

PR #1716, opened 2026-07-29
perf(dory): prime the prepared-G2 pairing cache generation-safely

0xAndoroid:
Stacked on #1714 (`perf/kernels-optimized`).

The modular stack deliberately skipped dory's global prepared-point cache because `init_cache`'s "smart reuse" keeps any size-≥ cache while URS generator **values** are seeded on the exact size — two setup sizes in one process would silently pair against wrong generators (dory's consumers prefix-match blindly). The footgun is real, but every dory cache consumer slices the setup `g1/g2` prefix from index 0, so the cache is safe iff it holds exactly the current setup's vectors.

This PR maintains that invariant with a process-wide generation guard in `DoryScheme::setup_prover`:

- first canonical setup size seen → `init_cache` (full win for single-size production processes)
- any different size later → `invalidate_cache` + priming disabled for the process lifetime (per-call G2 prep, i.e. the previous behavior — multi-size test processes stay correct; `property_based_round_trip` exercises this)

**Numbers** (pure-modular process, M4 mini, min-of-66 per-call over 3 interleaved runs): `multi_pair_g2_setup` @1024 pairs **23.7 → 18.0 ms/call (−24%)**. Scales with tier-2 pairing count (~328k pairings @2^23 ⇒ ~1.5–2 s thread time). Note the `modular_benchmark` harness masks this win: its legacy preprocessing already primes the cache via `DoryGlobals::init_prepared_cache`; the guard makes the modular stack self-sufficient and size-safe instead of relying on that.

Proof bytes unchanged — prepared points are deterministic from the same affine inputs: byte_diff 19/19, jolt-dory 41/41, clippy `host` + `host,zk`, fmt.

**Dropped companion candidate:** the traffic model also flagged `combine_hints` as "328k plain double-and-add muls, no GLV". Implemented GLV vector ops and benched them **~15% slower** — the premise was wrong: the arkworks fork's G1 `Mul` is already GLV (`GLVConfig::glv_mul_projective`), so the existing row-major loop is the measured optimum. Documented at the call site.


0xAndoroid:
Superseded by a16z/dory#29. The prepared-point cache now binds entries to generator identity and validates every setup-specialized pairing upstream, so Jolt does not need a process-local generation guard.

PR #1718, opened 2026-07-30
feat(jolt-prover): port the Akita integration to the modular prover

moodlezoup:
Ports the Akita (lattice) integration from `jolt-prover-legacy` (#1676) to the modular `crates/jolt-prover`. Like `jolt-verifier`, one compiled prover proves exactly one protocol: the `akita` feature swaps the wire types to the packed envelope, so exactly one of the two path modules (`dory`, `akita`) compiles into any given build.

## Acceptance

All nine tests from the scaffolding commit (2090501) pass un-ignored, in CI via the extended `test-prover-akita` job:

- **5 e2e analogs** of the legacy packed suite (`tests/akita_e2e.rs`): muldiv (with live tampers on the fused-inc claim wires), forced K=256, advice, full-capacity advice, committed-program × {1,2} chunks (with ProgramOneHot / reconstruction-wire tampers).
- **4 byte-diff ratchets** (`tests/akita_byte_diff.rs`): the modular proof equals legacy's `prove_packed` output **wire-for-wire** (structural equality on the shared `AkitaJoltProof` types) on muldiv, advice-consumer, and committed-program × {1,2} chunks, and verifies end-to-end.

## Structure

- **Shared stages** (`src/stages/`): recipes 1–7 and the generated driver expansions are protocol-agnostic; the protocol differences ride `jolt-verifier`'s feature-swapped batch internals plus two small cfg blocks mirroring the verifier's own (the stage-6a lattice input wrapper folding the four reduced `Inc` claims, the stage-7 lattice hamming dimensions). `dory/` and `akita/` keep only their pipeline ends (stage 0 commit, stage 8 opening, prover).
- **`akita::prove`**: stage 0 assembles the `OneHotTrace` columns off the witness plane and commits them as one native group, commits the per-proof untrusted-advice byte object, and absorbs through the verifier's own promoted helper; the reconstruction phase proves through the verifier's promoted batch construction; stage 8 opens `OneHotTrace` as one native same-point batch and discharges the auxiliary objects via `prove_packed_openings`. Precommitted auxiliaries (trusted advice, ProgramOneHot) are transparently re-derived at prove time and cross-checked against the passed commitments.
- **Generic seams** so `jolt-prover` needs no `jolt-akita` dependency: `CommitmentScheme::commit_batch` (native group commit), `TransparentObjectSetup` (single-sourced with legacy's convention), `GroupSetupMetadata` on the prover setup. Concrete Akita types bind at call sites.
- **`jolt-kernels` stays feature-free** (per the `CANONICAL_INSTRUCTION_ADDRESS` doctrine): the reference kernels adapt to the packed jolt-claims shape at runtime — the read-raf address kernel grows the nine-stage fused-inc fold (fused-delta-weighted pushforwards against the staged store column and its complement, RAF/entry weights shifted to γ⁹..¹¹), booleanity extends its masses over the inc columns, and the cycle/hamming kernels serve extra tables per the relations' own expression leaves. The four fused consumer cycle points ride `BytecodeStagePoints`.
- **`jolt-witness`** serves the lattice columns feature-free: `FusedInc`, the `UnsignedIncChunk`/`UnsignedIncMsb` one-hot grids, and the advice byte one-hot cell tables.

## Notable fixes the ratchets caught

1. The packed trace floor: `ProverConfig::derive` pads to 2¹² on the akita build (Akita cannot schedule the OneHotTrace group below 16 variables), matching legacy's `PCS::MIN_PADDED_TRACE_LENGTH`.
2. The lookup-selector block pads the non-power-of-two table count **up** (ceil, matching `jolt_poly::Math::log_2`).
3. Reconstruction leg convention: the verifier folds the `Π(1−vᵢ)` zero-pin into the **derived weight**, so the column tables must **replicate** across the missing high variables — zero-extending both squares the pin, passes every driver self-check, and fails only at the packed PCS opening.

## Review

A multi-agent adversarial review over the diff (base regressions / akita soundness / kernel math / witness encoding, each finding adversarially verified) confirmed two fail-closed hardening gaps, both fixed: stage 0 now rejects a ProgramOneHot argument disagreeing with the preprocessing-held commitment the verifier absorbs (was an opaque FS divergence at verification) and a packed setup whose declared dimensions disagree with the canonical OneHotTrace shape. Also restored from legacy: the advice-length guard, the store/rd disjointness debug-assert with the fused-encoding unit tests (±(2⁶⁴−1) extremes), and a compile-time pin of `INSTRUCTION_FLAG_ORDER` to the enum's discriminant order.

## Gates

clippy `-D warnings` on `jolt-prover` × {default, `prover-fixtures`, `akita`, `akita,prover-fixtures`} and the workspace under `host` / `host,zk`; nextest on `jolt-prover` × {default, `prover-fixtures` (Dory byte-diff), `akita,prover-fixtures` (17 tests)}, `jolt-{kernels,witness,openings,akita}`, `jolt-verifier` × {base, `akita`}, and legacy `muldiv` under `host` / `host,zk` plus the legacy akita suite.

Remaining port follow-ups (tracked in the test TODOs): derive the OneHotTrace setup params from `ProverConfig` + program shape (tests currently size the setup through a legacy prover instance), a packed committed-program prover-data shape, and the stage-granular packed byte-diff ratchet.


crates/jolt-prover/src/akita/prover.rs
```diff
@@ -0,0 +1,269 @@
...
+        .map(|commitments| {
+            let chunk_count = checked
+                .precommitted
+                .bytecode
+                .as_ref()
+                .map(|layout| layout.chunk_count())
+                .ok_or(ProverError::InvariantViolation {
+                    reason: "committed-program mode without a bytecode schedule",
+                })?;
+            let program =
```
markosg04: This recreates the setup, witness, commitment, and opening hint for each precommitted object during every proof. The commitment cross-check keeps the path correct, but moves preprocessing work into proof time. Packed prover-data should eventually retain the object setup, witness, and hint so `prove` can consume them directly.

crates/jolt-prover/src/akita/reconstruction.rs
```diff
@@ -0,0 +1,634 @@
...
+    let mut table = vec![F::zero(); 1usize << (BYTE_BITS + place_bits)];
+    for byte in 0..(1usize << BYTE_BITS) {
+        for place in 0..(1usize << place_bits) {
+            table[(byte << place_bits) | place] = F::from_u64(byte as u64) * F::pow2(8 * place);
+        }
+    }
+    table
+}
+
+impl<F, PCS> PrepareKernel<F, UntrustedAdviceReconstructionInstance<F>> for JoltAkitaBackend<F, PCS>
```
markosg04: These reconstruction kernels are implemented directly on `JoltAkitaBackend`, unlike the replaceable slots used by the shared backend. Before adding an optimized packed backend, it would be cleaner to give reconstruction its own slots or otherwise place these reference kernels behind the same interface.

crates/jolt-prover/src/akita/witness.rs
```diff
@@ -0,0 +1,622 @@
...
+    fn num_vars(&self) -> usize {
+        self.num_vars
+    }
+
+    #[expect(
+        clippy::unimplemented,
+        reason = "hint-owned witnesses are never evaluated here"
+    )]
+    fn evaluate(&self, _point: &[F]) -> F {
+        unimplemented!("hint-owned one-hot witness is evaluated by the Akita backend")
```
markosg04: This shape stand-in relies on Akita opening from the hint, although the `open_batch` contract still permits an implementation to call `evaluate` or `for_each_row`. Another implementation would panic here. The generic seam should eventually express hint-owned opening directly, or receive a real polynomial.

markosg04 (approved):
Looks good overall—just left a few small follow-up comments.

crates/jolt-prover/src/akita/prover.rs
moodlezoup: Done in 6f564708f: `CommittedProgramProverData` now retains the `ProgramOneHot` objects on the akita build (witnesses, plans, digest-seeded setups, hints — built once at preprocessing via `commit_program_one_hot`), and `akita::prove` takes the precommitted trusted-advice object (`Option<&AdviceOneHot>`) instead of its bare commitment, mirroring legacy's `prove_packed`. All per-prove re-derivation is gone; stage 0 cross-checks the retained commitments against the verifier preprocessing fail-closed. Byte-diff ratchets still match legacy wire-for-wire.

crates/jolt-prover/src/akita/reconstruction.rs
moodlezoup: Done in 6f564708f: `JoltAkitaBackend` now derives `KernelSlots` over four reconstruction slot fields (`Box<dyn PrepareKernel<F, …>>`, same delegating-impl pattern as the shared registry), and the reference kernels moved onto a `ReferenceReconstruction` marker that `reference()` boxes into the slots. An optimized packed backend now swaps boxes rather than reimplementing the type.

crates/jolt-prover/src/akita/witness.rs
moodlezoup: Done in 6f564708f, taking the first of your two directions: the seam now expresses hint-owned opening directly — `open_batch` (polynomial args, hint by value) became `open_batch_from_hint` with no polynomial argument, and the panicking `CommittedOneHotShape` stand-in is deleted. We dug into why before choosing: the Akita hint is not a cache but the retained committed object — the upstream lattice opening consumes the Ajtai commit's own output (`CommitInnerWitness`: the `A·s` rows and their gadget decompositions), which can't be recreated without re-running the commit, while the logical→backend polynomial conversion is a trivial index re-wrap. So a real-polynomial contract could only be honest at the cost of re-committing inside open. The principled fix — a first-class `Committed` object type on `CommitmentScheme` replacing the `OpeningHint` side-channel — is filed as #1782 (it reshapes the trait across all schemes, so it gets its own PR).

PR #1725, opened 2026-07-31
docs(specs): byte-addressable memory design exploration

moodlezoup:
## Summary

Design exploration for sub-word and misaligned load/store support (Zicclsm), motivated by the EF's zkVM standard prescribing **RV64IM+Zicclsm** as the common compilation target, and by misaligned accesses from unsafe Rust / FFI being unprovable today.

Two measurements reframe the problem (btreemap at 2^24):
- All RAM sumchecks combined are **~0.6% of prover thread-time** — commitment work and the Dory opening dominate, so trace length is the master cost lever, not the memory argument.
- **22.3% of the entire btreemap trace is sub-word emulation overhead** (a traced LB is 8 cycles, SW is 15).

## Recommendation (tiered)

Keep Twist doubleword-addressable and unchanged; add byte semantics at the instruction layer:

- **Tier 0** — fused extract/merge lookup tables (no protocol change): LB 8→4 cycles, SW 15→7.
- **Tier 1** — intra-dw offset via the memory cycle's *idle* lookup slot (`impl_lookup_table!(Ld, None)` today): width-specialized `AddressOffset` tables output `ea & 7` or an unsatisfiable marker; one product wire feeds `RamAddress − OffsetOrZero` into the RAF-evaluation input claim — the only Twist-side change, zero new committed columns. Sub-word loads 2 cycles, narrow stores 5.
- **Tier 2** — dual-expansion dispatch for the dw-crossing case: **both** variants compiled into the static bytecode at distinct expanded PCs; per-instance the PC stream enters one (like a trace taking both sides of a guest branch); wrong-variant execution is unsatisfiable via the marker. Full Zicclsm; aligned path unchanged; crossing ~5–11 cycles ("correct but possibly slow" is explicitly permitted).
- **Tier 3 (optional)** — rv→lookup-operand fusion for 1-cycle loads / 2-cycle stores.

Projected: btreemap trace −17% (Tier 2), −21% (Tier 3); fibonacci ~0%; full alternatives analysis in the spec (8-port byte Twist, 1-byte-per-cycle, packed-lane Twist, dual-port, branchy sequences, trap-and-emulate) with measured cost arguments.

## Notes for review

- Every load-bearing mechanism is anchored to file:line in the spec (RAF-eval unmap identity, constraint 17/18 sequence rules, `BytecodePCMapper` keying, operand-mux additivity, bytecode lane budget 447/512, LC 5-term cap).
- Acceptance criteria include flipping arch-test `misaligned-ldst-01` to passing and fixing two latent MMU trace-recording bugs currently masked by alignment asserts.
- Open questions at the end: mapper variant keying, extract-table prefix/suffix placement, marker-constant bound, bytecode-K growth on Ethproofs-class guests.

### DIFF
--- a/specs/byte-addressable-memory.md
+++ b/specs/byte-addressable-memory.md
@@ -200,6 +200,27 @@ variant dispatch. Tier 0 may relax the asserts from "naturally aligned" to "does
 doubleword boundary" (a new assert table) as an intermediate Zicclsm step: this covers all
 misaligned accesses except crossing ones.
 
+**Tier 0 as implemented (PRs #1761, #1762, #1768).** The fused single-lookup extract sketched
+above was rejected during implementation: with the offset delivered in *binary* form, the
+per-offset lane functionals are linearly independent at every phase-boundary cut, so the table
+has intrinsic prefix rank ≥ 8 per width (~16 new prefix/suffix families per width, per registry
+side). The implementation instead follows the codebase's own SRL/SRA binary-to-positional
+pattern, at one extra cycle per access: a rank-1 mask lookup converts the effective address to
+the addressed lane's byte mask (`WindowMask{B,H,W}`, each reading only the offset bits its
+alignment class allows, which also keeps every output in u64 range), and a constant-rank
+extract pulls the lane through the mask (`PextSigned` = `pext(x,y) + σ·(2^64 − 2^popcount(y))`,
+width-independent, full-domain testable; unsigned loads use plain `Pext`). Stores mirror this:
+`ANDN` clears the lane and width-specialized `ShiftData{B,H,W}` tables
+(`(rs2 mod 2^8w)·2^{8·offset}`, a rank-2 product over disjoint variables) place the store data;
+a general `Pdep` table was analyzed and rejected (deposited bit positions depend on the suffix
+popcount: prefix rank ~57). Total new machinery: 11 tables, 11 virtual instructions, 5 prefix
+and 9 suffix families across both registry sides. Realized traced cycles: LB/LBU 8→5,
+LH/LHU 9→6, LW 8→6, LWU 9→6, SB 13→8, SH 14→9, SW 15→9. Measured trace deltas:
+`btreemap` −8.9%, `sha2-chain` −7.2%, `fibonacci` ~0%. Measured prover wall-time: parity at
+fixed padded trace length (trace savings inside a power-of-two instance are invisible; the two
+extra globally-materialized prefixes cost nothing measurable); −27% at a padding-boundary
+crossing. The benefit unit is guest cycles per padded instance, not per-input wall time.
+
 ### Tier 1 — Offset via the memory cycle's idle lookup slot; floor addressing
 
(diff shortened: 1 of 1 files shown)


Acentelles:
Taking this over, starting with Tier 0 as discussed.

Before touching anything I adversarially verified the spec's load-bearing claims against the code. The core mechanisms all check out: the idle lookup slot on load/store cycles and the unconditional one-hot commitment (`witness.rs:253-262`), the traced cycle counts (all six re-derive exactly from the recipes plus the `SLL`/`SRL` sub-expansions), the non-interleaved `AddOperands` path for alignment asserts, the operand mux being a plain sum with no flag-exclusivity constraint, RAF evaluation's `input_claim_constraint` already supporting sum-of-products, and the temp/sequence/lane budgets (447/512 confirmed term by term).

Three corrections I'd like to fold into the spec, flagging them here first:

1. **Three latent MMU trace-recording sites, not two.** `trace_store_halfword` (`tracer/src/emulator/mmu.rs:590-629`) has the same defect as the two listed: `word_address = (ea >> 2) << 2` truncation plus a 32-bit mask (`value | (pre_value & 0xffff0000)`) against an 8-byte `pre_value`. There's also a fourth inconsistency in `trace_store` (`mmu.rs:632-663`, used by `store_word`): it records an 8-byte write of a 4-byte value. The acceptance checklist should cover all of them.
2. **The `misaligned-ldst-01` framing.** The README line cited as evidence of a "tracked known-failure" (`tests/arch-tests/README.md:113`) is a fenced sample-output block in the "Interpreting output" section (`add-01` and `misa-01` sit alongside it as placeholders). The test itself lives in the ACT4 submodule and there's no failure registry. The acceptance criterion (flip `sail.json` to `misaligned: supported`) stands; the justification sentence needs rewording.
3. **`remap_address` naming and divergence.** The `common/src/jolt_device.rs` copy is named `remap_word_address` (a grep for `remap_address` misses it), and the three copies diverge on the below-lowest-address case (legacy panics, jolt_device returns `Err(AddressBelowLowest)`, jolt-prover asserts). Only the `/8` truncation is uniform, which is what Tier 1 needs, but the divergence is worth noting for the Tier 2 guard-doubleword work.

Two smaller notes for the Tier 1 section, no action needed now: the RAF input claim carries a `mul_pow_2(phase3_cycle_rounds())` scaling (`raf_evaluation.rs:125-133`) that the `−OffsetOrZero` change must preserve, and the non-zk `input_output_claims()` path (`raf_evaluation.rs:461-475`) isn't discussed alongside the zk constraint.

Plan: push these spec edits to this branch, then start Tier 0 with a thin vertical slice (one table end-to-end, `ExtractW` through the full ring wired into the LW expansion) to validate the extract-family prefix/suffix decomposition (open question 3) before fanning out to the rest of the family. On that question: `shift_right_bitmask.rs` already has the `Σ_o eq(y_low, o) · lane_o(x)` shape, and under the interleaved convention y's low 3 bits sit at index bits 0/2/4, inside the final suffix chunk at both d=4 and d=8, so I expect it to decompose cleanly without a new prefix family per width.

0xAndoroid (approved):

PR #1726, opened 2026-07-31
fix(tracer): production boundary verification + CI golden gate for the parallel mode

Acentelles:
## Summary

Two additions to the parallel mode, per the handoff discussion — the serial path needs nothing.

**1. Production boundary verification** (`fix(tracer): verify chunk boundary state in production parallel runs`)

The per-chunk row-count assert misses divergences that preserve row counts (a value-level advice mismatch under replay, or a guest violating the append-then-read advice assumption): the trace would assemble silently corrupted. `diff_vs_cpu` was built for exactly this comparison but only ran in tests and examples.

Pass-1 already captures a full `ChunkCheckpoint` at every boundary, and every boundary is both chunk k's end and chunk k+1's start — so jobs now carry an `Arc`'d start checkpoint plus a `OnceLock` slot for the end boundary, published when pass-1 captures the next chunk (or the final state at termination). Workers verify their replay landed exactly on it. Cost: one state comparison per chunk (~33 KB memcmp per 2^20 rows); the `OnceLock` wait is a no-op in the steady state. Regression test: fault-injected register corruption after replay (row counts preserved) must panic instead of assembling.

**2. CI enforcement of the golden gate** (`ci: enforce the golden-trace gate`)

`trace_golden` existed but ran manually; "bit-exact on every commit" was author discipline rather than enforcement. New CI job runs `check` against the recorded fixtures in serial mode and in parallel mode (4 workers, 128-row chunks). Fixture re-records change what the gate proves and should be review-flagged events.

All six parallel-mode tests pass, including both fault-injection tests.


Acentelles:
Pushed a third commit that replaces the golden fixtures from this PR with a fixture-free equivalence check. The fixtures were wrong and I'd rather not have you merge them.

**Why they can't work.** Guest ELFs are rebuilt from source, and their `.rodata` embeds absolute build paths — `strings` on `fibonacci-guest` shows dependency panic locations under `$CARGO_HOME`. Whoever records writes `/Users/<name>/.cargo/...`; a runner writes `/home/runner/.cargo/...`. The differing lengths shift everything after them, so the final memory image differs for every guest, and any row that loads a shifted address differs too. The instruction stream is untouched, which is what makes the failure signature so recognisable: on a branch where I had the job enabled, all five guests came back with **equal row counts, every memory hash different, and three of five cycle streams different**. It's tighter than "machine-specific" — the same machine in a different checkout directory produces a third set of hashes.

So `golden_traces.json` (already on `perf/tracer-100mhz` via the earlier cherry-pick) is only ever valid for the person and directory that recorded it. Since this PR is what would have started *enforcing* it in CI, merging as it stood would have turned the job red on every runner.

**What replaces it.** `trace_equivalence` compares configurations inside one run rather than against stored values: serial as the reference, then parallel at 1 and 4 workers over 128-row and 64k-row chunks, each in a child process (the tracer picks its pipeline up from the environment, so spawning avoids mutating a process that already has worker threads). Nothing to re-record, portable, and it targets the property the two-pass pipeline actually has to guarantee. Verified locally: 4/4 configurations byte-identical across all five guests, and the row counts match the deleted fixtures exactly (925,216 / 919,991 / 785,996 / 625,461 / 473).

What it gives up is "output unchanged versus main", which the fixture version did not really deliver either — as soon as a toolchain or inline-crate bump forced a re-record, it proved "re-recorded" rather than "unchanged". If you want that property, it needs a reproducible guest build (`--remap-path-prefix` or equivalent), which felt out of scope here.

The boundary-verification commit is unchanged and still the substantive half of this PR. Also worth deleting `crates/jolt-prover-legacy/tests/fixtures/golden_traces.json` from `perf/tracer-100mhz` if you'd rather not carry a file nothing reads.

0xAndoroid (approved):
lgtm, thanks

moodlezoup (approved):

PR #1727, opened 2026-07-31
test(jolt-eval): trace-generation benchmarks and baseline harness (x86-tracer slice 0)

Acentelles:
## Summary

Slice 0 of `specs/x86-tracer-backend.md` (#1713): backend-generic trace-generation measurement infrastructure, landed before any backend work so baselines exist first (AC1).

- `TraceGenObjective` + `build_trace_setup` in jolt-eval: measures eager `ExecutionBackend::trace` through the modular seam, with guest build and `JoltProgram` construction excluded from the timed region.
- Criterion targets `trace_gen_fibonacci` / `trace_gen_sha2_chain` (the `prover_time_*` precedent: registered via `sync_targets.sh`, deliberately not in the optimizer enum), rows/s via `Throughput::Elements`; `x86`/`x86_fast` ids join with the backend.
- Guest inputs pinned to the `e2e_profiling.rs` defaults; new profile-guest configs (sha2, sha3, sha3-chain, btreemap) and an objective-neutral `GuestConfig::label()`.
- `trace-gen-baseline` bin: subprocess-per-guest throughput + peak RSS (`VmHWM` on Linux, `getrusage` fallback), markdown table output.
- `TRACER_PARALLEL` pinned in all harnesses (the tracer env-dispatches to the two-pass pipeline; measurements must be environment-independent).

Stacked on #1717 (base: `perf/tracer-100mhz`) and includes the two commits of #1726 (cross-repo PRs cannot chain fork branches as bases; the diff shrinks as the stack merges). Review from commit `c568c227d` onward.

## Note on numbers

The modular seam includes the per-row `Cycle` to `TraceRow` conversion and final-memory materialization, so seam MHz sits well below raw `trace_bench` MHz by construction. Post-#1717 that conversion pass dominates seam throughput and is a first-class optimization target of its own.


0xAndoroid (approved):
lgtm, thanks. I'm just going to run AI review as well

jolt-eval/bin/trace_gen_baseline.rs
```diff
@@ -0,0 +1,161 @@
...
+        println!(
+            "| {} | {} | {:.3} | {:.3} | {:.1} |",
+            v["guest"].as_str().unwrap_or(guest),
+            v["rows"],
+            v["median_wall_s"].as_f64().unwrap_or(f64::NAN),
+            v["mhz"].as_f64().unwrap_or(f64::NAN),
+            v["peak_rss_bytes"].as_f64().unwrap_or(f64::NAN) / (1024.0 * 1024.0),
+        );
+    }
+    println!(
```
0xAndoroid, line 153: The footer prints only `{ARCH} {OS}`, but this table is the AC8/AC10 denominator quoted at #1729 acceptance. Worth adding: CPU model, core count, commit hash, and the run spread (you already have all wall times — min/max next to the median makes outliers visible). Consider also stating the memory config: these guests run with `GuestConfig`'s heap = 32 KiB, not the 32 MiB `Program::new` default that `e2e_profiling.rs` uses — doesn't move the numbers (measured: plane is ~46 KB, fixed costs ≤0.2 ms), but "reference" should say what it ran.

jolt-eval/bin/trace_gen_baseline.rs
```diff
@@ -0,0 +1,161 @@
...
+#[cfg(target_os = "linux")]
+fn peak_rss_bytes() -> u64 {
+    let status = std::fs::read_to_string("/proc/self/status").unwrap_or_default();
+    status
+        .lines()
+        .find_map(|line| line.strip_prefix("VmHWM:"))
+        .and_then(|rest| rest.trim().strip_suffix("kB"))
+        .and_then(|kb| kb.trim().parse::<u64>().ok())
+        .map(|kb| kb * 1024)
+        .unwrap_or(0)
```
0xAndoroid: `unwrap_or(0)` (and the `cfg(not(unix))` arm) turn "couldn't measure" into a `0.0 MiB` table cell that reads as a measurement. `Option<u64>` → `n/a` keeps the table honest on parse failures / unsupported platforms.

jolt-eval/bin/trace_gen_baseline.rs
```diff
@@ -0,0 +1,161 @@
...
+        let output = std::process::Command::new(&exe)
+            .args(["--guest", guest, "--runs", &runs.to_string()])
+            .output()
+            .expect("failed to spawn measurement subprocess");
+        if !output.status.success() {
+            eprintln!("{}", String::from_utf8_lossy(&output.stderr));
+            panic!("measurement subprocess for {guest} failed");
+        }
+        let stdout = String::from_utf8_lossy(&output.stdout);
+        let line = stdout
```
0xAndoroid, line 130: `lines().last()` assumes the JSON is the final line, but guest prints land on this same stdout (`handle_jolt_print` — `cpu.rs:1284` is `print!`, no trailing newline), so a guest ending with `print!` corrupts the parse. None of the six baseline guests print today and the failure is a loud panic, so this is hardening: prefix the line (`BASELINE_JSON: {…}`) and parse the last line carrying the marker.

jolt-eval/benches/trace_gen_fibonacci.rs
```diff
@@ -0,0 +1,28 @@
...
+    // measurements are environment-independent.
+    std::env::remove_var("TRACER_PARALLEL");
+    let objective = TraceGenObjective::new(Fibonacci(400000));
+    let setup = objective.setup();
+    let mut group = c.benchmark_group(objective.name());
+    group.sample_size(10);
+    group.sampling_mode(SamplingMode::Flat);
+    group.measurement_time(std::time::Duration::from_secs(60));
+    group.throughput(Throughput::Elements(setup.trace_len as u64));
+    group.bench_function("reference", |b| b.iter(|| objective.run_reference(&setup)));
```
0xAndoroid: `run_reference` already returns the row count — asserting it equals `setup.trace_len` here is free and guards `Throughput::Elements` when the `x86`/`x86_fast` ids join this group: a backend emitting a different row count would otherwise silently misreport MHz. (Row equality is AC5's job, but the bench shouldn't depend on it to report correct units.)

jolt-eval/src/guests/mod.rs
```diff
@@ -114,6 +122,10 @@ pub trait GuestConfig: Default + Send + Sync {
     /// Cargo package name (e.g. "fibonacci-guest").
     fn package(&self) -> &str;
 
+    /// Objective-neutral guest label including parameters
+    /// (e.g. "fibonacci_400000"). Objectives prefix it with their own name.
+    fn label(&self) -> String;
+
     fn memory_config(&self) -> MemoryConfig {
```
0xAndoroid, line 129: Four of the six baseline guests are ≤10⁵ rows at their pinned inputs, so their MHz rows are fixed-cost dominated (per-trace `Vec::with_capacity(2^24)` reserve, ELF re-parse, device setup) — locally they read 20–35% below the two big guests on the same box. Fine for AC1 fidelity to `e2e_profiling`; suggest one footnote that MHz is only comparable within a guest, so nobody reads the table cross-guest.

0xAndoroid (approved):
Ran the harness locally at the PR head and instrumented the seam to decompose the timed region (M4 Max, so shares are indicative — structure carries to the target box):

| | fibonacci (4.40M rows) | sha2-chain (13.62M rows) |
|---|---:|---:|
| raw `crate::trace` | 37 ms (26%) → 118 MHz | 59 ms (14%) → 229 MHz |
| `Cycle→TraceRow` conversion | 108 ms (74%) | 377 ms (86%) |
| seam total | 145 ms → 30 MHz | 436 ms → 31 MHz |

The PR body already says the conversion pass dominates — quantified, it's 74–86% of the denominator on both AC8 guests. Since `X86TracerBackend` emits `TraceRow` directly, skipping conversion alone contributes 3.9× (fibonacci) / 7.3× (sha2-chain) of the ≥10× gate. Same shape on AC10: during conversion `Vec<Cycle>` (96 B/row) and `Vec<TraceRow>` (160 B/row) coexist, so the reference `VmHWM` is ~1.6× the 160 B/row floor an x86 arm would have.

Ask: make the baseline decomposable so #1729's ratios are interpretable — either add a `reference_raw` Criterion id (timing `crate::trace` only, same group/throughput) or include the `trace_bench` serial-arm MHz as a column in the baseline table, plus one footnote on the RSS composition. Not blocking slice 0; I'd want it before AC8/AC9 numbers are graded.

Also: AC1 wants the reference numbers recorded in the PR description before codegen lands — can you run `trace-gen-baseline` on the linux-x86_64 target box and paste the table here, with CPU model, core count, and commit hash? These numbers get quoted months from now; `aarch64 macos`-style provenance won't be enough to reproduce them.

Acentelles:
All five inline comments addressed.

I don't have a linux-x86_64 machine, so I haven't pasted numbers.

PR #1729, opened 2026-07-31
feat(jolt-tracer-x86): experimental AOT x86-64 tracer backend (x86-tracer slices 2-5)

Acentelles:
## Summary

Slices 2 to 5 of `specs/x86-tracer-backend.md` (#1713), **experimental**: `crates/jolt-tracer-x86`, an AOT x86-64 transpiling execution backend with fast and record modes over every final-bytecode kind, plus the differential-test infrastructure the spec puts before optimization.

- Whole-program AOT via dynasm-rs: one straight-line template per expanded-bytecode row, halfword-granular jump table (JALR dispatch + branch label source), PC-stall termination, explicit in-process bounds checks (deliberate departure from sp1-jit's no-checks/child-process model), sysv64 helpers for device I/O, host IO, and faults; fail-fast `TraceError` on unsupported kinds.
- `NativeBackend` alias: the AOT backend on x86_64-linux, the reference interpreter everywhere else; the crate compiles on all targets.
- **Fast-pass coverage of every final-bytecode kind.** Most of what looks missing from a base-ISA backend is source-only: shifts, W-arithmetic, the M and A extensions, CSR ops, ECALL and the advice-load kinds all expand into final rows the backend already emits (finality per `expand/grammar.rs` `is_source_only`). What remained was 28 templates: `Slt`, `Andn`, the Pow2 family, `MovSign`, `VirtualRev8W`, `Rotri`/`Rotriw`, `ShiftRightBitmaskI`, `VirtualSra`, the eight XOR-rotates, four asserts, both `ChangeDivisor` variants, the two advice-tape kinds, and `VirtualAdvice`. clippy independently confirms the coverage: the fail-fast match arm now reaches only `Noop`.
- **Per-group runtime advice.** The expanded bytecode erases source kind and inline key, so source instructions are recovered by re-decoding the ELF and keying by address; each group needing advice resolves at compile time into a job table, and one helper call at the group's entry fills the advice slots before any of its rows execute (matching the interpreter, which computes advice from the same pre-group state). DIV/REM mirrors the tracer's four variant formulas; **inline advice reuses the registered `build_advice` functions unchanged** through `InlineAdviceContext` implemented over the guest state — the seam introduced in the slice-1 PR, now with its second implementor. A `VirtualAdvice` row in a group without an advice computation is a compile-time error rather than a stale slot read.
- Per-instruction differential tests: 66 kinds x 1000 random operand/state instances against the reference `Cpu` (full register file, PC, memory, advice tape), with compile-time-exhaustive kind classification (`for_each_jolt_instruction_kind!`, no wildcard) pinned to what the transpiler accepts — 68/68 green. Whole-guest fast-pass equivalence across five guests (fibonacci, sha2-chain, sha3-chain, btreemap, muldiv — the last exercising the DIV/REM advice groups) against the post-#1717 interpreter: rows, device outputs, panic flag, final memory and advice tape all bit-exact.
- **`RowEmitter` seam** (`compile/emitter.rs`, per the updated spec): emission sits behind a trait whose production implementor is the dynasm templates. Emitters report `Unsupported` rather than erroring, so an ordered `EmitterSet` falls through and a hybrid (stencils for some kinds, templates for the rest) needs no extra plumbing; the fail-fast invariant-7 error moves to the set, keeping the authoritative supported-kind list in the emitter's own match. `compile_with(program, &EmitterSet)` plus the seam types are re-exported through the harness module, so the Alternative 11 stencil spike can live outside the crate. Prologue, dispatch stubs, jump table and chunk accounting stay in the driver — execution-model properties, not row encoding.
- **`jolt-emu-x86`** (AC4): the ACT4 runner. Arch tests halt by storing to `tohost` then spinning in `j .`, which the backend's PC-stall termination already ends, so the result word and signature region come out of final memory — no mid-execution HTIF hook. Matches `run.sh`'s contract and `jolt-emu`'s signature format; pipeline/transpiler rejection gets its own exit code so the x86 skip list can be regenerated mechanically ("skipped and listed").
- **`tracer_backend_equivalence`** (AC5): registered in `JoltInvariants` (Test, Fuzz, RedTeam) over a five-guest corpus with bounded fuzzable parameters. Scoped to what the fast pass observes — row count, device outputs and panic flag, final memory, advice tape — since full `TraceRow`-stream equality needs record mode; row count already catches control-flow divergence and final memory catches value divergence reaching RAM. The reference backend's eager trace is also compared against its own chunked fast pass on every platform, so the invariant is not vacuous where `NativeBackend` is the interpreter.
- **Record mode**: the backend produces `TraceRow`s, not just execution. Generated code writes a fixed 64-byte `Observation` per row (row index plus dynamic values) rather than constructing `TraceRow` directly, since its `Option` fields have no guaranteed layout; a Rust pass joins each observation with `expanded_bytecode[row_index]` for the static half. Two bodies are compiled from the same templates as separate assemblies (branch labels are per guest address and cannot be shared). `trace()` runs the fast pass first to size the observation buffer exactly, which also cross-checks the two bodies' row counts against each other.
- **Row-stream equality**: every `TraceRow` field-for-field against the reference on fibonacci, muldiv (DIV/REM advice) and sha2-chain (inlines), plus scale tests at 4.4M and ~800k rows. Since a proof is a deterministic function of the trace, byte-identical rows imply byte-identical proofs; literal AC7 additionally wants two proofs compared, which needs `byte_diff.rs`'s helpers extracted into a shared module (proposed, not done here).
- **Chunked execution** (`ChunkedExecutionBackend`, slice 4), so this can serve as an alternative fast first pass feeding per-chunk replay. Pausing is the crux: generated code does not maintain PC in straight-line rows, so it can only stop where a resumable PC is statically known, i.e. a group boundary — a row limit plus a two-instruction check per group start, with the group's address published as the resume PC. A boundary may precede the requested mark, which is what `skip_rows` covers and what makes chunk size 1 work. Checkpoints carry a memory image rather than an access-value log (the same tradeoff #1717's parallel machinery makes); the `Checkpoint` type is opaque, so switching to logs later is invisible to consumers. Gate: composition over chunk sizes 1, 7, 100, 2^18 and 1, 13, 1000, replayed in reverse order, row-for-row equal to the eager trace.
- **Safety** (AC11): the generated code mapping is asserted executable-and-not-writable by locating it in `/proc/self/maps` in the running process, and out-of-bounds loads and stores each surface the defined fault exit with the faulting address rather than a host segfault.
- **Slice 5, partial**: ALU rows take their second operand straight from the state plane rather than via a scratch register (one fewer instruction and register on the most common row shape); the chunk-boundary pause check is confined to chunk-capable bodies so eager traces do not pay for it (pausability is its own compile axis, four bodies at ~13ms each); AOT compile cost is reported as a tracked column. Register pinning into XMM lanes and fused fast-mode groups are implementable but deliberately deferred: they cannot be evaluated in this environment, and unmeasurable micro-optimization is not progress.
- Per-kind iai-callgrind microbenchmarks behind an `iai` feature, and `x86`/`x86_fast` Criterion ids in the jolt-eval `trace_gen` targets.
- Cross-engine baseline harness (`phase3_baseline`, ignored test). Provisional numbers, x86-64 Linux container under Rosetta, median of 3:

| guest | rows | ref serial (seam) | ref fast pass (execute mode) | AOT fast pass | ratio | AOT compile |
|---|---:|---:|---:|---:|---:|---:|
| fibonacci_400000 | 4.40M | 11.0 MHz | 77.3 MHz | **346.3 MHz** | 4.5x | 13ms |
| sha2_chain_4446 | 13.6M | 12.3 MHz | 131.8 MHz | **799.9 MHz** | 6.1x | 13ms |

**Treat these as indicative only.** Run-to-run variance on an unchanged binary reaches 30% here (the reference fast pass reads 60 to 79 MHz across runs), because the container translates x86 through Rosetta. The ratios are stable across runs; the absolute numbers are not. AC8/AC9 need a bare-metal measurement.

The untuned AOT fast pass is 5.4-6.1x the optimized interpreter's execute mode on the same platform: the two-pass pipeline's pass-1 ceiling is where AOT pays. Note the inline-heavy guest is AOT's *better* case, not its worse one — inline expansions are long straight-line runs of exactly the arithmetic rows the transpiler emits without dispatch. Gate-platform numbers (bare-metal linux-x86_64, where Rosetta translation no longer taxes the generated code) to follow; AC8/AC9 renegotiation is proposed on #1713.

Stacked on #1728 (slice 1); base is `perf/tracer-100mhz` for the cross-repo reason noted there. Review from commit `02c76a2ff` onward. Still to come: the remainder of slice 5 (register pinning, fused groups) and the AC8/AC9 gate measurements, both of which want native hardware; the x86 arch-test Makefile lane and skip list (needs generated ACT4 ELFs); literal AC7 (needs `byte_diff.rs`'s helpers extracted into a shared module). The blocker that dominates all of them: **nothing here has executed on real x86-64 silicon** — every result comes from a Rosetta-translated container. Record mode is deliberately de-emphasized per the measurements above — the fast pass is where AOT pays.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -212,6 +212,43 @@ jobs:
       - name: Serial/parallel trace equivalence
         run: cargo run --profile ci -p tracer --example trace_equivalence
 
+  x86-tracer:
+    runs-on: ubuntu-latest
+    name: x86 AOT tracer backend (native x86-64)
+    steps:
+      - uses: actions/checkout@v7.0.1
+      - uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          components: rust-src
+          cache: false
+      - uses: Swatinem/rust-cache@v2
+        with:
+          shared-key: release-tests
+          save-if: false
+      - name: Install nextest
+        uses: taiki-e/install-action@nextest
+      # Guest ELFs link ZeroOS, so building them needs its musl toolchain.
+      - name: Cache ZeroOS musl toolchain
+        id: cache-zeroos-musl
+        uses: actions/cache@v6
+        with:
+          path: ~/.zeroos/musl
+          key: zeroos-musl-toolchain-musl-1.2.3-gcc-9.4.0-Linux-x86_64
+      - name: Install ZeroOS musl toolchain
+        if: steps.cache-zeroos-musl.outputs.cache-hit != 'true'
+        run: |
+          curl -fsSL https://github.com/LayerZero-Labs/ZeroOS/releases/download/musl-toolchain-musl-1.2.3-gcc-9.4.0/zeroos-musl-toolchain-musl-1.2.3-gcc-9.4.0-Linux-x86_64.tar.gz -o /tmp/zeroos-musl.tar.gz
+          mkdir -p ~/.zeroos
+          tar -xzf /tmp/zeroos-musl.tar.gz -C ~/.zeroos
+      # The guest-dependent tests skip themselves without the CLI, so this
+      # install is what makes this job the backend's real coverage: it is the
+      # only place the differential and chunk-composition tests execute
+      # against native x86-64 hardware.
+      - name: Build and install jolt CLI
+        run: cargo install --path . --locked --force --profile ci --target-dir target
+      - name: Test jolt-tracer-x86
+        run: cargo nextest run --cargo-profile ci -p jolt-tracer-x86
+
   test-prover-zk:
     runs-on: ubuntu-latest
     name: Test jolt-prover-legacy (zk)
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3917,6 +3917,7 @@ dependencies = [
  "dynasm",
  "dynasmrt",
  "iai-callgrind",
+ "jolt-inlines-keccak256",
  "jolt-inlines-sha2",
  "jolt-platform",
  "jolt-program",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3916,6 +3916,7 @@ dependencies = [
  "common",
  "dynasm",
  "dynasmrt",
+ "iai-callgrind",
  "jolt-platform",
  "jolt-program",
  "jolt-riscv",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2291,6 +2291,33 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d0881ea181b1df73ff77ffaaf9c7544ecc11e82fba9b5f27b262a3c73a332555"
 
 [[package]]
+name = "dynasm"
+version = "3.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7f7d4c414c94bc830797115b8e5f434d58e7e80cb42ba88508c14bc6ea270625"
+dependencies = [
+ "bitflags 2.11.0",
+ "byteorder",
+ "lazy_static",
+ "proc-macro-error2",
+ "proc-macro2",
+ "quote",
+ "syn 2.0.119",
+]
+
+[[package]]
+name = "dynasmrt"
+version = "3.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "602f7458a3859195fb840e6e0cce5f4330dd9dfbfece0edaf31fe427af346f55"
+dependencies = [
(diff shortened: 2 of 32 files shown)


0xAndoroid (approved):
Looks great and clean. I don't quite understand why running in fast mode before the record mode makes it faster. I'm going to run it on a linux machine and check and then merge.

PR #1730, opened 2026-07-31
fix: address v12 security audit findings

0xAndoroid:
Dispositions for 44 findings from an external security audit: **31 fixed / 5 false-positive / 6 wontfix / 2 deferred** pending protocol decisions. This branch is the 31 fixes plus regression tests.

Notable real bugs:

- **secp256k1/p256 division inlines — aliasing tautology:** the advised quotient was stored to `rs3` before the `VirtualAssertEQ` checks ran; with `rs3 == rs1` the store overwrote the dividend and the checks compared the result against itself, accepting an arbitrary quotient. Stores now emitted after all checks; aliased-division regression tests added.
- **`unsafe_allocate_zero_vec` UB:** `alloc_zeroed` called with zero-size layouts (empty vec / ZST), and the all-zero-repr safety check was debug/test-only — fixed in both jolt-poly and the legacy copy; the repr check now runs in all builds.
- **sha2/keccak256 inline UB batch:** `&mut [u8]` buffer view held across `&mut self` calls (aliasing UB) and unaligned u64 absorption/digest stores — pointers now derived per unsafe block, rate buffers `repr(align(8))`, digest assembled without pointer casts.
- **Dory deserialization DoS:** proof byte-size cap, round-count and verifier-setup structure pre-validation before the upstream parser's `Vec::with_capacity` on attacker-controlled lengths, trailing-byte (canonicality) rejection, and commit-time SRS-capacity checks returning `Err` instead of panicking.
- **`double_and_add` 2P+Q=O:** the divisor-zero branch was gated behind a per-curve opt-in const (off for secp256k1) — now unconditional; reachable from `ecdsa_verify` as a host-side panic / unprovable trace.
- **GT/G2 untrusted-point hygiene:** exact-size, zero, and unitarity pre-filters ahead of the exact r-torsion check (rejects malformed input before the attacker-triggerable 254-bit exponentiation); on-curve-non-subgroup rejection pinned by tests.
- **MSM length equality** enforced in all builds (backends silently truncate to the shorter slice); **R1CS key/witness dimensional invariants** validated at construction and deserialization.

Gates: 2518/2518 tests; `cargo clippy --all --all-targets -- -D warnings` under `host` and `host,zk`; muldiv e2e 3/3 in both standard and ZK modes; two independent review passes over the batch.


markosg04 (approved):
**Approving — Dory changes only** (types.rs / scheme.rs / urs_lock.rs). Inlines/verifier/tracer need a separate stamp.

Fixes look right. Verified the `Validate::No` pre-scan is safe: the real parse re-validates every element (dory-pcs 0.4.1 `ark_serde.rs:435-437`), so subgroup checks run exactly once.

Two later follow ups:

1. Root cause is dory's parser trusting wire lengths (`Vec::with_capacity(num_rounds)`, `ark_serde.rs:446/471`; same for the setup's u64-prefixed `Vec<GT>`s). We should upstream the fix to dory itself — it's a couple of lines — and TODO these format-mirroring pre-scans away on the next bump.
2. Longer term dory-pcs should just expose serde like the rest of the stack, which kills this bug class at the wire layer, and run subgroup checks once at the verifier boundary instead of during parsing.

PR #1731, opened 2026-08-01
feat(akita): specialize prefix-packed commitments

markosg04:
## Summary

Akita can omit the digit-zero row of a one-hot polynomial when the memory activation is the public constant $M_\mu = 1$. This applies that specialization to instruction RA, bytecode RA, and the balanced-increment digit and carry columns. Their stage-7 claims use the paper's reconstructed coefficient $\widetilde{eq}(r_{address}, k_i) - \widetilde{eq}(r_{address}, 0)$.

RAM is intentionally outside this optimization. `RamRa` remains fully committed and keeps the base `RamHammingBooleanity`, Hamming-weight, Booleanity, and virtualization checks. A separate change can evaluate the note's nonconstant RAM activation construction.

## Changes

- Replace the generic packing path with protocol-owned fixed-prefix layouts and bind the resulting layout in the Akita digest.
- Add the mixed stage-7 reduction: two reconstructed legs for instruction/bytecode, three base legs for RAM, one Booleanity leg per increment column, then the fused-increment decode.
- Align the implementation notes and identifiers with the paper's digit-zero and $M_\mu$ notation; update Akita layouts and generated schedules for the new statement.

## Testing

Akita prover tests pass (457/457), Akita claims tests pass (187/187), and Akita verifier tests pass (82/82). Standard and ZK `muldiv` pass; both workspace clippy modes and the Akita verifier/prover clippy lanes are clean.

## Security Considerations

The omitted row is used only where activation is publicly one. When Booleanity and bytecode read-RAF openings share a runtime point, the clear verifier now requires their scalar claims to match before deduplicating the opening. RAM retains its existing committed-data one-hot check; this PR adds no prover-supplied or `Load + Store` RAM activation. Prover, verifier, transcript inventory, packing metadata, and tamper fixtures move together.

## Breaking Changes

Akita proof/setup compatibility and the packed-opening API change. The layout digest is now `jolt/akita/one_hot_trace/digit-zero-mu-one-full-ram/v7`.


quangvdao:
@markosg04 I reviewed the latest commit, 950d04f2c. I think the new binding argument identifies the right issue, but the Schwartz–Zippel step still has a gap.

The spec says that when the committed row sums are 0 or 1, the possible values of `Ã² - Ã` are exactly `{0}`. This is not true when every chunk has committed row sum `s_i = 1`. Booleanity and reconstruction then permit

```
Ã ∈ {1, 2}
Ã² - Ã ∈ {0, 2}
```

More generally, fixing a small set of possible values before `r₁` does not fix which value the prover selects after seeing `r₁`. If that selection remains adaptive, the prover is choosing `q_j` from a fixed set at every cycle while trying to satisfy

```
Σ_j eq(r₁, j) q_j = 0.
```

That is not evaluation of one fixed nonzero polynomial, so ordinary Schwartz–Zippel does not apply. Across many cycles there can be exponentially many selections.

I think there may be a way to complete the proof using RAF and Spartan. For each cycle, reconstruction permits at most two consecutive activation values. Moving to the larger value activates the digit-zero row in every chunk and strictly increases the RAF sum because every `unmap(k)` is positive and the integer bound prevents field wraparound. If `RamAddress` is already fixed by stage-0-bound data, RAF can select at most one activation value before `r₁`. Then `Ã² - Ã` is a fixed table and the Schwartz–Zippel step works. The proof needs to establish that uniqueness and the fact that `RamAddress` is fixed before `r₁`.

I do not yet have a forged proof, so I am treating this as a soundness proof gap rather than a confirmed protocol break. As your commit notes, the same adaptivity question may also apply to the previous virtual `A² = A` gadget.

The original concern about algebraically shifting the stage-7 linear combination looks resolved by transcript order. The stage-6b claims are absorbed before stage 7 draws `γ`, so the prover cannot choose coordinated claim shifts after learning `γ`. The remaining question is whether the activation polynomial itself is fixed early enough for the zero-check.

moodlezoup (approved):

PR #1732, opened 2026-08-01
perf(prover): port optimized Akita to the modular prover

markosg04:
## Summary

The modular prover lacked an Akita path and materialized trace-sized data that the packed backend can derive or stream. Stacked on #1731, this ports the optimized kernels, adds Akita proving and reconstruction, streams packed commitments, compacts witness storage, and releases large tables after their final use.

The digit-zero scope matches #1731: instruction RA, bytecode RA, and balanced increment digit/carry columns use the public `M_mu = 1` optimization. RAM retains the base three-leg Akita protocol and explicitly commits active zero digits.

## Performance

These are same-commit `sha2-chain` proving results on an M4 Max with 128 GiB RAM. Both backends use the optimized modular prover; each observation ran in a fresh process after a warm-up, and all 18 timed proofs verified. Akita is pinned to `1d48114e6f0aefd1384a57be167037c57de99d22`, the current upstream `main`.

| Padded trace | Dory prove | Akita prove | Dory / Akita | Dory peak RSS | Akita peak RSS |
|---:|---:|---:|---:|---:|---:|
| 2^20 | 3.84 s | 1.42 s | 2.70x | 1.11 GiB | 0.93 GiB |
| 2^21 | 5.59 s | 2.28 s | 2.45x | 2.02 GiB | 1.69 GiB |
| 2^22 | 10.02 s | 4.23 s | 2.37x | 4.72 GiB | 3.09 GiB |
| 2^23 | 17.23 s | 9.00 s | 1.91x | 9.00 GiB | 5.96 GiB |
| 2^24 | 32.03 s | 17.38 s | 1.84x | 12.40 GiB | 10.47 GiB |
| 2^25 | 56.93 s | 26.52 s | 2.15x | 28.45 GiB | 17.20 GiB |
| 2^26 | 99.41 s | 39.23 s | 2.53x | 43.47 GiB | 29.84 GiB |
| 2^27 | 171.00 s | 96.92 s | 1.76x | 66.24 GiB | 53.35 GiB |
| 2^28 | 314.27 s | 146.24 s | 2.15x | 86.68 GiB | 78.43 GiB |

Akita is 1.76x to 2.70x faster and uses less peak RSS at every measured scale. At 2^28 it reaches about 1.84 million trace cycles/s, saves 168.03 seconds, and uses 8.25 GiB less peak RSS than Dory. These are single observations rather than confidence intervals; the [full results and method](https://github.com/a16z/jolt/blob/120152d95a59ad71b689317ac1714a13a7b9c07a/specs/akita-pr-1732-benchmarks.md) include the prior-revision comparison and adaptive schedule selections. The full benchmark sweep was run on `098e2ddc6285bebcb992192ccc308b7d17ad9cc0`; it has not been rerun after the subsequent integration and cleanup commits.

## Changes

- Add the modular Akita prover path and shared Dory/Akita stage recipes.
- Apply digit-zero virtualization to instruction RA, bytecode RA, and increment columns while leaving RAM on the base protocol.
- Stream packed commitment/opening traversal and reduce peak witness lifetime.

## Reviewer guide

#1731 already landed the protocol and verifier changes. Against current main, this PR has no effective diff in `jolt-verifier`, `jolt-claims`, `jolt-openings`, `jolt-sumcheck`, `jolt-poly`, `jolt-r1cs`, `jolt-transcript`, `jolt-crypto`, `jolt-dory`, or `jolt-lookup-tables`.

The Akita prover integration is concentrated in 404 changed lines:

- `crates/jolt-prover/src/akita/witness.rs` assembles the packed semantic columns and RAM activation mask.
- `crates/jolt-prover/src/akita/stage0.rs` commits that source through Akita and releases reconstructible NTT state.
- `crates/jolt-prover/src/akita/{mod.rs,prover.rs}` selects the optimized backend and threads its trait bound.

The compact trace handoff is separate: `tracer/src/execution_backend.rs` builds the proof rows directly, `crates/jolt-prover/src/config.rs` derives the config from them, and `crates/jolt-witness/src/backend/trace/mod.rs` retains the same `Arc`.

The rest is implementation support: the packed Akita kernels, compact witness rows, field accumulators, and differential tests. Cleanup reduced the non-kernel/non-prover support diff from 1,789 to 1,360 changed lines.

## Testing

- [x] Optimized-kernel, Akita/Dory parity, verifier-fixture, legacy host/ZK, and modular end-to-end tests pass.
- [x] `cargo clippy` and `cargo fmt` pass.

## Security Considerations

The protocol and verifier changes are covered in #1731. Differential tests compare optimized kernels with the reference backend, and end-to-end tests verify complete proofs. RAM digit-zero virtualization deliberately remains out of scope.

## Breaking Changes

Inherits the Akita proof/setup changes from #1731. The modular prover's internal backend and stage APIs change to support Akita and streamed witness sources.


crates/jolt-witness/src/backend/trace/mod.rs
```diff
@@ -117,15 +125,60 @@ impl<T: TraceSource> ProgramSource for TraceBackend<T> {
...
+            trace: mut source,
+            device,
+            final_memory,
+            advice_tape,
+        } = inputs.trace;
+        let mut trace_rows = Vec::new();
+        let mut trailing_padding = 0;
+        #[cfg(feature = "field-inline")]
+        let mut raw_rows = Vec::new();
+        while let Some(row) = source.next_row() {
```
0xAndoroid, line 179: This runs a second full pass over the trace and materializes another O(T) row vector. At production sizes, that adds O(T) work and a transient second trace-sized allocation. We should construct or share the compact rows in the original trace-production path instead of walking the trace again here.

markosg04: should be addressed with 3c62478647c44a11dfcf1c8cc22397fdd06a56f6

crates/jolt-kernels/src/optimized/ram_ra_virtualization.rs
```diff
@@ -163,7 +183,7 @@ impl<F: JoltField> RamRaVirtualizationKernel<F> {
         let num_committed = self.folded_ra.num_polys();
         let points = num_committed + 2;
 
-        let mut q_evals = self.gruen.par_fold_out_in(
+        let q_evals = self.gruen.par_fold_out_in(
```
0xAndoroid: **[1/15] `ram_k = 1` panics with index-out-of-bounds; guard and regression test deleted**

With `ram_k = 1` (`log_k = 0`, `num_committed_ra_polys() = 0` — a shape jolt-claims' `supports_zero_bytecode_and_ram_d` test explicitly supports), `prepare()` passes and this fold executes `let mut product = evals[0]` (line 201) on an empty vec → panic. `origin/main` had an explicit `num_committed == 0` branch whose comment said verbatim "the round loop used to index out of bounds here (evals[0] of an empty vector)", plus the regression test `zero_committed_chunks_prove_in_parity_and_fail_closed` (fixture `ram_k: 1`) — both deleted. The reference kernel returns a clean `StageClaimPublicInputFailed` for the same input, so the two backends now diverge. Not reachable from an e2e guest proof (min ram_k >= 2), but in-contract for the kernel API.

0xAndoroid (commented):
## Adversarial review (recall mode)

Reviewed against the merge-base `origin/main` (6ff8921da). Process: 11 independent finder passes over the full diff, one verification pass per surviving candidate, one gap sweep. The 15 findings are posted as inline comments on their exact lines, ranked [1/15] (most severe) to [15/15].

**Refuted during verification** (so nobody re-litigates them): the `1u64 << column` digit-zero-mask overflow is blocked by the plan's 64-column capacity check running first (max index 63); the `TraceBackend::new` panic path never reaches production provers — all `new` call sites are test/profiling-gated, production goes through `try_new`.

**Cut for the 15-finding cap, one line each**: deleted `optimized/support.rs` helpers (`bind_pairs`, `fmadd_u64_split` incl. its overflow-bound comment, `gamma_power_pairs`, `map_indices`, `FirstErrorLatch`) pasted verbatim into up to 20 kernel files; `spartan_outer.rs` forks the still-living `BundleStore` as `RowsStore`; `trace_onehot.rs` carries a dead `Arc<RwLock<Option<...>>>` release layer with no writer plus an undocumented `JOLT_AKITA_DECOMPOSE_MODE` env override; `specs/akita-pr-1732-benchmarks.md` is a dated one-off benchmark log whose documented command targets a retired harness.

crates/jolt-kernels/src/optimized/registers_read_write.rs
```diff
@@ -0,0 +1,1972 @@
...
+    pub rs2: Option<(u8, u64)>,
+    /// `(register, pre-write value, post-write value)`.
+    pub rd: Option<(u8, u64, u64)>,
+}
+
+impl WitnessBundle for RegisterCycleRow {
+    // `TraceRow` is nameable from this crate only through the doc-hidden
+    // re-export the bundle derive uses; jolt-kernels deliberately has no
+    // jolt-program dependency.
+    fn from_row(
```
0xAndoroid: **[2/15] Register-domain guard dropped: ids 128–254 panic deep in bind()**

A trace row with register id 128..=254 is accepted by jolt-riscv's `checked_register_id` (`MAX_REGISTER_ID = 254`, documented as "a storage-format detail, not a protocol fact") and unvalidated by `TraceBackend::try_new` (the documented constructor for untrusted traces). `from_row` maps it straight through; the cycle-to-address transition then indexes `ra[entry.col as usize]` into a 2^7-length array → panic (lines ~1348/1356). The deleted `rows.rs::from_row` returned `InvalidWitnessData` ("register index {register} exceeds 7-bit register read-write domain"); the test `collect_rejects_out_of_domain_register_indices` was deleted with it. The reference tier (`jolt-witness/src/backend/trace/registers.rs:36-64`) still rejects with the identical error — another backend divergence.

crates/jolt-kernels/src/optimized/registers_read_write.rs
```diff
@@ -0,0 +1,1972 @@
...
+        rd_indices,
+    })
+}
+
+/// The entry count [`cycle_entries`] will produce for one cycle — the
+/// counting pass's cheap twin (kept adjacent so the merge rules stay in
+/// sync: rs2 folds into rs1's entry, rd into either read's).
+#[cfg(feature = "parallel")]
+fn cycle_entry_count(cycle: &RegisterCycleRow) -> usize {
+    let mut count = 0usize;
```
0xAndoroid: **[3/15] `u8::MAX` sentinel reintroduces a previously fixed, regression-tested bug**

`cycle_entry_count` uses `cols = [u8::MAX; 2]` as a sentinel. A `RegisterCycleRow` with rs2/rd carrying raw index 255 is counted as absent while `cycle_entries` emits an entry → the pass-2 scatter writes past its window (`window[written].write`) next to a `set_len` on partially-initialized memory. The old `sparse.rs` used Option-typed `entry_count` with the comment "no sentinel value can collide with a raw index", and the deleted test `cycle_entry_count_matches_cycle_entries` swept `Some(255)` naming this exact collision. Shielded today only by jolt-riscv's `REGISTER_NONE = 255` encoding one crate away — an invariant whose pinning test no longer exists.

crates/jolt-kernels/src/optimized/ram_trace.rs
```diff
@@ -93,35 +248,47 @@ impl RamAccessColumns {
...
-        // wrong RA claim from a prefix-covering table, so hard-error like
-        // the `PcRow::shared` twin instead of a release-compiled-out assert.
-        if columns.addresses.len() != 1usize << log_t {
-            return Err(KernelError::TableSizeMismatch {
-                table: "session-shared RAM access columns".to_owned(),
-                expected: 1usize << log_t,
-                got: columns.addresses.len(),
-            });
-        }
+        debug_assert_eq!(
```
0xAndoroid: **[4/15] Hard `TableSizeMismatch` error downgraded to `debug_assert_eq!`**

The deleted comment prohibited exactly this: "Five kernels across stages 2–6b reclaim these columns by type alone; a wrong-domain reclaim means OOB indexing or a silently wrong RA claim from a prefix-covering table, so hard-error … instead of a release-compiled-out assert." In release, a future caller reclaiming the TypeId-keyed session carry with a different `log_t` gets a shorter table (OOB panic mid-round) or a longer one (silently wrong RAM RA/val claim, surfacing only as an opaque final-round mismatch). The sibling carries (`pc_rows` in bytecode_read_raf.rs:165, `shared_instruction_rows`) kept their release-mode length guards, so the twin seams are now inconsistent.

crates/jolt-kernels/src/optimized/commitment.rs
```diff
@@ -22,68 +22,42 @@
...
 
 impl<F, PCS> CommitWitness<F, PCS> for OptimizedBackend
 where
     F: JoltField,
     PCS: CommitmentScheme<Field = F> + ModeStreamingCommitment,
 {
-    // The backend-neutral `commit_witness` span lives at the stage-0 call
-    // boundary (see the reference kernel) — no per-impl span, or the label
-    // would double-count under this backend.
+    #[tracing::instrument(
```
0xAndoroid: **[5/15] `commit_witness` span double-counts in the versioned taxonomy**

This per-impl `#[tracing::instrument(name = "commit_witness")]` sits inside the stage-0 seam span of the same name (`crates/jolt-prover/src/dory/stages/stage0.rs:198`). jolt-profiling aggregates `total_ns` per label across instances, so every `profile --backend optimized` run reports count = 2 and ~2x wall time for `commit_witness` in summary.json, corrupting the canonical jq queries and any `telemetry:*` objective on this label. The branch deleted the WARNING comment stating "no per-impl span, or the label would double-count under this backend"; `reference/commitment.rs:43` still honors the contract. Span labels are a versioned schema per CLAUDE.md, and the smoke test checks only label presence, so this ships silently.

crates/jolt-kernels/src/optimized/lazy_ra.rs
```diff
@@ -48,18 +48,24 @@ pub(crate) trait ChunkIndexSource: Send + Sync {
...
+    }
 }
 
-/// `N` address-folded selector columns bound `LowToHigh`, lazily until the
-/// fourth bind materializes dense.
+/// `N` address-folded selector columns bound `LowToHigh`, lazily for the
+/// first three binds.
 pub(crate) enum LazyFoldedRa<F: JoltField, S> {
-    /// Fewer than four binds: per-polynomial branch scale tables (the base
+    /// Fewer than three binds: per-polynomial branch scale tables (the base
```
0xAndoroid: **[6/15] Lazy-RA docs rewritten to a three-binds/T-8 model the code does not implement**

`origin/main`'s docs said "index-encoded for the first four binds, dense at `T/16` after" (booleanity.rs:34). The branch rewrote six sites to a three-binds/`T/8` model (booleanity.rs:34, ram_val_check.rs:12, this line "Fewer than three binds", instruction_ra_virtualization.rs:142, ram_ra_virtualization.rs:141, bytecode_read_raf.rs:721) with zero change to `LazyFoldedRa::bind` — `width < 8` still materializes at the fourth bind, at `T/16`. Either an intended earlier-materialization change never landed, or six docs are wrong by 2x. Concrete damage already landed: the collapsed booleanity parity test at `log_t = 4` was justified by the three-bind model ("three staged binds", booleanity.rs:1279), but under the real fourth-bind machine the materializing bind arrives only at finish — no remaining test exercises any dense-state round message.

crates/jolt-kernels/src/optimized/instruction_read_raf.rs
```diff
@@ -356,21 +345,11 @@ impl<F: JoltField> PrepareKernel<F, InstructionReadRaf<F>> for OptimizedInstruct
...
-        // first cycle bind, and a strong session carry would keep the
-        // 48 B × T rows resident through the stage-5 staging peak — later
-        // stages re-derive index-parallel instead. Re-emulating sources keep
-        // the strong carry (re-deriving means a full re-emulation walk).
-        if witness.rows().is_some() {
-            session.park(SharedInstructionRowsWeak(Arc::downgrade(&rows)));
-        } else {
-            session.park(SharedInstructionRows(Arc::clone(&rows)));
-        }
+        session.park(SharedInstructionRows(Arc::clone(&rows)));
```
0xAndoroid: **[7/15] Unconditional strong `SharedInstructionRows` park: ~670 MB extra residency at 2^24**

`prepare()` now parks a strong session carry where `origin/main` parked weak for slice-backed witnesses; the deleted comment: "a strong session carry would keep the 48 B x T rows resident through the stage-5 staging peak — later stages re-derive index-parallel instead." The kernel drops its own Arc at the first cycle bind, but the parked strong copy survives until the first stage-6 consumer takes it. Production witnesses are always slice-backed after this branch (`TraceBackend::random_access` always `Some`), so the weak path is dead and the row vector (40 B x T, ~670 MB at T = 2^24) stays resident through the stage-5 cycle-round tail. Introduced as unmentioned collateral of the bucket-refactor commit; `shared_instruction_rows` (lines ~311-326) still parks weak — the asymmetry suggests this was not a deliberate policy change.

crates/jolt-witness/src/backend/trace/mod.rs
```diff
@@ -117,15 +125,82 @@ impl<T: TraceSource> ProgramSource for TraceBackend<T> {
...
+    ) -> Self {
+        let TraceOutput {
+            trace,
+            device,
+            final_memory,
+            advice_tape,
+        } = inputs.trace;
         Self {
             config,
             program: inputs.program,
```
0xAndoroid, line 151: **[8/15] `TraceBackend::try_new` eagerly materializes the full trace, defeating streaming sources**

`while let Some(row) = source.next_row()` drains every `TraceSource` into an owned full-trace Vec at construction — a single-threaded per-row compaction pass (`JoltInstruction::try_from` + PC lookup) ahead of the parallel prover, plus a second full raw-`TraceRow` copy retained under the field-inline feature even for programs with no field-inline instructions. A chunked re-emulating source at 2^28 cycles now costs multi-GB residency where `origin/main` streamed chunk-sized buffers; the crate doc still advertises checkpointed sources. Side effect: `random_access()` now always returns `Some`, so the dory_byte_diff `chunk_boundary` "forced-walk arm" (tests/dory_byte_diff.rs:2143) silently exercises the same random-access path as the other arm while its comment claims it covers the sequential `visit_chunks`/`commit_streamed` fallback.

crates/jolt-kernels/src/optimized/commitment.rs
```diff
@@ -22,68 +22,42 @@
...
-/// threads.
-#[cfg(feature = "parallel")]
-const SUPERCHUNK_CYCLES_MAX: usize = 1 << 21;
+/// Cycles per superchunk. The dominant stage-0 wall cost on a many-core
+/// host is the per-superchunk join (its critical path is one window's
+/// serial MSM), so fewer, larger superchunks win: 2^17 → 2^21 measured
+/// 59.3s → 41.1s whole-prove at 2^25 on 64 threads. The extracted bundle
+/// buffer (64 B/cycle, two reused buffers) is the only staging that scales
+/// with the superchunk — columns feed the commit windows by closure.
+const SUPERCHUNK_CYCLES: usize = 1 << 21;
```
0xAndoroid: **[9/15] Fixed `SUPERCHUNK_CYCLES = 2^21` replaces the pool-scaled clamp**

Stage-0 streaming commit now stages two reused `CommittedColumnsWitness` buffers of 2^21 x 64 B (~268 MB total) regardless of core count. The deleted comments called exactly this "a needless reservation on a 10-core host" and noted "serial builds have no join to amortize; the floor minimizes staging" (old floor 2^17, ~16 MB). Superchunk size is wire-invisible per the module's own test, so nothing forces the fixed value; pure peak-RSS regression on small/serial hosts.

crates/jolt-kernels/src/optimized/bytecode_read_raf.rs
```diff
@@ -276,30 +314,82 @@ impl<F: JoltField> PrepareKernel<F, BytecodeReadRafAddressPhase<F>>
...
+            stage_cycle_points,
+            &rows,
+            addresses,
+            push_pc,
+            |_| F::one(),
+        );
+        #[cfg(feature = "akita")]
+        let pushforwards = {
+            let mut pushforwards = pushforwards;
+            pushforwards.extend(stage_pushforwards::<F, _, true>(
```
0xAndoroid: **[10/15] Akita address phase walks the trace twice for pushforwards**

Under `akita`, lines 348 and 358 each run `stage_pushforwards` as a complete O(T) walk over the same `Arc<Vec<InstructionCycleRow>>` (duplicate PC decode, seen/touched bookkeeping, per-block zeroing/fold). The module doc (lines 10-14) claims the four fused pushforwards "use the same walk", and the legacy prover it cites does one loop serving all 9 stages. Cost: roughly 15-35% of the stage-6a pushforward phase on the packed path.

crates/jolt-witness/src/consumer.rs
```diff
@@ -136,12 +134,50 @@ pub trait RowSource {
...
+#[derive(Clone)]
+pub struct RandomAccessRows {
+    rows: std::sync::Arc<Vec<TraceRow>>,
+    cycles: usize,
+    preprocessing: std::sync::Arc<jolt_program::preprocess::JoltProgramPreprocessing>,
+    padding: TraceRow,
+}
+
+impl RandomAccessRows {
+    pub(crate) fn new(
```
0xAndoroid, line 155: **[11/15] Over-long physical trace is silently truncated instead of rejected**

The relocated `RandomAccessRows` dropped the deleted rows.rs check `rows.len() > padded cycle domain` ("physical trace has N rows but the cycle domain has {cycles}"). With a mis-derived config (`2^log_t < trace.len()`), `window()` now serves the first `2^log_t` rows and the prover spends a full run producing a proof of a truncated execution that the verifier rejects at the output checks. The pinning test `view_collection_matches_the_chunked_walk` was deleted; no replacement check exists in `RandomAccessRows::new` or `TraceBackend`.

crates/jolt-prover/src/akita/witness.rs
```diff
@@ -282,8 +381,8 @@ where
             reason: "advice bytes exceed the configured maximum advice size",
         });
     }
-    let words = max_advice_bytes / 8;
-    let word_vars = words.next_power_of_two().max(1).ilog2() as usize;
+    let words = (max_advice_bytes / 8).next_power_of_two().max(1);
```
0xAndoroid: **[12/15] `commit_advice_one_hot` diverges from the legacy byte-parity oracle at non-pow2 shapes**

`let words = (max_advice_bytes / 8).next_power_of_two().max(1);` then `for word_index in 0..words` — padding words committed hot-at-byte-0. Legacy oracle (`jolt-prover-legacy/src/zkvm/packed.rs:481,496`): `vec![0u64; max_advice_bytes / 8]` iterated raw; `origin/main`'s modular prover matched it. For `max_advice_bytes` whose /8 is not a power of two — or < 8 bytes, where legacy commits zero positions and this code commits `WORD_BYTES` rows — the committed advice one-hot polynomial differs. Latent in production (`MemoryLayout::new` asserts pow2-or-0 advice sizes) but live for literally constructed layouts, which in-repo fixtures use; the akita byte-diff ratchets only exercise pow2 shapes, so the divergence ships unobserved.

crates/jolt-kernels/src/optimized/booleanity.rs
```diff
@@ -93,63 +95,121 @@ enum ColumnSelector {
...
         }
     }
 }
 
+struct BooleanityColumns {
+    openings: Vec<JoltOpeningId>,
+    selectors: Vec<ColumnSelector>,
+}
+
+fn booleanity_openings<F: JoltField>(
```
0xAndoroid: **[13/15] Booleanity opening-set guard replaced by compile-time cfg selection**

The runtime fail-closed comparison of the relation's symbolic leaf openings against the served opening set was deleted; `booleanity_openings` now picks the set purely by `cfg(feature = "akita")`. An akita-compiled jolt-kernels asked to prove a Booleanity relation whose summand is not the lattice variant silently serves the wrong opening set — wrong round polynomials with no error at prepare — where `origin/main` returned `KernelError::Unsupported`. Masked today only because the akita feature compiles the dense pipeline out.

crates/jolt-prover/src/akita/witness.rs
```diff
@@ -129,131 +133,227 @@ struct OneHotTraceSourceRow {
...
+                }
             }
-        }));
+            OneHotTraceColumn::Ram(selector) => row
+                .ram_address
+                .0
+                .map_or(0, |address| selector.chunk_usize(address as usize)),
+            OneHotTraceColumn::Increment(column) => row.fused_inc.selected_row(*column),
+        };
+        debug_assert!(row_index <= u8::MAX as usize);
```
0xAndoroid, line 231: **[14/15] Checked `hot_u8` conversion replaced by `debug_assert` + `as u8`**

`debug_assert!(row_index <= u8::MAX as usize); *selected_row = row_index as u8;` — a selector domain > 255 (`log_k_chunk > 8`) wraps 256 to 0 in release, which the digit-zero mask reads as "omitted", committing a wrong `OneHotTrace` with no error. `origin/main` returned `ProverError::InvariantViolation` for every such value. Currently unreachable only because jolt-claims' plan constructor rejects `log_k_chunk` outside {4, 8}; the fail-closed check moved from the point of fault to a distant crate invariant, and the public `TraceOneHotCommitment::commit_trace_one_hot` seam lacks stage0's `one_hot_k == 1 << log_k_chunk` guard.

crates/jolt-prover/tests/akita_byte_diff.rs
```diff
@@ -336,18 +329,17 @@ mod muldiv {
...
+            JoltVmWitnessInputs::new(&jolt_program, &program_preprocessing, trace_output),
         );
         let prover_preprocessing = JoltProverPreprocessing::<AkitaScheme, AkitaVc> {
             verifier: verifier_preprocessing,
             pcs_setup: object_setup,
             committed_program: None,
         };
 
-        let backend = akita::JoltAkitaBackend::reference();
+        let backend = akita::JoltAkitaBackend::optimized();
```
0xAndoroid: **[15/15] Acceptance-suite failure signals narrowed**

The muldiv byte-diff ratchet and all akita_e2e proves switched reference to optimized backend, so a regression in the reference packed kernels on the widest-coverage workload no longer fails CI (only the advice/committed ratchets at lines 481/617 still pin reference). The booleanity cycle parity tests collapsed from {log_t 4, 5} to log_t 4 only (see finding 6: zero dense-round coverage remains). The `stage_pushforwards_membership_is_value_independent` test — the only probe against the historical double-count bug its WARNING comment described — was deleted just as the helper went generic with legitimately-zero fused-inc weights. The e2e also hardcodes `auxiliary.len() == 2` (akita_e2e.rs:519, 772) instead of the plan-derived invariant.

0xAndoroid:
This is just a repost of a claude review, implement at your own descretion

crates/jolt-akita/src/trace_onehot.rs
```diff
@@ -0,0 +1,2823 @@
...
+        assert_eq!(sparse, materialized);
+        assert_eq!(compact, materialized);
+    }
+
+    #[test]
+    fn d64_k256_2e28_dense_rotations_fit_the_table_budget() {
+        let table_bytes = 1024 * 29 * 64 * std::mem::size_of::<[i16; 64]>();
+        assert_eq!(table_bytes, 243_269_632);
+        assert!(table_bytes <= ROTATED_CHALLENGE_TABLE_BUDGET);
+    }
```
0xAndoroid: Delete this test. It hand-recomputes the worst-case dense table size (`1024 * 29 * 64 * size_of::<[i16; 64]>()`) instead of calling the production sizing used by `prepare_rotations` (the `dense_bytes` computation around line 1666), so if that formula changes this stays green while the budget invariant it claims to guard drifts. The `assert_eq!(table_bytes, 243_269_632)` half is a pure arithmetic tautology. If the D=64/K=256/2^28-fits-the-budget invariant is worth pinning, derive `table_bytes` from the same code path production uses; otherwise drop the test.

0xAndoroid:
Please restore RoundProgress across all affected optimized sumcheck kernels instead of open-coding rounds and rounds_bound. RoundProgress was added on main in af32f2e34 specifically to centralize advance, total, and require_complete; the feature branch replacement appears to be rebase-integration fallout. The current code duplicates the same invariant across roughly 17 files and 104 rounds_bound occurrences, with no Akita-specific semantic or performance difference. Keep the Akita-specific relation and column changes, but adapt them to RoundProgress.

crates/jolt-akita/src/adapters.rs
```diff
@@ -184,6 +186,23 @@ impl AkitaProverSetup {
         self.verifier.one_hot_k
     }
 
+    /// Releases transformed setup slots after the trace commitment. Later
+    /// opening work rebuilds the slots on first use.
+    pub fn release_post_commit_ntt_residency(&self) -> Result<(), OpeningsError> {
```
moodlezoup: note to self: check this

crates/jolt-kernels/src/optimized/parity.rs
```diff
@@ -59,10 +58,11 @@ pub(crate) fn probe_input_claim<F: JoltField, R>(
...
 where
-    R: ConcreteSumcheck<F>,
-    SumcheckInputClaims<F, R>: InputClaims<F>,
-    SumcheckOutputClaims<F, R>: OutputClaims<F>,
-    ConcreteSumcheckChallenges<F, R>: SumcheckChallenges<F, JoltChallengeId>,
+    R: jolt_verifier::stages::relations::ConcreteSumcheck<F>,
+    jolt_verifier::stages::relations::SumcheckInputClaims<F, R>: jolt_claims::InputClaims<F>,
+    jolt_verifier::stages::relations::SumcheckOutputClaims<F, R>: jolt_claims::OutputClaims<F>,
+    jolt_verifier::stages::relations::ConcreteSumcheckChallenges<F, R>:
+        jolt_claims::SumcheckChallenges<F, jolt_claims::protocols::jolt::JoltChallengeId>,
```
moodlezoup: nit: `use` instead of these fully qualified paths

crates/jolt-witness/src/backend/trace/mod.rs
0xAndoroid: Addressed in a0a9f7221: raw normalization now requires a slice-backed TraceSource and rejects iterator/replaying sources instead of draining and retaining them. Replaying producers must emit compact rows at the producer boundary and use from_compact; the false forced-walk test was removed.

crates/jolt-kernels/src/optimized/bytecode_read_raf.rs
0xAndoroid: Addressed in a0a9f7221: base and fused-increment pushforwards now share one row traversal. Added direct combined-vs-separate coverage plus K16/K256 reference parity.

crates/jolt-prover/src/akita/witness.rs
0xAndoroid: Addressed in a0a9f7221: commit_advice_one_hot now rejects zero, sub-word, and non-power-of-two maximum sizes before constructing the object. Added focused shape-rejection coverage.

crates/jolt-kernels/src/optimized/booleanity.rs
0xAndoroid: Rechecked against the current types: under akita, Booleanity selects the lattice symbolic relation and dimensions at compile time, so the base cycle variant cannot reach this kernel. Both the symbolic relation and kernel derive openings from the same lattice helper. Resolving as a false positive.

crates/jolt-prover/src/akita/witness.rs
0xAndoroid: Addressed in a0a9f7221: assembly now validates the supported 4/8-bit chunk width and checks the plan logical arity before the hot loop, keeping the per-row cast branch-free and proven to fit u8. Added mismatch coverage.

0xAndoroid:
TODO (follow-up): move the generic packed row-source commitment/opening machinery in `crates/jolt-akita/src/trace_onehot.rs` into the upstream Akita repository. Jolt should retain only its protocol-specific column extraction and digit-zero/activation mapping; the generic commit, opening-fold, tensor, rotation, and decomposition kernels should live behind an Akita row-source interface.

0xAndoroid (approved):
LGTM

crates/jolt-kernels/src/optimized/parity.rs
moodlezoup: nit: `use` instead of fully-qualified paths

PR #1737, opened 2026-08-05
perf(preprocessing): parallelize CPU setup passes

0xAndoroid:
## What changed

- Convert `Cycle` → `TraceRow` with Rayon above a 2^14-row threshold; small traces stay serial. Error capture stays out of band so indexed collection writes directly into one output allocation while preserving ordering, error propagation, and panic behavior.
- Run `ProverConfig::derive`'s touched-address reduction through Rayon under the existing default `parallel` feature; small/no-default builds stay serial.
- Use the released crates.io `dory-pcs 0.4.1`. Companion a16z/dory#30 carries the prepared-point cache portion of this optimization; Jolt no longer pins an unreleased repository commit.
- Remove all campaign env selectors. These are unconditional defaults on the plain CPU prover path.

## CPU measurements

Originating locked 2^26 campaign pair, same proof bytes:

| preprocessing block | serial | parallel |
|---|---:|---:|
| `Cycle` → `TraceRow` | 7.54 s | 0.68–0.80 s |
| `ProverConfig::derive` | 0.50 s | 0.04 s |
| end-to-end elapsed | 26 s | 17 s |
| prove span | 4.24 s | 4.30 s (noise) |

Prepared-point initialization (companion a16z/dory#30):

| measurement | serial | parallel |
|---|---:|---:|
| campaign 2^26 prep tail | 16.2 s | 502 ms |
| Apple M4, `log_n=34`, 131,072 points/group | 6.449 s | 1.071 s median |

The campaign's 2^28 traces put serial conversion at ~30–43 s and serial prepared-point initialization above 30 s.

## Scope

Excluded W7 `PAD_INTERLEAVE`: it is implemented inside the CUDA trace-store/store-early pipeline and has no standalone plain-CPU path to upstream here.

## Verification

- workspace: 2508/2508
- legacy muldiv, host: 3/3
- legacy muldiv, host+zk: 3/3
- modular fixtures: 18/18
- modular fixtures+zk: 12/12
- clippy `-D warnings`: host and host+zk
- rustfmt; taplo; no-default-features check

Authored with Claude Code.


a16z-bsaari (approved):
test

markosg04 (approved):

PR #1745, opened 2026-08-06
refactor: complete the Akita shared-field cutover

Acentelles:
## Summary

**DRAFT, stacked on #1684 and on Akita #307; cannot land before both merge.** Until #1684 merges, this PR's diff against `main` includes #1684's commits; the commits belonging to THIS PR are the six from the bootstrap-edge deletion onward (`0516c2b8..6b5d3ff0`), equivalently the diff against `Acentelles:feat/solinas-field-stack`. Every stacked pin carries a `WARNING` comment and must be re-pointed to the LayerZero-Labs/akita merge revision of #307 before landing.

This PR discharges every "removed in the final migration PR" promise left in the tree by #1684 (the shared-field landing) and Akita #307 (the Akita-side cutover). Post-cutover, Akita's field type IS `jolt_field::Fp128`, so the bootstrap machinery is not replaced — it is deleted as meaningless.

## Changes

- **Delete the akita bootstrap edge from `jolt-field`**: `src/akita.rs`, the `akita` feature, and the optional `akita-config`/`akita-field` dependencies.
- **Restore `JoltField`'s serde bounds** (`Field + CanonicalEncoding + WithAccumulator + Serialize + DeserializeOwned`), reverting the recorded deviation that existed only because the orphan rule barred serde impls for the foreign bootstrap type.
- **Re-point the akita dependencies to the post-cutover API** (Akita #307 head) and adapt `jolt-akita` plus the legacy packed prover path; `akita-field` disappears from the dependency graph entirely.
- **Single-identity wiring**: a root `[patch]` table redirects Akita's `jolt-field` git dependency onto the workspace member, so exactly one `jolt-field` package identity resolves; the shared-field identity gate flips to reject every `akita-field` identity (structural `cargo metadata` check, color-independent).
- **Jolt-owned schedule catalogs regenerated and extended.** Post-cutover Akita resolves runtime schedules from generated catalogs only (the planner DP is offline-only), and upstream's generated tables are gitignored bootstrap artifacts a Git-dependency checkout does not contain. The catalogs now cover the production grids plus an explicit test-support grid (vars 12–16, 1–4 polynomials) and the forced-K256 e2e band (20 vars); test fixtures moved off shapes that are no longer schedulable, with dense-only/one-hot-only setups so neither flavor is built at unsupported shapes.
- **Per-field scalar-challenge convention.** The legacy transcripts hardcoded the digest-bytes-reversed (big-endian) scalar-challenge decode — BN254's convention — while the modular verifier decodes with the field's own `from_scalar_challenge_bytes`, unreversed little-endian for the shared Solinas `Fp128`. Pre-cutover the bootstrap adapter carried the reversal, so both sides agreed; the cutover silently flipped the verifier side. Both decodes consume the same 16 bytes, so the transcript stream stayed synchronized while every `challenge_scalar` value diverged — surfacing as a deterministic `StageClaimOutputMismatch` at the first batched-claim check, packed pipeline only. Fixed by giving the legacy `JoltField` a `from_scalar_challenge_bytes` whose default preserves the reversed convention (BN254/Dory byte-identical by construction), delegating the blake2b/keccak/prover-native transcripts to it, and overriding it on `AkitaFp128` to the shared unreversed convention.
- `jolt-prover-legacy`'s `akita` feature now enables `dep:jolt-field` directly, so `-p jolt-verifier --features akita` builds standalone.

## Testing

- [x] Ran tests for modified crates: `jolt-prover-legacy --features akita` 445/445 (including all three packed muldiv e2e variants); `jolt-akita` 38 passed, 3 skipped; `muldiv` e2e 3/3 under `host` and 3/3 under `host,zk` (the BN254 proving path is untouched).
- [x] `cargo clippy` and `cargo fmt` pass: fmt `--check` clean; clippy `--all-targets -- -D warnings` under `host`, `host,zk`, `-p jolt-verifier --features akita`, and `-p jolt-prover-legacy --features akita`, all with `RUSTFLAGS=-Dwarnings`.
- [x] Shared-field identity gate (post-cutover semantics): exactly one `jolt-field` identity, zero `akita-field`.

## Security Considerations

- **Challenge derivation** changes for the packed (Akita) pipeline only: `AkitaFp128` scalar challenges now decode unreversed little-endian, aligning the legacy prover with the modular verifier and with Akita #307's own transcript convention. BN254 challenge derivation is byte-identical by construction (the default preserves the reversed convention), witnessed by the Dory `muldiv` e2e passing in both modes. Packed proofs are not byte-frozen across the cutover.
- **Schedule catalogs** affect completeness, not soundness: a lookup miss is a hard error (fail-closed), and catalog identity is validated against the config policy on every lookup. One production boundary shape (16 vars, 81 polynomials — the widest K=16 group at the minimum padded trace) is unschedulable by the post-cutover planner although the pre-cutover runtime planner scheduled it; the coverage test pins it as an explicit exception that fails the moment the planner covers it, and the question is raised on the Akita #307 review thread.
- **Dependency-graph integrity**: the identity gate enforces a single `jolt-field` package identity and zero `akita-field`, closing the dual-identity hazard the `[patch]` wiring exists to prevent.

## Breaking Changes

- `jolt-field` loses its `akita` feature and optional akita dependencies; `JoltField` regains `Serialize + DeserializeOwned` supertraits (types implementing the component traits without serde no longer satisfy it).
- Packed (Akita-mode) proof transcripts change: `AkitaFp128` scalar-challenge decoding switches to the shared unreversed convention, so packed proofs produced before this PR do not verify after it (and vice versa). BN254/Dory proof formats are unchanged.
- `jolt-prover-legacy`'s `akita` feature now enables `dep:jolt-field`.


quangvdao:
Superseded by #1796. The replacement is refreshed onto the current field stack and Akita #429 API. Alberto’s six original cutover commits, from 0516c2b8 through 6b5d3ff0, are preserved unchanged in the replacement branch history.

PR #1746, opened 2026-08-06
perf(sdk): O(1) size-class guest allocator replacing first-fit linked list

0xAndoroid:
## Problem

The no_std guest allocator (`linked_list_allocator` first-fit, selected via ZeroOS's `alloc-linked-list` feature) walks its free list on every `alloc` **and** `dealloc`. In a zkVM every executed instruction is a proving-cost cycle, so this walk is pure trace-length overhead — and it grows with the number of live blocks.

**Measured on a real workload** (full stateless Ethereum mainnet block validation — reth/revm + MPT witness reveal over a 1.5 GiB heap, [jeth](https://github.com/0xAndoroid/jeth)): PC-sampled instruction profile attributed **85.7% of all executed instructions to the allocator**:

```
58.62%  <linked_list_allocator::hole::HoleList>::allocate_first_fit
27.10%  <linked_list_allocator::hole::HoleList>::deallocate
 5.57%  <k256::arithmetic::projective::ProjectivePoint>::add   (actual workload…)
```

Swapping in this allocator cut total trace length **~6×** (29.76B → 4.94B cycles for a 58M-gas block; 10.79B → 2.30B for a 31.8M-gas block), with identical guest outputs.

In-repo evidence (`examples/btreemap`, execute-only row counts): **−8.9%** rows at n=6000, **−21%** at n=20000 — the win scales with live-block count exactly as expected from the O(free-list) walk.

## Design — `jolt_platform::size_class_alloc`

- Allocations round up to a power-of-two size class (min 8 B); each class keeps a singly-linked free list (freed blocks store `next` in word 0). Alloc pops the head or bumps the arena cursor; dealloc pushes. **Every operation is a few dozen instructions; nothing searches.**
- Bump blocks are class-size aligned, so any `align ≤ size` is satisfied; larger aligns round the class up. Free-list reuse re-checks the requested alignment (in-place-grown blocks only guarantee their original class alignment) and falls through to the bump path on mismatch.
- `realloc` grows the **newest** bump block in place (catches Vec doubling), and treats same-class or shrinking requests as pointer-stable no-ops.
- No splitting/coalescing: bounded internal fragmentation (<2× per live block) + per-class high-water free lists, deliberately traded for O(1) — appropriate for single-run guest heaps.
- Registered in `__platform_bootstrap` **before** `kinit`, for `target_os = "none"` guests only. std/musl guests keep the ZeroOS allocator (musl's malloc manages its own arenas over mmap/brk).

## Testing

- Host-runnable unit tests (`cargo nextest run -p jolt-platform -E 'test(size_class)'`): class rounding, alignment-guarded reuse, LIFO recycling, in-place growth, copy-growth preserving contents, shrink stability, oversize/exhaustion nulls, alloc/free churn.
- `btreemap-guest` builds and executes to completion with matching (non-panicking) results on the emulator; the jeth workload additionally self-checks via its post-state-root assertion (any allocator bug ⇒ wrong state root ⇒ panic).

Note: GitHub Actions may be slow to schedule (ongoing incident) — happy to rebase/retrigger as needed.


jolt-platform/src/size_class_alloc.rs
```diff
@@ -0,0 +1,259 @@
...
+    let new_class = class_of(new_layout);
+    if new_class as usize >= NUM_CLASSES {
+        return ptr::null_mut();
+    }
+
+    // Same class (or shrink) → the existing block already fits. Shrinks keep
+    // the larger footprint; the tail is orphaned when the block is later freed
+    // into the smaller class (rare — `shrink_to_fit` — and bounded).
+    if new_class <= old_class {
+        return ptr_in;
```
moodlezoup: The "rare — `shrink_to_fit` — and bounded" claim doesn't hold: `Vec::into_boxed_slice`, `String::into_boxed_str`, and `collect::<Box<[T]>>()` all realloc-shrink to exact length, and the orphaned tail is footprint minus shrunk class — a Vec bump-grown in place to a 1 GiB footprint then boxed at small len orphans ~1 GiB forever. That also breaks the module doc's "< 2× per live block" bound, since the leak is per shrink-then-free event, not per live block. It's leak-not-corruption (the footprint invariant holds on every path), but on the long-running allocation-heavy guests this PR targets it can exhaust the arena — worth correcting both comments at minimum. If you want to close it: on a shrink, `old_layout`'s class never exceeds the true footprint, so the tail `[ptr + 2^new, ptr + 2^old)` can be split into power-of-two chunks and pushed onto the free lists in ≤ `old_class − new_class` operations.

---

jolt-platform/src/size_class_alloc.rs
```diff
@@ -0,0 +1,259 @@
...
+        // bump path (the head stays put for smaller-alignment requests).
+        let head = a.free[class];
+        if head != 0 && head & (layout.align() - 1) == 0 {
+            a.free[class] = *(head as *const usize);
+            return head as *mut u8;
+        }
+
+        // Bump path: blocks are class-size aligned.
+        let start = (a.cursor + size - 1) & !(size - 1);
+        let Some(new_cursor) = start.checked_add(size) else {
```
moodlezoup: `a.cursor + size - 1` wraps before the `checked_add` below ever gets a say — if the sum could overflow, `start` is already corrupt, so as written the `checked_add` can never fail meaningfully. Unreachable on riscv64 (cursor < 2^40, size ≤ 2^33), but on a 32-bit target (the toolchain installs riscv32imac and this crate carries riscv32 cfgs) cursor ≥ 0x8000_0000 and a class-28 request wraps in release, computing `start` below the heap and passing the `new_cursor > a.end` check. Either make the sequence fully checked (`a.cursor.checked_add(size - 1)`) or drop the `checked_add` and state the invariant — the current half-and-half is the one shape that's wrong. Same unchecked pattern at `addr + grown_size` in `realloc` and `heap_start + heap_size` in `init`.

---

jolt-platform/src/size_class_alloc.rs
```diff
@@ -0,0 +1,259 @@
...
+
+        // Fast path: recycle from the class free list. Blocks grown in place
+        // (see `realloc`) are only guaranteed their ORIGINAL class alignment,
+        // so re-check the requested alignment; on mismatch fall through to the
+        // bump path (the head stays put for smaller-alignment requests).
+        let head = a.free[class];
+        if head != 0 && head & (layout.align() - 1) == 0 {
+            a.free[class] = *(head as *const usize);
+            return head as *mut u8;
+        }
```
moodlezoup, line 90: Since the list is only ever popped at the head, one under-aligned head (an in-place-grown block freed into its new class) diverts every subsequent higher-alignment request in that class to the bump path — correctly-aligned blocks deeper in the list are unreachable for those requests, so over-aligned churn in that class grows the arena despite plentiful free blocks. Correctness is fine (a misaligned pointer is never returned); if the behavior stays, the comment should say the deeper entries are stranded too, not just that the head stays put.

---

jolt-platform/src/lib.rs
```diff
@@ -21,6 +21,8 @@ pub use advice::*;
 mod spoil;
 pub use spoil::{spoil_proof, UnwrapOrSpoilProof};
 
+pub mod size_class_alloc;
```
moodlezoup, line 25: This compiles the guest allocator — `static mut` arena included — into every target (host and wasm too) as safe `pub` fns from which UB is reachable in safe code: two host threads calling `alloc` is a data race, and an `init` mid-run resets the cursor over live blocks. Nothing off-guest uses it besides the unit test, so consider `#[cfg(any(target_arch = "riscv32", target_arch = "riscv64", test))]`. Related: `dealloc` writes through its raw pointer but carries no contract doc — mirror the one on `realloc`.

---

jolt-sdk/src/runtime/boot.rs
```diff
@@ -34,6 +34,21 @@ pub extern "C" fn __platform_bootstrap() {
...
+    // size-class allocator cut total trace length ~6×. See
+    // `jolt_platform::size_class_alloc`. std/musl guests keep the ZeroOS
+    // allocator (musl's malloc manages its own arenas over mmap/brk).
+    #[cfg(target_os = "none")]
+    zeroos::foundation::register_memory(zeroos::foundation::ops::MemoryOps {
+        init: jolt_platform::size_class_alloc::init,
+        alloc: jolt_platform::size_class_alloc::alloc,
+        dealloc: jolt_platform::size_class_alloc::dealloc,
+        realloc: jolt_platform::size_class_alloc::realloc,
+    });
```
moodlezoup, line 54: Every no_std guest still compiles and registers zeroos's linked-list ops (`jolt-sdk/Cargo.toml:121-124` enables `alloc-linked-list` for all riscv64 targets) only for this to overwrite them one call later — the `linked_list_allocator` crate rides along as dead weight in the guest binary (fn pointers stored into a static, so unlikely to be stripped). Since this comment already says musl guests are the only remaining users, moving `alloc-linked-list` into the existing `cfg(target_arch = "riscv64", target_os = "linux")` dependency block (with zeroos's standalone `memory` feature in the plain riscv64 block) would make this registration the only one and drop the dead allocator from no_std guests.

---

moodlezoup (commented):
Reviewed with a focus on allocator correctness (overlap/aliasing, in-place growth, boot wiring, the malloc-shim realloc rewrite) and on 32-bit/overflow edge cases. No corruption or soundness bugs found: the cursor-monotonicity and footprint invariants hold on every realloc path (in-place growth can only claim virgin bump space, shrink no-ops degrade to leaks, never overlap), the register-before-`kinit` ordering is correct against zeroos at the pinned rev (no allocation can flow through the displaced linked-list ops), and the malloc-shim header round-trip keeps the layout passed to `alloc::alloc::realloc` exactly equal to the block's current allocation layout under both allocators.

Five comments below — fragmentation/leak behavior that the in-code docs currently understate, overflow-check consistency, and two packaging/API-surface items. None block the core design.

---

jolt-platform/src/size_class_alloc.rs
0xAndoroid: Fixed in 39ca3c484: the no-op path is now same-class only. The newest bump block resizes in place (the cursor retreats, returning the tail to the arena); everything else moves via allocate-copy-free, so a block's class always equals its physical footprint. Went with the move rather than tail-splitting — on a shrink the copy is min(old, new) = new_size bytes, and the old ZeroOS realloc was alloc-copy-free on every call anyway. Both doc claims updated.

jolt-platform/src/size_class_alloc.rs
0xAndoroid: Fixed in 7a1b8725a — the rounding add is checked now, and so is addr + new_class_size in realloc's in-place resize (the nastier one: its wrap passes the <= end check and retreats the cursor below the heap). One correction to the scenario as written: the line-90 wrap can't return a below-heap pointer — wrapped < size − 1 forces start = 0, so the wrapping call itself returns null — but it stores the wrapped cursor first and corrupts every later alloc, so checked either way. init stays unchecked: it takes linker bounds, and a wrapped end < cursor just makes every alloc return null.

jolt-platform/src/size_class_alloc.rs
0xAndoroid: Comment fixed in 7a1b8725a — it now says an under-aligned head strands aligned blocks deeper in the list for over-aligned requests. Keeping the head-only check: scanning reintroduces the O(list) walk this allocator exists to avoid, and align > class-size churn hasn't shown up in any guest workload.

jolt-platform/src/lib.rs
0xAndoroid: Done in 39ca3c484 — gated exactly as suggested, and dealloc now documents its pointer contract (realloc's lives on its #[expect(not_unsafe_ptr_arg_deref)]).

jolt-sdk/src/runtime/boot.rs
0xAndoroid: Done in 7a1b8725a — alloc-linked-list moved to the riscv64+linux block; plain riscv64 now gets zeroos's standalone memory feature. The rebuilt no_std guest's dep-info has zero linked-list references and the muldiv e2e passes; the musl target's feature union is unchanged.

moodlezoup (approved):

PR #1747, opened 2026-08-07
perf(inlines): fuse Keccak theta-apply into rho via VirtualXORROT tables for all rho rotations — 3,260 → 2,652 rows/perm (−19%)

0xAndoroid:
Keccak-f[1600] is the dominant row consumer for EVM-adjacent guests (in an Ethereum block-validation guest, keccak is ~33% of all trace rows). This PR cuts the keccak256 inline from **3,260 to 2,652 rows per permutation (−18.7%)** by letting the rho step absorb the theta-apply XOR, one row per lane.

### What

1. **23 new virtual instructions** `VirtualXORROT{2,3,8,9,19,20,21,23,25,28,36,37,39,43,44,46,49,50,54,56,58,61,62}` (`rd = rotr64(rs1 ^ rs2, N)`). Together with the existing `VirtualXORROT63` these cover all 24 nonzero rotations of the Keccak rho step. Every instruction is an instance of the existing const-generic `VirtualXORROTTable<XLEN, R>` / `XorRotPrefix<R>` / `XorRotSuffix<R>` construction introduced for Blake — no new table *shape*, only new rotation amounts — registered in both the modular lookup-table stack and the legacy prover, with the table enums extended at the end to keep index ABI stable (asserted by `prover_lookup_table_abi`).
2. **Keccak sequence builder** computes `B[pi(x,y)] = rotl64(A[x,y] ^ D[x], r[x,y])` in one row per lane (was XOR + ROTRI), aliases B[0,0] to A[0,0] (rho offset 0; the old copy row was dead weight), and cuts virtual registers 66 → 58 (5 were never used; D[3..5] reuse dead C registers) — each register costs one reset row per inline invocation.

Per round: 131 → 106 rows (theta-C 20 XOR, theta-D 5 ROTRI + 5 XOR, fused apply+rho+pi 24 XORROT + 1 XOR, chi 25 ANDN + 25 XOR, iota 1 XORI). Per permutation: 24×106 + 50 I/O + 58 resets = 2,652.

### Cost of 23 more tables

- Lookup-table flags are virtual openings — **no new committed polynomials**.
- Verifier: +23 constant-size table-MLE evaluations in stage 5.
- Prover: the suffix pass runs only for tables used by the trace; the address-phase prefix pass grows 46 → 69 prefixes (trace-length-independent work).
- Guests that don't use keccak see no row change; blake, sha2 etc. untouched.

### Measured impact (Ethereum block-validation guest, block 25,698,189, 41.9M gas, 141,301 permutations)

| | rows | cycles/gas |
|---|---|---|
| main | 1,289.17M | 30.74 |
| this PR | 1,203.30M | **28.70 (−6.7%)** |

### Correctness

- `sha3_e2e_dory` proves + verifies in both standard and `zk` modes; the guest trace exercises all 24 XORROT amounts, so the instruction read-RAF sumcheck closing end-to-end-checks every new table.
- Per-amount `mle_random`, `prefix_suffix`, and 8-bit exhaustive hypercube tests (modular + legacy, 138 tests).
- New differential fuzz harness: the expanded inline sequence executed in the tracer emulator vs `tiny_keccak::keccakf`; run with 1,000,000 random permutations (plus 1M digest-level sponge checks across rate-straddling lengths). `KECCAK_FUZZ_ITERS` scales it; CI default stays fast.
- muldiv e2e both modes, byte-diff ratchets 18/18, modular zk_e2e 12/12, full legacy suite 513/513, clippy clean in both modes.

### Non-goals / follow-ups

- A `rd = rs1 ^ rotl64(rs2, 1)` table would fuse theta-D to 5 rows/round (−120/perm) but needs a new prefix/suffix shape (the skewed bit pairing splits across phase-dependent chunk boundaries) — left out to keep this PR to the audited construction.
- A fused absorb variant (`state ^= block; permute`) would cut ~50–85 guest-side rows per block.
- The architectural floor for any inline-sequence keccak is ~2,450–2,550 rows/perm (each chi output lane's 192-bit input cone needs ≥2 lookups ⇒ ≥50 rows/round before theta). Reaching SP1-class (~500-equivalent) requires a dedicated keccak precompile at the protocol level.


quangvdao:
Seems bad to be adding 23 new virtual instructions :/

0xAndoroid:
Closing — the 23 new XORROT instruction tables inflate the instruction read-raf sumcheck; not merging this shape. The follow-up (#1749) will be rebuilt against main with the table-lean subset.

PR #1749, opened 2026-08-07
perf(inlines): Keccak theta-D, in-place scheduling, and fused absorb (3,260 -> 3,087 rows)

0xAndoroid:
Standalone on `main`. Three reviewable commits.

## 1. `VirtualXORROTL1` — fuse theta-D into one row (−120/perm)

New virtual instruction `rd = rs1 ^ rotl64(rs2, 1)`, exactly `D[x] = C[x-1] ^ rotl64(C[x+1], 1)`. Theta-D goes 10 → 5 rows/round.

Distinct from the existing `VirtualXORROT{R}` (`rotr(rs1^rs2, R)`): only one operand is rotated, so output bit i pairs `x_i` with `y_{(i-1) mod 64}` — a *skewed* pairing needing a new prefix/suffix shape. No same-cost rewrite exists with shipped tables (pre-rotating costs the same 10 rows).

**Soundness / decomposition.** 3 new prefixes (accumulator, straddle coefficient, wrap coefficient) × 4 suffixes (`One`, local pairs, top-y-bit, bottom-x-bit); combine is bilinear: `P_acc·S_one + S_pairs + P_straddle·S_topY + P_wrap·S_x0`. At any phase-boundary split exactly one pair straddles, plus the `x_0`/`y_63` wrap pair handled by a checkpointed coefficient. The table is total; `materialize_entry == evaluate_mle` is enforced on the 8-bit exhaustive hypercube, random-64, and phased prefix/suffix harnesses at both phase geometries, asserted every round, in both prover stacks.

**Prover cost.** +1 used lookup table ≈ ~40K field mults/proof (present-table-gated read-raf; ~0.001–0.01% at 2^22), +3 prefix variants ≈ ~30K O(1) ops; verifier +~300 mults, proof +32 B. Zero new committed polynomials; no prover/verifier protocol changes.

## 2. Main-native in-place rho/pi/chi scheduling (−53/perm)

Walks pi's single 24-cycle backward and keeps one rotated lane in a temporary register; chi consumes rows in place. Rho remains the existing `VirtualROTRI` path; no per-rotation lookup tables are added. Registers drop 66 → **37** (−29 reset rows), and keeping the fixed `(0,0)` lane in place removes one identity-copy `XORI` per round (−24).

## 3. Fused absorb entry point

New inline entry `keccak256_absorb_permute`: 17 block loads + 17 XORs folded into the permutation load prologue. First absorb (zero state) and padded final absorb stay on the pure path; materializing them separately regresses short inputs.

On the aligned 2,048-byte hash benchmark (14 fused blocks after the first): guest RV rows **1,523 → 527** (−996, −71.1/block); the fused inline adds 34 rows/block, so total trace rows are **51,082 → 50,562** (−520, −37.1/block net).

## Row-count ledger

| point | rows/perm |
|---|---:|
| `main` baseline | 3,260 |
| + fused theta-D | 3,140 |
| + in-place scheduling | **3,087** |
| fused absorb+permute entry | 3,121 |

## Testing

- Table trio: exhaustive 8-bit hypercube, random MLE, phased prefix/suffix at both geometries with per-round assertions, in both stacks; append-only ABI gate
- Row-count ratchets: 3,087 permutation / 3,121 fused entry; registered-inline expansion fixture regenerated
- Differential fuzz vs tiny-keccak: 1,000,000 permutation cases + 1,000,000 fused-absorb cases
- `sha3_e2e_dory` standard + zk; `muldiv` standard + zk
- clippy `host` + `host,zk` with `-D warnings`; fmt


0xAndoroid:
Restacked as a standalone `main` PR after #1747 closed. Branch history was rebuilt to three commits; new ledger is 3,260 → 3,140 → 3,087 rows/perm, with a 3,121-row fused absorb entry. Full standalone gates pass.

0xAndoroid (commented):
Reviewed the full diff at `451e93cbd` against `main` (`e789b9f5f`).

- No soundness defect found in the `VirtualXORROTL1` relation, MLE, phase-boundary decomposition, or Keccak in-place schedule.
- The #1768 union retains ShiftDataB/H/W at `0x009e`–`0x00a0`; XORROTL1 is unique at `0x00a1`; modular and legacy table ordinals/dispatch match.
- Four test/maintenance gaps are inline. Expansion hashes match the changed rows, while independent harness fuzz covers sequence semantics but not the guest asm ABI.

AI-generated review: GPT-5.6/Codex via the user's tooling.

jolt-inlines/keccak256/src/sdk.rs
```diff
@@ -174,29 +176,13 @@ impl Keccak256 {
...
-        }
-
-        #[cfg(target_endian = "big")]
-        {
-            // For big-endian, convert each word from little-endian bytes
-            for i in 0..RATE_IN_U64 {
-                let word =
-                    u64::from_le_bytes(self.buffer.0[i * 8..(i + 1) * 8].try_into().unwrap());
-                self.state[i] ^= word;
+        if self.initialized {
```
0xAndoroid: [P1] No test calls this changed `update`/`finalize` state machine: the new differential test only calls `digest`, so `buffer_len`, `initialized`, and split-update transitions are uncovered. Add chunked vectors across the 136/272-byte boundaries against `tiny-keccak`.

jolt-inlines/keccak256/src/sdk.rs
```diff
@@ -240,98 +226,111 @@ fn to_bytes(state: [u64; 25]) -> [u8; HASH_LEN] {
...
+/// - `block` must point to 136 readable bytes and be 8-byte aligned.
+/// - The two memory regions must not overlap.
+#[cfg(all(
+    not(feature = "host"),
+    any(target_arch = "riscv32", target_arch = "riscv64")
+))]
+pub unsafe fn keccak256_absorb_permute(state: *mut u64, block: *const u8) {
+    use crate::{INLINE_OPCODE, KECCAK256_ABSORB_PERMUTE_FUNCT3, KECCAK256_FUNCT7};
+    core::arch::asm!(
+        ".insn r {opcode}, {funct3}, {funct7}, x0, {rs1}, {rs2}",
```
0xAndoroid: [P1] No CI guest/prover test executes this fused RISC-V ABI. Harness fuzz constructs funct3=1 directly, while every SHA3 proof fixture hashes 32 bytes and stays on the pure final-block path; a wrong funct3 or rs1/rs2 encoding here would pass. Add a guest vector of at least 272 bytes to clear and ZK coverage.

jolt-inlines/keccak256/tests/differential_fuzz.rs
```diff
@@ -0,0 +1,123 @@
...
+//!
+//! Two layers:
+//! - The expanded inline sequence, executed instruction-by-instruction in the
+//!   tracer emulator (`InlineTestHarness`), vs `tiny_keccak::keccakf`. This is
+//!   the layer that exercises `VirtualXORROTL1` and the in-place rho/pi/chi schedule.
+//! - The one-shot `Keccak256::digest` sponge (padding, absorption, squeezing)
+//!   vs `tiny_keccak::Keccak::v256` over random lengths straddling the rate.
+//!
+//! Iteration counts scale with `KECCAK_FUZZ_ITERS` (default keeps CI fast);
+//! the soundness gate runs one million iterations for each inline entry point.
```
0xAndoroid: [P2] No repository workflow sets `KECCAK_FUZZ_ITERS`, so the stated million-case gate is not automated; CI executes 2,000 cases for each inline entry. Wire a scheduled/manual gate or label the million-case run as one-off evidence.

jolt-inlines/keccak256/tests/row_count.rs
```diff
@@ -0,0 +1,65 @@
...
+    let _ = KECCAK256_NAME;
+
+    let instr = InlineTestHarness::create_default_instruction(
+        INLINE_OPCODE,
+        KECCAK256_FUNCT3,
+        KECCAK256_FUNCT7,
+    );
+    let sequence = instr.inline_sequence(&VirtualRegisterAllocator::default());
+
+    let mut histogram: BTreeMap<String, usize> = BTreeMap::new();
```
0xAndoroid: [P3] This histogram is an unasserted diagnostic: it parses `Debug` text and only prints under `--nocapture`. The repository handoff policy says to remove diagnostic counters/histograms or make them an intentional tool/benchmark; delete lines 29–43 and keep the two row assertions.

0xAndoroid (commented):
Adversarial review of the full diff vs `main` @ 451e93cb. Merge-base is the current main tip (e789b9f5f, #1768), so today's union-merge heal is fully contained in this diff. Every claim below was checked against the code at this head, not the PR text.

**Verified sound:**
- **Tag table / heal:** `0x00a1` unique (no duplicate discriminants in `for_each_jolt_instruction_kind`); #1768's ShiftData `0x009e`–`0x00a0` untouched; all four `kind.rs` marker macros populated; `per_instruction_diff` SUPPORTED 87→88; `prover_lookup_table_abi` pins modular↔legacy table indices with a COUNT-exhaustiveness assert. Swept every ShiftDataW/XorRot63 registration site: no double dispatch, no missed mirror (`jolt-program/src/asm.rs` intentionally untouched — keccak doesn't use that macro).
- **Semantics agree across all five surfaces** (tracer exec, x86 AOT `rol rax,1; xor`, `to_lookup_output`, `materialize_entry`, `evaluate_mle`): `out[p] = x[p] ^ y[(p−1) mod 64]`; the MLE's `y_{(i+1) % XLEN}` pairing is the correct MSB-first form of that.
- **Prefix/suffix decomposition** (both stacks, derived by hand at arbitrary splits): in-chunk pair weights `2^{suffix_pairs+local_bit}` match global output weights; straddle checkpoint `2^{suffix_pairs}(1−2·x_last)` is consumed by the next chunk's top y-bit; wrap = `y_0` seed + `(1−2y_0)·x_63`; the `Pairs` suffix's `& !1` correctly excludes the wrap x-bit; `TopYBit`/`BottomXBit` are genuinely {0,1} for `is_01_valued`. Legacy per-pair `bind_pair` recursion checked against the real schedule: `update_checkpoints` fires at `r.len() % 2 == 0` → `j` odd → `bit = 63 − j/2` is the right pair, and all variants update unconditionally, so the `.unwrap()`s hold.
- **In-place rho/pi/chi:** the backward walk of pi's 24-cycle writes only already-consumed registers ((0,2)'s B-value parked in `pi_temp`, (0,0) fixed point untouched); chi pre-captures the `b[0]`/`b[1]`-dependent ANDNs in `t3`/`t4` before those registers are overwritten; theta-D's reuse of `c[1]`/`c[2]` for `D[3]`/`D[4]` respects last-reads (C[1]@x=2, C[2]@x=3).
- **Row/canary math:** 24×(70 XOR + 5 XORROTL1 + 24 ROTRI + 25 ANDN + 1 XORI) + 25 LD + 25 SD + 37 resets = 3087; absorb +17 LD +17 XOR = 3121; ROTRI canary 576 = 24×24. Parity jsonl re-baseline touches only keccak rows.
- **Fused absorb:** builder XORs exactly `RATE_IN_U64 = 17` lanes before round 0; sponge padding indices stay in-bounds (`buffer_len ≤ 135` at finalize since `update` absorbs at exactly RATE); funct3=1 is collision-free in the opcode-0x0B/funct7 namespace.

**Findings this pass:** 2 inline comments — one P2 (the modular prefix's `first_phase && suffix_pairs == 0` arm is dead code that also computes the wrong value: it double-counts the wrap seed `y_0`), one P3 (the digest fuzz lane never takes the unaligned path). The four existing comments from the earlier pass (streaming `update`/`finalize` untested; fused ABI never executed in a proven guest; `KECCAK_FUZZ_ITERS` not wired in CI; row_count histogram) all still apply at this head.

No soundness issue found in any reachable path: the byte-parity e2e (`proof == legacy_proof` on a sha3 trace asserted to contain the new expansion) plus the phased prefix/suffix harnesses in both stacks pin the surfaces that matter.

*This review was AI-generated (Claude/Fable via the author's tooling).*

crates/jolt-lookup-tables/src/tables/prefixes/xor_rotl1.rs
```diff
@@ -0,0 +1,90 @@
...
+        let x_bit = (x >> local_bit) & 1;
+        let y_bit = (y >> (local_bit - 1)) & 1;
+        acc += F::from_u64((x_bit ^ y_bit) << output_bit);
+    }
+
+    let x_bottom = x & 1;
+    if suffix_pairs == 0 {
+        acc += if first_phase {
+            let y_top = (y >> (phase_pairs - 1)) & 1;
+            F::from_u64(x_bottom ^ y_top)
```
0xAndoroid: [P2] This `first_phase` arm double-counts the wrap seed. `acc` is already initialized to `y_top = y_0` (L27–29); adding the full `xor(x_bottom, y_top)` makes the wrap pair contribute `y_0 + xor(x_63, y_0)` instead of `xor(x_63, y_0)`. XLEN=2 check: acc = `y_0 + 2·xor(x_0,y_1) + xor(x_1,y_0)` vs table = `2·xor(x_0,y_1) + xor(x_1,y_0)` — off by `y_0`. The multi-phase arms are consistent (the seed is consumed by `ckpt[Wrap]·x_bottom`); only this arm re-adds it. Unreachable in-tree: `suffix_pairs == 0 && first_phase` forces `phase_pairs == XLEN`, i.e. one 128-bit phase = a 2^128-point materialization (kernels use 8-bit chunks, harnesses 16/8) — so no test can hit it, but as written the arm is dead *and* wrong. Mirror the checkpoint arm instead: `acc += (F::one() - F::from_u64(2 * y_top)) * F::from_u64(x_bottom)`. The legacy `bind_pair` recursion has no such shortcut and handles this geometry correctly.

jolt-inlines/keccak256/tests/differential_fuzz.rs
```diff
@@ -0,0 +1,123 @@
...
+
+    for i in 0..iters {
+        // Cover empty inputs, sub-rate, rate-straddling, and multi-block sizes.
+        let len = match i % 4 {
+            0 => (rng.next_u32() % 16) as usize,
+            1 => 120 + (rng.next_u32() % 32) as usize,
+            2 => (rng.next_u32() % 600) as usize,
+            _ => (rng.next_u32() % 4096) as usize,
+        };
+        let mut input = vec![0u8; len];
```
0xAndoroid: [P3] This lane never takes the unaligned path: `vec![0u8; len]` data is ≥8-aligned on every mainstream allocator, so `is_aligned` in `digest` is always true and `absorb_unaligned_first` / `absorb_unaligned` (copy-to-local + fused inline) are fuzzed zero times. They are covered only by the fixed-size vectors in `sdk.rs::test_keccak256_aligned_vs_unaligned`. Hash `&input[1..]` on alternating iterations to fuzz both branches.

Acentelles (approved):
Reviewed the current head (`10d0f00`) against `main`. The three optimizations compose cleanly:

- `VirtualXORROTL1` agrees across tracer execution, x86 AOT emission, lookup output, table materialization, and the MLE. The prefix/suffix decomposition accounts for the local, straddling, and circular-wrap pairs; the single-phase wrap double-count found during review is fixed and now has a targeted regression test.
- The in-place rho/pi/chi schedule respects the 24-lane pi cycle and preserves chi dependencies before overwriting lanes. The 3,087-row ledger and 576-ROTRI canary match the emitted schedule.
- The fused absorb ABI is now exercised by the 300-byte clear and ZK prover tests, while the sponge tests cover chunked updates around the 136/272-byte boundaries and aligned/unaligned paths.

The added lookup-table overhead is a reasonable trade for the 173-row permutation reduction, and the repeated-block fused path provides an additional net trace reduction. All current CI checks are green. No remaining correctness or soundness concerns from my review.

Acentelles (commented):
Review of head `10d0f00ed` against `main`. **No critical or correctness bugs found.** I checked the prefix/suffix algebra for `VirtualXORROTL1` (straddle and wrap terms), the theta-D register reuse, the backward walk of pi's 24-cycle, the chi ANDN/XOR ordering, the 3087/3121 row arithmetic, and the sponge control flow; the modular lookup-table suite passes in the PR worktree.

Everything below is non-critical. Two are worth addressing (an untested aligned-absorb path with a false comment, and a guest-row regression from the `initialized` design); the rest is cleanup and can be deferred or ignored.

crates/jolt-prover/tests/dory_byte_diff.rs
```diff
@@ -1855,12 +1855,18 @@ mod inline_sha3 {
 
     use super::support;
 
-    const KECCAK_ROTRI_ROWS: usize = 696;
+    // 24 rounds x 24 ROTRI per Keccak-f permutation (theta-D XORs use VirtualXORROTL1).
+    const KECCAK_ROTRI_ROWS: usize = 576;
+    // 300 bytes drive every `Keccak256::digest` inline path: a first block
+    // (plain permutation), one fused absorb-permute block, and a padded final block.
```
Acentelles: **[Non-critical, worth addressing]** **Coverage gap / comment is false.** This comment (and the twin in `zk_e2e.rs:336`) says 300 bytes drive *every* `Keccak256::digest` inline path, but the guest borrows the postcard-encoded `Vec<u8>` zero-copy, and the 2-byte varint length prefix puts the message at offset 2 mod 8. `digest` therefore takes `is_aligned == false` and only `absorb_unaligned_first` / `absorb_unaligned` (stack-copy) / `absorb_final` run. The aligned branch, `absorb_aligned_first` (the `&[u8] -> &[u64; 17]` cast) and `absorb_aligned` (the fused inline reading caller memory directly through `rs2`), is never executed on RISC-V by any in-repo test, and the ROTRI canary cannot tell the two branches apart (same 576 rows per permutation). A regression confined to the aligned path (wrong `rs2` operand, LD base/offset over caller-owned memory) would pass both sha3 e2e suites. Suggest feeding an 8-aligned message (e.g. hash from a `[u64; N]`-backed buffer in the guest, or pad the serialized input) so the fused-inline-on-caller-memory path is covered, and fix the comment.

jolt-inlines/keccak256/src/sdk.rs
```diff
@@ -3,23 +3,20 @@
...
 pub struct Keccak256 {
     /// The 25-word (1600-bit) Keccak state.
     state: [u64; 25],
     /// Buffer for incomplete blocks.
-    buffer: AlignedBlock,
+    buffer: [u64; RATE_IN_U64],
     /// Number of bytes in the buffer.
     buffer_len: usize,
+    /// Whether at least one block has been permuted into the state.
+    initialized: bool,
```
Acentelles: **[Non-critical, worth addressing]** **`initialized` adds a second absorb path that is more expensive than the fused inline.** The fused sequence is state-agnostic (25 LD state, then 17 x (LD rs2, XOR) = 3087 + 34 rows), so it is correct on a zero state. The software path it replaces (`absorb_words_first` / `absorb_final`: 17 x (LD state, LD block, XOR, SD state) = 68 rows + `keccak_f` 3087) is 34 rows *worse* whenever `state` is not a compile-time constant: the first `absorb_buffer` of every streaming hasher, every `finalize`, and every `digest` final block (short messages, the common case). The only place the shortcut can win is the `digest` first full block, where LLVM may fold `0 ^ v` for about a 17-row gain, so the per-hash sum is negative in every case. The sha2 `initial` precedent selects a genuinely cheaper inline (IV baked as immediates); this flag selects the slower path. Simpler and cheaper: drop `initialized` and the three `*_first` helpers, have `absorb_buffer` and `absorb_final` call `keccak256_absorb_permute` unconditionally, loop `0..full_blocks` in `digest`, and delete the private `keccak_f` wrapper.

jolt-inlines/keccak256/src/sdk.rs
```diff
@@ -68,7 +66,7 @@ impl Keccak256 {
             unsafe {
                 core::ptr::copy_nonoverlapping(
                     input.as_ptr().add(offset),
-                    self.buffer.0.as_mut_ptr(),
+                    self.buffer.as_mut_ptr().cast(),
```
Acentelles: **[Non-critical, cleanup]** **Redundant guest memcpy per full block in `update`.** This loop stages each 136-byte block through `self.buffer` before `absorb_buffer` calls the fused inline, which then re-loads the same 17 words from `self.buffer`. Pre-PR this copy was unavoidable; with the fused entry taking the block pointer in `rs2` it is pure staging (a `compiler_builtins` memcpy call, tens of rows, roughly 1-3% of the 3121-row inline per block). `digest` already has the exact dispatch this needs (`is_multiple_of(8)` then `absorb_aligned` / `absorb_unaligned` on the input slice); reuse it here once `self.initialized` (tracer `load_doubleword` asserts 8-byte alignment, so the check on the actual pointer is required). `hash-bench` `keccak_s` exercises exactly this streaming path.

jolt-inlines/keccak256/src/sdk.rs
```diff
@@ -96,19 +94,18 @@ impl Keccak256 {
...
         // Pad the message. Keccak uses `0x01` padding.
         // If buffer_len == RATE_IN_BYTES-1 both markers land in the same byte (0x01 | 0x80 = 0x81)
-        self.buffer.0[self.buffer_len] = 0x01;
-
-        // Zero the remaining bytes (including the last byte if needed)
-        if self.buffer_len + 1 < RATE_IN_BYTES {
-            unsafe {
+        unsafe {
+            let buffer = self.buffer.as_mut_ptr().cast::<u8>();
+            *buffer.add(self.buffer_len) = 0x01;
```
Acentelles: **[Non-critical, cleanup]** **Lost bounds check and dropped SAFETY comments.** The old `self.buffer.0[self.buffer_len] = 0x01` was a bounds-checked index; this raw write relies on `buffer_len < RATE_IN_BYTES`, which is now enforced only implicitly by `update` absorbing on full. Today that invariant holds, so no live bug, but a future edit that lets `buffer_len` reach 136 (e.g. deferring `absorb_buffer` to `finalize`) turns a deterministic panic into a silent 1-byte write past `buffer` into a neighbouring field and a wrong digest with no diagnostic. A `debug_assert!(self.buffer_len < RATE_IN_BYTES)` at the top of this block costs nothing in release. Separately, the PR deletes every `// SAFETY:` comment in this file while adding several new `unsafe` blocks (here, `absorb_aligned_first`, `absorb_final`, the host `keccak256_absorb_permute`); CLAUDE.md lists SAFETY on unsafe blocks as the one comment class worth writing.

crates/jolt-prover-legacy/src/zkvm/lookup_table/prefixes/xor_rotl1.rs
```diff
@@ -0,0 +1,165 @@
...
+use super::{PrefixCheckpoint, Prefixes, SparseDensePrefix};
+
+#[derive(Clone, Copy)]
+struct PrefixState<F> {
+    acc: F,
+    straddle: F,
+    wrap: F,
+}
+
+fn bind_pair<F: JoltField>(state: &mut PrefixState<F>, x: F, y: F, bit: usize, first_pair: bool) {
```
Acentelles: **[Non-critical, cleanup]** **Field multiplications on binary tail bits in a legacy hot path.** `prover_msg_read_checking` (`read_raf_checking.rs:1235-1256`) calls `prefix_mle` for all 62 `Prefixes` variants, at every point and both `c` values, in each of the 128 address rounds; the three XorRotL1 variants each re-run the whole `evaluate_state`. In the tail loop (lines 84-93) `x`, `y` are 0/1, yet `bind_pair` does `2^bit * x`, `2^bit * (1 - 2x)` and `straddle * y` as full BN254 muls (2-3 per pair, up to 7 pairs, x3 variants, x2 `c`). Sibling `xor_rot.rs:66` folds the binary tail into a single `F::from_u64(...)`, and the modular twin (`crates/jolt-lookup-tables/.../prefixes/xor_rotl1.rs:41-54`) accumulates `(x_bit ^ y_bit) << output_bit` as integers with zero muls in the loop. Materiality is low (well under 1% of prover time), but this is the one place the two stacks diverge in cost for no reason; accumulate the tail as a u64, make `straddle * y` a conditional add, and pick the straddle sign with a conditional negation.

jolt-inlines/keccak256/src/sequence_builder.rs
```diff
@@ -38,229 +30,242 @@ pub(crate) const ROUND_CONSTANTS: [u64; 24] = [
...
+            first_lane,
+        );
+    }
+}
+
+fn pi_destination((x, y): (usize, usize)) -> (usize, usize) {
+    (y, (2 * x + 3 * y) % 5)
+}
+
+fn inverse_pi(destination: (usize, usize)) -> (usize, usize) {
```
Acentelles: **[Non-critical, cleanup]** **pi map has a closed-form inverse and a second owner; `pi_temp` lane coupled to `first_source` only by a `debug_assert`.** `inverse_pi` brute-forces the preimage over 25 cells (552 searches per expansion, negligible cost but needless code with an `unreachable!`). Since `pi((x, y)) = (y, (2x + 3y) mod 5)` and `2^-1 = 3 (mod 5)`, `inverse_pi((X, Y)) = ((X + 3Y) % 5, X)`; check: `(1, 0) -> (0, 2) -> (1, 0)`. Also `exec.rs` `execute_rho_and_pi` already open-codes the same map, so `pi_destination` is a second owner; make it `pub(crate)` and call it from the reference. Finally, `rho_pi_lane` (line 128) hard-codes `(0, 2)` as the lane living in `pi_temp` while `rho_and_pi` independently picks `first_source = (1, 0)`; the two agree only because `pi_destination((1, 0)) == (0, 2)`, checked solely by the post-loop `debug_assert_eq!`, so changing `first_source` would give chi a stale register in release builds with no diagnostic. Derive one from the other (`const PI_TEMP_LANE = pi_destination(FIRST_SOURCE)` with a `const fn`) so the assert becomes unnecessary.

jolt-inlines/keccak256/tests/row_count.rs
```diff
@@ -0,0 +1,46 @@
...
+
+    let instr = InlineTestHarness::create_default_instruction(
+        INLINE_OPCODE,
+        KECCAK256_FUNCT3,
+        KECCAK256_FUNCT7,
+    );
+    let sequence = instr.inline_sequence(&VirtualRegisterAllocator::default());
+    // 24 rounds x 125 rows (70 XOR + 5 XORROTL1 + 24 ROTRI + 25 ANDN + 1 XORI)
+    // + 25 LD + 25 SD + 37 register resets.
+    assert_eq!(
```
Acentelles: **[Non-critical, cleanup]** **Duplicate ratchet.** `registered_inline_expansions_match_golden_fixture` (jolt-inlines/fixtures/src/lib.rs:222-233) already asserts `row_count` exactly for all eight KECCAK256 / KECCAK256_ABSORB_PERMUTE cases, runs in CI (`test-inlines` job), and is re-blessed via `JOLT_UPDATE_INLINE_EXPANSION_FIXTURES`. This test pins the same 3087/3121 with a second, hand-edited bless mechanism and a static breakdown comment that nothing checks, so it adds no failure signal the fixture lacks (CLAUDE.md: keep permanent tests only when they add a distinct signal beyond golden fixtures). No other inline crate has a `tests/row_count.rs`. Either delete it, or follow the sha2 precedent and put the per-mnemonic histogram in the failure message so the breakdown cannot rot.

crates/jolt-prover-legacy/src/zkvm/lookup_table/prefixes/xor_rotl1.rs
```diff
@@ -0,0 +1,165 @@
...
+) -> PrefixState<F>
+where
+    C: ChallengeFieldOps<F>,
+    F: JoltField + FieldChallengeOps<C>,
+{
+    let suffix_len = LOG_K - j - b.len() - 1;
+    let suffix_pairs = suffix_len / 2;
+    let current_bit = 63 - j / 2;
+    let first_pair = current_bit == 63;
+    let mut state = if first_pair {
```
Acentelles: **[Non-critical, cleanup]** **Seeding block duplicated verbatim.** Lines 46-60 and 108-122 both compute `bit = 63 - j / 2`, `first_pair = bit == 63`, and build the same `PrefixState` from zeros or the three `checkpoints[..].unwrap()` reads (13 lines each), and `bind_pair` repeats the `2^bit * x` / `2^bit * (1 - 2x)` tail in both arms. Siblings seed with a one-line `checkpoints[idx].unwrap_or(F::zero())` in both `prefix_mle` and `update_prefix_checkpoint`. A behavior-preserving fold: seed via `unwrap_or` with `straddle` defaulting to `F::one()` (so `acc += straddle * y` supplies the `+ y` on the first pair), keep a single `if first_pair { state.wrap = F::one() - y - y }`, then fall through to the shared `bit == 0 / else` tail. Note the checkpoints are `None` exactly when `first_pair` (`read_raf_checking.rs:564`, `:1024-1031`), which is why the branch exists at all.

jolt-inlines/keccak256/tests/row_count.rs
```diff
@@ -0,0 +1,46 @@
...
+    INLINE_OPCODE, KECCAK256_ABSORB_PERMUTE_FUNCT3, KECCAK256_FUNCT3, KECCAK256_FUNCT7,
+    KECCAK256_NAME,
+};
+use tracer::utils::inline_test_harness::InlineTestHarness;
+use tracer::utils::virtual_registers::VirtualRegisterAllocator;
+
+#[test]
+fn count_keccak256_rows() {
+    // Force inline registration (inventory) by touching the crate.
+    let _ = KECCAK256_NAME;
```
Acentelles: **[Non-critical, cleanup]** **Dead code with a misleading comment.** `KECCAK256_NAME` is a `pub const &str`; referencing it is inlined into the test CGU and emits no symbol into the keccak rlib, so it cannot affect whether the `inventory::submit!` statics are linked. The `use` of `INLINE_OPCODE` / `KECCAK256_FUNCT3` / `KECCAK256_FUNCT7` in the same file already makes the crate a linked dependency; this statement adds nothing. Same at `differential_fuzz.rs:42` and `:71`. If an explicit registration anchor is wanted, the repo idiom is `extern crate jolt_inlines_keccak256 as _;` (jolt-inlines/fixtures/src/lib.rs:1-8, tracer/examples/trace_*.rs). Otherwise delete the three lines and the comment before someone copies the incantation.

jolt-inlines/keccak256/src/lib.rs
```diff
@@ -5,10 +5,14 @@
 pub const INLINE_OPCODE: u32 = 0x0B;
 
 pub const KECCAK256_FUNCT3: u32 = 0x00;
+pub const KECCAK256_ABSORB_PERMUTE_FUNCT3: u32 = 0x01;
 pub const KECCAK256_FUNCT7: u32 = 0x01;
 pub const KECCAK256_NAME: &str = "KECCAK256_INLINE";
+pub const KECCAK256_ABSORB_PERMUTE_NAME: &str = "KECCAK256_ABSORB_PERMUTE_INLINE";
 
 pub const NUM_LANES: usize = 25;
+pub const RATE_IN_BYTES: usize = 136;
```
Acentelles, line 14: **[Non-critical, cleanup]** **Canonical constant created, shadow copy left behind.** With `RATE_IN_BYTES` / `RATE_IN_U64` now crate-level, `exec.rs:8` still carries a function-local `const RATE_IN_BYTES: usize = 136` in the reference sponge `execute_keccak256` (untouched by the PR), and the PR's own new tests spell the same geometry as literals (`single_input(136, 200)` at `differential_fuzz.rs:52` while importing `RATE_IN_BYTES` at `:18`; `spec.rs:33` vs the new harness at `spec.rs:76-79` which uses the constants). Point `exec.rs` at `crate::RATE_IN_BYTES` and use `RATE_IN_BYTES` / `NUM_LANES * size_of::<u64>()` in the harness calls so the rate and state size have one owner.

crates/jolt-prover/tests/dory_byte_diff.rs
0xAndoroid: Fixed in 96eee7a7f. The clear-mode byte-diff now runs a new `sha3_aligned([[u64; 17]; 2])` entry in the sha3 example guest, so both fused blocks are read from caller memory through `rs2`, followed by the pad-only final block. The ZK e2e keeps the `&[u8]` stack-copy path, and both comments now say which path they cover. `Vec<u64>` was not an option: the no_std guest has no `Vec` deserialization and serde caps fixed arrays at 32 elements.

jolt-inlines/keccak256/src/sdk.rs
0xAndoroid: Fixed in 96eee7a7f. `initialized` and the three `*_first` helpers are gone; `absorb_buffer`, `absorb_final`, and the block loop all call the fused inline. Guest trace for the 300-byte digest: 11,469 -> 11,071 rows.

jolt-inlines/keccak256/src/sdk.rs
0xAndoroid: Fixed in 96eee7a7f. `update` and `digest` share `absorb_full_blocks`, which absorbs complete blocks straight from the input (aligned: fused inline over caller memory; unaligned: one stack copy per block). Only the tail is staged in the buffer.

jolt-inlines/keccak256/src/sdk.rs
0xAndoroid: Fixed in 96eee7a7f. Padding writes go through a `&mut [u8; RATE_IN_BYTES]` view of the buffer, so `bytes[buffer_len] = 0x01` is bounds-checked again. Every unsafe block carries a SAFETY comment.

crates/jolt-prover-legacy/src/zkvm/lookup_table/prefixes/xor_rotl1.rs
0xAndoroid: Fixed in f18c36358. `bind_tail` folds the Boolean pairs into one u64 and reduces the `straddle` / `wrap` coefficients to conditional adds; the outgoing straddle sign is a conditional negation. Phased prefix-suffix tests and the sha3 byte-diff (proof == legacy) pass.

jolt-inlines/keccak256/src/sequence_builder.rs
0xAndoroid: Fixed in e2f0e0f43. `pi_source((X, Y)) = ((X + 3Y) % 5, X)`; `pi_destination` is a `const fn` shared with the reference in `exec.rs`; `PI_TEMP_LANE = pi_destination(RHO_PI_FIRST_SOURCE)` replaces the debug_assert.

jolt-inlines/keccak256/tests/row_count.rs
0xAndoroid: Fixed in e2f0e0f43. `tests/row_count.rs` is deleted; the inline expansion fixture is the only ratchet.

crates/jolt-prover-legacy/src/zkvm/lookup_table/prefixes/xor_rotl1.rs
0xAndoroid: Fixed in f18c36358. One `PrefixState::bound` seeds via `unwrap_or` (straddle defaults to one so `straddle * y` supplies the wrap seed) and binds the pair; `prefix_mle` and `update_prefix_checkpoint` both call it.

jolt-inlines/keccak256/tests/row_count.rs
0xAndoroid: Fixed in e2f0e0f43. The anchor lines are gone: `row_count.rs` is deleted and `differential_fuzz.rs` dropped its two.

jolt-inlines/keccak256/src/lib.rs
0xAndoroid: Fixed in e2f0e0f43. `exec.rs` uses the crate `RATE_IN_BYTES`; the emulator layouts in `spec.rs` and `differential_fuzz.rs` use `RATE_IN_BYTES` and `NUM_LANES * size_of::<u64>()`.

PR #1750, opened 2026-08-07
perf(instructions): fuse RV64 word arithmetic into single lookups — ADDW/ADDIW/SUBW/MULW/SLLIW 2→1 rows, SLLW 3→2

0xAndoroid:
## What

Promotes the W-suffix arithmetic kinds from 2-row virtual sequences (`op` + `sext.w`) to first-class single-row lookups, with **zero new lookup tables**:

| Instruction | Before | After | Sequence today |
|---|---|---|---|
| ADDW / ADDIW / SUBW / MULW | 2 | **1** | op + VirtualSignExtendWord |
| SLLIW | 2 | **1** | VirtualMULI(2^shamt) + sext.w |
| SLLW | 3 | **2** | Pow2W + MUL + sext.w → Pow2W + MULW |

These are among the hottest sequence-bearing kinds in RV64 guests (−1 row per dynamic execution each).

## Why this is sound with zero new tables

- The lookup **index construction is copied verbatim from the 64-bit siblings** (ADD/SUB/MUL/MULI): the operand combination (including the add/sub carry) is materialized into the integer index by the existing linear R1CS instruction-input leg *before* the table is consulted — the fused kinds change only which table projects the index to an output.
- The output projection reuses the shipped `SignExtendHalfWord` table (reads index bits 0..32, index-shape-agnostic, bilinear ≤4-suffix decomposition already in-tree) — the same pattern as the existing {MUL, MULHU} table sharing.
- jolt-riscv already carried `AddW/AddiW/SubW/MulW/SllIW` instruction structs with correct fused-flag sets (unwired until now); `SllIW` additionally gains `MultiplyOperands` (mirroring `MulI`).
- Read-raf sumcheck impact: none — no new tables, no new prefixes/suffixes, no new committed polynomials.

Implementation note: SLLIW's expansion still rewrites `shamt → 2^shamt` in its single row (one-line normalization); the tracer converts multiplier↔shamt directly when tracing the promoted kind to avoid re-entering expansion.

## Testing

- Lookup suites 323/323 (incl. per-kind materialize/constraint-input/trace-output macro tests for the 5 kinds); legacy focused mirror tests
- Tracer 134/134 (differential sequence tests for all six touched kinds); jolt-program 39/39 with expansion parity fixture regenerated — **exactly the 6 kinds' non-rd0 entries changed**, all unrelated and rd=0 hashes byte-identical
- Modular prover acceptance: clear 20/20, ZK 13/13; legacy `muldiv` e2e 3/3 in both `host` and `host,zk`
- `cargo clippy --all --all-targets -- -D warnings` in both feature modes; fmt clean


tracer/src/instruction/mod.rs
```diff
@@ -1000,6 +1007,23 @@ macro_rules! impl_final_jolt_row_data {
 
     (@from_row VirtualAdvice) => {};
 
+    (@from_row SLLIW) => {
```
0xAndoroid: why are we adding this?

tracer/src/instruction/mod.rs
```diff
@@ -983,12 +988,14 @@ macro_rules! impl_final_jolt_row_data {
...
+                        jolt_riscv::instructions::$marker(())
+                    );
                     JoltInstructionRow {
-                        instruction_kind: jolt_riscv::JoltInstruction::$marker(
-                            jolt_riscv::instructions::$marker(())
-                        ),
+                        instruction_kind,
                         address: instr.address as usize,
-                        operands: instr.operands.into(),
+                        operands: jolt_riscv::NormalizedOperands::from(instr.operands)
```
0xAndoroid: why did we change tihs?

crates/jolt-riscv/src/row.rs
```diff
@@ -17,6 +17,17 @@ pub struct NormalizedOperands {
     pub imm: i128,
 }
 
+impl NormalizedOperands {
+    /// Converts decoded source operands into the representation bound by the final Jolt row.
+    #[inline]
+    pub fn materialize_for_jolt(mut self, instruction_kind: JoltInstructionKind) -> Self {
```
0xAndoroid: why do we need this?

tracer/src/instruction/mod.rs
0xAndoroid: Row construction now routes operands through `NormalizedOperands::materialize_for_jolt(instruction_kind)` — the final row must bind the operand representation the lookup consumes, not the decoded one. Identity for every kind except SLLIW (see row.rs comment). The `let` hoist is just because `instruction_kind` is consumed twice (materialization + struct field).

crates/jolt-riscv/src/row.rs
0xAndoroid: It's the one seam mapping decoded source operands → proof-bound operands per kind. Only SLLIW is non-identity: `imm = 1 << (shamt & 0x1f)`. Promoted SLLIW reuses the MULI-shaped multiplication index (rs1 × imm) against SignExtendHalfWord — binding the multiplier `2^shamt` keeps the lookup-query construction byte-identical to MULI's, which is what makes this PR zero-new-tables. All three final-row producers call it (tracer mod.rs:722, mod.rs:998, jolt-riscv row.rs:65), so tracer and jolt-program expansion can't diverge on the bound operand.

tracer/src/instruction/mod.rs
0xAndoroid: Inverse of `materialize_for_jolt` for the one kind where that mapping is non-identity. Rows bind imm = 2^shamt (multiplier form); the tracer's `SLLIW::exec` shifts by shamt — the generic `@from_row` arm would copy the multiplier straight into the tracer instruction and exec would shift by 2^shamt. This arm recovers `shamt = trailing_zeros(imm)` (debug_assert power-of-two) on the row→instruction path, so materialize ∘ from_row = identity for all shamt ∈ [0,32). Without SLLIW-as-final-kind none of this existed because the expansion emitted a real MULI row; promotion to a single fused row is what created the two representations.

0xAndoroid:
- Removed the cross-cutting `materialize_for_jolt` operand-rebinding seam and its inverse/special-case plumbing; zero call sites remain.
- SLLIW is source-only again and expands to a one-row `VirtualMULIW` sequence. The expander materializes `1 << (shamt & 0x1f)`; `VirtualMULIW` reuses `SignExtendHalfWord` with the same one-row dynamic cost.
- Regenerated expansion parity hashes: only SLLIW entries changed; ADDW/ADDIW/SUBW/MULW remain 1 row and SLLW remains 2.
- Green: jolt-lookup-tables, tracer, jolt-program; legacy muldiv host + host,zk; modular prover fixtures clear + zk; clippy host + host,zk; fmt.

0xAndoroid:
Adversarial audit verdict: **SOUND**. Both landing conditions are addressed.

- Ported the reviewer-authored symbolic-exec fixes with attribution: ADDW/ADDIW/SUBW/MULW arms plus low-word truncation before `VirtualMULIW` sign extension. The 13 fused-word Z3 obligations pass, formally covering all inputs across the six expansions.
- Added the seven-test adversarial differential suite (~3.3k boundary cases) and trace-census harness.
- Removed acceptance-gate vacuity: muldiv now emits two directed raw instances of ADDW, ADDIW, SUBW, MULW, SLLIW, and SLLW; clear byte-diff and ZK paths assert the canonical final-row counts before proving. Observed counts: 2/2/2/4/2/2 for ADDW/ADDIW/SUBW/MULW/VirtualMULIW/VirtualPow2W.
- Final gates are green: z3, adversarial, tracer, program, lookup, legacy clear+ZK, modular clear+ZK, clippy clear+ZK, fmt. Modular ZK used `--test-threads=1` after reproducing the known parallel stack overflow.

Proposed follow-up: add a vendored-Z3 CI job, nightly or per PR, because this harness had no running gate and its symbolic arms rotted. I can send that as a separate PR; this PR does not touch `.github/workflows`.

crates/jolt-prover-legacy/tests/word_op_trace_census.rs
```diff
@@ -0,0 +1,68 @@
+//! Review evidence for PR #1750: do the acceptance-suite guests actually
```
0xAndoroid: remove this file

crates/jolt-prover/tests/zk_e2e.rs
```diff
@@ -107,6 +107,28 @@ mod support {
             .expect("modular trace")
     }
 
+    pub fn assert_fused_word_op_coverage(trace_output: &TraceOutput<OwnedTrace>) {
```
0xAndoroid: this seems like a very bespoke low quality test. Either remove or make it nicer

examples/muldiv/guest/src/lib.rs
```diff
@@ -1,12 +1,164 @@
 #![cfg_attr(feature = "guest", no_std)]
 use core::hint::black_box;
 
+#[cfg(target_arch = "riscv64")]
```
0xAndoroid: revert this example back to its original state

crates/jolt-lookup-tables/src/instructions/riscv/mod.rs
```diff
@@ -33,5 +36,18 @@ pub mod slti;
 pub mod sltiu;
 pub mod sltu;
 pub mod sub;
+pub mod subw;
 pub mod xor;
 pub mod xori;
+
+#[inline]
+pub(in crate::instructions) fn sign_extend_half_word<const XLEN: usize>(value: u64) -> u64 {
```
0xAndoroid: its a bad place for this stuff. Also feels like there should be a simpler solution to this sign extension that we can just inline into the functions that use it

crates/jolt-lookup-tables/tests/adversarial_word_ops.rs
```diff
@@ -0,0 +1,470 @@
+//! Adversarial differential tests for the fused RV64 word-arithmetic lookups
```
0xAndoroid: once again. Very bespoke test just for this PR. Should be removed.

crates/jolt-program/src/expand/tests.rs
```diff
@@ -412,12 +412,40 @@ fn source_only_expanders_are_not_target_legal() {
...
         CSRRW, CSRRS, EBREAK, ECALL, MRET,
-        SLL, SLLI, SLLW, SLLIW, SRL, SRLI, SRA, SRAI,
+        SLL, SLLI, SLLIW, SLLW, SRL, SRLI, SRA, SRAI,
         SRLIW, SRAIW, SRLW, SRAW,
     }
     assert_eq!(SourceInstructionKind::Inline.jolt_kind(), None);
 }
 
+#[test]
+fn fused_word_ops_have_expected_row_counts() -> Result<(), ExpansionError> {
```
0xAndoroid: seems bespoke test as well

crates/jolt-prover-legacy/tests/word_op_trace_census.rs
0xAndoroid: Removed word_op_trace_census.rs; no Cargo/test wiring existed. Fixed in 0d3c03658f.

crates/jolt-prover/tests/zk_e2e.rs
0xAndoroid: Removed the canonical-name coverage helper/call from zk_e2e.rs and its byte_diff.rs sibling for consistency after the guest revert. Fixed in 0d3c03658f.

examples/muldiv/guest/src/lib.rs
0xAndoroid: Restored examples/muldiv/guest/src/lib.rs byte-for-byte from origin/main. Fixed in 0d3c03658f.

crates/jolt-lookup-tables/src/instructions/riscv/mod.rs
0xAndoroid: Deleted the shared sign-extension helpers in both lookup stacks. Each corresponding output now inlines the operation followed by as u32 as i32 as i64 as u64. Fixed in 0d3c03658f.

crates/jolt-lookup-tables/tests/adversarial_word_ops.rs
0xAndoroid: Removed adversarial_word_ops.rs. Fixed in 0d3c03658f.

crates/jolt-program/src/expand/tests.rs
0xAndoroid: Removed the bespoke fused row-count test. The golden parity fixture already pins these expansion outputs by hash, so the extra test was redundant. Fixed in 0d3c03658f.

moodlezoup (approved):

PR #1751, opened 2026-08-07
perf(zkvm): magnitude reformulation of signed division — DIV 24->12, REM 24->9, DIVW 21->14, REMW 21->11 rows

0xAndoroid:
## What

Rewrites the signed-division virtual sequences over unsigned magnitudes, halving the family:

| Instruction | Before | After |
|---|---|---|
| DIV | 24 | **12** |
| REM | 24 | **9** |
| DIVW | 21 | **14** |
| REMW | 21 | **11** |
| **Σ family** | 90 | **46 (−49%)** |

Two commits: (A) magnitude reformulation using existing instructions only; (B) a new `VirtualNegateIf` virtual instruction (`rd = rs1 < 0 ? −rs2 : rs2`) that collapses the remaining sign-handling rows. **Net lookup-table count: −1** (adds NegateIf, deletes ChangeDivisor + ChangeDivisorW, which the reformulation makes dead — plus five orphaned signed-remainder prefix variants, removed after a zero-reference audit).

## Why the old chain was fat

42% of the old DIV sequence was a signed no-overflow gadget (MULH + MUL + SRAI + assert-EQ = 10 rows). Restating the division contract over unsigned magnitudes replaces it with the existing 1-row `MulUNoOverflow` assert — and the RISC-V MIN/−1 overflow case falls out of mod-2^64 wraparound for free, which is what kills the ChangeDivisor guards entirely.

## Soundness

The chain witnesses one advice value per instruction (quotient magnitude; previously two) and pins every output against adversarial advice:

- **d = 0**: `VirtualAssertValidDiv0` forces q = 2^64−1; REM's result path forces r = dividend via the LTE + exact-SUB + valid-unsigned-remainder chain.
- **MIN / −1**: |d| = 1 forces the unsigned quotient magnitude to 2^63; the sign recombination wraps to exactly MIN (DIV) / 0 (REM) mod 2^64.
- General case: uniqueness of (q, r) with `0 ≤ r < |d|` is enforced by the MulUNoOverflow + exact-MUL + LTE + valid-unsigned-remainder rows over magnitudes — the same shape the (already lean, untouched) unsigned family uses.
- `VirtualNegateIf` decomposes bilinearly with 2 new prefix variants and only shipped suffixes (One / RightOperand / RightOperandIsZero) — within the 4-suffix budget; negation-at-MIN wraps correctly (2^63 fixed point). Implemented in both the modular (`crates/jolt-lookup-tables`) and legacy stacks with index parity.

No BlindFold/claim-derivation surface is touched (no sumcheck instances change).

## Testing

- Tracer inline-sequence differential tests: 1,000 random cases per instruction **plus directed edge vectors** — (d=0), (MIN,−1), (MIN,1), (x,MIN), (0,d) and word equivalents with upper-word garbage (random u64s never hit these)
- `VirtualNegateIf`: mle_full_hypercube / mle_random / phased prefix_suffix, both stacks
- **The muldiv e2e guest gains 8 raw signed ops** (DIV/REM/DIVW/REMW × {d=0, MIN/−1} under black_box) — the signed paths now have proving+verifying coverage in both `host` and `host,zk`, and through the modular prover acceptance suites (clear 20/20, ZK 13/13)
- Expansion parity fixture re-baselined for the 4×3 division entries only; unrelated hashes unchanged
- Clippy both feature modes with `-D warnings`; fmt clean

Rebased on current main (post-#1748).


markosg04:
Could we wire the signed DIV/REM Z3 obligations into the nightly at reduced width? Z3_VERIFIER_BV_BITS=8 is supported, but these tests are #[ignore] and the nightly never opts into ignored tests, so they currently do not run at any width. Since the tracer/e2e coverage uses honest advice, a filtered 8-bit correctness + consistency run for DIV/REM/DIVW/REMW seems like the most useful mechanical check that arbitrary advice cannot produce a wrong or non-unique result.

markosg04 (approved):
we should really add the lean certificates for each of these soon

0xAndoroid:
Wired in 32339cf34: the nightly now has a second step running the eight signed DIV/REM/DIVW/REMW correctness+consistency proofs with `Z3_VERIFIER_BV_BITS=8` (`--run-ignored=only` + a name filter, `!cancelled()` so a main-suite failure doesn't hide its verdict; the existing failure-issue step covers both). All eight discharge in <1s locally at that width. Also refreshed the workflow's stale scope census (86/106 tests, LW now modeled) and the book's nightly-CI note.

PR #1754, opened 2026-08-07
perf(zkvm): read-raf hygiene — present-gated prefix materialization + single-pass legacy bucketing

0xAndoroid:
## What

Three instruction read-RAF cost reductions across the modular and legacy provers. This is a performance-only refactor: proof bytes remain unchanged, and the clear modular-vs-legacy byte-diff suite passes without fixture changes.

### 1. Modular: materialize only prefixes referenced by present tables

The optimized kernel previously materialized and phase-bound every `ALL_PREFIXES` chunk polynomial, including families absent from the trace. It now derives the referenced-prefix set from non-empty table buckets and materializes/binds only those prefixes.

Per-thread scratch rows (`p0`/`p2`) remain full-size and discriminant-indexed because `combine` reads prefix evaluations by the `Prefixes` discriminant. Compacting those rows would misalign table combines. Checkpoints for absent prefixes remain at their defaults and are never read.

### 2. Modular + legacy: ordered parallel table bucketing

Both provers need cycle indices grouped by lookup table. Before this PR:

- Modular scanned all `T` rows serially.
- Legacy scanned all `T` rows once per table: `Θ(N_tables × T)` total work.

Parallel builds now use two ordered `O(T)` passes:

1. Count each table's entries inside indexed trace chunks in parallel.
2. Allocate exact `MaybeUninit` buckets, avoiding zero-fill and a second full index set.
3. Fill disjoint bucket slices in parallel, preserving ascending cycle order.

The modular crate retains its serial implementation when built without the `parallel` feature.

### Scope

These changes apply only to the instruction read-RAF sumcheck. Bytecode read-RAF and other sumchecks are unchanged.

## Performance

Focused A/B runs on an M4 Max with 128 GiB RAM and 16 Rayon threads. Inputs used production-sized row layouts and identical synthetic table distributions; every comparison produced equal bucket checksums.

| Prover | Trace | Input rows | Serial `O(T)` | Ordered parallel `O(T)` | Speedup |
|---|---:|---:|---:|---:|---:|
| Modular | `2^30` | 48 GiB | 1.627 s | 391 ms | **4.16×** |
| Modular | `2^31` | 96 GiB | 14.70 s | 9.72 s | **1.51×** |
| Legacy | `2^31` | 64 GiB | 3.769 s | 640 ms | **5.89×** |

The modular `2^31` case crosses this machine's memory-compression threshold, reducing the parallel gain. `2^30` is the largest tested size below that threshold. The absent-prefix win scales with the number of unused prefix families.

## Testing

- `jolt-kernels`: **158/158**, including exact bucket-order and read-RAF reference-parity tests.
- Modular clear acceptance: **20/20**, including byte-diff checks with no fixture changes.
- Modular ZK acceptance: **13/13** serial.
- Legacy instruction lookups: **57/57**; legacy muldiv: **3/3** in both `host` and `host,zk`.
- Workspace clippy passes in `host` and `host,zk` modes with `-D warnings`; `cargo fmt` passes.


markosg04 (approved):
no changes to the verifier codepaths

PR #1755, opened 2026-08-07
perf(zkvm): mask-mediated lane framework for sub-word memory — loads 8-9 -> 4-5 rows, stores 13-15 -> 9-10

0xAndoroid:
## What

Rewrites the nine sub-word memory expansions around table-built lane masks — the hottest sequence family in the ISA drops **Σ93 → 57 rows (−38.7%)**:

| Instruction | Before | After | New shape |
|---|---|---|---|
| LB / LBU | 8 | **4** | AlignAddr + LD + LaneMaskB + extract |
| LH / LHU / LWU | 9 | **5** | + halfword/word masks (LWU zero-extends via the generalized SRL) |
| LW | 8 | **5** | sign-extends via LaneExtractS |
| SB | 13 | **9** | masked-XOR merge (was a 5-row mask build + RMW) |
| SH / SW | 14 / 15 | **10** | same framework |

Atomics are untouched at the source level; LR.W/SC.W fixture entries move only because their expansions nest the shared LW/SW recipes (−8 rows ride along).

## How

Four stages, one commit each:

1. **C0 algebra pack** (zero tables): duplicate-Pow2 hoisting in stores, cheaper LBU/LWU tails.
2. **SRL-is-PEXT generalization**: the shipped `VirtualSRL` table is already a general parallel-bit-extract (`Σ x_k·y_k·2^{pop(y<k)}`); its prefix/suffix closed forms (`trailing_zeros`/`leading_ones`) are generalized to window-PEXT/popcount forms that are **value-identical on every previously-reachable input** — gated by the full existing shift suite passing with a byte-identical expansion fixture before any new table lands. Unsigned lane extraction then reuses this table with a lane-mask operand.
3. **Five trivial tables** (AlignAddr, LaneMaskB/H/W, Pow2Lane) + **LaneExtractS** for signed extraction — exactly 4 prefixes / 4 suffixes, the shipped `VirtualSRA` arity ceiling. Sign-fill algebra: `PEXT(D,M) + T·2^64 − 2S` (T = lane-top detector, S = top weight) ≡ exact two's-complement fill for every contiguous lane, with a straddle-cancellation term at phase boundaries.
4. **Nine expansion rewires** to the recipes above (alignment asserts preserved; misalignment still panics on both execute and trace paths).

## A found-and-fixed production bug (disclosure)

During e2e gating, clear-mode muldiv **rejected at stage 5**: the shared prefix-checkpoint updater was missing dependencies for the three lane-mask tables — their standalone and phased tests passed while production checkpointing diverged. Fix: each mask's base constant enters the first final-phase checkpoint exactly once (after the last pair binds). A regression test now pins the **post-final-round checkpoint state that production actually consumes**, closing the test-shape gap that let this slip. Reviewers may want to eye `prefixes/lane_mask.rs` closely — the fix was independently re-derived and confirmed at both phase widths during internal review.

## Soundness notes

- Store merge `D' = D ⊕ ((D ⊕ rs2·2^{8o}) & M)` is bit-local: out-of-lane bits of the shifted value land above the lane or wrap past 2^64; adversarial old content and sign bits are killed by `&M`.
- **z3 formal coverage**: `symbolic_exec` arms added for all six new virtual instructions — the z3 suite proves all nine rewritten expansions ∀-inputs (vendored Z3: 83/83).
- Read-raf cost: +6 used tables ≈ 180–360K field mults/proof (~0.01–0.02% at 2^22, present-gated — guests without sub-word ops pay zero), verifier +~1.8K mults, proof +192 B. A fixed set serving nine architectural instructions in essentially every real guest.

## Testing

- Lookup suites 346/346 (incl. full-hypercube materialize==MLE for every new table, phased prefix/suffix at both geometries, cross-stack MLE parity, ABI append-only); tracer 136/136 with directed vectors (all offsets × sign-boundary payloads × misalignment) and exact per-instruction row locks
- Legacy muldiv e2e 3/3 both modes; BTreeMap proof e2e; modular acceptance clear 20/20 + ZK 13/13 (dynamic coverage: sha2-chain exercises LB/LBU/LW/SB/SW; BTreeMap exercises LB/LBU/LHU/SB/SH; directed tests cover the remainder)
- Expansion parity fixture: exactly 102 hashes moved (9 rewritten families + LR.W/SC.W nesting), all others byte-identical
- Clippy both feature modes `-D warnings`; fmt. Rebased on current main.


0xAndoroid:
This is being replaced by #1761 and stack.

PR #1757, opened 2026-08-07
perf(inlines): SHA-256 big-endian ingest + weak-sigma polish - 2,436 -> 2,300 rows/block, -428/digest e2e

0xAndoroid:
## What

SHA-256 inline optimization with no new lookup tables.

| Metric | Baseline | This PR | Delta |
|---|---:|---:|---:|
| Custom-IV inline rows/block | 2,436 | **2,300** | **-136** |
| Fixed-IV inline rows/block | 2,396 | **2,260** | **-136** |
| `sha2-chain` guest rows/digest | 3,063 | **2,635** | **-428 (-14%)** |

- **Big-endian ingest:** the existing `SHA256_INLINE` and `SHA256_INIT_INLINE` entries now read raw 64-byte blocks in SHA-256 big-endian byte order. Each `LD` packs two words, `VirtualRev8W` reverses both 32-bit lanes, and `SRLI` extracts the upper lane. Eight added inline rows replace the guest byte-swap pass, removing about 284 guest rows/block.
- **Weak sigma reduction:** sigma pairs use the existing Blake `VirtualXORROTW7` table where the rotation algebra permits it, plus `ROTRIW` + `ANDI` for logical word shifts. This saves 144 inline rows/block; net savings are 136 after big-endian ingest.

## Direct-inline ABI change

The two existing opcode/funct selectors changed in place: their input pointer now addresses raw big-endian bytes, not 16 native-endian `u32` words. Direct callers that pre-swap words must pass the original block bytes instead. The safe `Sha256` API is unchanged.

## Testing

- `cargo nextest run -p jolt-inlines-sha2 --features host --cargo-quiet` - 12/12
- `cargo nextest run -p jolt-inlines-fixtures --cargo-quiet` - 3/3, including registered expansion golden fixtures
- Modular-vs-legacy `sha2-chain` e2e at 2^17 - byte-identical


markosg04 (approved):

0xAndoroid:
Adversarial review pass complete. Verdict: code sound, including the Rev8W I-shaped heal (verified against decode normalization; modular-vs-legacy sha2-chain e2e byte-identical at 2^17). Findings:

- PR body is materially stale vs head c24fc3d6: it describes additive BE entry points with 4 variants at 2,292/2,252 rows and a committed probe/ratchet-test/100K-fuzz suite ("18/18") — the head instead rewrites the existing two inlines to BE-only (fixture counts 2,300/2,260), and the probe/ratchet/fuzz were removed in the Aug-17 cleanup (shipped suite = 12 tests). The in-place ABI break is unmentioned. Body needs a rewrite before review.
- jolt-inlines-sha2 stays at 0.1.0 while the memory contract for the same opcode changed — a stale out-of-tree guest against a new host yields silently wrong digests; consider a version bump.
- jolt-riscv doc comment on Rev8W says "lower 32 bits" — stale/wrong (both 32-bit halves are reversed).
- Sequencing: if #1753 (SRLIW 3→1) lands, word_shr becomes a σ₁ pessimization (−48/block) and fixture counts go stale (loud CI failure; known follow-up).
- byte_diff.rs calibration comment "~3396 cycles/hash" is now ~2968 (cosmetic).

— adversarial review by Claude (Fable), posted via Pika

PR #1761, opened 2026-08-10
feat: fused sub-word extraction for LW (byte-addressable memory, Tier 0 slice 1)

Acentelles:
Stacked on #1725. First implementation slice of the spec's Tier 0: the `LW` expansion drops from 8 traced cycles to 6 by replacing the shift/mask extraction tail (`SLLI` + two-row `SRL` + `VirtualSignExtendWord`) with two fused lookups.

## New tables

- **`WindowMaskW(ea)`** = `0xFFFFFFFF << 32*ea_2`: byte mask of the addressed word lane. Reads only bit 2 (bits 0-1 are zeroed by the word-alignment assert), so it decomposes rank-1 via one new `Pow2OffsetW` prefix/suffix pair.
- **`PextSigned(x, mask)`** = `pext(x, mask) + sign * (2^64 - 2^popcount(mask))`: width-independent sign-extending parallel extract. One table will serve LB/LH/LW once the remaining mask tables land. Its decomposition (2 new prefixes, 4 new suffixes, reusing `RightShift` and `RightOperandIsZero`) is valid on the full index domain, so its tests run unrestricted.

## Design note (spec open question 3)

The spec's fused per-offset extract tables have intrinsic prefix rank >= 8 per width, so this follows the existing SRL/SRA binary-to-positional pattern instead: one rank-1 lookup produces the window mask, one constant-rank lookup extracts through it. Costs one extra cycle per access vs the spec's fused projection; saves ~14 prefix/suffix families per width. Details in the commit message.

## x86 AOT backend

The two new final kinds are lowered natively in `jolt-tracer-x86` (they appear in every word-load, so the AOT backend cannot decline them): `WindowMaskW` as four ALU ops, `PextSigned` as BMI2 `pext` + `popcnt` followed by a `shl`/`sar` pair by `64 - popcount` whose mod-64 `cl` masking makes the `popcount ∈ {0, 64}` edges fall out branch-free. Compilation refuses pre-BMI2/POPCNT hosts, gated per program so pext-free guests keep compiling (the BMI1 gate above it stays unconditional, as before). Both kinds join the per-instruction differential suite (1000 random instances each vs the reference interpreter).

## Review fixes

- **`Pow2OffsetW` is phase-boundary-agnostic.** The prefix previously assumed phase boundaries never fall inside the low three index bits (debug-assert only; a future config with a 1-2 bit suffix window would silently drop the `2^(32*ea_2)` factor in release). Both stacks now locate bit 2 relative to the suffix / window / checkpoint split, the same way the `WindowSign` prefixes already do, and the legacy checkpoint update keys on the bound bit's position instead of the final-phase round number. Both prefix-suffix harnesses gained a phase-size parameter, and WindowMaskW pins the behavior with two-round phases (`suffix_len` hits 2); the new tests fail against the old prefix in both stacks.
- **The rewritten LW sequence is formally verified.** `test_sequence!(LW)` is enabled in z3-verifier: `SymbolicCpu` gained a doubleword-granular symbolic RAM array (shared between sequence and expected models, equated between the consistency pair), an `LD` arm, and the LW expected model (sign-extended word lane of `RAM[ea & !7]` selected by bit 2 of `ea`). Both obligations discharge in ~0.3s at the 64-bit model width, so they run un-ignored; the new `VirtualWindowMaskW`/`VirtualPextSigned` symbolic models are live code.
- **`MAX_SUFFIXES` is pinned structurally.** Hoisted to module scope and asserted equal to the actual maximum suffix count over `LookupTables::iter()`, so the next table exceeding it fails a unit test instead of a release-mode bounds check mid-proof.
- **One `pext` and one sigma convention per stack.** Each stack gains shared `pext(x, y)` (O(1) shift-and-truncate for contiguous masks, set-bits iteration otherwise) and `window_sign_bit(x, y)` helpers: the eight verbatim window-sign copies collapse onto the latter, the legacy `RightShiftPrefix` evaluates through the `Pext` suffix pair instead of hand-inlining it (the pairing must stay bit-identical for `PextSignedTable::combine` to be sound), and the five `pext_signed` copies reduce to one helper per crate, all in the `2^XLEN - 2^popcount` formulation. The tracer interpreter drops its unconditional 64-iteration loop for the same shape (equivalence checked exhaustively at XLEN=8 and on 250k random 64-bit inputs).
- **The decomposition's XLEN=64-boundness is a compile-time fact.** `WindowMaskW`'s `materialize_entry`/`evaluate_mle` stay genuinely XLEN-generic, but inline const asserts in `suffixes()`, `combine()`, and the legacy `Pow2OffsetW` prefix now make any non-64 instantiation of the decomposition a compile error instead of a silent release-mode unsoundness.

## Notable side changes

- Legacy `RightShiftPrefix` generalized from the tail-mask shortcut to the faithful recursion (identical on all tail bitmasks; required for window masks; matches the modular implementation).
- Legacy `MAX_SUFFIXES` 4 -> 5 (`PextSigned` is the first 5-suffix table; caught by the muldiv e2e, now also pinned by the unit test above).
- 18 golden expansion-fixture hashes re-baselined (LW, plus LRW/SCW which embed the word-load expansion).

## Validation

muldiv e2e in `host` and `host,zk`; jolt-lookup-tables (incl. exhaustive XLEN=8 hypercube for both tables and the two-round-phase boundary tests); tracer 127/127; jolt-program 38/38; prover-abi-tests; jolt-tracer-x86 per-instruction difftests; z3-verifier LW correctness + consistency; fmt + clippy both modes.

Follow-ups (next slices): unsigned `Pext` + `WindowMaskB/H` for the remaining loads (each also needs its z3 expected model to un-comment its `test_sequence!`), then the store side.


crates/jolt-lookup-tables/src/tables/prefixes/pow2_offset_w.rs
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    fn evaluate(checkpoints: &[PrefixEval<F>], b: LookupBits, suffix_len: usize) -> F {
+        // Bit 2 of the raw index stays in the suffix until the final phase.
+        if suffix_len != 0 {
+            return F::one();
+        }
+        let bits: u128 = b.into();
+        let bit2 = ((bits >> 2) & 1) as u32;
+        checkpoints[Prefixes::Pow2OffsetW] * F::from_u64(1u64 << (32 * bit2))
```
Acentelles: The `32` here (and in the suffix, and the legacy twin's `(1u64 << 32) - 1`, which meanwhile keys its checkpoint on the generic `2 * XLEN - 3`) is really `XLEN/2`, but `WindowMaskWTable` is XLEN-generic and gets instantiated at 8 in its hypercube test. Only the XLEN=64 decomposition path is exercised today so nothing fails, but a small-XLEN prefix_suffix test in slice 2 would be silently wrong. Either derive it from `XLEN/2` or add the `debug_assert_eq!(XLEN, 64)` guard `VirtualXORROTWTable` uses in `suffixes()`/`combine()`. Related: if WindowMaskB/H land as planned, a single const-generic `Pow2Offset<LOW_BIT>` (W=2, H=1, B=0) would avoid a third and fourth copy of this state machine, at the cost of one rename now.

crates/jolt-lookup-tables/src/tables/suffixes/pow2_offset_w.rs
```diff
@@ -0,0 +1,16 @@
+use super::SparseDenseSuffix;
+use crate::lookup_bits::LookupBits;
+
+/// `2^(32·ea_2)` where `ea_2` is bit 2 of the (non-interleaved) lookup index.
+pub enum Pow2OffsetWSuffix {}
+
+impl SparseDenseSuffix for Pow2OffsetWSuffix {
+    fn suffix_mle(b: LookupBits) -> u64 {
+        if b.len() < 3 {
```
Acentelles, line 9: This pairs with the prefix's `suffix_len != 0` gate only because suffix lengths are multiples of the chunk size: the suffix returns 1 for `b.len() < 3` while the prefix returns 1 whenever `suffix_len != 0`, so a config that ever put the boundary at suffix_len 1 or 2 would drop the `2^(32·ea_2)` factor entirely. Unreachable with log_m ∈ {8, 16}, but cheap to pin with a `debug_assert!(suffix_len == 0 || suffix_len >= 3)` on the prefix side (both stacks and the legacy twin share the assumption).

z3-verifier/src/virtual_sequences.rs
```diff
@@ -380,6 +402,22 @@ fn symbolic_exec(instr: &Instruction, cpu: &mut SymbolicCpu) {
...
+        Instruction::VirtualPextSigned(VirtualPextSigned { operands, .. }) => {
+            // Faithful for contiguous nonzero masks (the only shape the
+            // window-mask instructions produce): sign-extending extract via
+            // shift-left then arithmetic shift-right.
+            let rs1 = cpu.x[operands.rs1 as usize].clone();
+            let rs2 = cpu.x[operands.rs2 as usize].clone();
+            let tz = trailing_zeros(&rs2, cpu.bv_bits);
+            let lz = leading_zeros(&rs2, cpu.bv_bits);
+            cpu.x[operands.rd as usize] = rs1.bvshl(&lz).bvashr(lz + tz);
+        }
```
Acentelles, line 475: Two latent issues in these models, both currently unobservable because the memory-op sequences are commented out of `test_sequence` and z3-verifier isn't in CI: (1) the shl/ashr PextSigned model is only faithful for contiguous masks, and since `cpu.asserts` are consumed as solver assumptions, a contiguity assert would vacuously exclude exactly the misuse cases — an `ite` on `contiguous(rs2)` falling back to a fresh unconstrained BV would make any future non-contiguous use fail verification loudly instead of certifying wrong semantics. (2) `0xFFFF_FFFF` is hardcoded while the shift above it scales with `cpu.word_bits`; under a reduced-width solver model (the tractable way to ever test the LW sequence here) the mask is wrong — worth deriving from `word_bits`.

Acentelles (commented):
Deep review (semantic consistency, independent MLE re-derivation, soundness, maintainability — plus test runs at this head: jolt-lookup-tables 320/320, legacy prefix_suffix/mle filter 183/183 incl. srl/sra/rotr, tracer 127/127, jolt-program 38/38). No soundness or completeness bugs found. The LW constraint chain is fully pinned (right operand forced to 0 by the flag combination, lane select and load address both derive from the same v0 register write), the PextSigned/WindowMaskW MLEs and the 5-term combine split verify independently including non-contiguous masks and the sign-bit-straddles-boundary case, the RightShiftPrefix generalization agrees with the old shortcut on every reachable SRL/SRA/ROTR input (their materialize_entry was already the total recurrence), and the Pow2OffsetW round bookkeeping checks out against both stacks' calling conventions (bit 2 = odd round 125 in legacy; the modular kernel only calls evaluate at phase starts with full binary chunks). Three latent-hazard comments inline.

One nit that can't anchor inline (file not in the diff): the "46 table prefixes" doc comments in crates/jolt-kernels/src/reference/instruction_read_raf.rs (lines 14, 176, 179) are stale now that this PR brings the count to 49.

Acentelles:
Addressed the review findings in 8461ab592 + 101a45d2e:

- `debug_assert_eq!(XLEN, 64)` in `WindowMaskWTable::suffixes()`/`combine()` (both stacks) guarding the hardcoded 32-bit lane width in the Pow2OffsetW pair; the const-generic `Pow2Offset<LOW_BIT>` refactor is deferred to the WindowMaskB/H slice.
- `debug_assert!(suffix_len == 0 || suffix_len >= 3)` pinning the phase-boundary assumption in both stacks' Pow2OffsetW prefixes.
- z3: the PextSigned model now ite-guards on symbolic mask contiguity and havocs rd with a `BV::fresh_const` otherwise, so non-contiguous misuse fails verification loudly; the WindowMaskW mask derives from `word_bits` instead of a hardcoded `0xFFFF_FFFF`.
- Kernel docs no longer embed the prefix count (was stale at 46).

Validated: jolt-lookup-tables 320/320, legacy prefix_suffix/window_mask/pext filter 55/55, `cargo check`+clippy clean on jolt-kernels and z3-verifier (tests target), fmt applied.

0xAndoroid:
Merged origin/main to resolve the conflicts from #1725/#1748/#1750 landing (merge commit 69915a5dc, no rebase/force-push — done with Claude Code on Andrew's behalf).

Resolution notes:
- `specs/byte-addressable-memory.md`: took main's version (the final #1725 squash supersedes the draft this branch was stacked on).
- Registry files: union of both sides; kept main's removal of the dead `VirtualLW`/`VirtualSW` (#1748).
- Tag collision with the merged wordops: main owns `0x0089`–`0x008e`, so `VirtualWindowMaskW`/`VirtualPextSigned` moved to `0x008f`/`0x0090`, and the expansion-parity fixtures were re-baselined for the new tags (same 18 LW/LRW/SCW entries in the main fixture + 4 inline-fixture lines; all wordops entries match main byte-for-byte).

Local gates: muldiv e2e `host` + `host,zk`; jolt-prover `prover-fixtures` 21/21 and `prover-fixtures,zk` 14/14; jolt-lookup-tables 335 + ABI; jolt-program 38; jolt-riscv 37; tracer 137; jolt-kernels(zk) 156; fmt + clippy both modes clean.

0xAndoroid:
This is currently put on pause as we decide whether we should integrate #1755 ideas into this PR.

Acentelles:
With #1755 now closed in favor of this stack, here is a concrete idea-by-idea assessment to unpause the decision.

**Row accounting** (the nine sub-word expansions, spec-branch baseline Σ93): this stack lands at Σ60 (loads 5/5/6/6/6/6, stores 8/9/9); #1755 was Σ57 (loads 4/4/5/5/5/5, stores 9/10/10). The two designs win on opposite halves, so the best combination is Σ54.

**Worth integrating:**

1. **Address fusion on loads** (#1755's AlignAddr, the entire source of its load advantage). In this stack it means: (i) a new rank-1 `AlignAddr` table indexed by `rs1+imm` via `AddOperands + LeftOperandIsRs1Value + RightOperandIsImm` (the exact `AssertWordAlignment` flag pattern, so no new binding machinery) with output `(sum mod 2^64) & !7` and MLE `Σ_{k=3..63} 2^k·b_k`, replacing the ADDI+ANDI pair with one row; (ii) the WindowMask instructions take the real imm instead of 0, indexing `rs1+imm` directly (they only read the low 3 bits, where the unwrapped sum and the wrapped ADDI agree, same argument as the asserts). Loads drop to 4/4/5/5/5/5. Stores keep the ADDI because ShiftData consumes raw ea as a register operand, staying at 8/9/9, already one row under #1755's stores. Net Σ93 -> 54, better than either PR alone.
2. **z3 enablement**: #1755 turned on forall-input z3 proofs for all nine rewritten expansions (83/83). This stack added the z3 arms but the memory sequences are still commented out of `test_sequence` (flagged in the reviews). Porting their test enablement closes that gap at near-zero cost.
3. **The checkpoint regression-test shape** from their disclosed production bug (standalone and phased prefix tests green while production checkpointing diverged). This stack's Pow2Offset/OffsetScale/ShiftData checkpoint machinery is exactly that shape; it is validated here by round-by-round simulation and the e2e suites, but pinning the post-final-round checkpoint state that production consumes, as their regression test does, cheaply hardens against the same test-shape gap.
4. **Optional, follow-up tier — SRL-is-PEXT reuse**: dropping this stack's `PextTable` by reusing the generalized SRL table for unsigned extraction. The prefix half is already generalized here (#1761); the remaining step is generalizing the RightShift/RightShiftHelper suffixes under the same value-identity gate #1755 used. Saves one table and two suffixes but touches shipped SRL/SRA paths; not a blocker.

**Not worth taking:** the store merge (their XOR/AND/XOR masked-merge costs 9/10/10 vs ANDN + ShiftData + ADD at 8/9/9), and LaneExtractS (its sign-fill needs straddle-cancellation terms at phase boundaries, while PextSigned is total on the full index domain with no straddle handling and one width-independent table serves LB/LH/LW).

**Suggested path:** unpause and land the stack as reviewed, then a follow-up slice for items 1-3 — or fold item 1 into #1762 if we would rather not re-baseline the load fixtures twice. Either way item 1 re-baselines every load fixture once more.

0xAndoroid (approved):
Everything looks great, thanks

PR #1762, opened 2026-08-10
feat: fused extraction for the remaining loads (LB/LBU/LH/LHU/LWU)

Acentelles:
Stacked on #1761. Extends its WindowMaskW/PextSigned pattern to the full load side.

- **New tables**: `Pext` (unsigned parallel extract; the unsigned sibling of `PextSigned`, which is width-independent and reused as-is for signed LB/LH), `WindowMaskB` (reads the low 3 address bits), `WindowMaskH` (reads bits 1-2 only; bit 0 is zeroed by the halfword-alignment assert). Each mask table is a rank-1 decomposition via its own `Pow2Offset` prefix/suffix pair, total on the full domain with no u64 clipping.
- **New instructions**: `VirtualPext` (0x008b), `VirtualWindowMaskB` (0x008c), `VirtualWindowMaskH` (0x008d), across the usual ring including legacy mirrors and z3-verifier.
- **Expansions**: LB/LBU 8 -> 5 traced cycles, LH/LHU 9 -> 6, LWU 9 -> 6 (further reduced by the address-fusion update below).
- **Fixtures**: 60 golden parity hashes re-baselined (12 each). The registered-inline parity fixture needs no changes (verified: no inline embeds these loads).

Validation: modular + legacy table/binding tests (30 + 25), ABI test, jolt-program 38/38, muldiv e2e in `host` and `host,zk`. Trace-length deltas to follow in a comment.

## Update: address fusion (folded from #1755)

A new rank-1 `AlignAddr` table (`VirtualAlignAddr`, 0x0091; 0x008e-0x0090 reserved by the stores slice) computes `(rs1 + imm) & !7` in one `AddOperands + RightOperandIsImm` lookup, replacing every load's ADDI + ANDI pair; the WindowMask instructions take the immediate directly, indexing the unwrapped `rs1 + imm` sum (its low 3 bits agree with the wrapped ADDI, same argument as the alignment asserts). Loads drop to **LB/LBU 4, LH/LHU/LW/LWU 5** traced cycles, matching #1755's load counts. The decomposition is the LowerWord accumulator with bits 2-0 cleared and the carry dropped, `prefix * one + suffix`. 78 further golden hashes re-baselined (all six loads plus LRW/SCW). Re-validated: modular tables 344/344, legacy filter 75/75, jolt-program 38/38, tracer 127/127, muldiv e2e in `host` and `host,zk`, clippy both modes.


Acentelles:
Trace-length deltas vs the spec-branch baseline (`8a013391f`), deterministic counts with identical inputs:

| workload | before | after | delta |
|---|---|---|---|
| btreemap (400 ops) | 370,269 | 354,453 | **-4.27%** |
| sha2-chain (30 iters) | 92,969 | 92,237 | -0.79% |
| fibonacci (100k) | 1,101,216 | 1,101,184 | -0.003% |
| memory-ops | 924 | 891 | -3.6% |

btreemap matches the static estimate (-3 cycles x ~5.3k sub-word loads).

Acentelles:
Prover wall-time measurements (M3 laptop on legacy prover):

**1. Fixed padded size: parity.** btreemap at the default input (both sides pad to 2^16): baseline 1.47/1.49/1.50/1.49/1.51 s vs loads 1.48/1.51/1.52/1.47/1.47 s (N=5 interleaved, ±1.5%). A trace reduction that stays inside the same power-of-two instance is invisible in wall time, as expected — and the two extra globally-materialized prefixes cost nothing measurable.

**2. Padding-boundary crossing: 1.37x.** btreemap with `--target-trace-size 3217000` (2075 ops, identical input both sides): baseline traces 2,130,815 cycles -> pads to 2^22; loads traces 2,041,907 -> pads to 2^21. Result over 3 interleaved rounds: baseline 10.99/11.30/11.06 s vs loads 8.02/8.02/8.13 s, i.e. **-27% wall time** for the same guest computation, with the padded-kHz output confirming the halved instance. The gap to 2x is the prover's sub-linear components (Dory tier-2 ~ sqrt(T)), which are ~45% of the baseline at this scale and amortize further at larger instances.

**Framing for this stack**: the reliable benefit is guest cycles per padded instance (btreemap -4.27%, growing with the store-side follow-up); wall-time wins appear when workloads cross padding boundaries, and are large when they do.

Acentelles (commented):
Deep review (semantic consistency, independent MLE re-derivation, soundness, maintainability, plus test runs at this head: jolt-lookup-tables 338/338, legacy prefix_suffix/window_mask/pext filter 70/70, jolt-program 38/38, tracer 127/127 after one load-induced flake of tests::test_checkpoints that passes standalone and on suite rerun). No soundness or completeness bugs. The constraint chains for all five loads are fully pinned (LB needs no alignment assert since WindowMaskB reads all 3 low bits; the halfword assert covers WindowMaskH's ignored bit 0 on the same un-wrapped rs1+imm; LWU keeps its word-alignment assert), all three verifier-side MLEs were re-derived independently (including the WindowMaskB per-bit product form and non-contiguous Pext cases), the legacy Pow2OffsetB/H checkpoint state machines were verified branch-by-branch against the round conventions (the `r_x.unwrap()` is unreachable with None since round 127 is always odd), PextSigned's sign-bit selection is correct for 0xFF<<8k and 0xFFFF<<16m masks, and wiring/flags/opcodes/ABI/fixtures are complete and collision-free.

One stack-management note: the base branch received review-fix commits 8461ab592 + 101a45d2e after this PR was cut. The z3 hunk will conflict on both edges (the new arms sit between the two arms those commits rewrote), and the guard patterns those commits added must be mirrored by hand into this PR's new files since they produce no textual conflict. Suggested order: rebase first, consolidate the Pow2Offset prefixes, then mirror the asserts once into the consolidated impl. Inline comments carry the specifics.

crates/jolt-lookup-tables/src/tables/window_mask_b.rs
```diff
@@ -0,0 +1,86 @@
...
+        result
+    }
+}
+
+impl<const XLEN: usize> PrefixSuffixDecomposition<XLEN> for WindowMaskBTable<XLEN> {
+    fn prefixes(&self) -> &'static [Prefixes] {
+        &[Prefixes::Pow2OffsetB]
+    }
+
+    fn suffixes(&self) -> &'static [Suffixes] {
```
Acentelles, line 53: Same latent mismatch the base branch pinned for WindowMaskW in 8461ab592: the table is XLEN-generic but the Pow2OffsetB pair hardcodes 8-bit byte granularity (`8 * offset` shifts, `2^8/2^16/2^32 - 1` factors), so any XLEN != 64 decomposition silently disagrees with the MLE. Mirror the `debug_assert_eq!(XLEN, 64)` guard here and in `combine()`, and in the WindowMaskH + legacy twins (4 files total).

crates/jolt-lookup-tables/src/tables/prefixes/pow2_offset_b.rs
```diff
@@ -0,0 +1,25 @@
...
+pub enum Pow2OffsetBPrefix {}
+
+impl<F: Field> SparseDensePrefix<F> for Pow2OffsetBPrefix {
+    fn default_checkpoint() -> F {
+        F::one()
+    }
+
+    fn evaluate(checkpoints: &[PrefixEval<F>], b: LookupBits, suffix_len: usize) -> F {
+        // The low 3 index bits stay in the suffix until the final phase.
+        if suffix_len != 0 {
```
Acentelles: This gate pairs with the suffix's `b.len() < 3` guard only because phase boundaries never split the low 3 index bits; for B all three bits must stay in the suffix together, so the assumption is more load-bearing than the W case the base pinned in 8461ab592. Mirror `debug_assert!(suffix_len == 0 || suffix_len >= 3)` here, in the H twin, and in the legacy prefixes (`prefix_mle` and `update_prefix_checkpoint`).

crates/jolt-prover-legacy/src/zkvm/lookup_table/prefixes/pow2_offset_b.rs
```diff
@@ -0,0 +1,102 @@
...
+
+use super::{PrefixCheckpoint, Prefixes, SparseDensePrefix};
+
+/// `2^(8·(ea mod 8))` where `ea` is the (non-interleaved) lookup index: the
+/// byte-offset scale of a doubleword effective address.
+///
+/// The offset is `4·ea_2 + 2·ea_1 + ea_0`, so the prefix is the product of
+/// per-bit factors `2^(8·2^k·ea_k) = 1 + (2^(8·2^k) − 1)·ea_k` for
+/// `k = 2, 1, 0`.
+pub enum Pow2OffsetBPrefix<const XLEN: usize> {}
```
Acentelles: With this file the legacy Pow2Offset checkpoint state machine now exists in three hand-written copies (W 72 lines, H 94, B 102) differing only in which of bits {2,1,0} contribute. A single `Pow2OffsetPrefix<XLEN, LOW_BIT>` (W=2, H=1, B=0) covers all three exactly: the per-bit factor is uniform (`1 + (2^(8*2^k) - 1)*bit_k` for `k = 2..=LOW_BIT`), W's degenerate passthrough branches fall out of the generic form, and the dispatch cost is the `XorRotWPrefix<ROTATION>` pattern the enum already uses. The 1761 review deferred this to exactly this slice; doing it now means the phase-boundary assert gets mirrored into one impl instead of four sites. The modular side (three near-identical 25-line `evaluate()` files) takes the same const generic nearly for free.

crates/jolt-program/src/expand/memory/lwu.rs
```diff
@@ -32,21 +33,19 @@ pub(in crate::expand) fn expand_lwu(
...
+        SourceInstructionKind::VirtualWindowMaskW,
         v0.operand(),
+        v0.operand(),
+        0,
     );
-    asm.expand_i(
-        SourceInstructionKind::SRLI,
+    // rd = zero-extended word lane of the loaded doubleword.
+    asm.expand_r(
+        SourceInstructionKind::VirtualPext,
```
Acentelles: expand_lw and expand_lwu are now line-for-line identical except VirtualPextSigned vs VirtualPext, the exact pair this PR parameterizes with `signed: bool` for bytes and halfwords in shared.rs. Fold into a shared `expand_word_load(instruction, signed)` with lw.rs/lwu.rs as thin wrappers; the emitted sequences are unchanged so no fixture churn.

z3-verifier/src/virtual_sequences.rs
```diff
@@ -408,6 +411,28 @@ fn symbolic_exec(instr: &Instruction, cpu: &mut SymbolicCpu) {
...
+            cpu.x[operands.rd as usize] = cpu.bv_u64(0xFFFF).bvshl(shift);
+        }
+        Instruction::VirtualPext(VirtualPext { operands, .. }) => {
+            // Faithful for contiguous nonzero masks (the only shape the
+            // window-mask instructions produce): zero-extending extract via
+            // shift-left then logical shift-right.
+            let rs1 = cpu.x[operands.rs1 as usize].clone();
+            let rs2 = cpu.x[operands.rs2 as usize].clone();
+            let tz = trailing_zeros(&rs2, cpu.bv_bits);
+            let lz = leading_zeros(&rs2, cpu.bv_bits);
```
Acentelles, line 503: Three things here, all interacting with the base's post-cut fix 101a45d2e (which rewrote the two arms flanking this insertion, so the rebase conflicts on both edges): (1) the VirtualPext model is a copy of the pre-fix unguarded shl/lshr pattern; mirror the base's contiguity `ite` with a fresh-BV havoc (bvlshr in place of bvashr) so non-contiguous misuse fails verification loudly instead of certifying wrong semantics. (2) `0xFF`/`0xFFFF` and the 8/16 shift units are hardcoded while the rewritten W arm now derives from `cpu.word_bits`; a `byte_bits = bv_bits / 8` derivation keeps reduced-width solver models faithful. (3) When resolving the conflict, keep the base's rewritten W and PextSigned arms.

Acentelles:
Addressed the review findings:

- Rebased onto the updated base (the z3 hunk auto-merged cleanly; the base's rewritten WindowMaskW/PextSigned arms are intact alongside this PR's new arms).
- 212866a5c: the three Pow2Offset prefix state machines consolidated into one const-generic `Pow2OffsetPrefix<LOW_BIT>` per stack (W=2, H=1, B=0), with the phase-boundary `debug_assert!(suffix_len == 0 || suffix_len >= 3)` pinned once per stack instead of per copy.
- 873f528f0: `debug_assert_eq!(XLEN, 64)` in WindowMaskB/H `suffixes()`/`combine()` (both stacks), mirroring the base's WindowMaskW guard.
- b4c1dffc0: z3 `VirtualPext` gets the same contiguity-`ite`-with-havoc guard as the base's `VirtualPextSigned` fix; B/H lane masks and shift units now derive from `bv_bits` instead of hardcoded `0xFF`/`0xFFFF`/8/16.
- 12474919d: `expand_lw`/`expand_lwu` folded into `shared::expand_word_load(instruction, signed)`; emitted sequences unchanged.

Validated at the new head: jolt-lookup-tables 338/338, legacy prefix_suffix/window_mask/pext filter 70/70 (both exercise all three consolidated variants across the full 128-round binding trajectory), jolt-program 38/38 incl. the 360-case golden-fixture parity test (confirms the lw/lwu fold is emission-identical), `cargo check` on z3-verifier tests, clippy `-D warnings` on jolt-lookup-tables/jolt-program/z3-verifier and jolt-prover-legacy (host), fmt applied.

Acentelles:
Folded #1755's address fusion into this PR (4319d987e), per the integration plan on #1761 — credit to @0xAndoroid for the AlignAddr idea.

- New rank-1 `AlignAddr` table + `VirtualAlignAddr` (0x0091): `(rs1 + imm) & !7` in one lookup via the `AddOperands + RightOperandIsImm` pattern the alignment asserts already use; the decomposition is the LowerWord accumulator with bits 2-0 cleared and the carry bit dropped.
- The three WindowMask instructions now take the immediate directly and index the unwrapped `rs1 + imm` sum (low 3 bits agree with the wrapped ADDI). The stores slice's register-operand uses (imm = 0) are unaffected.
- Loads: **LB/LBU 8 -> 4, LH/LHU 9 -> 5, LW 8 -> 5, LWU 9 -> 5**. Combined with the stores slice this puts the nine sub-word expansions at Σ54 vs the Σ57/Σ60 of either design alone.
- 78 golden hashes re-baselined (all six loads, 12 each, plus LRW/SCW); the 360-entry truncation guard holds.
- Validation: modular tables 344/344 (incl. the new table's hypercube/prefix-suffix/binding tests), legacy filter 75/75, jolt-program 38/38, tracer 127/127, muldiv e2e in `host` and `host,zk`, clippy `-D warnings` in both feature modes, z3 arm added.

Remaining #1755 integrations land separately: one test-only hardening PR on top of #1768 (z3 forall-proof enablement for the nine sequences + the post-final-round checkpoint regression test), since both cut across all three slices; SRL-is-PEXT stays out of the stack as a later standalone PR (it touches shipped SRL/SRA paths). #1768 gets restacked on this head next.

Acentelles:
Resolved the merge conflicts against the updated base (through 2739c9a3d) and re-reviewed the result.

**Conflict resolution (e2c3afd63, ec96a0e01):**
- The base's RV64 word-arithmetic fusion claimed tags 0x0089-0x008e and moved WindowMaskW/PextSigned to 0x008f/0x0090, colliding with this PR's assignments. New numbering: AlignAddr keeps 0x0091 (matching the stores slice), Pext/WindowMaskB/WindowMaskH move to 0x0092-0x0094, and the narrow-store reservation shifts to 0x0095-0x0097 (ShiftDataB/H/W will need renumbering on rebase).
- `JoltInstructionKind` serializes as its u16 tag, so the renumbering re-blessed the same 78 expansion parity hashes (six loads x12 plus LRW/SCW) the AlignAddr fusion touched; the word-arith entries from the base were already correct. Documented in the fixture test comment.
- `VirtualSw`/`VirtualLw` modules dropped along with the base: main removed them as dead and no references remain.

**Re-review findings, fixed in fb87c1d41:**
- The AOT x86 backend (#1729 + c070b9f0a, both post-cut) was missing templates for the four PR-new kinds — E0004 on x86_64-linux, invisible on macOS where the native module is cfg-gated out. Added Pext/WindowMaskB/WindowMaskH/AlignAddr templates, extended the per-program BMI2 gate to VirtualPext, and moved the five imm-bearing kinds to `alu_ri` difftests so the immediate path is exercised.
- The base's WindowMaskW template still read the pre-AlignAddr-fusion rs1-only form, silently ignoring the immediate the expansion now carries (a semantic mis-merge with no textual conflict). Folded the immediate in; verified against the interpreter semantics, and the difftest generator change would now catch a regression.
- Added the phase-boundary `debug_assert` to the modular `AlignAddrPrefix`: zeroing carry-range chunks is only sound when no chunk straddles index bit XLEN, matching the Pow2Offset prefix's documented invariant. (Latent only: current phase widths of 8/16 never straddle.)

**Noted, not changed:** the z3 load `test_sequence!`s stay commented out ("memory operations are not tested", pre-existing since 12c2777f4), so the new load models rely on the difftests and e2e coverage; and the `debug_assert_eq!(XLEN, 64)` lane-granularity guards in WindowMaskB/H follow the existing sibling-table idiom rather than a compile-time assert.

Verification at fb87c1d41: fmt + clippy (host, host+zk) clean; jolt-program/jolt-lookup-tables/tracer 541/541; legacy prefix/window-mask/pext filter 71/71; jolt-inlines fixtures 3/3; muldiv e2e both modes; `cargo check -p jolt-tracer-x86 --all-targets --target x86_64-unknown-linux-gnu` passes. An independent removed-behavior audit (20k-trial randomized simulation of the new load sequences against RISC-V reference semantics, plus numeric checks of the consolidated Pow2Offset prefix against the deleted per-width variants) came back clean.

0xAndoroid (approved):
lgtm, thanks

0xAndoroid:
Github is down again. Pika will merge this once all checks pass.

PR #1763, opened 2026-08-10
feat(isa): implicit-carry ISA layer — ADDC and MULC instructions

Acentelles:
## Summary

First of three PRs implementing the implicit-carry spec (#1710, `specs/implicit-carry-handling.md`): the ISA layer only. Adds `ADDC` and `MULC` as guest-emittable final Jolt instructions and the emulator/trace plumbing that records each row's incoming carry, behind a new `implicit-carry` cargo feature mirroring the `field-inline` lane. No protocol changes: with the feature off, builds and proof geometry are unchanged (the `muldiv` e2e passes in both standard and ZK modes); with it on, traces containing `ADDC`/`MULC` are not yet provable — constraint support lands in the follow-up.

## What's included

- **Instructions**: `ADDC` (`rd = low64(rs1 + rs2 + carry)`) and `MULC` (`rd = low64(rs1 * rs2 + carry)`), custom opcode `0x5B`, funct3 `0`, funct7 `0x06`/`0x07`, Jolt tags `0x0110`/`0x0111`. Decoded by both the tracer and `jolt-program` decoders; rejected when the feature is off. Gated behind a new `SourceExtension::ImplicitCarry` and the `RV64IMAC_JOLT_IMPLICIT_CARRY` profile.
- **Circuit flags**: `UsesCarry` and `ProducesCarry` (14 → 16, exactly filling `CircuitFlagSet(u16)`; a new const assert makes a 17th flag a build error instead of a silent bit collision). `ADD`/`MUL` gain `ProducesCarry` under the feature via `jolt_instruction!` support for per-flag cfg attributes.
- **Emulator semantics**: `cpu.carry` holds the last instruction's carry-out; producers set it in `exec`, everything else clobbers it to zero in the macro-generated `execute`, so trace mode and execute-only mode stay in lockstep for chunked parallel replay. Carry is captured in `ChunkCpuState` and both replay-divergence checks.
- **Trace**: `RISCVCycle.carry` / `TraceRow.carry` record the row's incoming carry (zero on padding rows).
- **Lookups**: both instructions reuse the `RangeCheck` table with combined operands `rs1 + rs2 + carry` (< 3·2^64) and `rs1·rs2 + carry` (< 2^128), matching the spec's soundness argument that `NextCarry` inherits its range check from the 128-bit lookup domain.
- **CI**: clippy + nextest lanes for `implicit-carry` and the `field-inline,implicit-carry` composition, mirroring the existing `field-inline` jobs.

## Semantics worth reviewing

- `ADDC`/`MULC` declare `HAS_SIDE_EFFECTS = true`: an `rd = x0` occurrence is rewritten to a virtual register during expansion rather than noop-eliminated, which would silently break a carry chain. `ADD`/`MUL` remain side-effect-free, so `add x0, a, b` does **not** produce a usable carry.
- The legacy prover compiles with the feature on but ignores the two new flag bits (`circuit_flags_from_riscv` reads only the low 14) and maps `ADDC`/`MULC` to `RangeCheck`. Fail-closed protocol-config rejection arrives with the constraint PR.

## Testing

- Randomized `{ADD, MUL} → {ADDC, MULC}` pair tests over edge-case operands (0, 1, `u64::MAX`, large high halves; 864 cases), carry-clobber by intervening instructions, row-0 zero carry, and a trace-vs-execute lockstep check (`tracer/src/instruction/mod.rs`).
- Standard lookup-table macro tests for both instructions; the shared trace-match harness now seeds `cpu.carry` from the cycle, so future carry-aware instructions get correct coverage for free.
- Full matrix locally: workspace clippy (`host`, `host,zk`), per-crate clippy/nextest with the feature off, on, and composed with `field-inline`; `muldiv` e2e in both modes.

## Out of scope (per spec §Non-Goals)

Committed `Carry` column, `CarryUsed` product term, uniform R1CS rows, shift-relation extension, `carry_init` opening, BlindFold sync, modular-stack geometry, jolt-eval invariant, inline (`BigintMul256`) rewrites, guest-crate feature forwards, z3-verifier/lean-gen updates.

### DIFF
--- a/crates/jolt-program/src/execution/trace.rs
+++ b/crates/jolt-program/src/execution/trace.rs
@@ -251,6 +251,28 @@ pub struct TraceRow {
     pub carry: u64,
 }
 
+impl TraceRow {
+    /// Builds a row from the always-present fields, defaulting the cfg-gated
+    /// extensions. Downstream constructors stay immune to future cfg-gated
+    /// fields, and unlike `..Default::default()` at fully-specified sites this
+    /// stays clean under clippy's `needless_update` with `-D warnings`.
+    pub fn new(
+        instruction: JoltInstructionRow,
+        registers: RegisterState,
+        ram_access: RamAccess,
+    ) -> Self {
+        Self {
+            instruction,
+            registers,
+            ram_access,
+            #[cfg(feature = "field-inline")]
+            field_inline: None,
+            #[cfg(feature = "implicit-carry")]
+            carry: 0,
+        }
+    }
+}
+
 impl JoltCycle for TraceRow {
     type Instruction = JoltInstructionRow;
 
--- a/crates/jolt-program/src/execution/trace.rs
+++ b/crates/jolt-program/src/execution/trace.rs
@@ -245,6 +245,10 @@ pub struct TraceRow {
     pub ram_access: RamAccess,
     #[cfg(feature = "field-inline")]
     pub field_inline: Option<Arc<FieldInlineTraceData>>,
+    /// The row's incoming implicit carry (the previous row's carry-out).
+    /// Zero on padding rows.
+    #[cfg(feature = "implicit-carry")]
+    pub carry: u64,
 }
 
 impl JoltCycle for TraceRow {
@@ -296,6 +300,12 @@ impl JoltCycle for TraceRow {
             RamAccess::Write(write) => Some(write.post_value),
             RamAccess::Read(_) | RamAccess::NoOp => None,
         }
+    }
+
+    #[cfg(feature = "implicit-carry")]
+    #[inline]
+    fn carry(&self) -> u64 {
+        self.carry
     }
 }
 
--- a/crates/jolt-prover-legacy/src/zkvm/instruction/mod.rs
+++ b/crates/jolt-prover-legacy/src/zkvm/instruction/mod.rs
@@ -416,6 +416,15 @@ impl<const XLEN: usize> LookupQuery<XLEN> for JoltTraceCycle<'_> {
 }
 
 fn circuit_flags_from_riscv(flags: RiscvCircuitFlagSet) -> [bool; NUM_CIRCUIT_FLAGS] {
+    // jolt_riscv's flag set can be wider than the legacy layout (implicit-carry
+    // adds UsesCarry/ProducesCarry); silently truncating a set carry flag would
+    // prove wrong semantics. This runs at preprocessing time, not per cycle.
+    assert!(
+        flags.bits() >> NUM_CIRCUIT_FLAGS == 0,
+        "circuit flags {:#018b} exceed the legacy prover's {NUM_CIRCUIT_FLAGS} flags; \
+         ADDC/MULC are not supported by the legacy prover",
+        flags.bits()
+    );
     let mut converted = [false; NUM_CIRCUIT_FLAGS];
     for (index, value) in converted.iter_mut().enumerate() {
         *value = flags.bits() & (1 << index) != 0;
@@ -424,6 +433,7 @@ fn circuit_flags_from_riscv(flags: RiscvCircuitFlagSet) -> [bool; NUM_CIRCUIT_FL
 }
 
 fn instruction_flags_from_riscv(flags: RiscvInstructionFlagSet) -> [bool; NUM_INSTRUCTION_FLAGS] {
+    assert!(flags.bits() >> NUM_INSTRUCTION_FLAGS == 0);
     let mut converted = [false; NUM_INSTRUCTION_FLAGS];
     for (index, value) in converted.iter_mut().enumerate() {
         *value = flags.bits() & (1 << index) != 0;
--- a/crates/jolt-prover-legacy/src/zkvm/instruction/mod.rs
+++ b/crates/jolt-prover-legacy/src/zkvm/instruction/mod.rs
@@ -358,6 +358,13 @@ impl<const XLEN: usize> InstructionLookup<XLEN> for JoltInstructionRow {
             | JoltInstruction::FieldLoadFromX(_)
             | JoltInstruction::FieldStoreToX(_)
             | JoltInstruction::FieldLoadImm(_) => return None,
+            // Same combined-operand range-check lookup as ADD/MUL; the legacy
+            // prover has no implicit-carry constraint support yet, so proving
+            // a trace containing these is unsupported until then.
+            #[cfg(feature = "implicit-carry")]
+            JoltInstruction::AddC(_) | JoltInstruction::MulC(_) => {
+                LookupTables::RangeCheck(Default::default())
+            }
             JoltInstructionKind::NoOp
             | JoltInstructionKind::LD
             | JoltInstructionKind::SD
--- a/crates/jolt-riscv/src/instructions/implicit_carry.rs
+++ b/crates/jolt-riscv/src/instructions/implicit_carry.rs
@@ -4,7 +4,7 @@
 //! `Carry` column) and, like `ADD` and `MUL`, export the high 64 bits of the
 //! true arithmetic result as the next row's carry. The carry is
 //! non-architectural: it is not part of the memory-checked register file.
-//! See `specs/implicit-carry-handling.md`.
+//! See the implicit-carry spec in <https://github.com/a16z/jolt/issues/1710>.
 
 use crate::jolt_instruction;
 
--- a/crates/jolt-riscv/src/instructions/implicit_carry.rs
+++ b/crates/jolt-riscv/src/instructions/implicit_carry.rs
@@ -1,0 +1,34 @@
+//! Implicit-carry arithmetic instructions.
+//!
+//! `ADDC` and `MULC` consume the previous row's implicit carry (the committed
+//! `Carry` column) and, like `ADD` and `MUL`, export the high 64 bits of the
+//! true arithmetic result as the next row's carry. The carry is
+//! non-architectural: it is not part of the memory-checked register file.
+//! See `specs/implicit-carry-handling.md`.
+
+use crate::jolt_instruction;
+
+jolt_instruction!(
+    /// Jolt ADDC: `rd = low_64(rs1 + rs2 + carry)`, carry-out = `high_64(rs1 + rs2 + carry)`.
+    AddC,
+    circuit flags: [
+        AddOperands,
+        WriteLookupOutputToRD,
+        UsesCarry,
+        ProducesCarry,
+    ],
+    instruction flags: [LeftOperandIsRs1Value, RightOperandIsRs2Value]
+);
+
+jolt_instruction!(
+    /// Jolt MULC: `rd = low_64(rs1 * rs2 + carry)`, carry-out = `high_64(rs1 * rs2 + carry)`,
+    /// over the unsigned 128-bit widening of the raw 64-bit words.
+    MulC,
+    circuit flags: [
+        MultiplyOperands,
+        WriteLookupOutputToRD,
+        UsesCarry,
+        ProducesCarry,
+    ],
+    instruction flags: [LeftOperandIsRs1Value, RightOperandIsRs2Value]
+);
--- a/crates/jolt-riscv/src/kind.rs
+++ b/crates/jolt-riscv/src/kind.rs
@@ -632,7 +632,10 @@ macro_rules! source_side_effects_for_marker {
         true
     };
     (Add) => {
-        false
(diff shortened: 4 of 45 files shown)


crates/jolt-tracer-x86/Cargo.toml
```diff
@@ -16,6 +16,13 @@ field-inline = [
   "jolt-riscv/field-inline",
   "tracer/field-inline",
 ]
+# Passthrough only: the AOT backend has no carry register; the feature exists
+# so `for_each_jolt_instruction_kind!` expansions resolve their cfg entries.
+implicit-carry = [
```
Acentelles, line 21: **Compile break (verified):** this lane does not compile on the crate's only real platform. `cargo check -p jolt-tracer-x86 --features implicit-carry --target x86_64-unknown-linux-gnu` fails with two errors:

- `E0063`: `TraceRow` literal in `reassemble_rows` (`src/native/mod.rs:310`) is missing the new cfg-gated `carry` field.
- `E0004`: non-exhaustive match over `JoltInstruction` (`src/native/compile/emit.rs:608`), `AddC(_)`/`MulC(_)` not covered.

It passes on darwin only because `mod native` is gated `#[cfg(all(target_arch = "x86_64", target_os = "linux"))]`, and the new CI clippy/test lanes omit `-p jolt-tracer-x86`, so nothing catches it. Fix both sites and add the crate to the implicit-carry clippy lane.

crates/jolt-prover-legacy/src/zkvm/instruction/mod.rs
```diff
@@ -358,6 +358,13 @@ impl<const XLEN: usize> InstructionLookup<XLEN> for JoltInstructionRow {
             | JoltInstruction::FieldLoadFromX(_)
             | JoltInstruction::FieldStoreToX(_)
             | JoltInstruction::FieldLoadImm(_) => return None,
+            // Same combined-operand range-check lookup as ADD/MUL; the legacy
+            // prover has no implicit-carry constraint support yet, so proving
+            // a trace containing these is unsupported until then.
+            #[cfg(feature = "implicit-carry")]
+            JoltInstruction::AddC(_) | JoltInstruction::MulC(_) => {
```
Acentelles, line 365: **Feature-on legacy lane is internally inconsistent, and "unsupported" is enforced nowhere:**

1. `circuit_flags_from_riscv` (this file, ~line 418) bit-truncates the feature-on 16-flag jolt-riscv set to the legacy hand-maintained 14-variant enum, silently dropping `UsesCarry`/`ProducesCarry` with no width assertion.
2. Genuine count drift: `claim_reductions/bytecode_reconstruction.rs:59` imports `jolt_riscv::NUM_CIRCUIT_FLAGS` (16 feature-on) while the committed-bytecode lane layout in `zkvm/bytecode/chunks.rs` uses the legacy 14 — the loops at `bytecode_reconstruction.rs:369-375` and `:781-786` then index `BytecodeCircuitFlag(chunk, 14/15)` into lanes aliasing the instruction-flag lanes (missing-opening panic in committed-program mode).
3. An honest ADDC trace never reaches this arm's lookup: `Cycle::ADDC`/`MULC` are absent from `define_rv64imac_trait_impls!` (~line 493), so witness gen dies in `_ => panic!("Unexpected instruction")`. The only guard, `is_supported_instruction`, is consulted solely in a test.

CI runs clippy-only on prover-legacy with this feature — none of this is compile-visible. Either add a hard rejection at trace ingestion or drop the prover-legacy lane until the constraint support lands (PR 2/3).

crates/jolt-riscv/src/instructions/implicit_carry.rs
```diff
@@ -0,0 +1,34 @@
+//! Implicit-carry arithmetic instructions.
+//!
+//! `ADDC` and `MULC` consume the previous row's implicit carry (the committed
+//! `Carry` column) and, like `ADD` and `MUL`, export the high 64 bits of the
+//! true arithmetic result as the next row's carry. The carry is
+//! non-architectural: it is not part of the memory-checked register file.
```
Acentelles, line 6: **Contract violation elsewhere: MULH/MULHSU become accidental carry producers.** The tracer routes guest `MULH` through the jolt-program expansion whose final row is `ADD rd, v_tmp, v_sy` (`crates/jolt-program/src/expand/arithmetic/mulh.rs:~55`; same in `mulhsu.rs:~75`). ADD has `PRODUCES_CARRY = true`, so after a guest `MULH` retires, `cpu.carry = high64(v_tmp + v_sy)` instead of 0.

Scenario: `mulh t0,a,b; addc t1,x,y` — the ADDC consumes the correction-add's carry-out. This violates the clobber rule this PR documents and tests (the test only covers single-row XOR). Nothing enforces "expansion-final row is a non-producer"; consider a structural check in the expansion builder or an explicit trailing clobber row.

tracer/src/instruction/mod.rs
```diff
@@ -2010,6 +2052,176 @@ mod tests {
...
+                                );
+                            }
+                        }
+                    }
+                }
+            }
+        }
+
+        #[test]
+        fn intervening_instruction_clobbers_carry_to_zero() {
```
Acentelles, line 2212: **This invariant is violated by SLL:** the variable-shift expansion ends in `MUL rd, rs1, 2^(rs2&63)` (`crates/jolt-program/src/expand/shifts/sll.rs:~21`), so after a guest `SLL`, `cpu.carry = high64(rs1 * 2^shamt)` — the shifted-out bits — instead of 0, and a following ADDC consumes them.

Inconsistently, `SLLI` lowers via `VirtualMULI` (clobbers to 0) and `SLLW`/`SLLIW` end in `VirtualSignExtendWord`, so `sll` and `slli` with identical shift amounts leave different carry states. This test passes only because XOR is a single-row instruction; multi-row expansions ending in ADD/MUL (SLL here, MULH/MULHSU per the other comment) escape it.

crates/jolt-riscv/src/kind.rs
```diff
@@ -1186,6 +1204,12 @@ macro_rules! jolt_side_effects_for_marker {
     (FieldLoadImm) => {
         true
     };
+    (AddC) => {
```
Acentelles, line 1211: **rd=x0 asymmetry between the four carry producers.** AddC/MulC are marked side-effecting here (correct: their carry-out must survive an rd=x0 write via the virtual-register rewrite), but ADD and MUL remain `false` — so `add x0, a, b` is rewritten to `ADDI x0,x0,0` by the noop rewrite (`jolt-program/src/expand/materialize.rs:79-90`, reached via `takes_rd0_expansion`), a non-producer that clobbers carry to 0. That contradicts ADD's unconditional `ProducesCarry` circuit flag and its emulator exec.

Scenario: a guest computing only the carry bit with `add x0,a,b; addc rd,x,y` gets 0 instead of `high64(a+b)`. Either make Add/Mul side-effecting under the feature, or specify rd=x0 as carry-destroying in the spec.

crates/jolt-program/src/execution/trace.rs
```diff
@@ -245,6 +245,10 @@ pub struct TraceRow {
     pub ram_access: RamAccess,
     #[cfg(feature = "field-inline")]
     pub field_inline: Option<Arc<FieldInlineTraceData>>,
+    /// The row's incoming implicit carry (the previous row's carry-out).
+    /// Zero on padding rows.
+    #[cfg(feature = "implicit-carry")]
+    pub carry: u64,
```
Acentelles, line 251: **Latent E0063 under feature unification.** This cfg-gated required field breaks every `TraceRow` literal in crates that have no implicit-carry lane: `jolt-witness/src/testing.rs:48` (plus `backend/trace/tests.rs`) and 7 sites in `jolt-kernels` (e.g. `optimized/spartan_shift.rs:582`). Any cargo invocation co-building one of them with a crate enabling `jolt-program/implicit-carry` fails to compile — the same failure mode already reproduced in jolt-tracer-x86. Invisible today only because current CI lanes never co-build them; it will bite when the modular prover grows its lane in PR 2/3. Consider updating those literals (cfg-gated `carry: 0`) now, or giving `TraceRow` a constructor so new fields have one seam.

tracer/src/utils/instruction_macros.rs
```diff
@@ -74,8 +84,18 @@ macro_rules! declare_riscv_instr {
...
+
             fn execute(&self, cpu: &mut $crate::emulator::cpu::Cpu, ram: &mut Self::RAMAccess) {
-                self.exec(cpu, ram)
+                self.exec(cpu, ram);
+                // Non-carry-producing instructions clobber the implicit carry
+                // to zero; producers set `cpu.carry` inside `exec`. Living in
+                // `execute` (not `RISCVTrace::trace`) keeps trace and
+                // execute-only modes in lockstep for chunked replay.
+                #[cfg(feature = "implicit-carry")]
+                if !Self::PRODUCES_CARRY {
```
Acentelles: **Clobber sits one level too low.** The only call site of `RISCVInstruction::execute` in the tracer is the default `RISCVTrace::trace` (enum-level `Instruction::execute` also dispatches through `trace(cpu, None)`, so the chunked-replay justification in this comment is moot). Placing `if !Self::PRODUCES_CARRY { cpu.carry = 0; }` once after `self.execute(...)` in `RISCVTrace::trace` would cover every impl — this macro, plus the hand-written copies in `virtual_advice.rs:63-66` and `field_inline.rs:69-72`, already three in this PR. As written, the next manual `impl RISCVInstruction` that forgets the clobber silently leaks a stale producer carry into a downstream ADDC with no compile error.

Related: `PRODUCES_CARRY` is dead plumbing as a trait const — nothing generic reads it (this macro could use `$produces_carry` directly), and the manual impls it nominally serves ignore it. Moving the clobber to `trace` would make the const genuinely load-bearing.

crates/jolt-riscv/src/instructions/implicit_carry.rs
```diff
@@ -0,0 +1,34 @@
...
+//! See `specs/implicit-carry-handling.md`.
```
Acentelles: **Dangling citation:** `specs/implicit-carry-handling.md` does not exist anywhere on this branch (checked the full `specs/` listing). The normative carry contract — who produces, who clobbers, rd=x0 behavior, exactly the ambiguities behind the MULH/SLL/rd=x0 comments — has no findable document. Cite issue #1710 instead, or add the spec file in this PR.

tracer/src/utils/instruction_macros.rs
```diff
@@ -7,9 +7,19 @@ macro_rules! declare_riscv_instr {
...
+        declare_riscv_instr!(@inner $name, $mask, $match_, $format, $ram, false);
     };
-    (@inner $name:ident, $mask:expr, $match_:expr, $format:ty, $ram:ty) => {
+    (
+      name    = $name:ident,
+      mask    = $mask:expr,
+      match   = $match_:expr,
+      format  = $format:ty,
+      ram     = $ram:ty,
+      produces_carry = true $(,)?
```
Acentelles: **Duplicated macro arm:** the two top-level arms are byte-identical except this trailing `produces_carry = true` and the `false`/`true` forwarded to `@inner` (17 duplicated header lines). One arm with an optional tail does the same job and const-folds identically:

```rust
$(, produces_carry = $pc:literal)? $(,)?
// forward: { false $(|| $pc)? }
```

Every future grammar change currently has to be made in both arms or they silently diverge.

tracer/src/instruction/mod.rs
```diff
@@ -2010,6 +2052,176 @@ mod tests {
         assert!(Instruction::decode(word, 0x8000_0000, false).is_err());
     }
 
+    fn r_type_word(opcode: u32, funct3: u32, funct7: u32, rd: u8, rs1: u8, rs2: u8) -> u32 {
```
Acentelles, line 2064: **Magic numbers duplicating named constants from this same file:** the test encoders hardcode `0x5b`/`0b000`/`0x06`/`0x07` for which `FUNCT7_ADDC`/`FUNCT7_MULC` (added by this PR at the top of this file) and the custom opcode/funct3 constants already exist; `ADD_WORD`/`MUL_WORD` likewise re-derive `ADD::MATCH`/`MUL::MATCH` (existing precedent: `mulhsu.rs:31` builds test words from `Self::MATCH`). A funct7 reassignment would leave these literals stale but still compiling.

Acentelles:
Folded into #1764 per review: that branch is rebased on this one's final tip (26d9e21a1) and reviewed there as the combined ISA + proving diff.

PR #1764, opened 2026-08-11
feat(prover): implicit-carry proving via committed carry column (spec #1710)

Acentelles:
Part 2 of the implicit-carry implementation (spec: #1710, discussion there). Stacked on #1763 (ISA layer); only the last 13 commits are new here. Everything is behind the `implicit-carry` feature; the default build is unchanged.

## What this adds

Proving and verification for ADDC/MULC in the legacy prover, with the modular (typed) verifier extended to verify the resulting proofs.

Protocol changes, all per the spec:

- **Committed `Carry` column** (the row's incoming carry), absorbed after the BytecodeRa commitments.
- **`CarryUsed = UsesCarry x Carry`** as a fourth product-virtualization term.
- **`NextCarry`** as a new outer R1CS input, split out of the 128-bit lookup output: `RLO = LookupOutput + 2^64 * NextCarry`, plus `NextCarry = 0` when the row does not produce a carry. R1CS inputs go 35 to 39, uniform constraint rows 19 to 21 (uniskip domain 10 to 11), circuit flags 14 to 16 (`UsesCarry`, `ProducesCarry`).
- **Shift relation** `NextCarry(t) = Carry(t+1)` as a sixth gamma term on the existing shift sumcheck.
- **`CarryClaimReduction`**: Carry ends up with three openings (product point, shift point, and the `Carry(0) = 0` initialization). A dedicated stage-6b reduction, modeled on the increment reduction, folds them into the single final opening that joins the stage-8 batch.

Soundness of the split relies on `NextCarry` inheriting the 128-bit range check from the lookup domain; the carry-in needs no separate range check because every `Carry` opening is tied to a `NextCarry` opening by the shift relation, with `Carry(0)` pinned to zero.

## Testing

- `examples/carry-chain`: guest exercising all four producer/consumer pairings via `.insn` (including the `ADDC x0, x0` carry-read idiom), checked against a `u128` reference, with tamper rejection for the carry commitment and each new claim.
- `muldiv` e2e green in both modes; full legacy suite green feature-on (449) and feature-off; modular crate suites green feature-on (883) and feature-off; the verifier tamper manifest covers the eight new clear-claim leaves.
- CI: clippy lanes for `implicit-carry` and `field-inline,implicit-carry` across every crate that defines the lane, plus legacy and modular nextest jobs feature-on.

## Deviations from the spec (for review)

1. **ZK fails closed** (`compile_error!` on `implicit-carry` + `zk`, same for `akita`). The BlindFold constraint functions are written but untested, so ZK enablement is deferred to the next PR rather than landing unverified. The spec had ZK in the initial landing.
2. **`Carry(0) = 0` is enforced as a public input pair of the reduction** (all-zeros point, claim 0) instead of a named final-opening leaf. This costs nothing in proof size and keeps stage 8 unchanged; it is close to the spec's discussion of alternative 7, so flagging it explicitly.
3. **The modular prover does not prove implicit-carry programs yet**; its witness backend explicitly classifies the carry polynomials as not served. The modular verifier fully verifies legacy proofs feature-on.
4. The `implicit_carry_pair_soundness` jolt-eval invariant from the spec is not implemented yet; planned as a follow-up.

## Measured performance

- **Cycle counts** (the `mul256_carry_vs_baseline_trace_length` test, trace-only, noop-controlled): the 256x256 schoolbook multiplication costs **55 net cycles with MULC/ADDC vs 145 with the portable u128 sequence** (2.6x; both nets include the same ~15-cycle checksum fold, so the pure sequences compare at roughly 40 vs 130). Consistent with the spec's 142-to-39 sketch. The test ratchets the carry version at under half the portable cost.
- **Overhead on carry-free workloads** (legacy `profile --name sha3`, three interleaved runs per configuration, identical binaries): wall times overlap completely (off 21.3/21.4/24.5 s, on 21.2/22.0/22.6 s) and peak RSS is within the run-to-run spread (399/401 MiB off, 393 MiB on). No measurable regression, matching the spec's expectation.

### Reproduce

Every number above regenerates from this branch:

```bash
# Cycle counts (deterministic; also a CI ratchet asserting carry < baseline/2)
cargo nextest run -p jolt-prover-legacy mul256 --features implicit-carry --cargo-quiet --no-capture

# Overhead on a carry-free workload: stage both binaries, then interleave timed runs
cargo build --release -p jolt-prover-legacy && cp target/release/jolt-prover-legacy /tmp/legacy_off
cargo build --release -p jolt-prover-legacy --features implicit-carry && cp target/release/jolt-prover-legacy /tmp/legacy_on
for i in 1 2 3; do /usr/bin/time -p /tmp/legacy_off profile --name sha3 >/dev/null; /usr/bin/time -p /tmp/legacy_on profile --name sha3 >/dev/null; done
```

For the adoption outlook: tracing `examples/recover-ecdsa` (625,599 cycles per recovery) shows 26.0% of the trace in MUL/MULHU/SLTU, 26.3% in ADD, and 17.0% in limb packing (the 5x52 representation libsecp256k1 uses to defer carries). Applying the measured 2.6x to the carry cluster projects a 1.2x-1.5x end-to-end ECDSA speedup from mechanical adoption, more if the field representation moves to 4x64 once carries are free. Projection, not measurement; the adoption PR will carry the e2e number.


Acentelles:
The existing ECDSA precompile (secp256k1 inlines) gets **1.39x cheaper to prove** by rebuilding its field-arithmetic sequences on carry-chain instructions, that is, 206,143 → 148,034 trace cycles per verification, with prover cost scaling with trace length.

Per-op inline sequences (the precompile's building block)

| inline op                  | rows before | rows after | gain  |
|----------------------------|------------:|-----------:|------:|
| BIGINT256_MUL              |         142 |         69 | 2.06x |
| SECP256K1_MULQ  (Fq mul)   |         186 |         95 | 1.96x |
| SECP256K1_SQUAREQ          |         167 |         87 | 1.92x |
| SECP256K1_DIVQ             |         194 |        104 | 1.87x |
| SECP256K1_MULR  (Fr mul)   |         232 |        115 | 2.02x |
| SECP256K1_SQUARER          |         213 |        107 | 1.99x |
| SECP256K1_DIVR             |         240 |        124 | 1.94x |

The gap between ~2x per op and 1.39x end to end is the precompile's
non-arithmetic share (point-op control flow, loads/stores, advice rows).

Acentelles:
Also, the same 256-bit multiplication, in ordinary Rust (the compiler turns it into the usual mul/mulhu/add/sltu carry bookkeeping leads to 145 cycles of trace). The inline assembly using the new ADDC/MULC instructions directly takes on 55 cycles.

crates/jolt-verifier/Cargo.toml
```diff
@@ -19,6 +19,7 @@ allocative = ["dep:allocative", "jolt-claims/allocative"]
 prover-fixtures = ["jolt-prover-legacy/host"]
 fs-audit = []
 field-inline = ["jolt-claims/field-inline", "jolt-r1cs/field-inline"]
+implicit-carry = ["jolt-claims/implicit-carry", "jolt-r1cs/implicit-carry"]
```
Acentelles, line 22: `implicit-carry` does not forward to `jolt-prover-legacy` (the `prover-fixtures` fixture prover), so `--features prover-fixtures,implicit-carry` builds a feature-off fixture prover: its proofs lack the `isa` protocol-config field, the Carry commitment, and the carry claims, and the suite fails before any test logic runs. Consequently every carry tamper target marked `TamperCoverage::Active` (e.g. `stage6.claims.carry_claim_reduction.carry` in `tests/support/tamper_manifest.rs`) is dead in all CI lanes, and the new CI lane runs jolt-verifier feature-on *without* `prover-fixtures`, so neither the tamper suite nor the byte-diff ratchets ever execute feature-on. Suggested fix: add `jolt-prover-legacy?/implicit-carry` here and a `prover-fixtures,implicit-carry` CI lane.

crates/jolt-prover-legacy/Cargo.toml
```diff
@@ -77,9 +77,11 @@ field-inline = [
   "tracer/field-inline",
 ]
 implicit-carry = [
+  "jolt-claims?/implicit-carry",
   "jolt-lookup-tables/implicit-carry",
   "jolt-program/implicit-carry",
   "jolt-riscv/implicit-carry",
+  "jolt-verifier/implicit-carry",
```
Acentelles, line 84: Missing `?`: `jolt-verifier` is optional (`dep:jolt-verifier` under `prover`), and line 80 uses the weak form `jolt-claims?/implicit-carry`, but this line force-enables the dependency. `cargo build -p jolt-prover-legacy --no-default-features --features minimal,implicit-carry` (the recursion configuration `minimal` exists for, per its comment) pulls in the entire jolt-verifier stack. One-character fix: `jolt-verifier?/implicit-carry`.

crates/jolt-witness/src/backend/trace/oracle.rs
```diff
@@ -87,6 +87,8 @@ impl<T: TraceSource + Clone> TraceBackend<T> {
                         Compact,
                     ))
                 }
+                #[cfg(feature = "implicit-carry")]
+                C::Carry => Err(not_served(id, UNSERVED_REASON)),
```
Acentelles, line 91: Feature-unified modular prover compiles cleanly but cannot prove: `cargo check -p jolt-prover --features jolt-witness/implicit-carry,jolt-verifier/implicit-carry` succeeds, yet feature-on jolt-claims/jolt-verifier make the modular stack request Carry (committed_openings order, `Stage6bSumchecks` carry member, `V_NEXT_CARRY`), so every modular prove dies at runtime with `WitnessError::NotServed` — and `UNSERVED_REASON` ("no consumer on the modular stack") is now false feature-on. Fail-closed, so no soundness risk, but inconsistent with this PR's own `compile_error!` guards for implicit-carry+zk and implicit-carry+akita in `jolt-verifier/src/config.rs`. Suggest a matching jolt-prover-side guard, or at minimum an updated reason/WHY comment stating modular proving is a staged follow-up.

crates/jolt-prover-legacy/src/zkvm/instruction/addc.rs
```diff
@@ -0,0 +1,89 @@
...
+    }
+}
+
+impl Flags for ADDC {
+    fn circuit_flags(&self) -> [bool; NUM_CIRCUIT_FLAGS] {
+        let mut flags = [false; NUM_CIRCUIT_FLAGS];
+        flags[CircuitFlags::AddOperands] = true;
+        flags[CircuitFlags::WriteLookupOutputToRD] = true;
+        flags[CircuitFlags::UsesCarry] = true;
+        flags[CircuitFlags::ProducesCarry] = true;
```
Acentelles, line 20: ADDC/MULC set `ProducesCarry` here, and the jolt-riscv rows set it on ADD/MUL feature-on — but the legacy per-instruction `Flags` impls in `add.rs`/`mul.rs` (same directory) were not updated and never set it, contradicting the `ProducesCarry` doc comment ("ADD, MUL, ADDC, MULC") in `instruction/mod.rs`. Latent today (all production consumers derive flags from jolt-riscv rows via `circuit_flags_from_riscv`, and `jolt_trace_cycle_flags_match_final_row_flags` compares row-derived against row-derived, so no test catches the divergence), but any future consumer of the per-instruction trait computes `ProducesCarry=false` for ADD/MUL rows the constraint system treats as producing. Suggest gate-setting it in `add.rs`/`mul.rs`, or a test comparing per-instruction `Flags` impls against jolt-riscv rows.

crates/jolt-prover-legacy/src/zkvm/proof.rs
```diff
@@ -404,6 +405,21 @@ fn commitments_from_proof_payload_order<C>(
         .take(instruction_ra_count)
         .collect::<Vec<_>>();
     let ram_ra = commitments.by_ref().take(ram_ra_count).collect::<Vec<_>>();
+    // The Carry commitment is appended after the BytecodeRa family in
+    // `all_committed_polynomials`; pull it explicitly so the trailing collect
+    // cannot swallow it.
+    #[cfg(feature = "implicit-carry")]
+    let (bytecode_ra, carry) = {
+        let mut rest = commitments.collect::<Vec<_>>();
+        let Some(carry) = rest.pop() else {
```
Acentelles, line 414: Carry occupies a different position in each ordering: last (after BytecodeRa) in `all_committed_polynomials`/the commitment payload, but `insert(2)` after the Inc pair in `final_opening_polynomial_order` (jolt-claims `committed_openings.rs:56`) and `prove_stage8`. That divergence is what forces this collect-then-pop, the mirror WARNING comments, and this error arm — which is unreachable given the `minimum` check above (feature-on `minimum` includes the +1, so `rest` always has an element) and fabricates its count (`got: minimum - 1`). The next trailing committed column breaks the pop and the insert(2) independently, and since `minimum` is only a lower bound the failure mode is `rest.pop()` mis-assigning a BytecodeRa commitment as Carry — an opaque stage-8 batch failure, not a compile error. Suggest one canonical position sourced from jolt-claims' order tables for both the payload split and the final-opening batch, and deleting the dead arm.

crates/jolt-prover-legacy/src/zkvm/spartan/shift.rs
```diff
@@ -659,6 +746,10 @@ impl<F: JoltField> Phase1State<F> {
                             if is_first_in_sequence {
                                 v += params.gamma_powers[3];
                             }
+                            #[cfg(feature = "implicit-carry")]
+                            {
+                                v += params.gamma_powers[5].mul_u64(carry);
```
Acentelles, line 751: Hot-path cost feature-on for every program, including carry-free guests (carry is 0 on almost all rows even in carry-heavy code): (1) this line does an unconditional per-cycle `mul_u64` while the adjacent `is_virtual`/`is_first_in_sequence` terms are branch-guarded precisely because their flags are usually false — `if carry != 0` matches the surrounding pattern; (2) `Phase2State::new` (~line 951) re-walks the whole trace in a separate pass just for `carry_poly`, using the same chunks and `eq_evals` as the 5-column pass above it — a sixth accumulator in the existing pass (struct instead of tuple zip) removes the O(T) traversal; (3) `compute_message` (~line 1011) multiplies the carry term by `eq_plus_one_r_outer_evals[i]` separately from the bracketed sum that already carries that factor — pre-scaling `carry_poly` by `gamma_powers[5]` at construction lets `carry_evals[i]` join the existing bracket, dropping two muls per (j,i) per round to one add.

crates/jolt-prover-legacy/src/zkvm/witness.rs
```diff
@@ -131,6 +142,14 @@ impl CommittedPolynomial {
                     .collect();
                 PCS::process_chunk(&preprocessing.generators, &row)
             }
+            #[cfg(feature = "implicit-carry")]
+            CommittedPolynomial::Carry => {
+                let row: Vec<i128> = row_cycles
+                    .iter()
+                    .map(|cycle| cycle.carry() as i128)
```
Acentelles, line 149: `cycle.carry()` is a non-negative u64 (almost entirely zeros), but widening to `i128` dispatches the 128-bit signed MSM path for every tier-1 row commit — double scalar width, double window passes, 16-byte staging. RdInc/RamInc need i128 only because their diffs are signed. Collecting `Vec<u64>` lets `process_chunk` use the u64 SmallScalar path.

crates/jolt-prover-legacy/src/zkvm/claim_reductions/carry.rs
```diff
@@ -0,0 +1,281 @@
...
+{
+    fn get_params(&self) -> &dyn SumcheckInstanceParams<F> {
+        &self.params
+    }
+
+    #[tracing::instrument(skip_all, name = "CarryClaimReductionSumcheckProver::compute_message")]
+    fn compute_message(&mut self, _round: usize, previous_claim: F) -> UniPoly<F> {
+        let half_n = self.carry_poly.len() / 2;
+        let mut evals = [F::zero(); DEGREE_BOUND];
+        for j in 0..half_n {
```
Acentelles, line 195: This scan is single-threaded while every other stage-6 reduction (e.g. `IncClaimReductionSumcheckProver`) parallelizes with rayon — at T=2^24 the first round is 8.4M sequential iterations on one core while the batched round waits. Also `initialize` (~line 165) materializes three full 2^log_T field vectors (`eq_product`, `eq_shift`, collected `eq_combined`) for one table; accumulating the gamma-scaled shift table in place into the `eq_product` buffer drops one full-size allocation and the collect.

Three smaller verified cleanups elsewhere in the carry lane, folded here to keep this review to one submission:
- `jolt-verifier/src/stages/stage6b/carry_claim_reduction.rs` (~line 115): the `EqZeroSelector` arm allocates `vec![F::zero(); n]` and runs `try_eq_mle` — `EqPolynomial::zero_selector` (jolt-poly, already a dependency; used by this file's legacy counterpart) computes it with no allocation and no unreachable error path.
- `jolt-claims/src/protocols/jolt/relations/spartan/shift.rs` (~line 132): `output_expression` duplicates the entire four-term base sum under cfg(feature)/cfg(not(feature)), differing only in the added carry term; the same file's `input_expression` shows the compact single-base + cfg-rebind pattern.
- `jolt-verifier/src/config.rs`: the implicit-carry `compile_error!` comments cite `specs/implicit-carry-handling.md`, which does not exist in the repo.

crates/jolt-prover-legacy/src/zkvm/claim_reductions/carry.rs
```diff
@@ -0,0 +1,281 @@
...
+        params: CarryClaimReductionSumcheckParams<F>,
+        trace: Arc<Vec<Cycle>>,
+    ) -> Self {
+        let [gamma, gamma_sqr] = params.gamma_powers;
+        let t = 1 << params.n_cycle_vars;
+
+        let carry_coeffs: Vec<u64> = trace
+            .par_iter()
+            .map(|cycle| cycle.carry())
+            .chain(rayon::iter::repeat_n(0u64, t - trace.len()))
```
Acentelles, line 161: Two issues: (a) this inline map is a second definition of the committed Carry column alongside `CommittedPolynomial::Carry`'s `generate_witness` arm added in this PR (witness.rs) — they must agree bit-for-bit or stage-6b opens a different polynomial than the one committed in stage 1, and nothing ties them together; (b) the `.chain(repeat_n(0, t - trace.len()))` padding is dead code — the prover resizes the trace to the padded power-of-two with NoOps before any stage runs (`prover.rs`, `trace.resize(padded_trace_len, ...)`), and `n_cycle_vars = trace_len.log_2()` already assumes it. Suggest calling the `generate_witness` path, or at minimum dropping the padding chain.

crates/jolt-prover-legacy/src/zkvm/r1cs/evaluation.rs
```diff
@@ -1094,12 +1215,17 @@ impl ProductVirtualEval {
                 out_unr[5] = eq1_val.mul_to_product_accum(acc_branch_flag.barrett_reduce());
                 out_unr[6] = eq1_val.mul_to_product_accum(acc_next_is_noop.barrett_reduce());
                 out_unr[7] = eq1_val.mul_to_product_accum(acc_virtual_instr_flag.barrett_reduce());
+                #[cfg(feature = "implicit-carry")]
+                {
+                    out_unr[8] = eq1_val.mul_to_product_accum(acc_uses_carry_flag.barrett_reduce());
```
Acentelles, line 1220: The PR adds the right generic extension (`ProductFactorExpr::Committed` + a CarryUsed row in `PRODUCT_CONSTRAINTS`), but no consumer reads the table — the carry lane is re-special-cased with hardcoded indices: `out_unr[8]`/`out_unr[9]` here, `spartan/product.rs` binding Carry to `claims[claims.len() - 1]` (~line 828), and weight slot 3 hardcoded in both the verifier's fused factors and jolt-claims' `product_remainder`. Adding a fifth product constraint or reordering `PRODUCT_CONSTRAINTS` breaks the "carry is last / slot 3" convention with no compile error: `claims[len-1]` silently binds a different factor's claim to the Carry opening. Deriving factor lists, claim indices, and weight slots by iterating `PRODUCT_CONSTRAINTS` makes the table the single source of truth it was added to be.

mathmasterzach:
I've completed a careful pass over this and #1763. Once this is rebased into #1764 it looks sound to me. I just have a few small cleanup/clarity comments.

In `crates/jolt-prover-legacy/src/zkvm/instruction/add.rs` and `crates/jolt-prover-legacy/src/zkvm/instruction/mul.rs`, `impl Flags` appears to be missing: `flags[CircuitFlags::ProducesCarry] = true;`.

This doesn't introduce a correctness issue because `JoltTraceCycle::circuit_flags()` delegates to `JoltInstructionRow` where ADD and MUL are seperately marked as carry producers. If `impl Flags` is truly legacy, then I think don't think it strictly needs to be aligned, but given that `flags[CircuitFlags::ProducesCarry] = true;` is included in `crates/jolt-prover-legacy/src/zkvm/instruction/addc.rs` and `crates/jolt-prover-legacy/src/zkvm/instruction/mulc.rs`, I believe it should be included here.

Also, these instructions should be documented in the jolt book, including the explicit point that instructions like `add x0, a, b` produce a usable carry and thus are not optimized away to avoid footguns.

Acentelles:
Thanks for the careful pass. All three points are addressed, and #1763 is now folded into this PR as requested (branch rebased onto the ISA tip, base retargeted to `main`, #1763 closed).

- `impl Flags` for the legacy ADD/MUL now sets `flags[CircuitFlags::ProducesCarry] = true` (6abaf88e7), consistent with `addc.rs`/`mulc.rs`.
- Book page added: `book/src/how/optimizations/implicit-carry.md` (e298734fb), wired into SUMMARY. It documents ADDC/MULC and calls out the footgun explicitly: `add x0, a, b` / `mul x0, a, b` are real carry producers and are never optimized away.

The rebase also surfaced a real regression from the ISA branch's guest-boundary carry clobber, fixed in be2b81a1c: the appended `ADDI x0,x0,0` clobber row was the first-ever rd=x0 row inside a virtual sequence, and the tracer's rd0-expansion re-expanded it as a source at execution time, so the executed cycle disagreed with the bytecode row (stage-1 bytecode read-raf mismatch on the advice/stdlib e2e tests, feature-on). `takes_rd0_expansion` now applies only to top-level instructions. Validated: legacy feature-on e2e suite (muldiv incl. committed-program variants, carry_chain, advice x2, stdlib) plus feature-off muldiv, clippy clean both modes.

crates/jolt-program/src/expand/mod.rs
```diff
@@ -171,6 +172,61 @@ fn expand_source_instruction_with_provider<P: InlineExpansionProvider + ?Sized>(
     result
 }
 
+/// Feature-on, a guest instruction that is not a carry producer must leave
+/// carry = 0. Recipes may legitimately end in ADD/MUL rows (MULH, SLL) and may
+/// nest producer-ending helpers like SLL mid-sequence, so the guarantee is
+/// enforced once at the source-instruction boundary: append a non-producing
+/// noop and re-stamp the sequence metadata. Inline sequences are exempt — the
```
0xAndoroid, line 179: **Inline sequences bypass the clobber, so an inline ending in `ADD`/`MUL` exports its carry-out.** `clobber_trailing_carry` sits only on the non-inline arm of `expand_source_instruction_with_provider`; the `Inline` arm goes `materialize_inline → final_rows_to_instructions`, and the tracer mirrors that (`INLINE::trace_sequence` runs the rows with no trailing clobber). The book page says *every other instruction clobbers the carry to zero* and the spec's row-level rule is `NextCarry = 0` for all non-producers — an inline is a non-producer at the guest level.

Not a soundness gap (trace and bytecode agree, the proof is consistent either way), but it is an ISA-contract hole exactly at the boundary this feature targets: bigint/secp256k1 inlines rebuilt on ADDC/MULC will end in carry producers, and hand-written `asm!` placed right after one reads the inline's internal carry — the same footgun the trailing-noop rule exists to prevent for MULH/SLL.

Fix: run the same trailing-noop rule on the `Inline` arm (one more `.and_then(|rows| clobber_trailing_carry(...))`, using the inline's source row for the noop address), or make the exemption a stated contract in the book and require providers to end on a non-producer. Add the inline case to `expanding_non_producers_clobber_carry_to_zero` either way.

crates/jolt-verifier/src/config.rs
```diff
@@ -46,6 +58,20 @@ pub struct JoltProtocolConfig {
     pub zk: ZkConfig,
     pub commitment: CommitmentConfig,
     pub scalar_challenge_endianness: ScalarChallengeEndianness,
+    /// The ISA axis: present only in implicit-carry builds so the feature-off
+    /// wire format is unchanged. Cross-build proofs fail closed at
+    /// deserialization (the config frames differ).
+    #[cfg(feature = "implicit-carry")]
+    pub isa: IsaConfig,
```
0xAndoroid, line 65: **Cross-build rejection is by codec accident, not by `validate_proof_config`.** With `isa` present only feature-on, a feature-off verifier handed a feature-on proof fails wherever the deserializer happens to misalign (extra `carry` commitment, extra claim fields, this field), with whatever error that yields; nothing reaches the `ProtocolConfigMismatch` path. The stated motivation — an unchanged feature-off wire format — buys nothing this PR doesn't already give up: the feature-off proof bytes are byte-identical to pre-PR only because the feature is off, and the repo carries no wire-compat requirement.

Fix: make the axis unconditional, `pub enum IsaConfig { Rv64imac, Rv64imacImplicitCarry }` with the selected variant chosen by `cfg!`, and add the mirror of `rejects_scalar_challenge_endianness_mismatch`. One byte in every proof; fixtures regenerate.

crates/jolt-claims/src/protocols/jolt/relations/spartan/product_remainder.rs
```diff
@@ -92,12 +100,22 @@ impl SymbolicSumcheck for ProductRemainder {
...
+        #[cfg_attr(not(feature = "implicit-carry"), expect(unused_mut))]
+        let mut right = product_weight(0) * opening(right_instruction_input_product())
             + product_weight(1) * opening(branch_flag_product())
             + product_weight(2) * (JoltExpr::one() - opening(next_is_noop_product()));
+        // Weight slot 3 = the CarryUsed row's position in the legacy prover's
+        // `PRODUCT_CONSTRAINTS` table, pinned there by a const assertion.
+        #[cfg(feature = "implicit-carry")]
+        {
+            use crate::protocols::jolt::geometry::spartan::{carry_product, uses_carry_product};
+            left = left + product_weight(3) * opening(uses_carry_product());
```
0xAndoroid, line 116: **The lane position is pinned in one crate and open-coded in five.** `product_weight(3)` here, `product_uniskip_weight(3)`, `gamma.pow(5)` / `gamma_powers[5]` in both shift relations, `Challenge(5)`/`Challenge(6)` in the legacy BlindFold terms, `stage2_gammas[4]` in both read-RAF paths, and `claims[claims.len() - 1]` in `ProductVirtualRemainderProver::cache_openings` all encode "CarryUsed is the last product lane / the carry term is the last shift term". The const assertion in legacy `r1cs/constraints.rs` pins only the legacy `PRODUCT_CONSTRAINTS` table; nothing ties the jolt-claims, jolt-verifier, or jolt-r1cs literals to it, so a fifth lane or a reorder on one side compiles cleanly and fails only in the e2e.

Fix: give the two indices one owner and reference it everywhere — e.g. `pub const CARRY_USED_LANE: usize = SPARTAN_PRODUCT_BASE_LANES + SPARTAN_PRODUCT_FIELD_INLINE_LANES` next to `SPARTAN_PRODUCT_IMPLICIT_CARRY_LANES` in `jolt-r1cs/src/constraints/jolt.rs`, and `SHIFT_GAMMA_COUNT - 1` (already exists on the legacy side) for the shift term; then the legacy const assert can compare against the same constant instead of a literal 4.

crates/jolt-prover-legacy/src/zkvm/r1cs/constraints.rs
```diff
@@ -407,6 +431,27 @@ pub static R1CS_CONSTRAINTS: [NamedR1CSConstraint; NUM_R1CS_CONSTRAINTS] = [
         if { { JoltR1CSInputs::NextIsVirtual } - { JoltR1CSInputs::NextIsFirstInSequence } }
         => ( { 1i128 } ) == ( { JoltR1CSInputs::OpFlags(CircuitFlags::DoNotUpdateUnexpandedPC) } )
     ),
+    // if ProducesCarry {
+    //     assert!(RightLookupOperand == LookupOutput + 2^64 * NextCarry)
+    // }
+    // The lookup domain bounds RightLookupOperand < 2^128 and the RangeCheck
```
0xAndoroid, line 437: **The range argument rests on two facts that hold by inspection only.** `NextCarry ∈ [0, 2^64)` follows from (1) `RightLookupOperand` being bound to the 128-bit lookup index by the instruction read-RAF identity with `LeftLookupOperand = 0`, which the `Add|Sub|Mul|Advice ⇒ LeftLookupOperand == 0` guard provides for the four producers, and (2) every `ProducesCarry` kind using `RangeCheck`, so `LookupOutput = RLO mod 2^64`. Both are true for ADD/MUL/ADDC/MULC today and pinned nowhere: a later `ProducesCarry` instruction on any other table (or one without `AddOperands`/`MultiplyOperands`, which drops the RLO guard) leaves `NextCarry` unconstrained above bit 63 and no test fails, because the honest prover still satisfies the split.

Fix: a table-level test in jolt-lookup-tables/jolt-riscv iterating `JoltInstructionKind`: for every kind whose circuit flags contain `ProducesCarry`, assert `lookup_table() == Some(RangeCheck)` and `AddOperands || MultiplyOperands`. Cheap, and it is the only thing that turns this comment into an enforced invariant (CLAUDE.md: name the mechanism that pins each invariant).

crates/jolt-prover-legacy/src/zkvm/prover.rs
```diff
@@ -3617,6 +3652,209 @@ mod tests {
...
+            None,
+            None,
+        )
+        .unwrap();
+        let io_device = prover.program_io.clone();
+        let (jolt_proof, _debug_info) = prover
+            .prove()
+            .expect("prover should produce verifier-native proof");
+
+        // Tamper coverage across the distinct carry grounding paths: each
```
0xAndoroid, line 3733: **No test feeds a dishonest witness through the constraint system.** These five tampers and the eight manifest leaves in jolt-verifier all mutate *claims* of an honestly generated proof, so they exercise the verifier's algebra (any claim offset on a consistent proof fails whether or not a given constraint exists). Nothing checks that the constraints themselves reject a wrong carry: a wrong guard on `LookupSplitsIntoOutputAndNextCarry`, a shift term wired in the wrong direction, or a dropped `Carry(0) = 0` term would all pass this suite.

Fix (cheapest first): (a) a unit test on `R1CSEval` with a hand-built `R1CSCycleInputs` whose `next_carry`/`carry_used` disagree with `right_lookup`/`lookup_output`, asserting the two carry rows evaluate nonzero; (b) one e2e that mutates `cycle.carry` on the ADDC row of the traced `carry-chain` guest before proving (the CPU-computed `rd` then disagrees with `LookupOutput`) and asserts the prover's constraint check trips / the verifier rejects. (b) is the only test that would have caught a wrong `carry_witness_coeffs` → shift → claim-reduction chain end to end.

crates/jolt-prover-legacy/src/zkvm/spartan/shift.rs
```diff
@@ -168,19 +192,31 @@ impl<F: JoltField> SumcheckInstanceParams<F> for ShiftSumcheckParams<F> {
...
+        let mut values = vec![
             self.gamma_powers[1],
             self.gamma_powers[2],
             self.gamma_powers[3],
             self.gamma_powers[4],
             -self.gamma_powers[4],
-        ]
+        ];
+        #[cfg(feature = "implicit-carry")]
+        values.push(self.gamma_powers[5]);
```
0xAndoroid, line 218: **Every `#[cfg(feature = "zk")]` carry arm in this PR is unreachable code.** `implicit-carry` + `zk` is a `compile_error!` in jolt-verifier (and reaches jolt-prover-legacy through the feature forwarding), and no CI lane builds the pair, so this push, the `Challenge(5)`/`Challenge(6)` product terms, the `output_constraint_challenge_values` arm, the product.rs opening lists, and `CarryClaimReductionSumcheckParams::{input,output}_claim_constraint` are neither compiled nor tested anywhere. The CLAUDE.md sync invariant (`input_claim` ↔ `input_claim_constraint`) is precisely what an untested arm violates silently, and the complexity rule treats caller-less paths as blockers.

Fix: either drop the ZK arms from this PR and land them with the ZK-enablement PR where `muldiv` ZK e2e covers them, or lift the `compile_error!` now and add the `host,zk,implicit-carry` clippy + e2e lane so the arms are at least exercised by BlindFold R1CS satisfiability.

0xAndoroid (commented):
Reviewed the healed head (`999965d`, i.e. this branch merged with `origin/main` at `8e55acc`) for soundness of the implicit-carry scheme, transcript handling, spec conformance, tests, and complexity. Six inline comments; none is a soundness break of the scheme as implemented, three ask for enforcement/tests of properties that currently hold by inspection.

**Verified**
- Carry range: `NextCarry ∈ [0, 2^64)` from the split row, given `RightLookupOperand` is the 128-bit lookup index (read-RAF identity, `LeftLookupOperand = 0`) and `RangeCheck` yields `RLO mod 2^64`; `Carry(t+1) = NextCarry(t)` through the `EqPlusOne` shift term with no wraparound (row `T-1` forced to 0, satisfied by padding); `Carry(0) = 0` through the `γ²·eq(0,·)` public pair with claim 0. Both `Carry` openings (product cycle point, shift point) come from the single committed column, so the chain has one proof-visible source.
- `CarryUsed = UsesCarry·Carry` as the 4th product lane degrades to today's `RightLookupAdd`/`RightLookupEqProductIfMul` for every flag-bearer with `UsesCarry = 0`; MULHU/ADDI-class rows are untouched because only `ProducesCarry` gates the split.
- Transcript: the `Carry` commitment is absorbed after the `BytecodeRa` family in `absorb_transcript_commitments`, matching `all_committed_polynomials`; `carry_gamma` is drawn right after `inc_gamma` in `Stage6bDraws`, matching the legacy `challenge_scalar_powers(3)`; stage-8 batch order agrees entry-for-entry (`final_opening_polynomial_order` inserts `Carry` at index 2); FS inventories updated.
- Tracer/bytecode parity: virtual sequences expand through `jolt_program::expand`, so the trailing clobber row is in both trace and bytecode; x86 AOT declines `AddC`/`MulC` rows and models ADD/MUL carry-out in reassembly with the `skip_rows ≥ 1` resume rule.
- Cost: one extra `u64` committed column, product uniskip first-round degree 6→9, outer 27→30, one extra degree-2 stage-6b sumcheck over `T`, one extra shift column; no new lookup tables. Consistent with the reported flat sha3 wall time.
- Spec deviations listed in the description (ZK fail-closed, `Carry(0)=0` as a public pair, modular prover not proving, jolt-eval invariant deferred) match the code.

**Heal notes** (commits `b8147cd`, `999965d`, no rebase): `Carry` serialization tag renumbered 19→11 after main's `BalancedIncDigit`/`BalancedIncCarry` took 9/10; `RingCore→Ring`; verifier `CarryClaimReduction` bounds moved to `JoltField`; main's new `TraceRow` literals routed through `TraceRow::new`; `carry_chain_e2e_dory` ported to the fallible `gen_from_elf`; 37 nominal-import findings from the new `style` job fixed (two named rv64 row constants replace the open-coded `NUM_EQ_CONSTRAINTS - k`). No parity fixtures needed regeneration. Local: clippy in all five CI lanes, `muldiv` host/zk, `carry_chain_e2e_dory`, `mul256`, modular + tracer suites feature-on/off all green.

Review produced with Claude Code on Andrew's behalf.

0xAndoroid:
Healed after #1818 (`jolt-prover-legacy` removed): 17e04e2e9 merges `main` (d39bd518a) — the 25 legacy-prover files are dropped and the shift relation's sixth gamma term (`NextCarry(t) = Carry(t+1)`) now lives in `Shift::unbatched_relation()`.
The legacy prover was this PR's only producer of the `Carry` column, so fac11fe4d ports the proving side to the modular prover: `jolt-witness` serves `Carry`/`NextCarry`/`CarryUsed` (the backend no longer fails closed, so deviation 3 in the description is obsolete), `jolt-kernels` gains the outer/product/shift carry terms plus reference and optimized `CarryClaimReduction` kernels, and `jolt-prover` commits and opens `Carry` (stage 0 / stage 8) behind a new `implicit-carry` feature; `implicit-carry` × {`akita`, `zk`} still fails closed at compile time.
`standard_carry_chain_verifier_proof_is_accepted` now proves and verifies `examples/carry-chain` end-to-end on the modular prover (Dory, default feature set + `implicit-carry`), and the CI lanes cover `jolt-kernels`/`jolt-prover`/`jolt-host` feature-on.
The carry-chain fixture also exposed a latent overflow in the optimized product-remainder kernel (two full-range `field × u64` products folded into one small-scalar accumulator exceed its 2^318 Barrett budget, corrupting the left table for ~23 % of weight draws); fixed via `fmadd_u64_split` and pinned feature-independently by `wide_lane_parity_with_reference_kernels`.
Verified locally: clippy lanes (host / host,zk / akita / implicit-carry / sdk host,implicit-carry), the touched crates' suites feature-on and -off, verifier fixture + tamper tests on and off, akita e2e feature-off.

PR #1765, opened 2026-08-11
ci: PR diff classification comment

0xAndoroid:
Maintains one sticky comment per PR (created once, edited in place, keyed on a hidden `<!-- diff-classifier -->` marker) that splits the diff into actual code / tests / docs / fixtures / helper scripts, rendered as an aligned table in a `diff` fence so additions color green and removals red:

```diff
@@              added  removed @@
+ actual code   +1905
- actual code             -207
+ tests          +744
- tests                    -21
+ docs            +16
- docs                     -38
+ fixtures         +2
- fixtures                  -1
  helper scripts    0        0
  ----------------------------
+ total         +2667
- total                   -267
```

(Color is per-line in diff highlighting, so each category is a `+` line for additions and a `-` line for removals; zero sides are omitted and an all-zero category collapses to one neutral line.)

## Categories & precedence

Every added/removed line gets exactly one category (first match wins), so the categories always sum to the total:

1. **fixtures** — `Cargo.lock`, `*.snap` anywhere, and data-extension files (`.json`, `.jsonl`, `.bin`, `.txt`, `.elf`, …) under fixture-ish directories (`fixtures/`, `testdata/`, `golden/`, `snapshots/`, `corpus/`, …) or under `tests/` / `benches/` (test *inputs*, not test code).
2. **docs** — `*.md`/`*.mdx` anywhere, and anything under `docs/`, `specs/`, or `book/`.
3. **tests (file level)** — anything under a `tests/` or `benches/` directory, and source files named like tests in any language (`test.rs`, `tests.rs`, `test_*.py`, `*_test(s).rs`, `bench*.rs`, …) — e.g. `zkvm/lookup_table/test.rs` is test-helper infrastructure even though it isn't `cfg(test)`-gated. Tests beat helper scripts: a test-named file under `scripts/` counts as tests.
4. **helper scripts (path level)** — anything under `scripts/` or `tools/`, including Rust helpers living there (e.g. `scripts/ci/wasm-e2e/prove_and_save.rs`): scripts/ beats code.
5. **actual code** — Rust/jolt code only: `.rs` files analyzed line by line (lines inside `#[cfg(test)]` / `#[cfg(all(test, …))]` items or `#[test]`/`#[tokio::test]`-attributed fns count as tests, with a test region winning over doc comments inside it; `#[cfg(any(test, …))]` is *not* test-only since it can compile in non-test builds; remaining pure doc-comment lines `///`/`//!` count as docs), plus `Cargo.toml` manifests and native guest sources (`.c`/`.h`/`.S`/`.ld`, …).
6. **helper scripts (catch-all)** — everything else: CI workflows (`.yml`/`.yaml`), `.py`/`.sh` tooling, justfile/Makefile-type files, lint/deploy configs, and other non-Rust support files.

Added lines are resolved against the head revision, removed lines against the merge-base revision, so `cfg(test)` region membership is computed at the revision where the line actually exists. The lexer tracks block comments / strings / raw strings across lines, so braces inside string literals or comments don't corrupt region boundaries (approach ported from a standalone Rust LOC counter).

## Mechanics

- `pull_request` on `[opened, synchronize, reopened, ready_for_review]` — drafts included, refreshed on every push.
- Fast: `filter: blob:none` full-history checkout (commits + trees only; git batch-fetches just the blobs the diff touches), exact `merge-base` three-dot diff — totals match GitHub's Files-changed counts exactly.
- Non-blocking: classification errors degrade to a job-summary note; the job never fails the PR.
- Fork PRs: the default token is read-only there, so the comment step is skipped and the classification lands in the job summary instead.
- Classifier is a single stdlib-only Python file (`scripts/ci/diff_classifier.py`), no dependencies.

## Spec-workflow gating

`spec-tracking.yml`'s large-PR spec nudge now counts only **actual code** lines (the classifier's `code` category, added + removed, via `--format json`) instead of all non-spec lines:

- `code_lines <= 500` -> job passes with a summary note ("spec not required: N actual-code lines <= 500"); no comment.
- `code_lines > 500` with a `specs/*.md` in the diff -> passes ("spec present").
- `code_lines > 500` without a spec -> the existing warning comment, now stating the actual-code count (dedupe grep unchanged, so PRs already warned under the old wording are not re-warned).
- Classifier failure -> falls back to the previous all-lines count, so the nudge never silently disappears. The step runs the classifier from the base checkout (`pull_request_target`), never code from the PR head, and needs no token — fork PRs behave identically.

Gating examples (classifier run on real PRs): #1755 -> 2140 actual-code lines, no spec -> still warns; #1727 -> 515 actual-code lines (of 1031 total), has a spec -> "spec present"; this PR -> 0 actual-code lines of 993 total -> "spec not required" (under the old all-lines rule it was warned — see the stale warning comment below).

## Validation against real PRs

Totals match GitHub's official additions/deletions exactly:

| PR | GitHub | code | tests | docs | fixtures | helper |
|---|---|---|---|---|---|---|
| #1748 (remove dead VirtualLW/SW) | +7 −106 | +0 −94 | +2 −4 | +5 −8 | +0 −0 | +0 −0 |
| #1755 (sub-word lane framework) | +2667 −267 | +1933 −207 | +716 −21 | +16 −38 | +2 −1 | +0 −0 |
| #1758 (Z3 nightly workflow) | +135 −6 | +17 −0 | +16 −6 | +6 −0 | +20 −0 | +76 −0 |

Spot-checked by hand: `SourceInstructionKind::VirtualSW` removal inside `mod tests` → tests; `///` edits outside test regions → docs; `///` edits inside `mod tests` → tests; `main_expand_parity_hashes.json` golden update → fixtures; Cargo.lock churn → fixtures; `z3-nightly.yml` → helper scripts; `z3-verifier/Cargo.toml` → actual code.

This PR's own sticky comment below is the live demo (everything lands under helper scripts, as it should).

## Known limits

- Block doc comments (`/** */`, `/*! */`) count as code; the codebase uses line doc comments.
- A bare `#[test]` fn's *preceding* unrelated attributes (e.g. `#[serial]` above `#[test]`) count as code; stacked attributes *after* the region trigger are included.


markosg04 (approved):

scripts/ci/diff_classifier.py
```diff
@@ -0,0 +1,876 @@
...
+                continue
+            if scan.has_code:
+                test_lines.add(lineno)
+                new_item = _Item(keyword_item=bool(_ITEM_KEYWORD.match(code)))
+                if not new_item.consume_line(scan):
+                    item = new_item
+                await_item = False
+            else:
+                test_lines.add(lineno)  # blank/comment line between attr and item
+            continue
```
moodlezoup, line 532: The `await_item` flow only recognizes an attribute by `#` on its *first* line, so a multiline attribute's continuation lines get fed to `_Item` as if they were the item. Confirmed:

```rust
#[test]
#[should_panic(
    expected = "boom",
)]
fn dies() { helper(); }
```

marks only lines 1–3 as tests — the trailing `,` on line 3 ends the pseudo-item at group depth 0 (the paren opened on a line this `_Item` never saw), so `)]` and the whole fn leak to code. The same-line variant fails the other way: `#[cfg(test)] use foo::Bar;` completes attribute and item on one line, but `await_item` stays armed and swallows the *next* (production) item as tests. Neither pattern exists in the repo today (rustfmt splits attributes), but both skew the `code` count that now gates the spec nudge — worth fixing, or at least adding to the documented known limits (the current list only covers unrelated attrs *before* `#[test]`).

---

scripts/ci/diff_classifier.py
```diff
@@ -0,0 +1,876 @@
...
+                out.append(chr(int(body[i + 1 : i + 4], 8)))
+                i += 4
+                continue
+        out.append(c)
+        i += 1
+    return "".join(out)
+
+
+def _strip_prefix(path: str, prefix: str) -> str | None:
+    path = _unquote_git_path(path)
```
moodlezoup, line 594: Git appends a literal tab after the filename on `--- a/...`/`+++ b/...` lines when the path contains a space (unquoted, so `_unquote_git_path` never fires). The tab survives into the stored path, `file_extension()` returns `".rs\t"`, and a Rust file like `src/my file.rs` silently falls through every rule to the helper catch-all instead of line-level analysis (confirmed). No tracked paths with spaces exist in jolt today, but the failure is silent misclassification rather than an error. A path can't legitimately end in an unquoted tab (git c-quotes control characters), so stripping it is safe:

```suggestion
    path = _unquote_git_path(path.rstrip("\t"))
```

---

scripts/ci/diff_classifier.py
```diff
@@ -0,0 +1,876 @@
...
+    scanner = _Scanner()
+    test_lines: set[int] = set()
+    doc_lines: set[int] = set()
+
+    pending_cfg: _CfgAttr | None = None
+    await_item = False  # a test-only attribute was seen; consume attrs then item
+    item: _Item | None = None
+    rest_of_file_is_test = False
+
+    lines = content.splitlines()
```
moodlezoup, line 480: `splitlines()` splits on `\f`, `\v`, U+2028, etc., which git does not count as line terminators — a single U+2028 inside a string literal shifts every subsequent line number, so diff line numbers resolve against the wrong region for the rest of the file (confirmed: an appended production fn after a `#[cfg(test)]` mod came back as tests).

```suggestion
    lines = content.split("\n")
```

---

scripts/ci/diff_classifier.py
```diff
@@ -0,0 +1,876 @@
...
+        # enum variants, match arms. Item declarations (fn/mod/struct/...)
+        # never end at a comma — a multiline fn signature has `,`-terminated
+        # parameter lines but is either inside unclosed parens or followed by
+        # its braced body.
+        return not self.keyword_item and code.endswith(",")
+
+
+def _starts_cfg_attr(code: str) -> bool:
+    code = code.lstrip()
+    return code.startswith("#[cfg") or code.startswith("#![cfg")
```
moodlezoup, line 274: `#[cfg_attr(...)]` also starts with `#[cfg`, so `#[cfg_attr(all(test), ignore)]` on an always-compiled fn marks the entire fn as tests (confirmed): `_cfg_arg` finds the `cfg` inside `cfg_attr`, and `_call_args(expr, "all")` matches the leading `all(test)` while ignoring the trailing `, ignore`. `#[cfg_attr(test, ignore)]` only escapes by accident (its arg string isn't exactly `test`). Matching `#[cfg(` / `#[cfg(` here closes the hole for everything except the rare whitespace form `#[cfg (…)]`, which the current prefix check misses anyway.

---

scripts/ci/diff_classifier.py
```diff
@@ -0,0 +1,876 @@
...
+    parser.add_argument(
+        "--no-merge-base",
+        action="store_true",
+        help="diff base..head directly instead of merge-base(base, head)..head",
+    )
+    parser.add_argument(
+        "--format",
+        choices=("text", "markdown", "json"),
+        default="text",
+    )
```
moodlezoup, line 837: Neither workflow uses `--format text` (diff-classifier.yml passes `markdown`, spec-tracking.yml passes `json`), yet `text` is the default — `format_summary`, `CATEGORY_LABELS`, and `MINUS` exist only for that path. Same for `--no-merge-base` and `--repo`: no caller. Defaulting to `markdown` and dropping the text path trims ~25 lines and a second label dict. Relatedly, `--details` (which diff-classifier.yml does pass) prints only to stdout, so the per-file breakdown lands in the raw job log, not the step summary or comment — if that's intentional it may not be worth the `per_file` plumbing in `classify()`; if not, append it to the `--output` file.

---

.github/workflows/spec-tracking.yml
```diff
@@ -94,42 +94,65 @@ jobs:
...
-            awk '$1 != "-" && $2 != "-" { added += $1; removed += $2 } END { print (added + removed) ? added + removed : 0 }')
+          # The spec nudge counts only ACTUAL CODE lines (added + removed):
+          # the diff classifier's "code" category — Rust/jolt code. Tests,
+          # docs, fixtures, and helper scripts are excluded, so e.g. a large
+          # test-only or CI-only PR no longer trips the spec requirement.
+          # The classifier diffs merge-base(BASE, HEAD) against HEAD, same
+          # three-dot semantics as before. It runs from the BASE checkout
+          # (pull_request_target), never code from the PR head.
+          CODE_LINES=""
+          if JSON=$(python3 scripts/ci/diff_classifier.py --base "$BASE" --head "$HEAD" --format json); then
```
moodlezoup, line 115: This replaces one fast `git diff --numstat` with a per-changed-line Python pass that `git show`s and lexes every touched `.rs` file at two revisions — in a `pull_request_target` job with no `timeout-minutes` (default 360) and no `concurrency` group, running for fork PRs without the first-contribution approval gate. The failure containment is solid (everything degrades to the numstat fallback), but a crafted PR with enormous `.rs` diffs can pin a runner for hours, once per push, with no cancel-in-progress. `diff-classifier.yml` already sets `timeout-minutes: 10` and `cancel-in-progress: true`; this job should get the same.

---

.github/workflows/diff-classifier.yml
```diff
@@ -0,0 +1,88 @@
...
+            const fs = require('fs');
+            const marker = '<!-- diff-classifier -->';
+            const body = fs.readFileSync(process.env.COMMENT_PATH, 'utf8');
+            const comments = await github.paginate(github.rest.issues.listComments, {
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              per_page: 100,
+            });
+            const existing = comments.find(c => c.body && c.body.includes(marker));
```
moodlezoup, line 73: `find` matches *any* comment containing the marker, with no author filter — anyone who comments the marker string before the workflow's first run captures the slot: the bot edits their comment from then on and never creates its own, and between pushes they can edit "their" comment to display whatever numbers they like. Adding `c.user.type === 'Bot'` (or matching the actions login) to the predicate closes it. This step is also a near-copy of the sticky upsert in `.github/workflows/bench-crates.yml` (which shares the same gap) — worth extracting a shared composite action, or at least fixing both the same way.

---

moodlezoup (commented):
Reviewed the classifier's Rust region tracking, diff parsing, and both workflow integrations, with the line-level logic exercised directly against synthetic inputs. The core partition contract, diff-parsing edge cases (renames, `-U0` zero-count hunks, binary files, quoted/unicode paths), the `pull_request_target` base-checkout isolation, shell fallback paths, and injection surfaces all check out. Inline comments cover a handful of confirmed edge-case bugs in the Rust test-region state machine, two workflow hardening items, and some unused CLI surface.

---

PR #1768, opened 2026-08-11
feat: fused read-modify-write for narrow stores (SB/SH/SW)

Acentelles:
Stacked on #1762; completes the Tier 0 load/store family.

- **New tables**: `ShiftDataB/H/W` — the store value shifted into its lane, `(rs2 mod 2^8w) << (8·(ea mod 8 & m))` with the same restricted-offset-bit reads as the window masks (m = 7/6/4). A product over disjoint variables, so each decomposes at bilinear rank 2 (a joint lane-accumulator prefix plus a Pow2-style offset-scale pair). A general `Pdep` table was analyzed and rejected: prefix rank ~57 (the deposited bit index depends on suffix popcount).
- **Sequences**: all three stores unify under one helper — `[assert] ADDI ANDI LD WindowMask ANDN ShiftData ADD SD`. SB 13 -> 8 traced cycles, SH 14 -> 9, SW 15 -> 9. `ANDN` handles mask-old in one row (BitManipulation is profile-legal).
- **Fixtures**: 27 golden hashes re-baselined (SB/SH/SW, plus SCW which embeds a word store); inline fixture verified unchanged.

**Cumulative Tier-0 trace deltas** vs the spec-branch baseline: btreemap **-8.9%**, sha2-chain **-7.2%**, memory-ops -23%, fibonacci flat.

Validation: modular + legacy shift_data tests (18 + 15), ABI test, jolt-program 38/38, muldiv e2e in `host` and `host,zk`, clippy clean on touched crates.


Acentelles:
Prover wall-time parity check at fixed padded size (interleaved A/B, quiet machine, spec-branch baseline vs this branch): baseline 1.52/1.44/1.41 s vs stores 1.47/1.40/1.40 s — parity within noise. The cumulative Tier-0 table/prefix machinery costs nothing measurable at equal instance size; the benefit is the trace reduction (btreemap -8.9%, sha2-chain -7.2% across the stack), which converts to wall time at padding-boundary crossings (see the 1.37x measurement on #1762) and to guest capacity at fixed instance sizes.

Acentelles:
@0xAndoroid  Ethproofs-class measurement: traced mainnet block 25698189 (41.93M gas) with [0xAndoroid/jeth](https://github.com/0xAndoroid/jeth) against this stack merged onto current main, vs clean main:

| | main | + Tier 0 stack |
|---|---|---|
| total cycles | 1,289,309,699 | 1,203,812,048 |
| cycles/gas | 30.75 | **28.71 (-6.6%)** |

Identical block hash and passing post-state-root assertion on both sides — 1.2B cycles of real EVM execution over the new load/store expansions. This also partially answers the spec's open question 5 (Ethproofs-class impact): the interpreter's byte-granular memory/calldata handling benefits as projected, sitting between the btreemap (-8.9%) and sha2-chain (-7.2%) proxies' per-workload profile. Merge-onto-main preview for the stack: one trivial conflict pair (main's dead VirtualLW/VirtualSW removal vs our adjacent additions).

crates/jolt-lookup-tables/src/tables/suffixes/offset_scale.rs
```diff
@@ -0,0 +1,34 @@
...
+}
+
+impl SparseDenseSuffix for OffsetScaleSuffix<4> {
+    fn suffix_mle(b: LookupBits) -> u64 {
+        offset_scale_suffix(b, 4)
+    }
+}
+
+fn offset_scale_suffix(b: LookupBits, eighths: u64) -> u64 {
+    if b.len() < 6 {
```
Acentelles, line 31: This `b.len() < 6` gate pairs with the prefixes' `suffix_len != 0` gates: the joint assumption is `suffix_len == 0 || suffix_len >= 6` (the three offset y bits sit at interleaved index bits 0/2/4). A simulation of the combine identity confirms it breaks at suffix splits 2 and 4 (112 counterexamples at XLEN=8) — unreachable with log_m ∈ {8, 16}, but the same latent class the base branch pins for Pow2Offset with `debug_assert!(suffix_len == 0 || suffix_len >= 3)`. Mirror the assert into the four new prefixes (modular `evaluate` in offset_scale.rs and shift_data.rs; legacy `prefix_mle` + `update_prefix_checkpoint` in both), with `>= 6` here. ShiftDataSuffix correctly needs no guard (at an empty suffix it returns L_s·P_s = 0 while this one returns P_s = 1).

crates/jolt-lookup-tables/src/tables/shift_data_b.rs
```diff
@@ -0,0 +1,93 @@
...
+        result
+    }
+}
+
+impl<const XLEN: usize> PrefixSuffixDecomposition<XLEN> for ShiftDataBTable<XLEN> {
+    fn prefixes(&self) -> &'static [Prefixes] {
+        &[Prefixes::ShiftDataB, Prefixes::OffsetScaleB]
+    }
+
+    fn suffixes(&self) -> &'static [Suffixes] {
```
Acentelles, line 58: Same latent mismatch the base pins for the WindowMask tables: this table is XLEN-generic but its decomposition path (the ShiftData/OffsetScale prefixes and suffixes in this crate) hardcodes `crate::XLEN = 64`, so any other instantiation silently disagrees with the MLE. Mirror `debug_assert_eq!(XLEN, 64)` into `suffixes()`/`combine()` of the three modular ShiftData tables. The legacy side needs none — `ShiftDataPrefix<XLEN, EIGHTHS>` and friends are fully generic this time.

z3-verifier/src/virtual_sequences.rs
```diff
@@ -408,6 +414,49 @@ fn symbolic_exec(instr: &Instruction, cpu: &mut SymbolicCpu) {
...
+            let offset = ea.extract(2, 1).zero_ext(cpu.bv_bits - 2);
+            let shift = offset * cpu.bv_u64(16);
+            cpu.x[operands.rd as usize] = data.bvshl(shift);
+        }
+        Instruction::VirtualShiftDataW(VirtualShiftDataW { operands, .. }) => {
+            let data = cpu.x[operands.rs1 as usize].clone() & cpu.bv_u64(0xFFFF_FFFF);
+            let ea = cpu.x[operands.rs2 as usize].clone();
+            let offset = ea.extract(2, 2).zero_ext(cpu.bv_bits - 1);
+            let shift = offset * cpu.bv_u64(32);
+            cpu.x[operands.rd as usize] = data.bvshl(shift);
```
Acentelles, line 473: These three models are total (masked shifts, no contiguity caveat — unlike Pext they need no havoc guard), but the widths are hardcoded where the base's rewritten WindowMask arms now scale with `bv_bits`. For ShiftDataW this is more than style: at reduced `Z3_VERIFIER_BV_BITS`, `bvshl` by 32 on a 32-bit BV yields 0, so the model silently degenerates. Derive lane masks and shift units from `bv_bits / 8 * EIGHTHS` when resolving the (guaranteed) rebase conflict with the base's rewritten VirtualPext arm next door.

Acentelles (commented):
Deep review of the true delta (c41c9069c..d11c8e92a — note GitHub's diff view is currently polluted with #1762's content because the base branch was force-pushed with review fixes after this PR was cut). Four analysis passes plus test runs at this head: jolt-lookup-tables 356/356, legacy prefix_suffix/window_mask/pext/shift_data filter 85/85, jolt-program 38/38 incl. the golden-fixture parity test, tracer 127/127.

No soundness, completeness, or panic bugs. The RMW chain is fully pinned: ANDN clears exactly the WindowMask lane, the mask lane and ShiftData lane coincide for all 8 ea residues per width (so the ADD merge is carry-free unconditionally, not just under the asserts), and register/RAM read-write checking closes the ea-desync and lied-about-LD paths. The rank-2 decomposition with its dual-checkpoint invariant (ShiftData = L·P, OffsetScale = P, cross-read on the update snapshot) was verified by derivation and by exhaustive simulation of the legacy round machine at every round (log_m ∈ {8,16}, all three EIGHTHS, random challenges over GF(2^61−1)), including an XLEN=16 run covering the cross-phase lane-accumulation path (ShiftDataW's x bits 31..8 binding in non-final phases) that the XLEN=8 hypercube tests structurally cannot reach. All three MLEs match materialize_entry on the full 2^16 hypercube with multilinearity confirmed in all 128 coordinates. Wiring, flags, opcodes (0x008e-0x0090), ANDN profile legality, SCW's recursive re-baseline, and the 360-entry fixture guard all check out.

Rebase guidance (onto feat/byte-addressable-tier0-loads @ 12474919d): three files conflict at five sites — both stacks' prefixes/mod.rs dispatch matches (keep the base's consolidated `Pow2OffsetPrefix::<LOW_BIT>` arms, append the six new ShiftData/OffsetScale arms) and the z3 arms (keep the base's rewritten WindowMask/Pext neighbors, insert the three ShiftData arms). shared.rs auto-merges with expand_word_load; the re-baselined fixtures stay valid since the base changed no emitted sequences (rerun the parity test as insurance); the stale "46 table prefixes" kernel doc comments auto-fix. While rebasing, consider matching the base's `const VARIANT` style over `fn variant() + unreachable!()`, and a one-line comment that `OFFSET_MASK = 8 − EIGHTHS` clears the low bits only because EIGHTHS is a power of two. The AMO.W paths (expand_amo_post64's ORI/SRLI/SLL/SLL/XOR/AND/XOR chain) could adopt WindowMaskW/ShiftDataW in a later slice.

Acentelles:
Addressed the review findings:

- **Rebased onto feat/byte-addressable-tier0-loads @ 12474919d** (213784fcf) — this also fixes the polluted diff view. Of the mapped conflicts only the modular `dispatch_prefix!` needed a hand resolution (kept the consolidated `Pow2OffsetPrefix::<LOW_BIT>` arms, appended the six new arms); the legacy dispatch and z3 arms auto-merged correctly against the base's rewritten neighbors, and the golden-fixture parity test passes unchanged post-rebase.
- 168006ac6: `debug_assert!(suffix_len == 0 || suffix_len >= 6)` in all four ShiftData/OffsetScale prefixes (legacy: both `prefix_mle` and `update_prefix_checkpoint`), WARNING comments on both `b.len() < 6` suffix guards, plus the style alignment with the consolidated Pow2Offset: `const VARIANT` over `fn variant() + unreachable!()`, the power-of-two precondition note on `OFFSET_MASK = 8 − EIGHTHS`, and a raw-index/interleaved cross-reference.
- 0df62e7bf: `debug_assert_eq!(XLEN, 64)` in the three modular ShiftData tables' `suffixes()`/`combine()`; the legacy side is fully XLEN-generic and intentionally left unguarded.
- 090098d75: z3 ShiftData models scale with `bv_bits` (`bv_bits/8`, `bv_bits/4`, `word_bits`), fixing the word model's degeneration at reduced solver widths.

Validated at the new head: jolt-lookup-tables 356/356, legacy prefix_suffix/window_mask/pext/shift_data filter 85/85, jolt-program 38/38 incl. fixture parity, tracer 127/127, `cargo check` on z3-verifier tests, clippy `-D warnings` on jolt-lookup-tables/jolt-program/z3-verifier and jolt-prover-legacy (host), fmt applied.

Acentelles:
Restacked on the address-fusion head of #1762 (4319d987e -> new base): the store sequences now inherit the fused loads, and this branch resolves the expected append-adjacency conflicts (enum tails, dispatch matches, kind/lib registrations; ShiftData opcodes 0x008e-0x0090 slot in before AlignAddr's 0x0091, and the parity fixture was regenerated for the combined state — exactly the same 27 store hashes move, loads keep their fused hashes).

Validated at the new head 0548401bb: jolt-lookup-tables 362/362, legacy prefix_suffix/window_mask/pext/shift_data/align_addr filter 90/90, jolt-program 38/38 incl. fixture parity with the 360-entry guard, tracer 127/127, muldiv e2e in `host` and `host,zk` (first e2e of fused loads + fused stores together), z3 check, clippy `-D warnings` in both feature modes.

The nine sub-word expansions across the stack now stand at LB/LBU 4, LH/LHU/LW/LWU 5, SB 8, SH/SW 9: Σ54 vs the spec baseline's Σ93. Next: the test-only hardening PR on top of this branch (z3 forall-proofs for the nine sequences + the post-final-round checkpoint regression test).

Acentelles:
Post-address-fusion re-measurement (same machine, same block, same `main` @ a33d89a05 as the earlier numbers; post-fusion stack = 97de78c09 merged onto that main with the same VirtualLW/VirtualSW resolution):

| | main | + Tier 0 stack (pre-fusion, Σ60) | + Tier 0 stack (post-fusion, Σ54) |
|---|---|---|---|
| jeth block 25698189, total cycles | 1,289,309,699 | 1,203,812,048 | **1,182,552,470** |
| cycles/gas | 30.75 | 28.71 (−6.6%) | **28.20 (−8.3%)** |

The fusion contributes a further 21.3M cycles (−1.77% on top of the pre-fusion stack); the guest's post-state-root assertion passes, so the block still self-verifies. Trace-length proxies at pinned inputs, spec-branch baseline (8a013391f) vs post-fusion stack: btreemap 37,881 → 33,559 (**−11.4%**), sha2-chain 1,569,347 → 1,453,827 (**−7.4%**) — the fusion lands hardest on load-dense code, as expected. For the #1755 comparison: loads are now identical to its design and every store is one row cheaper, so its Σ57 would sit strictly between the two stack columns on any workload.

0xAndoroid (approved):
lgtm, thanks

PR #1770, opened 2026-08-12
feat(sumcheck): pluggable RoundScheduler for batch prove_round/finish_rounds

mtbadakhshan:
> **Note — two commits described below are not yet on this branch.** The `RoundScheduler` method defaults and the `ChaosTraversal` → `ReverseTraversal` rename are staged on [`a16z/jolt:pr1770-scheduler-cleanup`](https://github.com/a16z/jolt/tree/pr1770-scheduler-cleanup) (based on `49cd5a90a`, gates green). They could not be pushed here directly because maintainer edits are disabled on the fork — please cherry-pick or pull them in.

## Summary

Make batched sumcheck member visit order a pluggable policy instead of hard-wiring sequential host iteration in `prove_batch`. This lands the round-traversal seam from `specs/clean-slate-prover.md` so device backends can later issue all active members per round in one crossing without changing proof bytes (invariant 8).

## Changes

- Add `RoundScheduler` / `MemberRound` / `MemberFinish` and stock `SequentialRounds` in `jolt-sumcheck`; `prove_batch` builds the active-member work list and delegates `prove_round` / `finish_rounds` to the scheduler
- Both `RoundScheduler` methods default to declaration-order traversal, so an impl overrides only the half it changes — a backend that batches rounds into one crossing but has nothing to say about the terminal bind writes one method. `SequentialRounds` is the defaults, named
- Fold, running-claim updates, inactive padding, round-sum checks, and transcript ops stay in the engine — order/transport are free; skipping an active member is `MissingRoundMessage`, not silent padding
- Wire a `JoltBackend::round_scheduler` slot via `BuildRoundScheduler` (minted per stage from `ProofSession`); stage drivers pass the scheduler into `prove_batch`
- Reference/optimized backends keep `SequentialRounds` (behavior-preserving default)

## Delta vs `specs/clean-slate-prover.md`

The spec is left untouched by this PR; the deviations it records are listed here instead, so the landed spec stays the record of what was agreed rather than being rewritten by the PR that departs from it.

1. **Different mechanism than the spec's deferred sketch.** The spec parks the device problem under *"Known engine extension, deferred"* and sketches a **member group**: one `ProveRounds` serving N co-located members, returning the pre-folded round polynomial given the batch coefficients. This PR implements a `RoundScheduler` instead. The member group was rejected because it moves batching coefficients across the seam and still needs per-member claims back, while reshaping the Fiat-Shamir-critical member list. A backend that wants two-phase rounds owns both ends via the same `JoltBackend` value.
2. **Not deferred anymore.** The spec contracts per-member host iteration as *not* seam law but leaves the extension for later. This PR lands the seam now: round traversal is a policy, chosen like every other kernel slot.
3. **Invariant 8 gains a slot and a gate.** Invariant 8's "swapping any `JoltBackend` slot implementation" now covers the round-traversal slot. The byte-invariance gate correspondingly grows a third backend: `JoltBackend::reference()` vs a reversing traversal vs the optimized CPU backend.
4. **New engine rules the spec does not state.** A handle left unfilled is `SumcheckError::MissingRoundMessage` against that member (not silent `claim / 2` padding); the borrowed work slice cannot add or drop members, so the Fiat-Shamir-critical set stays engine law; uni-skip rounds stay outside this seam (single kernel, not a batch).
5. **Deferred here, and worth recording: no `Send` bound.** Intra-round *parallelism* needs `&mut [&mut dyn ProveRounds<F> + Send]` on `prove_batch`, which propagates to `PrepareKernel::prepare` and the generated driver — a future signature change, not taken here (it would constrain ~30 kernels for an unused traversal). So this PR buys batched-crossing transports, not threading. The bound belongs on the minted kernel, never on a `Box<dyn PrepareKernel<..>>` slot field (`#[derive(KernelSlots)]` matches single-bound and silently skips extras).

## Testing

- [x] Unit tests in `round_scheduler_tests.rs`: reordering traversal leaves proof bytes / transcript state unchanged; skipping an active member fails closed
- [x] `byte_diff` muldiv ratchet: `ReverseTraversal` (reverse-order) traversal still matches legacy wire bytes, with a counter so `SequentialRounds` cannot vacuously pass
- [x] Engine-twins / existing sumcheck tests updated for the new `prove_batch` signature

## Security Considerations

No change to verifier logic, Fiat–Shamir absorption order, or claim algebra. The engine still owns which members are active and how rounds are folded; a malicious/buggy scheduler can only reorder or omit work — omission is rejected (`MissingRoundMessage`), and reordering is required to be proof-byte-invariant (enforced by tests). Uni-skip remains outside this seam.

## Breaking Changes

Internal API only: `prove_batch` and `StageProver::prove` take an extra `&mut dyn RoundScheduler`. No proof-format or serialization change.


0xAndoroid:
I think you should not edit old specs

moodlezoup (approved):

PR #1777, opened 2026-08-14
fix: WASM verifier trusted-advice and build-wasm

crazywriter1:
## Summary
- Fix generated WASM verifiers to accept and forward trusted-advice commitments (live `jolt build-wasm` path + legacy macro path).
- Apply full memory config (including advice sizes) and skip mismatched clear/ZK CLI preprocessing.
- Stop `jolt new --wasm` from setting `[lib].path` before `wasm_verify.rs` exists.

## Test plan
- [x] `cargo nextest run -p jolt-sdk-macros --all-features`
- [x] `cargo nextest run -p jolt` (build_wasm generator tests)
- [x] Local WASM E2E: prove fib → `jolt build-wasm` → `node verify.mjs` → `verification result: true`


src/build_wasm.rs
```diff
@@ -26,29 +26,38 @@ use toml_edit::{value, Array, DocumentMut, Item, Table};
 struct FunctionAttributes {
     pub func_name: String,
     pub attributes: Attributes,
+    pub has_trusted_advice: bool,
```
0xAndoroid: Branching the exported verifier function signature based on the guest signature creates a second parser/type predicate plus duplicated Rust wrappers and JavaScript handlers. Give every verifier the same fourth byte-slice argument, with empty bytes meaning `None`; then `has_trusted_advice`, both generator branches, and their paired tests disappear.

src/build_wasm.rs
```diff
@@ -253,16 +407,14 @@ pub fn verify_{func_name}(preprocessing_data: &[u8], proof_data: &[u8], io_data:
...
 
-    let path = src_dir.join("wasm_verify.rs");
-    let mut file = File::create(path)?;
-    file.write_all(code.as_bytes())?;
-    Ok(())
+    code
 }
 
-pub fn modify_cargo_toml(name: &str) -> Result<()> {
+pub fn modify_cargo_toml(name: &str, set_lib_path: bool) -> Result<()> {
```
0xAndoroid: `set_lib_path = false` creates a hidden partial mode whose only effect is adding dependencies during scaffolding. `build_wasm()` adds those dependencies again before invoking wasm-pack. Remove the scaffold-time call and keep this function single-purpose, or generate the complete WASM library during scaffolding.

src/build_wasm.rs
```diff
@@ -331,17 +485,25 @@ pub fn modify_cargo_toml(name: &str) -> Result<()> {
 pub fn build_wasm() {
     println!("Building the project with wasm-pack...");
     let functions = extract_provable_functions();
-    let function_names: Vec<String> = functions.iter().map(|f| f.func_name.clone()).collect();
     let is_std = is_std().expect("Failed to check if std feature is enabled");
-    for function in &functions {
-        preprocess_and_save(&function.func_name, &function.attributes, is_std)
-            .expect("Failed to preprocess functions");
+    let project_zk = project_enables_zk();
+    let cli_zk = cfg!(feature = "zk");
```
0xAndoroid: The CLI is unified, so its compile-time `zk` feature is not a project mode and must not decide preprocessing behavior. A normally installed CLI must build both transparent and ZK projects. Run preprocessing through a helper compiled in the target project's Cargo graph; then delete `project_enables_zk`, `cli_zk`, and this mode branch.

src/build_wasm.rs
```diff
@@ -331,17 +485,25 @@ pub fn modify_cargo_toml(name: &str) -> Result<()> {
...
+    if project_zk != cli_zk {
```
0xAndoroid: This fallback still reports a successful build with missing or stale verifier preprocessing. The generated HTML always fetches that artifact. Continuing here can package no preprocessing or retain preprocessing for an older guest; fail the command unless the project-native generation step succeeds.

0xAndoroid (commented):
The trusted-advice forwarding path is fail-closed, but build-wasm currently violates the unified CLI contract and can package missing or stale preprocessing. The generator and scaffolding changes also add avoidable branching.

crazywriter1:
Thanks for the review. Addressed in the latest commit.

Every generated `verify_*` now uses the same fourth byte-slice argument (empty means `None`), so the guest-signature branch is gone. `jolt new --wasm` no longer mutates Cargo.toml; `modify_cargo_toml` is build-wasm-only and always sets `[lib].path`. CLI zk/clear preprocessing is gone: `build-wasm` compiles a project-native helper through `guest::compile_*` / `preprocess_*` and fails unless `preprocessed_*_verifier.bin` is produced and non-empty.

Local WASM E2E still passes with an empty fourth argument. @0xAndoroid

0xAndoroid (approved):
I removed a couple of useless tests, but otherwise lgtm

PR #1778, opened 2026-08-14
refactor(inlines): remove renamed instruction-alias facade, pass canonical kinds to emit_*

0xAndoroid:
## What

Removes the migration scaffolding left over from the inline recipe migration (8775247b6) and gives inline recipes an assembly-style surface:

- **`jolt-inlines-sdk`**: deletes `alias_instruction!` and the legacy-shaped `instruction::{add::ADD, andn::ANDN, ...}` re-export tree.
- **`jolt-program`**: deletes the `InlineInstruction` marker trait and its for-each impl macro. `InlineExpansionBuilder::emit_*` (and `bin`) now take canonical `SourceInstructionKind` values directly, deriving `JoltInstructionKind` via `jolt_kind()` with the same emit-vs-recursive-expand routing as before.
- **`jolt-riscv`**: adds a generated `kinds` module — value-level `SourceInstructionKind` constants, one per instruction, importable by name. The SDK re-exports it once (`host::kinds`), so inline crates keep a single SDK dependency.
- **`jolt_asm!`** (new, SDK): a `macro_rules!` assembly-style layer for straight-line recipe cores. Statements are `mnemonic dst, src..;` in RISC-V operand order, operands are ordinary Rust expressions, and each statement expands to exactly one builder call — zero runtime cost, byte-identical recipes. ~30 mnemonics covering the R/I/load/store/lui/advice/assert shapes in use.

  ```rust
  // blake2 G-function core
  jolt_asm!(self.asm, {
      add temp1, va, vb;      // v[a] = v[a] + v[b] + m[x]
      add va, temp1, mx;
      xorrot32 vd, vd, va;    // v[d] = rotr64(v[d] ^ v[a], 32)
      add vc, vc, vd;
      xorrot24 vb, vb, vc;    // v[b] = rotr64(v[b] ^ v[c], 24)
      ...
  });
  ```

## jolt_asm! scope

Converted where recipes are straight-line and read as assembly: SDK `MulAccExt` bodies (mac/m2ac/adc) and paired-u32/advice-store helpers; blake2 + blake3 G-functions and init/load tails; secp256k1 local mac helpers; the secp256k1/p256 top-limb check blocks (`add` + `assert_eq` + `assert_lte`).

Deliberately left as plain `emit_*` calls: sha2 (its round core runs live `Value` constant-folding that fixed mnemonics cannot express), and the bigint/keccak/p256/secp256k1 reduction loops, where emits interleave with Rust control flow one or two at a time — a macro block per statement would add noise, not remove it. Remaining plain call sites read `emit_r(MUL, t, a, b)` via imported kind constants.

## Why

- The marker-generic API had no type-safety benefit — every instruction implemented the same trait, so e.g. `emit_r::<Ld>` was accepted. Value arguments match the central pipeline's `ExpansionBuilder` style.
- Why a `kinds` module instead of the existing associated constants: `SourceInstructionKind::VirtualAdvice` is unreachable — the same-named enum variant constructor wins path resolution — and associated constants cannot be `use`-imported at all. Plain module constants have neither problem and are generated from the same `for_each_instruction_kind!` table, so they cannot drift.
- No overhead concern: kind dispatch is one enum-discriminant match per emitted row during recipe construction, and inline expansions are cached per inline (`with_cached_inline_sequence`), so recipes are built once, not per trace row. `jolt_asm!` compiles away entirely.

## Notes

- Branch is merged up to current `main` (merge commit, no rebase; includes the v12 audit changes to p256/secp256k1 advice stores and the new `InlineAdviceContext` API — the store-after-checks recipes are preserved exactly).

## Verification

- `cargo nextest run -p jolt-inlines-fixtures` — **registered inline expansion golden fixtures pass** (SHA-256 byte parity of expanded sequences ⇒ recipe bytes unchanged through the API change, the kind-constant conversion, and every `jolt_asm!` conversion)
- `cargo nextest run -p jolt-program` — expand pipeline tests pass (42)
- `cargo nextest run -p jolt-inlines-{sdk,sha2,keccak256,blake2,blake3,bigint,p256,secp256k1,grumpkin} -p tracer` — 166 tests pass, incl. reference-vs-harness execution parity for every inline
- `cargo clippy --all --features host --all-targets -- -D warnings` and `--features host,zk` — clean
- `cargo fmt`


markosg04 (approved):

0xAndoroid:
Adversarial review pass complete. Verdict: sound, behavior-preserving. Two untouched golden fixtures pass at head 7e3dbc3b4 (24-inline expanded-rows hash + 360-hash expander parity); dest-order call-site flips hand-verified. Nits:

- rotl64 jolt_asm! arm is dead code (zero users; keccak calls the builder method directly) and is the one arm violating the macro's "exactly one builder call" doc claim.
- Two coexisting Kind aliases (SourceInstructionKind in inlines, JoltInstructionKind in expand/*) — compiler-enforced but a future-confusion vector.
- __submit_inline_op! macro body misindentation (cosmetic).

— adversarial review by Claude (Fable), posted via Pika

PR #1783, opened 2026-08-19
test: z3 forall-proofs and checkpoint pinning for the byte-addressable stack

Acentelles:
Stacked on #1768; the test-only hardening slice closing the two verification gaps flagged across the stack's reviews, both ported from #1755's ideas per the integration plan on #1761.

## z3 forall-input proofs for the nine memory sequences

`SymbolicCpu` gains a doubleword-granular memory model (a Z3 array from addresses to `bv_bits`-wide values; the sequences only access the containing aligned doubleword, so keys never partially alias), LD/SD symbolic arms, and memory-inclusive equivalence checks in both `test_correctness` and `test_consistency`. The previously commented-out memory entries are enabled for LB/LBU/LH/LHU/LW/LWU/SB/SH/SW against width-scaled lane semantics (`bv_bits/8`-wide lanes, matching the scaled instruction models), so Z3 proves each expansion equivalent to the architectural semantics for **all** inputs, under the alignment-assert assumptions where the ISA traps. All 18 proofs complete in under a second at the default 64-bit model — no solver-heavy gating needed. Suite: 83 -> 101 passing.

## Post-final-round checkpoint regression test

Pins the exact prefix-checkpoint state the production prover consumes: binds all 128 address rounds through `Prefixes::update_checkpoints` with `read_raf_checking`'s `suffix_len` schedule, then checks every byte-addressable prefix family (Pow2Offset B/H/W, AlignAddr, OffsetScale B/H/W, ShiftData B/H/W, plus the generalized RightShift and WindowSign/WindowSignPow2) against closed forms at the challenge point, at both phase geometries (log_m 8 and 16). The per-table prefix_suffix tests do not pin this state on their own; #1755 shipped a production bug in exactly this gap.

## Negative controls (run locally, then reverted)

- LB spec flipped to zero-extension and SB spec widened to a halfword: both z3 proofs fail.
- Checkpoint loop skipping the final pair's update: both regression tests fail.
- A too-weak control (early `suffix_len = 0`) was verified to be genuinely unobservable by the updaters, not missed by the test.

## Validation

z3-verifier 101/101 (non-ignored), checkpoint tests 2/2, clippy `-D warnings` on z3-verifier and jolt-prover-legacy (host), fmt. Not in scope: SRL-is-PEXT table reuse (touches shipped SRL/SRA paths; separate PR outside this stack).


Acentelles (commented):
Two adversarial review passes (spec/math with independent re-derivation and a Python re-simulation of the checkpoint updater recurrences; harness/semantics). No wrong-spec or blocking findings: all nine z3 lane specs match the tracer semantics byte-for-byte, the memory model's single-key argument holds for every tested sequence, the harness changes are provably inert for the 83 pre-existing tests (the mem disequality is structurally false when neither side stores), the twelve checkpoint closed forms and the round-to-bit mappings are correct against the head implementations, and the alignment assumptions were verified satisfiable and address-covering (rs1 free surjects onto every aligned ea). Sensitivity analysis: mid-phase update corruption is caught for the accumulating families (any wrong pair contribution shifts the pinned value; false-pass probability ~2^-57 per seed); the resetting families' pre-final state is invisible to the test but also unread by production. Four actionable items inline; fixes follow on the branch. Noted for later, not this PR: word AMOs and LR/SC.W sequences now consist entirely of instructions with symbolic arms and are likely enableable with just spec closures.

z3-verifier/src/virtual_sequences.rs
```diff
@@ -611,13 +644,14 @@ fn test_correctness<I: RISCVInstruction + RISCVTrace>(
...
-    // We don't care if virtual registers differ
-    solver += cpu.x[..RISCV_REGISTER_COUNT as usize]
+    // We don't care if virtual registers differ; memory must match.
+    let registers_differ = cpu.x[..RISCV_REGISTER_COUNT as usize]
         .iter()
         .zip(cpu_expected.x[..RISCV_REGISTER_COUNT as usize].iter())
         .map(|(x1, x2)| x1.ne(x2))
         .reduce(|acc, t| acc | t)
         .unwrap();
+    solver += registers_differ | cpu.mem.ne(&cpu_expected.mem);
```
Acentelles, line 666: Asserts are consumed as assumptions with no satisfiability control: an UNSAT assumption set would make every correctness proof pass vacuously. The current sets are satisfiable (verified by reasoning: the alignment constraints touch rs1 only), but a cheap standing guard belongs here — check the solver is Sat after adding the assumptions and before adding the disequality, in both test_correctness and test_consistency.

z3-verifier/src/virtual_sequences.rs
```diff
@@ -409,6 +433,15 @@ fn symbolic_exec(instr: &Instruction, cpu: &mut SymbolicCpu) {
                 .bvshl(shift.zero_ext(cpu.bv_bits))
                 .extract(cpu.bv_bits - 1, 0)
         }
+        Instruction::LD(LD { operands, .. }) => {
```
Acentelles, line 441: Tracer LD truncates the immediate through `as i32` (ld.rs: `imm as i32 as u64`) while this arm adds the full i64. Divergent only for |imm| >= 2^31, which a 12-bit decode can't produce, and the sequences emit imm 0 — but matching the tracer exactly costs one cast. Also worth a WARNING on the arms (or the mem field doc at line 171): they key the array by the raw `rs1+imm`, and the never-partially-alias invariant is guaranteed by the expansions, not the arms — `expand_amo_d` does `LD rs1, 0` on a raw address, so a future AMO.D entry would need care.

z3-verifier/src/virtual_sequences.rs
```diff
@@ -749,6 +785,42 @@ fn test_consistency(instr: &Instruction) {
     }
 }
 
+/// The scaled sub-word load semantics: the `eighths`-byte lane of the
+/// containing doubleword at `rs1 + imm` (lanes are `bv_bits/8` wide so
+/// reduced solver widths stay faithful), sign- or zero-extended.
+fn lane_load(cpu: &SymbolicCpu, rs1: u8, imm: i64, eighths: u32, signed: bool) -> BV {
```
Acentelles, line 808: All nine proofs run with the single template imm = 1234, so the negative-immediate path through format_i_imm/normalize_imm in the expansions is unexercised, and only one alignment residue class is covered. One load and one store with a negative imm (a different residue) covers the shared mechanism. Separately, completeness is out of scope by construction: a sequence that added a spurious assert would still pass, since asserts are assumptions.

crates/jolt-prover-legacy/src/zkvm/lookup_table/prefixes/mod.rs
```diff
@@ -796,3 +796,114 @@ impl Prefixes {
...
+    /// path while those tests stay green (this exact gap produced a
+    /// production bug in the parallel #1755 implementation).
+    fn final_checkpoints_match_closed_forms(log_m: usize, seed: u64) {
+        let mut rng = StdRng::seed_from_u64(seed);
+        let r: Vec<Fr> = (0..LOG_K).map(|_| Fr::from_u64(rng.next_u64())).collect();
+
+        let mut checkpoints: Vec<PrefixCheckpoint<Fr>> = vec![None.into(); Prefixes::COUNT];
+        for j in (1..LOG_K).step_by(2) {
+            // Same schedule as read_raf_checking's update site.
+            let suffix_len = LOG_K - (j / log_m + 1) * log_m;
```
Acentelles, line 836: This duplicates the schedule formula from read_raf_checking.rs; if production's schedule ever changes, this test keeps validating the old one and stays green. A comment pinning the sync requirement (or a shared constant) prevents silent drift. Same for the 0x1761/0x1768 seeds — one line saying they're PR mnemonics.

Acentelles:
Review findings addressed in the follow-up commit:

- **Vacuity guard**: both `test_correctness` and `test_consistency` now assert the assumption set alone is Sat before adding the disequality, closing the in-principle vacuous-pass hole for all current and future asserted sequences.
- **Negative-immediate coverage**: `test_LB_negative_imm_correctness` (imm −8) and `test_SH_negative_imm_correctness` (imm −6, a different alignment residue) prove the shared immediate sign-extension path; both pass.
- **LD arm** now truncates the immediate through `i32`, matching tracer `ld.rs` exactly; the raw-address keying WARNING (with the `expand_amo_d` example) is on the memory-model doc.
- **Checkpoint test**: schedule-sync WARNING against silent drift from `read_raf_checking`, seeds documented as PR mnemonics.

Validation at the new head: z3-verifier 103/103, checkpoint tests 2/2, clippy clean, fmt applied. Left as noted follow-ups: the word-AMO/LR-SC z3 entries (now enableable with just spec closures) and the store-test failure message printing a nonexistent rd (failure-path cosmetics).

Acentelles:
CI root causes found and fixed (97de78c09 on the base branch; this PR restacked on it):

1. **Test modular crates (shard 1)**: the 1768 restack's conflict resolution had fused the AlignAddr and ShiftDataB rows of `prover_lookup_table_abi.rs` into a single 4-tuple — a compile error. It slipped through locally because that test is gated behind `required-features = ["prover-abi-tests"]`, which the local validation runs never enabled; it now runs locally with the feature (363/363, so the cross-stack table order is verified at runtime too).
2. **Test inlines**: the address fusion shortened every LW by one row, and BLAKE3's inline embeds four of them (705 -> 701 rows per case); the registered-inline fixture is re-blessed via `JOLT_UPDATE_INLINE_EXPANSION_FIXTURES` (only the four BLAKE3 cases changed; the CI dump showing SHA256 entries was just assert_eq printing the full vector).

Neither failure originated in this PR's commits (CI does not run z3-verifier); they surfaced here because this PR got the stack's only full CI run. Note for the stack owners: origin/feat/byte-addressable-tier0-loads has since gained tracer-x86 AOT templates and a SUPPORTED-pin bump for the four load kinds; the three ShiftData kinds (and the pending opcode renumbering) will need the same treatment when the stores branch is restacked onto it.

0xAndoroid (approved):

PR #1793, opened 2026-08-21
spec(sdk): define aligned borrowed byte inputs

quangvdao:
## Summary

This draft proposes an explicit byte codec for `#[jolt::provable]` parameters. The codec gives guest code a borrowed byte slice whose payload starts at an address divisible by eight. Guest startup can decode the frame without copying or scanning the payload.

Postcard remains the default for existing arguments. A guest must opt in with `#[jolt::bytes]`.

Proposed specification: [`specs/aligned-borrowed-inputs.md`](https://github.com/quangvdao/jolt/blob/codex/aligned-bulk-input-spec/specs/aligned-borrowed-inputs.md)

### Diff metadata

| Item | Value |
|---|---|
| Base | `6ff8921da235231c16a23dc7a7021ca28f10c3a8` on `a16z/jolt:main` |
| Head | `f6efb689dc05ab3ef3ec6fb54d4acffcd67762a0` |
| Commits | 1 |
| Files changed | 1 |
| Diff | 492 insertions and 0 deletions |
| Status | Proposed specification only |

## Motivation

Jolt currently uses Postcard for each function argument. Postcard is a good default for typed values, including byte slices. Its variable size slice length prefix means that an application cannot rely on the payload address being aligned for RV64 word loads.

Applications can work around this by understanding Postcard framing and adding their own padding. That puts Jolt transport details in guest code. It also makes each application responsible for keeping native execution, tracing, proving, generated verification, advice commitment, and Wasm input construction consistent.

The proposal moves byte framing and alignment into Jolt. Applications continue to own their semantic byte format and all field validity checks.

## Changes

This PR specifies the following design:

- `#[jolt::bytes] input: &[u8]` selects the new codec.
- The same codec works with public input, `TrustedAdvice<&[u8]>`, `UntrustedAdvice<&[u8]>`, and `PrivateInput<&[u8]>`.
- Each frame contains zero alignment bytes, a fixed little endian `u64` length, the payload, and a fixed little endian `!length` footer.
- The payload address is divisible by eight.
- The guest decoder checks framing and bounds before it returns a borrowed slice.
- One SDK framing implementation serves every generated host path and every input channel.
- One macro argument descriptor records input provenance and codec choice.
- Native execution uses aligned temporary storage so it follows the same function contract as RV64 execution.

### Ownership

| Concern | Current behavior | Proposed behavior |
|---|---|---|
| Ordinary typed values | Postcard | Postcard, unchanged |
| Large byte payload framing | Application workarounds around Postcard | Jolt owned opt in frame |
| Input provenance | Jolt advice wrappers | Unchanged |
| Payload alignment | Not guaranteed | Eight byte alignment for the new codec |
| Field decoding | Application | Application, unchanged |
| Generated input paths | Repeated Postcard calls in macro output | One argument model and one framing owner |

## Frame and failure behavior

The proposed frame is:

```text
zero padding | u64 length | payload | u64 !length
```

The decoder checks zero padding, integer conversion, checked bounds, the complete payload, and the footer before it creates the borrowed slice. It performs constant work with respect to payload length.

The footer is needed because the guest sees the configured channel region. `JoltDevice` returns zero for configured addresses after the supplied host vector. A length prefix alone cannot detect every truncated payload when the missing suffix contains zero bytes. The nonzero complement footer detects that truncation without changing the memory map for existing programs.

Canonicality applies to each declared frame at its cursor. The proposal does not add an end of channel marker. Generated APIs emit no unused suffix. Any suffix supplied through low level `JoltDevice` construction remains covered by the existing public input or advice binding.

## Security Considerations

This PR changes no proof code, verifier code, transcript code, or memory checking code. It only proposes future work.

The specification requires the implementation to treat every frame as untrusted for memory safety. It requires checked arithmetic and complete bounds checks before slice construction. It also requires public input and advice commitments to cover the complete encoded frame.

Jolt will return bytes, not typed field elements or Rust object layouts. Applications must continue to check canonical encodings, field moduli, lengths, and domain rules.

## Breaking Changes

None in this PR because it contains only a proposed specification.

The future implementation will preserve byte identical Postcard behavior for every parameter that does not use `#[jolt::bytes]`. Adding the attribute will intentionally change that guest's program identity and input representation.

## Testing

- [x] `git diff --check`
- [x] Confirmed every local Markdown reference in the specification exists at the pinned base
- [x] Confirmed the draft contains every section required by `specs/TEMPLATE.md`
- [ ] Rust tests and Clippy were not run because this PR changes only `specs/**`, which the Rust workflow excludes

The specification defines independent frame vectors, malformed frame tests, macro compile tests, standard and zero knowledge proof tests, a `jolt-eval` invariant, compatibility fixtures, and separate RV64 main and virtual cycle measurements for the implementation PR.

## Specification status

The specification is proposed. This draft is for design review before implementation.

The main review questions are:

1. Should Jolt expose an explicit parameter codec such as `#[jolt::bytes]`?
2. Is eight byte alignment the right first guarantee for RV64 guests?
3. Is the fixed length and complement footer frame acceptable, or should Jolt expose actual channel lengths to the guest instead?
4. Should the native wrapper copy into aligned temporary storage so native and RV64 calls share one contract?

## Reviewer map

Suggested reading order:

1. Summary, goal, and invariants.
2. Canonical byte frame and failure behavior.
3. Macro argument model and SDK framing owner.
4. Proof binding, application boundary, and compatibility.
5. Acceptance criteria, tests, and performance requirements.

### DIFF
--- a/specs/aligned-borrowed-inputs.md
+++ b/specs/aligned-borrowed-inputs.md
@@ -1,0 +1,492 @@
+# Spec: Aligned Borrowed Byte Inputs
+
+| Field       | Value                          |
+|-------------|--------------------------------|
+| Author(s)   | @quangvdao                     |
+| Created     | 2026-08-21                     |
+| Status      | proposed                       |
+| PR          |                                |
+
+## Summary
+
+Jolt currently uses Postcard for every function argument. This is a good default for small typed values, but it is not a complete input model for large encoded objects. A borrowed byte slice receives a variable size Postcard length prefix, so the address of its payload depends on its length. Applications that need aligned word loads must understand this outer frame and repair alignment inside their own wire format. This spec adds an explicit byte codec for `#[jolt::provable]` parameters. Jolt will encode a fixed width length and footer, place the payload at an eight byte aligned address, and give the guest a borrowed slice without copying or allocating. The same codec will work for public input, trusted advice, untrusted advice, and private input. Postcard will remain the default.
+
+## Intent
+
+### Goal
+
+Add one Jolt owned input codec that gives guest code a bounded, borrowed byte slice with an eight byte aligned payload while keeping input provenance separate from input representation.
+
+### Terms
+
+An **input channel** is one of the existing Jolt memory regions for public input, trusted advice, or untrusted advice. `PrivateInput<T>` uses the untrusted advice channel and adds the existing zero knowledge requirement.
+
+A **codec** defines how one function argument becomes bytes in its channel and how the generated guest entry point reconstructs the argument. Postcard is the existing default codec. The byte codec in this spec is an explicit second codec.
+
+A **byte frame** is the canonical length, padding, and payload representation defined below. It is part of the bytes that Jolt commits or exposes. It is not application data.
+
+### Invariants
+
+1. **Postcard remains the default.** A parameter without the new attribute must produce the same channel bytes and generated guest behavior as it does before this change.
+
+2. **The byte codec is explicit.** The SDK must not automatically change the representation of every `&[u8]` parameter. Existing programs must opt in.
+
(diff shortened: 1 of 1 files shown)


markosg04:
stale

PR #1794, opened 2026-08-21
perf(field): port subword Fp64 kernels

quangvdao:
## Summary

This ports the field arithmetic kernel work from [Akita #427](https://github.com/LayerZero-Labs/akita/pull/427) into the canonical `jolt-field` implementation landed by #1792.

It fixes carry handling for valid generic subword `Fp64<P>` moduli, adds the fused packed `FpExt2<TwoNr>` path, preserves optimized NEON, AVX2, and AVX-512 execution, and adds independent regression coverage. This is the Jolt-side prerequisite for refreshing and merging [Akita #429](https://github.com/LayerZero-Labs/akita/pull/429).

### Scalar Fp64 arithmetic

* Keep the first Solinas fold as an exact two-word value when `C * (product >> BITS)` does not fit in one `u64`.
* Use the carry-preserving reducer for generic 63-bit moduli such as `2^63 - 259` and `2^63 - 25`.
* Simplify subword addition and subtraction using the fact that `2P < 2^64` for every subword modulus.
* Retain a dedicated AArch64 63-bit multiply and reduction sequence.

### Extension and packed arithmetic

* Replace the boolean `Ext2Config::IS_NEG_ONE` marker with `Ext2NonResidueKind::{Generic, NegOne, Two}` so specialized formulas are selected explicitly.
* Use a correct reduced-product fallback for custom non-residues in delayed `FpExt2<Fp64>` accumulation.
* Fuse the four raw base products and two final reductions for packed `FpExt2<TwoNr>` when the exact accumulation bound holds.
* Share the carry-preserving reduction through the existing `SimdWord` engine across NEON, AVX2, and AVX-512.
* Keep Jolt's full-low-word SIMD correction multiply. This removes an Akita-specific AVX fallback that was only required by its older low-32-bit implementation.

### Evidence and portability

* Add independent integer-oracle tests for two generic 63-bit moduli.
* Add fused extension regressions for wide subword products, first-fold carry, and large offsets.
* Add a custom non-residue delayed-accumulation regression.
* Add a reproducible scalar, packed, and extension benchmark for `2^63 - 259`.
* Add dedicated CI execution on AVX2, AArch64 NEON, and emulated AVX-512.

No public Fp31 or Fp63 preset is added in this PR.

## Validation

* `cargo fmt --all --check`
* `taplo fmt --check crates/jolt-field/Cargo.toml`
* `cargo clippy -p jolt-field --all-targets --no-default-features --features solinas -- -D warnings`
* `cargo nextest run -p jolt-field --no-default-features --features solinas --release`: 119 passed
* Native AArch64 NEON packed differential suite: 9 passed
* Native AVX2 packed differential suite: 9 passed
* Native AVX-512 packed differential suite: 9 passed
* AVX2 and AVX-512 cross-compilation with `-D warnings`

A short Apple M4 smoke run measured approximately 1.11 ns for scalar Fp64 multiply, 2.64 ns for a two-lane packed multiply, 7.21 ns for scalar Ext2 multiply, and 8.27 ns for a two-lane packed Ext2 multiply. These numbers confirm the benchmark and optimized paths execute. They are not a stable baseline comparison.

Repository-wide `taplo fmt --check` still reports three pre-existing vendored Lean package TOML files under `zklean-extractor/package-template/.lake/packages`. This PR does not touch those files.

### DIFF
--- a/.github/workflows/field-portability.yml
+++ b/.github/workflows/field-portability.yml
@@ -34,12 +34,14 @@ jobs:
     steps:
       - uses: actions/checkout@v7
       - uses: actions-rust-lang/setup-rust-toolchain@v1
+      - name: Install nextest
+        uses: taiki-e/install-action@nextest
       - name: Require AVX2
         run: grep -Eq '(^|[[:space:]])avx2([[:space:]]|$)' /proc/cpuinfo
       - name: Run packed field tests
         env:
           RUSTFLAGS: -D warnings -C target-feature=+avx2
-        run: cargo test --release -p jolt-field --no-default-features --features solinas --test solinas_packed_differential
+        run: cargo nextest run --release -p jolt-field --no-default-features --features solinas --test solinas_packed_differential
 
   aarch64-neon:
     name: AArch64 NEON field kernels
@@ -47,6 +49,8 @@ jobs:
     steps:
       - uses: actions/checkout@v7
       - uses: actions-rust-lang/setup-rust-toolchain@v1
+      - name: Install nextest
+        uses: taiki-e/install-action@nextest
       - name: Require native AArch64 NEON
         run: |
           test "$(uname -m)" = "aarch64"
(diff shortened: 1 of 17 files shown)


Acentelles:
The arithmetic changes look sound to me. I checked the scalar carry-preserving reduction, packed SIMD path, fused `FpExt2<TwoNr>` bounds, generic non-residue fallback, and AArch64 sequence. 

One request: repository policy requires cargo nextest while the new portability workflow uses cargo test in all three jobs (field-portability.yml:42,57,76).

Acentelles (approved):

PR #1796, opened 2026-08-22
refactor(akita)!: complete the shared field cutover

quangvdao:
## Summary

This PR completes Jolt's breaking cutover to one shared Solinas field implementation and the current Akita polynomial-commitment interface.

It now:

- makes the workspace `jolt-field` package the single field identity used by Jolt and Akita;
- removes the temporary reverse `jolt-field -> akita-field` adapter and its duplicate trait surface;
- ports dense, one-hot, advice, and committed-program openings to Akita's current grouped-commitment, resolved-schedule, setup, and coefficient-packing APIs;
- keeps the trace one-hot source allocation-free and non-materialized;
- validates prover-controlled commitment sizes, proof shapes, decode budgets, and transcript-grinding nonce widths from the same resolved schedule used by Akita verification; and
- separates the memory-heavy legacy, modular, and verifier Akita CI lanes and serializes only the Akita byte-diff binary through a scoped nextest test group.

The branch is pinned to the immutable upstream Akita `main` commit [`4505404b5bd9548970f0753b0118616a373c867e`](https://github.com/LayerZero-Labs/akita/commit/4505404b5bd9548970f0753b0118616a373c867e). The temporary fork/hotfix pin is gone.

Normative Jolt design: [`specs/jolt-field-rebuild.md`](specs/jolt-field-rebuild.md).

### Diff metadata

- Base: `756bddce3a3413b6ef6dcf9d9120d0b2054ab1d8` (`main`)
- Head: `7745e09904eeb02d8a157326847d4870fd07794a`
- Head branch: `quang/pr-1745-refresh`
- Commits: 79
- Files changed: 41
- GitHub diff: 2,298 insertions, 4,086 deletions

The history is intentionally longer than the final 41-file change surface. It preserves the first-principles field rebuild, trait consolidation, workspace replacement, audit fixes, performance ports, Akita cutover, CI hardening, and subsequent restacks on Jolt `main`.

## Dependency chain

The original three-PR dependency path is closed:

1. [Jolt #1810](https://github.com/a16z/jolt/pull/1810) merged allocation-free canonical coefficient access into `jolt-field`.
2. [Akita #447](https://github.com/LayerZero-Labs/akita/pull/447) moved Akita to Jolt's shared field identity.
3. [Akita #458](https://github.com/LayerZero-Labs/akita/pull/458) landed the grouped proof-shape fix that had previously existed as the temporary #450 fork hotfix.
4. [Akita #454](https://github.com/LayerZero-Labs/akita/pull/454) renamed expanded-setup `seed` state to its actual `descriptor` semantics.
5. [Akita #460](https://github.com/LayerZero-Labs/akita/pull/460) made transcript grinding protocol-bound end to end.
6. Current Jolt `main` also contains [Jolt #1817](https://github.com/a16z/jolt/pull/1817), so committed programs use the direct bytecode/program-image batch rather than the superseded reconstruction opening path.

This head is based on current Jolt `main` and pins current official Akita `main`; there is no remaining fork dependency.

## Motivation

Before this cut:

- the rebuilt Solinas field depended backward on Akita's former field crate through a temporary feature, giving the same scalar two package and API identities;
- the Jolt adapter targeted older Akita schedule, setup, polynomial-source, and proof-shape types;
- verifier-side allocation checks mirrored security-sensitive Akita formulas locally;
- grouped advice schedules and direct committed-program objects were not represented through the final shared batching boundary; and
- the modular Akita proof fixtures could exhaust hosted-runner memory when nextest launched several proof processes concurrently.

The resulting staged integration compiled, but it was not the final ownership, verification, or CI boundary.

## Architecture

```mermaid
flowchart LR
    F["workspace jolt-field<br/>Fp128 + accumulators"] --> A["Akita crates<br/>one package identity"]
    C["generated dense/K16/K256 catalogs"] --> R["preprocessing schedule registry"]
    G["advice and direct-program groups"] --> R
    T["TracePackedOneHot<br/>selected positions only"] --> K["Akita coefficient-packing kernel"]
    R --> P["grouped commitment and opening"]
    K --> P
    R --> V["Jolt pre-deserialization shape guard"]
    V --> S["Akita canonical proof shape<br/>+ grinding plan + decode budget"]
    S --> D["backend proof deserialization"]
```

## Change-surface overview

| Area | Before | After |
|---|---|---|
| Field ownership | Temporary `jolt-field -> akita-field` bootstrap edge | One workspace `jolt-field` identity consumed directly by Akita |
| Fp128 accumulation | Akita-specific adapter types | Const-generic `Fp128Accumulator` and `Fp128SignedAccumulator` owned by `jolt-field` |
| Akita schedules | Older fixed-profile fields and constructors | Current `GroupCommitPhaseParams`, resolved rows, adaptive dimensions, and protocol-epoch-3 catalogs |
| Grouped objects | Superseded advice/profile and committed-program reconstruction paths | Independently committed advice and direct-program objects joined with the main trace opening |
| Trace one-hot | Integration-specific tensor/sparse witness hooks | Canonical position-source coefficient packing without a trace-sized dense polynomial |
| Verifier sizing | Jolt-local proof-shape formula mirror | Akita `canonical_proof_shape`, canonical grinding plan, frozen group profiles, and decode budget |
| Transcript binding | Existing backend proof bytes only | Serialized backend proof contains and binds Akita's ordered transcript nonce stream; Jolt checks the scheduled nonce width before decoding |
| CI | Legacy and modular Akita tests shared a resource-heavy lane | Separate legacy, modular, verifier, and catalog-drift checks; only the Akita byte-diff binary is serialized by nextest |

## Single field identity and arithmetic ownership

The root manifest patches Akita's Git `jolt-field` dependency to the workspace member and pins all Akita crates to one immutable official upstream revision. `Cargo.lock` contains one `jolt-field` identity and no `akita-field` package.

Inside `crates/jolt-field`:

- the temporary `akita` feature, optional Akita dependencies, and `src/akita.rs` reverse adapter are deleted;
- `JoltField` regains its normal serde contract;
- the field-specific accumulator module is renamed and generalized as `Fp128Accumulator<P>` and `Fp128SignedAccumulator<P>`;
- `Fp128<P>` owns the wide and signed deferred-reduction implementations; and
- independent arithmetic tests cover general, small-scalar, signed-product, merge, zero, and reduction behavior.

The legacy prover retains `AkitaFp128` only as the Arkworks/orphan-rule wrapper around `jolt_field::Prime128OffsetA7F7`. Conversion, inversion, sampling, accumulation, serialization, and transcript challenges delegate to the canonical shared field.

## Current Akita integration

`crates/jolt-akita` is ported to Akita's current public boundary:

- dense and one-hot constructors use adaptive schedule dimensions instead of a caller-fixed source ring dimension;
- supported row-major K16/K256 one-hot inputs use the native one-hot backend, while unsupported sparse-unit shapes fail explicitly before dense conversion;
- one-hot commitment calls name the source type and CPU backend explicitly;
- schedule keys use frozen `GroupCommitPhaseParams` for the final and precommitted groups;
- recursive setup-prefix contributions remain rejected because Jolt's verifier boundary does not admit them; and
- K16/K256 configurations accept the current adaptive commitment geometry, including 512-dimensional matrices.

The dense, K16, and K256 generated tables are regenerated from Akita's current planner. Their catalog identity advances to protocol epoch 3, and the affected rows/digests reflect the transcript-grinding-aware objective.

Grouped advice rows remain preprocessing-owned rather than verifier-planned. Preprocessing installs the distinct ordered presence combinations for untrusted advice, trusted advice, or both, binds the resolved row to the public `OpeningScheduleSelection`, and passes only frozen profiles to verification.

Current Jolt `main`'s direct committed-program representation is preserved. Direct bytecode and program-image commitments are precommitted groups in the same Akita opening batch; the old reconstruction sumcheck and auxiliary program-opening path are not reintroduced.

## Allocation-free trace one-hot opening

`TracePackedOneHot` remains a row-major source of selected positions. It does not construct a trace-sized dense field polynomial.

The source implements Akita's canonical subring coefficient-packing contract:

- `visit_segment_ring_range` is the single row traversal used by the packed coefficient kernel;
- batch kernels consume `SubringCoefficientPackingPlan`;
- dense, ordinary one-hot, and trace one-hot groups dispatch through representation-homogeneous grouped kernels;
- K16/K256 paths retain fused commitment/opening kernels and bounded deferred-reduction accumulators; and
- source arity, dimensions, live positions, segments, and coefficient bounds are checked before kernel execution.

The superseded tensor projection, packed witness, root projection, sparse linear-combination, and extension-reduction hooks are deleted rather than maintained in parallel.

## Verifier, proof shape, and transcript grinding

Before payload-sized backend deserialization, Jolt now:

1. decodes the fixed-size public schedule selection;
2. resolves the dense, K16, K256, grouped-advice, or direct-program row;
3. derives exact commitment coefficient lengths from the row's frozen `GroupCommitPhaseParams` and `CompressionChainPlan`;
4. checks coefficient and byte lengths with checked arithmetic;
5. caps and decodes the small proof-shape descriptor;
6. derives Akita's canonical transcript `GrindingPlan` from the resolved schedule and opening layout;
7. compares the supplied shape with `canonical_proof_shape(schedule, layout, EXT_DEGREE, grinding_plan)`, including `nonce_stream_bits`;
8. validates the proof's byte decode budget; and only then
9. deserializes the backend proof payload.

Akita #460 places the ordered `TranscriptNonceStream` inside the serialized backend proof. Jolt's outer transcript already absorbs those serialized backend proof bytes, so no duplicate Jolt nonce field is introduced. The shape guard instead pins the exact scheduled nonce width before payload allocation, and Akita replays the nonce stream during backend verification.

Malformed sizes, truncated commitments, oversized shape descriptors, forged opening counts, forged nonce widths, and over-budget payloads fail before backend payload allocation. The former Jolt mirror of Akita's recursive, terminal, extension-reduction, and bounded-sumcheck shape formulas is deleted; Akita's canonical implementation is the single source of truth.

## Security, verifier safety, and compatibility

- The integrated Cargo graph contains one field package identity.
- Commitment and proof allocation bounds come from the same frozen profiles and canonical shape used by verification.
- Transcript grinding is schedule-derived, shape-bound, serialized in deterministic level order, and replayed by Akita verification.
- Prover-controlled byte sizing uses checked arithmetic before allocation.
- The verifier resolves preprocessing-installed rows and does not run the schedule planner.
- Direct committed-program commitments remain position- and order-bound; tamper tests cover commitment leaves, reordered groups, trace order, and direct bytecode claims.
- The main trace one-hot source remains non-materialized.
- Legacy and modular packed paths use the same Solinas scalar-challenge convention.

This is an intentional breaking package/API/proof cutover. It does not preserve the old `akita-field` package identity, pre-cutover Akita setup/proof wire formats, schedule catalog identity, or transcript bytes. The generic commitment API also rejects unsupported sparse-unit one-hot shapes instead of materializing them densely. No compatibility wrapper or dual representation is provided.

Preserved capabilities include dense commitments, K16/K256 one-hot openings, trusted and untrusted advice independently or together, direct committed programs, legacy/modular proof parity, and fail-closed tamper handling.

## Complete 41-file reviewer map

| Domain | Files | Reviewer-visible change |
|---|---|---|
| Workspace dependency graph | [`Cargo.toml`](Cargo.toml), [`Cargo.lock`](Cargo.lock) | Patch Akita onto the workspace field, pin official Akita `4505404b5`, remove `akita-field`, and resolve the current grinding-enabled dependency graph. |
| CI | [`.config/nextest.toml`](.config/nextest.toml), [`.github/workflows/rust.yml`](.github/workflows/rust.yml) | Rename the post-cutover identity gate; split legacy/modular/verifier Akita checks; serialize only the Akita byte-diff binary; add catalog drift coverage. |
| Shared field core | [`crates/jolt-field/Cargo.toml`](crates/jolt-field/Cargo.toml), [`src/akita.rs`](crates/jolt-field/src/akita.rs), [`src/algebra.rs`](crates/jolt-field/src/algebra.rs), [`src/fp128_accumulators.rs`](crates/jolt-field/src/fp128_accumulators.rs), [`src/lib.rs`](crates/jolt-field/src/lib.rs), [`src/solinas/fp128.rs`](crates/jolt-field/src/solinas/fp128.rs) | Delete the bootstrap feature/adapter, restore serde bounds, generalize Fp128 accumulators, bind them to the Solinas field, and retain wide-reduction tests. |
| Akita package and benchmark wiring | [`crates/jolt-akita/Cargo.toml`](crates/jolt-akita/Cargo.toml), [`benches/akita_paths.rs`](crates/jolt-akita/benches/akita_paths.rs) | Adopt the current Akita crates, field traits, constructors, and explicit zero semantics. |
| Akita runtime integration | [`lib.rs`](crates/jolt-akita/src/lib.rs), [`adapters.rs`](crates/jolt-akita/src/adapters.rs), [`configs.rs`](crates/jolt-akita/src/configs.rs), [`native_batching.rs`](crates/jolt-akita/src/native_batching.rs), [`planning.rs`](crates/jolt-akita/src/planning.rs), [`schedule_registry.rs`](crates/jolt-akita/src/schedule_registry.rs), [`scheme.rs`](crates/jolt-akita/src/scheme.rs), [`shape_guard.rs`](crates/jolt-akita/src/shape_guard.rs), [`trace_onehot.rs`](crates/jolt-akita/src/trace_onehot.rs) | Centralize schedule planning, provision grouped rows, dispatch typed backends, reject unsupported sparse shapes before materialization, join direct-program groups, derive grinding-aware canonical shapes, enforce decode budgets, and pack trace coefficients in one row traversal. |
| Generated schedules | [`jolt_fp128_dense_bounded.rs`](crates/jolt-akita/src/schedules/jolt_fp128_dense_bounded.rs), [`jolt_fp128_onehot_k16.rs`](crates/jolt-akita/src/schedules/jolt_fp128_onehot_k16.rs), [`jolt_fp128_onehot_k256.rs`](crates/jolt-akita/src/schedules/jolt_fp128_onehot_k256.rs), [`schedules/mod.rs`](crates/jolt-akita/src/schedules/mod.rs) | Regenerate dense/K16/K256 planner rows and protocol-epoch-3 identities for the current schema and objective. |
| Akita integration tests | [`tests/pathologies.rs`](crates/jolt-akita/tests/pathologies.rs), [`tests/prefix_packing.rs`](crates/jolt-akita/tests/prefix_packing.rs), [`tests/schedules.rs`](crates/jolt-akita/tests/schedules.rs), [`tests/support/mod.rs`](crates/jolt-akita/tests/support/mod.rs) | Rebind shared-field traits and cover catalog resolution, preprocessing-only grouped rows, two-advice profiles, setup capacity, unsupported one-hot rejection, shape rejection, and serialized proof tampering. |
| Legacy packed prover | [`crates/jolt-prover-legacy/Cargo.toml`](crates/jolt-prover-legacy/Cargo.toml), [`src/field/akita.rs`](crates/jolt-prover-legacy/src/field/akita.rs), [`src/zkvm/packed.rs`](crates/jolt-prover-legacy/src/zkvm/packed.rs) | Select `jolt-field/solinas`, wrap canonical Fp128, port conversions/challenges/accumulation, and retain advice and direct-program end-to-end proofs. |
| Modular prover and verifier fixtures | [`crates/jolt-prover/Cargo.toml`](crates/jolt-prover/Cargo.toml), [`tests/akita_e2e.rs`](crates/jolt-prover/tests/akita_e2e.rs), [`crates/jolt-verifier/tests/fs_attacks.rs`](crates/jolt-verifier/tests/fs_attacks.rs), [`tests/soundness/tampering/akita.rs`](crates/jolt-verifier/tests/soundness/tampering/akita.rs) | Select the shared Solinas backend and rebind e2e, parity, Fiat-Shamir, direct-program, and tamper fixtures to the shared ring traits. |
| Repository safeguards | [`scripts/check-shared-field-identity.sh`](scripts/check-shared-field-identity.sh), [`scripts/check_style_invariants.py`](scripts/check_style_invariants.py), [`scripts/tests/test_check_style_invariants.py`](scripts/tests/test_check_style_invariants.py) | Reject stale `akita-field` identities and scope the nominal-import exception to provenance-marked generated Rust, with regression coverage. |
| Design and constraints | [`specs/jolt-field-rebuild.md`](specs/jolt-field-rebuild.md), [`specs/byte-addressable-memory.md`](specs/byte-addressable-memory.md) | Record the field rebuild and final cutover, and keep the committed-program memory bound honest after deleting sparse-unit materialization. |

## Commit-history map

- `e25af1030` through `0516c2b87` — first-principles Solinas implementation, trait consolidation, oracle/golden tests, workspace replacement, CI repair, and deletion of the Akita bootstrap edge;
- `f5b76392e` through `eb93a0310` — first post-cutover Akita API port, shared-field identity gate, legacy feature/challenge repair, and exact-uniform sampling;
- `47914923c` through `a8880e362` — current Akita field capabilities, wide/subword kernels, portability lanes, allocation-free coefficient access, and refreshed shared-field pin;
- `06f9b8d6f` and `7cddb7358` — complete grouped schedule/setup/proof-shape integration and the temporary proof-shape hotfix pin;
- `5aa4cff41` and `5552507d4` — split and serialize the Akita CI lanes after hosted-runner memory failures;
- `f5e91c05e` — merge current Jolt `main`, preserving #1817's direct committed-program batching and resolving the three Akita conflicts in favor of current shared APIs; and
- `e76e8384c` — replace the temporary fork with official Akita `main`, derive the canonical transcript-grinding plan, bind nonce-stream width in the shape guard, and regenerate protocol-epoch-3 catalogs; and
- `7745e0990` — address the final review round: remove fake sparse materialization, centralize planner policy, type the shape guard, restore one-pass trace packing and fused field multiplication, scope nextest serialization, delete dead probes, and add regression coverage.

## Validation completed at `7745e0990`

The following passed locally on the exact pushed tree:

- `cargo fmt --all -- --check`, `git diff --check`, and `python3 scripts/check_style_invariants.py --base upstream/main`;
- `cargo clippy -p jolt-akita --all-targets -- -D warnings`;
- `cargo clippy -p jolt-prover-legacy --all-targets --features akita -- -D warnings`;
- `cargo clippy -p jolt-prover --all-targets --features akita,prover-fixtures -- -D warnings`;
- `cargo nextest run -p jolt-akita --cargo-quiet --no-fail-fast`: 56 passed, 1 skipped;
- focused shared-field accumulator and legacy Akita arithmetic tests: 5 passed;
- `cargo nextest run --cargo-profile ci -p jolt-prover --features akita,prover-fixtures --cargo-quiet`: 18 passed, 2 skipped, including four byte-diff ratchets, advice, forced K256, and committed-program proofs; and
- `cargo nextest run --cargo-profile ci -p jolt-akita --run-ignored all -E 'test(catalogs_match_planner_regeneration)' --cargo-quiet`: 1 passed, 56 skipped.

The fresh GitHub Actions matrix is running on this head. At publication time, no GitHub check is failing; required checks and review remain pending.

## Merge status

The code-level dependency gates are closed: Jolt #1810 and #1817 are in the base, and the branch pins official Akita `main` containing #447, #458, #454, and #460.

Remaining gates are:

1. the fresh GitHub Actions matrix on `7745e0990`; and
2. required maintainer review.

## Reviewer map

Suggested reading order:

1. [`specs/jolt-field-rebuild.md`](specs/jolt-field-rebuild.md) and the root dependency changes;
2. `crates/jolt-field` ownership and accumulator changes;
3. `crates/jolt-akita/src/schedule_registry.rs` and the regenerated catalogs;
4. `crates/jolt-akita/src/trace_onehot.rs`, `scheme.rs`, and `native_batching.rs`;
5. `crates/jolt-akita/src/shape_guard.rs`, especially grinding-plan derivation and pre-allocation validation;
6. legacy/modular direct-program, advice, parity, Fiat-Shamir, and tamper fixtures; and
7. `.github/workflows/rust.yml` plus repository identity/style safeguards.


crates/jolt-akita/src/adapters.rs
```diff
@@ -827,7 +822,7 @@ pub(crate) fn sparse_unit_polynomial(
     }
 
     let mut seen = BTreeSet::new();
-    let mut coeffs = Vec::new();
+    let mut evals = vec![AkitaField::zero(); domain_size];
```
Acentelles: **Prover memory blowup:** `sparse_unit_polynomial` replaces the deleted upstream `SparseRingPoly` (O(#ones) memory) with a fully materialized `vec![AkitaField::zero(); 2^num_vars]`, retained through `open()` via `AkitaHintPolynomials::SparseUnit`. Any `is_one_hot` polynomial that misses the `one_hot_polynomial` fast path (K != the setup `one_hot_k`, or ColumnMajor order) at trace scale allocates the dense vector (num_vars=30 is ~16 GiB, plus `DensePoly::from_field_evals`' own copy) before `validate_commit_shape` runs. The only gate, `supports_unit_sparse_dimension`, is a lower bound. This looks like the actual root cause of the CI memory exhaustion this PR works around with `--test-threads 1`. The indices are already in hand (`for_each_one`); consider routing them through the one-hot backend representation or pushing a sparse commit source back into the upstream API.

crates/jolt-akita/src/shape_guard.rs
```diff
@@ -310,197 +246,74 @@ fn validate_commitment_len(
...
-    shape: &TerminalLevelProofShape,
+fn validate_proof_shape(
+    shape: &AkitaBackendProofShape,
     schedule: &FoldSchedule,
+    layout: &OpeningClaimsLayout,
 ) -> Result<(), OpeningsError> {
-    validate_ext_reduction_shape(shape.extension_opening_reduction.as_ref())?;
-    let scheduled = &schedule.terminal.params.response_shape;
-    if !scheduled.admits_realized(&shape.terminal_response) {
+    let grinding_plan = derive_transcript_grinding_plan::<AkitaConfig>(schedule, layout)
```
Acentelles: **Shape guard hardcodes the dense config for all three flavors:** `validate_proof_shape` uses `derive_transcript_grinding_plan::<AkitaConfig>` and `AkitaConfig::EXT_DEGREE`, but both deserialize entry points resolve schedules per flavor (Dense / OneHotK16 / OneHotK256). Today the presets agree, but nothing pins the grinding policy across configs: if a future akita rev differentiates one-hot grinding plans or nonce budgets, honest one-hot proofs get rejected pre-deserialization, or a forged shape that is canonical-for-the-wrong-config passes this allocation guard, which is exactly the DoS this module exists to prevent. `resolve_schedule_row::<Cfg>` is already generic; thread the resolving `Cfg` into shape validation too.

crates/jolt-prover-legacy/src/field/akita.rs
```diff
@@ -321,24 +327,23 @@ fn elem_to_field(a: &BigInt<2>) -> AkitaField {
...
-        0..=3 => AkitaField::solinas_reduce(&a.mul_wide_limbs::<M, 5>(mag.0)),
-        4 => AkitaField::solinas_reduce(&a.mul_wide_limbs::<M, 6>(mag.0)),
-        _ => a * AkitaField::solinas_reduce(&mag.0),
-    }
+    // Reduce the magnitude first, then multiply in the field: equal modulo p
+    // to reducing the full 2+M-limb product. The shared `jolt_field::Fp128`
+    // dropped the pre-cutover `mul_wide_limbs` specializations (this glue was
+    // their only consumer); `solinas_reduce` accepts ≤10 limbs, covering the
+    // M ∈ {3, 4} magnitudes callers pass (S160/S192/S256).
+    a * AkitaField::solinas_reduce(&mag.0)
```
Acentelles: **Pessimization on a false premise:** the WHY comment claims the shared `jolt_field::Fp128` dropped the `mul_wide_limbs` specializations, but `mul_wide_limbs::<3,5>/<4,6>` exists at `crates/jolt-field/src/solinas/fp128.rs:719` and this PR's own `Fp128SignedAccumulator::product` uses exactly the old fused pattern. Based on that premise, the fused single-reduction path (`solinas_reduce(&a.mul_wide_limbs::<M,..>(mag.0))`) was replaced by `a * solinas_reduce(&mag.0)`: an extra full Solinas reduction plus a 128x128 field multiply per call in `mul_to_accum_mag`/`mul_to_product_mag`, which back the `FMAdd<F, S160/S192/S256>` impls in the packed prover's sumcheck accumulation inner loops. Congruent mod p, so byte-diff CI cannot catch it. Suggest restoring the M-matched fused arms and correcting the comment.

crates/jolt-akita/src/trace_onehot.rs
```diff
@@ -388,14 +387,67 @@ impl TracePackedOneHot {
...
+        let mut coefficients = [AkitaField::zero(); D];
+        let mut selected_rows = [NO_SELECTED_ROW; u64::BITS as usize];
+        let mut cached_row = usize::MAX;
+        let mut committed_zero_mask = 0u64;
+        for (coefficient_index, coefficient) in coefficients.iter_mut().enumerate() {
+            let field_offset = segment_offset + coefficient_index;
+            let row = field_offset / self.one_hot_k;
+            if row != cached_row {
+                self.rows
+                    .fill_row(row, &mut selected_rows[..self.num_columns]);
```
Acentelles: **Per-position traversal blowup:** this batch kernel calls `ring_coefficients` once per live ring position with a row cache that is a per-call local, so each trace row is re-filled roughly `num_columns * K/D` times, versus the single row-major fill per ring in `visit_segment_ring_row_range` used by the neighboring commit/fold kernels. `num_live_positions` is trace-scale and this path is live for all one-hot K16/K256 root folds, so this is a bounded but real opening-path regression. Hoisting the `(cached_row, selected_rows, committed_zero_mask)` cache across positions, or restructuring onto the row-major walk, would remove it.

crates/jolt-akita/src/trace_onehot.rs
```diff
@@ -388,14 +387,67 @@ impl TracePackedOneHot {
...
             num_vars: total_field_elems.trailing_zeros() as usize,
-            construction_ring_elems: total_field_elems / construction_ring_d,
         })
     }
 
     fn total_field_elems(&self) -> usize {
         1usize << self.num_vars
     }
 
+    fn ring_coefficients<const D: usize>(
```
Acentelles: **Second owner of the packing law:** `ring_coefficients` re-derives the position -> (row, column, coefficient) mapping that `visit_segment_ring_range` in this same file already owns (segment sizing, fill_row, committed_digit_zero_mask, K>=D vs K<D packing), and the two copies already disagree on selector validation: the commit path rejects a selector byte >= K as `InvalidInput`, while this path's `field_offset % one_hot_k == selected_row` silently yields an all-zero ring for the same corrupt row. A future packing-convention change updates one site and not the other, caught only by e2e. Implementing `ring_coefficients` via `visit_segment_ring_range` filtered to the target column would restore the one-owner rule.

crates/jolt-akita/src/shape_guard.rs
```diff
@@ -177,7 +152,16 @@ pub(crate) fn deserialize_checked_grouped_backend_payload(
     }
```
Acentelles, line 226: **Guard tail duplicated:** the cap-check + shape-deserialize + `validate_proof_shape` + `validate_decode_budget` sequence is copy-pasted between `deserialize_checked_backend_payload` and `deserialize_checked_grouped_backend_payload`, and `validate_commitment_len` duplicates `validate_commitment_profile_len` body-for-body (`expected_commitment_coeff_len(resolved)` is literally `expected_commitment_coeff_len_for_profile(&resolved.profiles().final_group)`). The next shape-guard hardening (including the config-threading fix above) has to land in two places or one deserialization route silently keeps the weaker guard, and no test distinguishes the paths. Extracting one checked-backend-proof helper would close this.

.github/workflows/rust.yml
```diff
@@ -372,11 +372,45 @@ jobs:
...
+        uses: taiki-e/install-action@nextest
       # The MODULAR packed-prover acceptance suites: the five e2e analogs
       # (with live tampers) plus the four byte-diff ratchets asserting
       # wire-for-wire equality with the legacy packed prover.
       - name: Run jolt-prover tests with akita feature
-        run: cargo nextest run --cargo-profile ci -p jolt-prover --features akita,prover-fixtures
+        # Each byte-diff arm retains both modular and legacy proof artifacts.
+        # Running four arms concurrently exhausts hosted-runner memory and
+        # causes the runner service to terminate the otherwise healthy suite.
+        run: cargo nextest run --cargo-profile ci -p jolt-prover --features akita,prover-fixtures --test-threads 1
```
Acentelles: **Lane-wide serialization instead of a scoped test group:** `--test-threads 1` serializes all nine tests to dodge memory exhaustion attributable to the four byte-diff ratchet arms (and the sparse-unit densification flagged in adapters.rs), roughly doubling lane wall time for the five e2e tests that do not need it. The repo already has the scoped mechanism: nextest test-groups with `max-threads` in `.config/nextest.toml` (used for dory and zk-core-fixtures). Also, the local acceptance commands in CLAUDE.md lack the flag, so developers still run the arms concurrently and can OOM locally.

Acentelles (commented):
Ran a deep review of this cutover (multi-angle scan plus adversarial verification of each candidate finding). Ten findings below as inline comments, ranked roughly by severity: two substantive (the sparse-unit densification and the dense-config hardcoding in the shape guard), one hot-path pessimization justified by a factually false comment, and the rest maintainability.

Areas checked and cleared, for the record: the shared fp128 port is a faithful move with randomized parity and production-modulus differential coverage; Fiat-Shamir challenge-byte conventions match between legacy and modular paths; the canonical-shape exact-equality check is complete at the pinned akita rev (honest realized shape equals the canonical shape by construction, so the loss of the old admits-style tolerance cannot reject honest proofs); `validate_decode_budget` passing the same width twice is correct because `ExtField = Field` is pinned by a compile-time trait bound for every config Jolt uses; no stale references to the deleted bootstrap feature remain; and the akita pin is an immutable LayerZero-Labs rev.

crates/jolt-akita/src/configs.rs
```diff
@@ -20,7 +20,7 @@ fn dp_planned_schedule<Cfg: CommitmentConfig>(
 ) -> Result<FoldSchedule, AkitaError> {
```
Acentelles: **Duplicate planner invocation:** `dp_planned_schedule` re-encodes `schedules::emit`'s planner call (find_schedule with `honest_fold_policy_of::<Cfg>`, policies, `policy_of::<Cfg>`, ring_challenge_config, then `validate_structure`) instead of calling the already-`pub(crate)` `emit::regen`. If the planner contract grows a parameter or validation step, catalog emission/drift-testing and setup capacity sizing diverge, sizing setup for a different schedule than the one emitted and verified.

crates/jolt-akita/src/adapters.rs
```diff
@@ -50,7 +50,7 @@ pub(crate) type AkitaBackendProofShape = AkitaBatchedProofShape;
 pub(crate) type AkitaBackendVerifier = AkitaBackendVerifierSetup<AkitaField>;
 pub(crate) type AkitaBackendDensePoly = DensePoly<AkitaField>;
 pub(crate) type AkitaBackendOneHotPoly = OneHotPoly<AkitaField, u8>;
-pub(crate) type AkitaBackendSparsePoly = SparseRingPoly<AkitaField>;
+pub(crate) type AkitaBackendSparsePoly = DensePoly<AkitaField>;
```
Acentelles: **Compatibility name for a deleted encoding:** `AkitaBackendSparsePoly` is now a second name for `DensePoly<AkitaField>`, identical to `AkitaBackendDensePoly` two lines above. Since `sparse_unit_polynomial` now builds a fully dense eval vector, a reader of the `SparseUnit(Arc<[AkitaBackendSparsePoly]>)` variant wrongly assumes sparse storage and sparse cost. Suggest deleting the alias; the variant name already carries the semantic distinction.

crates/jolt-field/src/fp128_accumulators.rs
```diff
@@ -1,25 +1,24 @@
...
+use core::marker::PhantomData;
 
 use crate::signed::S256;
-use crate::Accumulator;
+use crate::{Accumulator, Fp128, Fp128MulU64Accum, Fp128ProductAccum, Ring, Unreduced};
 
 /// Deferred-reduction accumulator for fp128 products.
 #[derive(Clone, Copy)]
-pub struct AkitaAccumulator(Fp128ProductAccum);
+pub struct Fp128Accumulator<const P: u128>(Fp128ProductAccum, PhantomData<Fp128<P>>);
```
Acentelles: Nit: the `PhantomData<Fp128<P>>` fields on both accumulators are unnecessary; unused const generic parameters do not trigger E0392, so the marker fields and their `Default` plumbing are dead state. `struct Fp128Accumulator<const P: u128>(Fp128ProductAccum);` does the same job.

Acentelles:
Overall, the PR looks good to me. It is a natural continuation of the previous two field-crate PRs, and the risky parts already held up under a deep pass. I can't see any changes to the architecture of Jolt, no new crates, and nothing soundness-breaking.
  
My AI found two of the inline comments are worth resolving before merge, since both are cheap relative to the size of this PR.

quangvdao:
@Acentelles thanks for the careful pass — I addressed the full review in [`7745e0990`](https://github.com/a16z/jolt/commit/7745e09904eeb02d8a157326847d4870fd07794a).

The substantive fixes are:

- deleted the fake sparse-unit fallback that materialized `2^num_vars`; unsupported one-hot shapes now fail explicitly before dense conversion, and the misleading sparse alias plus dead ignored benchmark are gone;
- threaded the actual Dense/K16/K256 config through proof-shape and grinding validation, and consolidated the shared commitment/proof guard tail;
- restored the fused `mul_wide_limbs` fp128 paths;
- replaced per-position trace refills with one row-major packing traversal, with independent K16/K256 × D64/128/256/512 parity coverage, a one-fill-per-row regression test, and explicit invalid-selector rejection;
- scoped serialization to the Akita byte-diff binary through a nextest test group instead of serializing the whole modular lane;
- centralized schedule planning behind one canonical function used by regeneration, setup sizing, and grouped-row provisioning; and
- removed the unnecessary accumulator `PhantomData` state.

I also ran an independent adversarial full-PR audit after the fixes. It found no Critical/High correctness, soundness, serialization, allocation, panic-path, or feature-gating defects. Its two Moderate follow-ups—one remaining duplicate planner owner and the broken ignored sparse benchmark—are included in the same commit.

On the exact pushed tree: all 56 `jolt-akita` tests passed, catalog regeneration matched exactly, all 18 modular Akita acceptance tests passed (including the four byte-diff ratchets and committed-program proof), and the Akita/legacy/modular clippy plus format/style checks passed. Fresh CI is running on the new head.

Acentelles (approved):

quangvdao:
### Maintainability follow-up

I addressed the local `trace_onehot.rs` debt in `0b88b660d` and removed the obsolete packed-witness surface in `afaf69220` (both are present at current head `bbce94ed5`).

What changed:

- Replaced the 3,351-line `crates/jolt-akita/src/trace_onehot.rs` with focused modules for the public source model, canonical traversal/geometry, commitment, opening fold, decomposition, backend kernel dispatch, heterogeneous grouped dispatch, and tests.
- Kept the existing `jolt_akita::{no_selected_row, TraceOneHotRows, TracePackedOneHot}` public surface stable.
- Kept each packing law with its existing canonical implementation; this is a source-layout refactor, not a rewrite of the hot-path formulas or trace traversal.
- Added explicit per-module imports and kept cross-module implementation visibility at `pub(super)` rather than creating new crate-wide APIs.
- Deleted `SparseUnitPolynomial` and its self-only test from `packed_witness.rs`. It had no production callers.
- Deliberately did **not** decompose or otherwise restructure legacy `packed.rs`; that prover is slated for deletion.

The resulting production files range from 114 to 777 lines, with `decomposition.rs` now the largest. The longer-term upstream sparse-position streaming API could still reduce how much packing machinery Jolt owns locally, but the giant local module boundary is gone.

Local validation completed before the clean synchronization merge:

- formatting, diff hygiene, and repository style-invariant check
- strict clippy for `jolt-akita`
- strict clippy for `jolt-prover-legacy --features akita`
- full `jolt-akita` nextest suite: 56/56 passed (one additional test skipped; nextest reported one passing fixture as slow/leaky)

Not claimed from the local pass: exact catalog regeneration and the modular prover Akita fixture rerun were interrupted before completion. Current-head CI remains authoritative for those gates.

PR #1797, opened 2026-08-22
feat(field): restore and prove Fp128 kernels

quangvdao:
## Summary

This PR restores architecture-specific `Fp128` kernels behind the explicit `jolt-field/asm` feature and proves the shared add, subtract, and baseline multiply instruction bodies generically for every supported pseudo-Mersenne offset.

It now:

- uses register-parameterized assembly for add, subtract, and multiply on AArch64 and x86-64 whenever `asm` is enabled;
- uses AArch64 assembly for square and fused multiply-add, with portable Rust retained as the differential oracle;
- proves add, subtract, and baseline multiply for every `p(c) = 2^128 - c` satisfying the same offset bounds enforced by Rust;
- retains complete byte-linked callable theorems and a primality certificate for the production A7F7 specialization;
- retains the separately proved A7F7 BMI2+ADX x86-64 multiply path;
- differentially tests the assembly and portable implementations for public offsets 275 and A7F7 plus test-only offset 173, so the generic path cannot silently become unreachable;
- keeps assembly opt-in: no product crate enables `jolt-field/asm` in this PR;
- records the transcript, portability, benchmark, and migration-audit repairs discovered while moving the field implementation into Jolt.

The former base PR, #1794, has merged. This PR is now based directly on `main`.

Normative design: [`specs/jolt-field-proof-continuity.md`](specs/jolt-field-proof-continuity.md).

### Diff metadata

- Base: `a43f881c7d6e63fbf4fecfc4e7d4fec8b777d961` (`main`)
- Head: `793f1432894332602f29c31c55a3b19c74a872a7`
- Commits: 29
- Files changed: 87
- Diff: 6,416 insertions and 342 deletions

## Motivation

The Jolt field rebuild replaced the old `Fp128` add and subtract assembly with portable Rust, while Akita later developed fixed A7F7 AArch64 kernels and HOL Light proofs on a separate branch. Removing `akita-field` without moving those assets would have lost both the optimized kernels and their theorem connection.

The first restoration proved only the A7F7 specialization. Review correctly identified two missing boundaries:

1. assembly should be explicitly opt-in and continuously compared with the portable implementation;
2. the arithmetic bodies take the Solinas offset as a register input, so their machine proofs should quantify over that input instead of treating A7F7 as the only proved modulus.

This PR now makes those boundaries explicit. A7F7 remains the exact full-symbol fixture and the only specialized BMI2+ADX path, while the common register bodies and their arithmetic proofs are offset-generic.

The migration audit also found a field-specific transcript byte-order issue, a missing fallible parallel helper, an invalid extension benchmark comparison, and incomplete portability coverage. Those repairs remain part of this change surface.

## Architecture and proof connection

```mermaid
flowchart TD
    F[asm feature enabled] --> D[architecture dispatch]
    D --> C[load const-generic offset c]
    C --> B[shared register instruction body]
    B --> G[generic HOL theorem for valid c]
    A[A7F7 standalone object and witness] --> X[exact byte checker]
    X --> H[full callable A7F7 theorem]
    P[portable Rust] --> T[differential tests and fuzzing]
    B --> T
```

| Surface | Without `asm` | With `asm` on AArch64 | With `asm` on x86-64 |
| --- | --- | --- | --- |
| Add/sub | Portable Rust | Generic register body | Generic register body |
| Multiply | Portable Rust | Generic register body | Generic baseline body; A7F7 may select BMI2+ADX |
| Square | Portable Rust | Parameterized assembly, differentially tested | Portable Rust |
| Fused multiply-add | Portable Rust | Parameterized assembly, differentially tested | Portable Rust |
| Full callable byte theorem | N/A | A7F7 fixture | A7F7 baseline and BMI2+ADX fixtures |

The shared body includes are the source for both Rust `asm!` and the standalone proof objects. The generic machine theorems start immediately after the fixture's constant-loading instruction and assume the same fixed register contract used by Rust: `c` is in `x4` on AArch64 or `r8` on x86-64. The A7F7 corollaries additionally prove the literal load, ABI result placement, and `ret` transition.

The artifact checker compares complete A7F7 callable Linux witness symbols with the proof objects. HOL Light imports those exact objects with `define_assert_from_elf`, so a one-byte mismatch prevents the proof from loading. Darwin x86-64 has an exact checked frame wrapper around the proved arithmetic/ABI sequence; the wrapper itself is not yet covered by a callable theorem.

## Generic formal claim

Define

```text
p(c) = 2^128 - c
valid(c) := 0 < c < 2^32 and c(c + 1) < p(c).
```

For every valid `c` and canonical `a, b < p(c)`, the AArch64 and baseline x86-64 instruction-body theorems prove:

- add returns `(a + b) mod p(c)`;
- subtract returns `(a + p(c) - b) mod p(c)`, equal to `(a - b) mod p(c)`;
- multiply returns `(a * b) mod p(c)`;
- each result is canonical;
- only the registers, flags, program counter, and events named by the frame condition may change.

These are precisely the arithmetic bounds used by the Rust `Fp128<P>` implementation. The public field aliases are intentionally limited to offset 275 and A7F7; test-only offset 173 exercises the generic path independently of both aliases.

For A7F7,

```text
p = 2^128 - 2^32 + 22537
  = 0xffffffffffffffffffffffff00005809.
```

The callable A7F7 corollaries prove the literal offset load, the architecture calling convention, result registers, and return. `JOLT_FP128_A7F7_PRIME` separately proves `prime p` with a checked Pocklington certificate. The optimized x86-64 BMI2+ADX theorem remains A7F7-specific because that instruction sequence embeds the specialization rather than accepting `c` as a register parameter.

## Testing and proof policy

Formal proof is one layer of the evidence, not a substitute for testing:

- deterministic unit tests compare assembly with portable Rust over edge cases and seeded random canonical values;
- libFuzzer compares all architecture kernels for offsets 173, 275, and A7F7;
- x86 fuzzing asserts which baseline or BMI2+ADX backend actually executed;
- native AArch64 and x86-64 CI run both differential fuzzing and clean HOL Light replays;
- the ordinary `jolt-field` fuzz target is part of the repository fuzz matrix;
- the portability workflow uses repository-policy `cargo nextest` in every job;
- fuzz CI pins `nightly-2026-08-24`, avoiding an upstream compiler regression introduced by the following nightly.

The AArch64 square and fused multiply-add kernels are differentially tested but do not yet have machine-code theorems. They do not inherit the add/sub/multiply theorem merely because they share reduction ideas.

## Safety and trust boundary

The assembly declares every input, output, temporary register, and condition flag. The bodies use no memory or stack and operate on caller-saved registers. Their machine theorems assume canonical inputs, matching the field representation invariant; no release-mode range check is added to the hot path.

Assembly is an explicit library option. No product crate enables it here. A downstream rollout must forward the feature deliberately and retain architecture validation.

BMI2+ADX selection is a compile-time deployment contract, not a runtime CPU check. Generic x86-64 builds use the baseline path. A binary built with both features may fault on a processor lacking them.

The remaining trusted components include:

- Rust's inline-assembly contract and compiler-generated code around inlined callers;
- the Darwin x86-64 frame wrapper;
- the linker and final build configuration;
- the HOL Light kernel, s2n-bignum architecture models, OCaml, and the checker host;
- the physical processor;
- downstream feature selection, reachability, optimization, and final linking.

The current legacy `akita` feature still uses the external Akita field type. These theorems establish the local Jolt cutover target; they do not prove the current Akita-enabled verifier binary end to end.

Unreduced arithmetic retains a separate caller obligation. Fp128's eight signed `i32` lanes permit 32,768 same-sign unit additions, and its wrapping `u128` product slots have `2^64 - 1` terms of documented nonnegative-product headroom. Release hot paths do not enforce those limits.

## Compatibility and migration repairs

This repository makes no backward-compatibility guarantee for the affected proof transcript.

- Solinas Fp128 and the temporary Akita adapter derive scalar challenges directly from little-endian bytes, matching Akita. BN254 and Dory retain Jolt's historical reversal. Existing packed Akita proofs produced with the old Jolt behavior are incompatible; no dual mode is added.
- Every `CanonicalEncoding` implementation must choose its scalar-challenge byte order explicitly.
- Public Fp128 aliases 159 and 2355 are removed. Offset 275 remains the portable/reference field and, when `asm` is enabled on a supported architecture, uses the same generic add/sub/multiply bodies as any other valid offset.
- `cfg_try_fold_reduce!` is restored, and `cfg_join!` reaches Rayon through a hidden Jolt re-export.
- The extension benchmark's generic leg now calls the generic schedules rather than the specialization under test.
- Scalar benchmarks cover add, subtract, multiply, and square for Fp32, Fp64, and Fp128.
- Portability CI reacts to Rust toolchain and Cargo-configuration changes and independently checks native AArch64 and x86 SIMD/scalar arithmetic.

## Performance

On an AMD Ryzen 9 9950X, the proved baseline multiply measured about 8.11 microseconds per 4,096 products versus 9.39 microseconds for portable Rust under `-C target-cpu=native`, a 13.6% reduction. The A7F7 BMI2+ADX path measured about 2.4% faster than the baseline in the production-shaped benchmark.

On Apple M4, the AArch64 multiply measured about 1.29x the portable throughput. Earlier measurements showed that moving hot add/subtract operations behind an out-of-line call cost 2.34x to 2.49x, so production keeps these bodies inline; standalone objects exist for proof import and inspection.

## Validation at `793f14328`

The following passed locally at the exact head:

- clean AArch64 HOL Light replay for add, subtract, multiply, callable A7F7 corollaries, and primality;
- clean x86-64 HOL Light replay for add, subtract, baseline multiply, BMI2+ADX multiply, callable A7F7 corollaries, and primality;
- `cargo nextest` for `jolt-field` with `solinas,asm` (119 tests) and with portable `solinas` (118 tests);
- warnings-denied Clippy for both assembly and portable feature sets;
- 10,000 local differential fuzz iterations across offsets 173, 275, and A7F7;
- formatting and whitespace checks.

All 51 GitHub checks passed at this exact head. The run includes both architecture differential-fuzz jobs, both HOL Light jobs, the shared field matrix, all five repository fuzz-crate jobs including `jolt-field`, Clippy, formatting, verifier fixtures, and every modular and legacy test job.

## Out of scope

This PR does not prove:

- AArch64 square or fused multiply-add at the machine-code level;
- the A7F7-specific BMI2+ADX theorem for arbitrary offsets;
- portable Rust reduction, packed SIMD arithmetic, extension-field schedules, inversion, decoding, or serialization at the machine-code level;
- compiler-generated code surrounding every inlined copy or the Darwin x86-64 frame wrapper;
- caller compliance with canonical-input and unreduced-headroom invariants;
- reachability in a final Akita-enabled executable or the complete prover/verifier protocol.

## Reviewer map

Suggested review order:

1. `specs/jolt-field-proof-continuity.md`
2. `crates/jolt-field/src/solinas/fp128.rs` and `fp128/add_sub.rs`
3. `proofs/hol-light/fp128_common.ml` and the six generic architecture theorems
4. the A7F7 callable corollaries and `fp128_prime.ml`
5. `crates/jolt-field/asm/` and `scripts/check_fp128_proof_artifacts.py`
6. differential unit tests, fuzz target, and proof/portability workflows
7. the formal-verification Book chapters
8. transcript, parallel-helper, benchmark, and migration-integration repairs


moodlezoup:
Let's put these kernels behind an `asm` feature in `jolt-field`, and add differential fuzzing comparing the assembly and portable implementations

Acentelles (commented):
Review findings (10), ranked most-severe first as inline comments. Headline: the hand-written A7F7 assembly itself was independently re-derived and cross-checked (carry chains, encodings, fold bounds, primality certificate) and no arithmetic bugs were found; the findings below are integration, coverage, and maintenance issues.

crates/jolt-field/src/akita.rs
```diff
@@ -100,13 +100,10 @@ impl CanonicalEncoding for AkitaField {
         <Self as akita_field::CanonicalBitLength>::num_bits(self)
     }
 
-    /// Legacy convention: digest bytes are interpreted as a big-endian
-    /// integer before reduction.
+    /// Akita transcripts interpret digest bytes directly as little-endian.
     #[inline]
     fn from_scalar_challenge_bytes(bytes: &[u8]) -> Self {
```
Acentelles, line 105: **Endianness flip has no protocol-version gate, and the two stacks' trait defaults now diverge.**

The 128-bit Fiat-Shamir challenge decoding flips from big-endian (byte-reversed) to little-endian. The break itself is spec-documented as intentional, but two pieces of residue look unhandled:

1. Any packed-Akita proof serialized before this PR still passes `validate_proof_config` (`crates/jolt-verifier/src/config.rs` has only the zk and commitment axes — no version or convention field) and then fails deep inside a sumcheck with a generic mismatch, instead of a clean "unsupported proof version" rejection.
2. The legacy `JoltField::from_scalar_challenge_bytes` default still byte-reverses (`crates/jolt-prover-legacy/src/field/mod.rs:232`) while the modular `CanonicalEncoding::from_scalar_challenge_bytes` default does not (`crates/jolt-field/src/algebra.rs:344`). A future field type implemented on both stacks that relies on the defaults gets BE challenges from the legacy prover and LE from the modular verifier — a transcript mismatch the type system cannot catch.

crates/jolt-field/Cargo.toml
```diff
@@ -31,8 +31,11 @@ default = ["bn254"]
 akita = ["dep:akita-config", "dep:akita-field"]
 bn254 = ["dep:ark-ff", "dep:ark-serialize", "dep:ark-bn254"]
 solinas = []
+asm = []
```
Acentelles, line 37: **The new opt-in `asm` feature is enabled by no workspace crate, so the proved kernels ship nowhere.**

The previously-unconditional AArch64 Fp128 mul/sqr kernels (the 1.29x M4 win) and every newly proved A7F7 kernel are compiled out of all consumer and default builds. A downstream consumer building with `features = ["solinas"]` (the pattern every existing consumer uses) silently loses the asm kernels versus the pre-PR tree, and the bench-crates regression gate cannot see it because CI benches pass `--features jolt-field/asm` — measuring a configuration nothing ships. The BMI2+ADX multiply is doubly unreachable: its `cfg(target_feature)` gate is compile-time and no build config sets `-C target-feature=+bmi2,+adx`. As it stands, a kernel + HOL proof + dedicated CI jobs are maintained for code that never executes in any real artifact.

crates/jolt-field/src/solinas/fp128/add_sub.rs
```diff
@@ -0,0 +1,560 @@
...
+    fn add_raw_aarch64_dispatch(a: [u64; 2], b: [u64; 2]) -> [u64; 2] {
+        // AArch64 add immediates are limited to 12 bits. List the built-in
+        // small offsets here and use a register for every other offset.
+        match Self::C_LO {
+            275 => Self::add_raw_aarch64_imm::<275>(a, b),
+            159 => Self::add_raw_aarch64_imm::<159>(a, b),
+            2355 => Self::add_raw_aarch64_imm::<2355>(a, b),
+            0xffff_a7f7 => Self::add_raw_aarch64_a7f7(a, b),
+            _ => Self::add_raw_aarch64_reg(a, b, Self::C_LO),
+        }
```
Acentelles: **The `_ =>` fallback `_reg` asm kernels are unreachable by every test and fuzz target.**

The dispatch match arms cover exactly `C ∈ {275, 159, 2355, 0xffff_a7f7}`, and the unit test plus fuzz target exercise exactly those four offsets — so the generic-register kernels (add/sub, both architectures) have zero runtime coverage anywhere. Registering any future Fp128 modulus with another offset routes hot-path field arithmetic through never-executed handwritten asm where a latent carry/borrow bug ships silently.

Related coverage gaps: `fp128-formal-verification.yml`'s BMI2 job relies on cargo-fuzz appending ambient `RUSTFLAGS` with no compiled-in probe asserting `mul_raw_x86_64_a7f7_bmi2_adx` was actually the code under test, and the AVX-512 portability lane switched its scalar oracle to asm without the `fp128_asm_matches_portable` cross-check step the other two lanes got.

crates/jolt-field/build.rs
```diff
@@ -0,0 +1,86 @@
...
+        return Ok(());
+    }
+
+    let asm_dir = Path::new("asm").join(&target_arch);
+
+    let out_dir = PathBuf::from(
+        env::var_os("OUT_DIR")
+            .ok_or_else(|| io::Error::new(io::ErrorKind::NotFound, "Cargo did not set OUT_DIR"))?,
+    );
+    let compiler = env::var("CC").unwrap_or_else(|_| "cc".to_owned());
```
Acentelles: **Cross-compilation assembles target-arch `.S` files with the host `cc`.**

With `fp128-proof-linkage` enabled, build.rs uses the plain host `cc` (env `CC` fallback), never consulting `CC_<target>` or passing `--target`; only Darwin-host Apple targets get `-arch` handling. `cargo build --target aarch64-unknown-linux-gnu -p jolt-field --features solinas,fp128-proof-linkage` on an x86_64 Linux host feeds `asm/aarch64/*.S` (with `.inst` directives) to the host x86 assembler — either hard-failing the build or, with a multi-target clang, emitting a wrong-arch/wrong-format object into `OUT_DIR` that `check.sh` and the HOL Light proofs then consume.

scripts/check_fp128_proof_artifacts.py
```diff
@@ -0,0 +1,527 @@
...
+        "standalone multiplication proof object",
+        mul_object_bytes,
+        mul_expected,
+    )
+    require_bytes(
+        "standalone BMI2 and ADX multiplication proof object",
+        bmi2_adx_mul_object_bytes,
+        bmi2_adx_mul_expected,
+    )
+    add_witness_expected = (
```
Acentelles, line 387: **Inspection witnesses require byte-for-byte equality including compiler-generated glue, and platform selection uses the host, not the built target.**

A `rust-toolchain.toml` bump where LLVM reorders the return moves, picks a different `mov` encoding, or changes frame emission breaks `fp128-formal-verification` CI on a PR that touched no fp128 code, with a raw byte-mismatch dump whose only fix is re-transcribing glue bytes into this script. Selecting the Darwin expectation by `sys.platform` applies the wrong frame expectation when the checker runs against a binary built for a different platform than the host (artifacts copied between machines, Linux container on a mac) — and the Darwin branch is never executed in CI, so it silently rots. The boundary-tolerant helpers written for exactly this problem (`require_bytes_once`, `require_instruction_sequence_once`) are dead code called only by their own unit tests.

proofs/hol-light/check.sh
```diff
@@ -0,0 +1,353 @@
...
+}
+trap cleanup EXIT
+
+if [[ -n "$TARGET_TRIPLE" ]]; then
+  PROFILE_ROOT="$PROOF_TARGET/$TARGET_TRIPLE/release"
+else
+  PROFILE_ROOT="$PROOF_TARGET/release"
+fi
+
+find_newest_object() {
```
Acentelles: **`find_newest_object` can select a stale object from a different build fingerprint in non-clean local runs.**

It picks `*/out/fp128_*.o` across all cargo build-fingerprint directories by mtime. Sequence: run check.sh with toolchain A; edit an `.inc` and run with toolchain B (new fingerprint dir, newer `.o`); revert and switch back to A — cargo reuses A's cached build without re-running build.rs, but B's object has the newest mtime, so the byte check and HOL proof validate B's bytes while the production binary embeds A's. The "proof matches production" claim is then made about the wrong artifact. (CI is safe because it always passes `--clean`.)

.github/workflows/fp128-formal-verification.yml
```diff
@@ -0,0 +1,144 @@
...
+      - "crates/jolt-field/**"
+      - "proofs/hol-light/**"
+      - "scripts/check_fp128_proof_artifacts.py"
+      - "scripts/tests/test_check_fp128_proof_artifacts.py"
+      - "Cargo.toml"
+      - "Cargo.lock"
+      - "rust-toolchain.toml"
+      - ".cargo/**"
+  pull_request:
+    branches: ["**", main]
```
Acentelles: **Trigger and job structure are wasteful and drift-prone.**

- `pull_request` fires on `branches: ["**"]` with paths including the workspace-wide `Cargo.toml`/`Cargo.lock`: a routine lockfile bump in an unrelated crate burns roughly two runner-hours rebuilding OCaml and HOL Light from source twice.
- The two ~45-minute proof jobs are verbatim copies (duplicated hol-light/s2n-bignum SHA pins, no `actions/cache` for the HOL Light build); a pin bump edited in one job but not its twin makes the two architectures silently verify against different prover libraries.
- The BMI2 job greps `/proc/cpuinfo` and hard-fails instead of skipping on a runner without ADX, turning the job red on unrelated PRs. `fp128_asm_differential` is also already fuzzed by `fuzz-crates.yml`.

scripts/check_fp128_proof_artifacts.py
```diff
@@ -0,0 +1,527 @@
...
+
+import argparse
+import re
+import shutil
+import subprocess
+import sys
+from pathlib import Path
+
+
+AARCH64_ADD_BODY = [
```
Acentelles, line 13: **Third hand-transcribed copy of every kernel's machine code.**

The `.inc` asm and the `proofs/hol-light/*_object.ml` `define_assert_from_elf` lists are the other two, and the `AARCH64_*_OBJECT_WORDS` / `*_WITNESS_WORDS` constant pairs here are element-for-element identical duplicates. A one-instruction change to a kernel body requires re-deriving the same hex in three hand-edited places per architecture; updating the object.ml and `.inc` but forgetting these constants fails CI on a correct kernel with a raw byte dump the author is tempted to "fix" by pasting the compiled bytes back in — blessing whatever was built. A single constant per operation (or parsing the object.ml lists as the one source of truth) removes the drift channel.

crates/jolt-field/src/solinas/fp128/add_sub.rs
```diff
@@ -0,0 +1,560 @@
...
+            if overflow | carry3 { r1 } else { s1 },
+        )
+    }
+
+    #[cfg(all(feature = "asm", target_arch = "aarch64"))]
+    #[inline(always)]
+    fn add_raw_aarch64_dispatch(a: [u64; 2], b: [u64; 2]) -> [u64; 2] {
+        // AArch64 add immediates are limited to 12 bits. List the built-in
+        // small offsets here and use a register for every other offset.
+        match Self::C_LO {
```
Acentelles: **Magic offsets duplicated across six dispatch sites; imm/reg asm bodies are near-identical twins.**

`0xffff_a7f7` and the small-moduli list `{275, 159, 2355}` are hand-duplicated across four match statements in this file and two `C_LO` comparisons in fp128.rs. Adding or changing a supported modulus means editing six sites, and a typo in one silently falls through to the generic register path — that architecture quietly stops using the proved/optimized kernel with no test failing (both paths are correct). Each imm/reg asm variant pair is instruction-for-instruction identical differing only in the constant's operand kind, so a fix to one body can drift from its twin unnoticed. A named shared const plus one dispatch predicate (and a macro for the imm/reg pairs) collapses this.

typos.toml
```diff
@@ -25,3 +25,8 @@ extend-exclude = ["patches/*.patch", "crates/jolt-field/tests/golden_bytes.rs"]
 # (A_tru, A_unt) in the packed (Akita) prove path docs.
 "tru" = "tru"
 "unt" = "unt"
+# HOL Light names and theorem type.
+"ASSUM" = "ASSUM"
+"thm" = "thm"
+# x86 instruction byte strings contain this immediate byte token.
+"ba" = "ba"
```
Acentelles, line 40: **Repo-global `"ba" = "ba"` exception disables a common-typo check everywhere to accommodate one script.**

Any future book chapter or spec writing "ba" where "by"/"be" was meant now passes CI forever (same for the global `"thm"`/`"ASSUM"` entries). The established scoped pattern already exists two lines up (`extend-exclude` for golden_bytes.rs): adding `scripts/check_fp128_proof_artifacts.py` to `extend-exclude` and scoping the HOL tokens to `proofs/hol-light` keeps the typos gate intact.

crates/jolt-field/src/akita.rs
Acentelles: Severity qualification: this is protocol hygiene, not a current correctness bug — backward compatibility for the flip is explicitly disclaimed and both current Akita implementations override the defaults consistently. The actionable residue is forward-looking: consider making transcript decoding mandatory to implement (no default) or adding a version/convention field to the proof config. Non-blocking.

crates/jolt-field/Cargo.toml
Acentelles: Qualification: "never executes in any real artifact" is too absolute — an external consumer can enable the `asm` feature, and `-C target-cpu=native` on an ADX host does light up the BMI2 path at compile time. The substantive point stands: no workspace consumer enables it and the benches measure an opt-in configuration nothing in-repo ships, so this needs an explicit rollout decision (wire `asm` into the intended consumer, bench both configurations, or document that this PR deliberately lands dormant proof infrastructure).

scripts/check_fp128_proof_artifacts.py
Acentelles: Qualification after discussion: exact byte equality is arguably the point of the proof-continuity claim, so treating glue tolerance as a defect was too strong — for bytes the HOL proofs cover, strict equality is correct. The residue worth resolving: (a) if the ABI moves/frame around the kernel body are NOT in the proved region, strict equality on them buys only toolchain-bump breakage — and the PR itself contains `require_bytes_once`/`require_instruction_sequence_once` written for exactly that and left dead; either use them for unproved glue or delete them and document the strict policy. (b) The host-vs-target `sys.platform` selection only matters if cross-checking is meant to be supported. Follow-up, not blocking.

proofs/hol-light/check.sh
Acentelles: Narrowing this: the independently pinned expected bytes mean a materially different stale object fails the byte check loudly — so the realistic failure mode is a confusing false failure in non-clean local runs, not silent false assurance as originally stated. Replacing the mtime search with the exact build output path is still the right fix, but as a robustness follow-up, not a blocker.

.github/workflows/fp128-formal-verification.yml
Acentelles: Two qualifications: (1) "skip instead of hard-fail" on missing ADX was the wrong prescription — a silently green job that tested nothing is worse than red; the right fixes are guaranteeing ADX-capable runners or making the skip visible (neutral status), never green-on-skip. (2) If fuzz-crates.yml's lane covers only the baseline x86 kernel while this lane covers BMI2+ADX, they are complementary, not duplicates — withdrawn. The trigger-path and duplicated-job-pin points stand as non-blocking CI maintenance.

scripts/check_fp128_proof_artifacts.py
Acentelles: Partly withdrawn: keeping the checker's byte lists as an independent audit anchor (rather than parsing them from the object.ml lists) is a defensible proof-policy choice, and deriving one copy from another would reduce the independence the model is buying. Retained: the independence is only as strong as the transcription discipline — the tempting failure mode (fixing a mismatch by pasting compiled bytes back in) defeats the cross-check exactly when it matters, so documenting the transcription procedure would help. The element-identical OBJECT_WORDS/WITNESS_WORDS alias pairs are still worth collapsing. Non-blocking.

crates/jolt-field/src/solinas/fp128/add_sub.rs
Acentelles: Downgrading to low-priority refactoring. One caveat on the mitigation: a dispatch typo is only caught by the witness byte check if that check verifies the dispatch/call site — if it only verifies the proved kernel's bytes are present in the artifact, the unused kernel still passes while dispatch routes to the generic path (unverified either way). A shared const for the offsets is cheap; macro-generating the imm/reg asm pairs may cost more readability than it saves.

Acentelles:
The changes I think should be addressed before merge are:

- Add differential coverage for an `Fp128<P>` offset outside {275, 159, 2355, 0xffff_a7f7} so the generic-register add/sub kernels, and the AArch64 generic multiply path—actually execute.
- Make build.rs use a target-aware assembler/compiler. The current plain CC/cc invocation is not sufficient for supported Cargo cross-compilation.
- Resolve the asm feature’s deployment story explicitly: either enable it in the intended consumer, or document that these kernels are deliberately dormant until a later cutover and ensure benchmarks distinguish shipped/default from opt-in performance.

Acentelles (approved):

quangvdao:
Thanks for the detailed review. Consolidated update at `793f1432894332602f29c31c55a3b19c74a872a7`:

### Review findings addressed

- **Explicit assembly opt-in:** the kernels are behind `jolt-field/asm`. No workspace product enables the feature in this PR; the Book and PR description now state that a downstream rollout is a separate decision, and validation enables the feature explicitly.
- **Independent differential coverage:** unit tests and libFuzzer compare assembly against portable Rust on AArch64 and x86-64 for public offsets 275 and A7F7 plus test-only offset 173. The extra offset forces the generic register path to execute. The x86 fuzz hook also asserts whether the baseline or BMI2+ADX multiply backend actually ran. AArch64 square and fused multiply-add are included in the differential checks even though they are not yet machine-proved.
- **Target-correct proof builds:** `build.rs` now obtains the assembler through `cc::Build` with the Cargo host and target, the checker requires an explicit target OS, and the proof runner selects the exact Cargo build output rather than the newest matching object.
- **Proof/CI maintenance:** the duplicated architecture proof jobs are one matrix job, trigger paths were narrowed, the portability jobs use repository-policy `cargo nextest`, and the fuzz target is pinned and exercised in both the dedicated architecture jobs and the repository fuzz matrix.
- **Transcript convention:** scalar-challenge decoding is now mandatory to implement rather than inherited from a divergent default. Solinas/Akita use direct little-endian decoding; BN254/Dory retain their historical reversal. The intentional incompatibility with old packed-Akita proofs is documented; no compatibility mode is claimed.
- **Dispatch and field surface:** unused public Fp128 aliases 159 and 2355 were removed. Offset 275 and A7F7 remain, while shared register-parameterized add/sub and baseline-multiply bodies handle every valid const-generic offset. The A7F7 specialization constant is centralized.
- **Artifact and spelling hygiene:** the checker no longer carries redundant object/witness aliases, the independent byte-transcription policy is documented, and spelling exceptions are scoped by file type instead of disabling checks repository-wide.

### Additional strengthening beyond the original requests

The HOL Light development now defines `p(c) = 2^128 - c` and proves the AArch64 and baseline x86-64 add, subtract, and multiply instruction bodies for every `c` satisfying the same bounds enforced by Rust:

```text
0 < c < 2^32
c(c + 1) < p(c)
```

The full byte-linked callable A7F7 corollaries still cover the literal constant load, ABI result placement, and `ret`, and the A7F7 primality certificate remains separate. The BMI2+ADX multiply theorem remains A7F7-specific because that optimized byte sequence is itself specialized.

This gives us three independent layers: generic machine proofs for the shared instruction bodies, exact callable proof linkage for the production specialization, and differential testing against portable Rust over multiple offsets.

All 51 GitHub checks pass at the final head, including both architecture HOL Light jobs, both assembly differential-fuzz jobs, the full repository fuzz matrix, portability, Clippy, verifier fixtures, and the modular/legacy test suite. The PR description has also been rebuilt against the current `main...head` diff with the exact proof scope, trust boundary, validation, and remaining exclusions.

PR #1798, opened 2026-08-22
feat(akita): batch dense advice with the packed trace

omibo:
## Summary

The Akita path now commits trusted and untrusted advice directly as dense polynomials of little-endian `u64` words. It no longer converts advice into a byte one-hot table, proves an advice-specific reconstruction relation, or opens advice through separate PCS proofs.

Advice remains independently committed from `OneHotTrace`: trusted advice can be committed out of band, while untrusted advice is committed before the trace during proving. Akita's heterogeneous batching combines every present dense advice group with the final one-hot trace group in one opening proof, even though the groups have different source representations, dimensions, setups, and evaluation points.

## Motivation

Before this change, Akita represented the same advice twice:

```text
raw advice bytes
  +-- little-endian word polynomial --> RamValCheck --> AdviceClaimReduction
  |
  `-- byte one-hot polynomial -------> advice reconstruction --> PCS opening

OneHotTrace ----------------------------------------------------> PCS opening
```

The PIOP already consumes advice as words, so the byte one-hot representation and its reconstruction proof were redundant. This change commits the word polynomial directly and removes the advice-only reconstruction and separate opening work.

After this change:

```text
raw advice bytes
  `-- canonical dense word polynomial
        |-- independent dense commitment
        `-- RamValCheck --> AdviceClaimReduction --> final advice claim

trace rows --> final one-hot commitment conditioned on advice profiles

[advice claim(s), OneHotTrace claim]
  `-- one heterogeneous Akita batch proof and verification
```

## Canonical advice representation

Advice bytes are packed eight at a time into little-endian `u64` words and zero-padded to the public configured capacity. Inputs larger than that capacity are rejected. `common::advice::canonical_advice_words` is shared by the witness and commitment paths.

The logical word polynomial is prefix-packed into Akita's supported dense arity floor when necessary. Trusted and untrusted advice retain distinct semantic IDs and layout digests even when their capacities and contents match. The dense source uses a bounded configuration large enough for every `u64` coefficient.

## Heterogeneous batch

The canonical Akita group order is:

```text
[UntrustedAdvice?, TrustedAdvice?, OneHotTrace]
```

Each advice object remains an independent dense commitment. `OneHotTrace` remains a streamed one-hot commitment and is committed after all present advice objects, using their frozen commitment profiles.

Stage 8 resolves the final advice claims produced by `AdviceClaimReduction`, reduces them through their prefix-packed layouts, and calls `CommitmentScheme::prove_batch` once with zero or more advice groups followed by the trace. Verification reconstructs the same ordered statement and calls `CommitmentScheme::verify_batch` once.

Committed-program bytecode and image objects keep their existing auxiliary reconstruction and opening path.

## Parameter lifecycle

An advice commitment is created before `log_T` is known, so its standalone setup depends only on its public capacity. The final grouped Akita schedule, however, depends on both the final trace layout and the exact ordered profiles produced by the independent advice commitments:

```text
AkitaScheduleLookupKey {
    final_group: OneHotTrace layout derived from log_T,
    precommitteds: exact ordered advice profiles,
}
```

Generating every trusted-capacity × untrusted-capacity × trace-arity combination globally would make the checked-in catalog grow as a Cartesian product, and future precommitted objects would add more dimensions.

Instead, preprocessing uses the program's known advice capacities to freeze the reachable advice profiles and provisions grouped rows for every reachable final trace arity before constructing the main setup. Setup sizing includes all provisioned rows and the complete group-batch capacity.

Commit, prove, verify, and proof deserialization only resolve previously provisioned rows. A registry or catalog miss is an error; none of those paths invokes the expensive planner.

## Verifier boundary

The verifier resolves the proof-selected row from trusted preprocessing data before deserializing Akita commitment and proof payloads. It validates commitment coefficient lengths, byte lengths, and proof shape against the resolved schedule before shape-backed allocation. This keeps verifier memory proportional to the supplied payload and prevents proof-controlled shape counts from causing large allocations.

Roles, order, commitment flavor, arity, evaluation count, layout digest, setup capacity, schedule selection, and auxiliary-proof count are checked fail-closed.

## Removed protocol work

- Advice byte one-hot commitments.
- Advice reconstruction relations and sumchecks.
- Advice byte-opening IDs and proof fields.
- Separate advice PCS opening proofs and verification.
- Advice participation in reconstruction; committed program reconstruction remains.

The end-to-end suites cover no advice, untrusted-only, both advice kinds, K=16/K=256, committed programs, byte parity, malformed statements, and tampered proof payloads.

## Non-goals

- Changing committed-program reconstruction.
- Changing the Dory path or adding `akita + zk` support.
- Running parameter generation during commit, prove, verify, or decode.


markosg04 (approved):
Looks good to me. As a follow-up, I think we should move more of the generic batching boundary into Akita. Akita could expose an extensible heterogeneous source wrapper—or a prepared-group API—so callers can combine dense, one-hot, and custom streaming sources without writing their own dispatch layer. It could also support setup-owned supplemental schedule rows for application-specific group shapes.

Jolt currently carries both of these adapters itself: the source dispatch and the runtime schedule registry. Those are generic PCS concerns rather than Jolt protocol logic, and they account for much of the added code here. With that support in Akita, Jolt could mostly assemble its ordered groups and call one batched prove/verify path, substantially reducing the Jolt-side diff and maintenance burden.

PR #1799, opened 2026-08-22
feat(field-fv): prove scalar Fp64 kernels

quangvdao:
## Summary

This PR proves the compiler-emitted scalar `Prime64Offset59` add, subtract, and multiply kernels at the instruction level on AArch64 and x86-64 without changing normal production dispatch.

It adds:

- callable HOL Light theorems for AArch64 and x86-64 add, subtract, and multiply;
- separate Darwin and Linux AArch64 schedules and separate baseline/BMI2 x86-64 multiplication coverage;
- a checked primality certificate for `2^64 - 59`;
- exact byte checks connecting certified Rust inspection symbols, standalone proof objects, and HOL imports;
- a fail-closed certified-build matrix for compiler, target, feature, profile, wrapper, and CI policy;
- differential fuzzing against independent `u128` modular arithmetic on Linux AArch64 and x86-64 BMI2;
- final-executable inspection tooling that distinguishes exact proved sequences from structural candidates.

The measured custom assembly experiment was slower than the generic Rust path, so this PR deliberately leaves production `Fp64` arithmetic in portable Rust. The proof objects are inspection artifacts, not a new assembly backend.

PR #1797 has merged. This PR is now based directly on `main` at `53ed6c31977d463e1e9061fd4b5569825d04e4e6`.

### Diff metadata

- Base: `53ed6c31977d463e1e9061fd4b5569825d04e4e6` (`main`)
- Head: `a1b748dddd3861c0d65c3b9e2c333dca2cf3b5ce`
- Commits: 49
- Files changed: 68
- Diff: 6,184 insertions and 88 deletions

## Why this shape

These operations are small enough that an out-of-line call or rigid inline-assembly constraints can cost more than the arithmetic. Native Ryzen 9 9950X measurements found the experimental assembly path neutral for base-field addition, about 9.4% slower for subtraction, about 6.0% slower for multiplication, and about 21% slower for extension-field subtraction. That production experiment was reverted.

The retained approach proves stable compiler inspection symbols that invoke the same generic `Prime64Offset59` operations as normal callers under a fixed build identity:

```mermaid
flowchart LR
    R[Prime64Offset59 Rust operation] --> W[certified inspection symbol]
    M[certified build matrix] --> W
    M --> O[standalone proof object]
    O --> B[exact byte comparison]
    W --> B
    O --> H[HOL Light execution theorem]
    H --> T[modular arithmetic result]
    E[linked executable] --> X[exact or structural sequence report]
```

The inspection symbol provides a stable compiler artifact without inserting a call boundary into normal field operations. The final-executable checker makes the remaining application boundary visible rather than silently promoting a proof-object theorem to an arbitrary optimized binary.

## Formal claims

For

```text
p = 2^64 - 59 = 0xffff_ffff_ffff_fc4b,
```

and canonical `a, b < p`, HOL Light proves that each exact imported instruction sequence returns the canonical result of addition, subtraction, or multiplication modulo `p`. The callable theorems also establish the relevant ABI and frame conditions.

The optimized x86-64 multiply uses `mulx` and requires BMI2 only; it does not require ADX, AVX2, or AVX-512. Addition and subtraction use the baseline x86-64 schedule. `JOLT_FP64_PRIME` separately proves that `p` is prime.

| Target | Proof scope | Operations | Native CI |
| --- | --- | --- | --- |
| `aarch64-apple-darwin` | Complete inspection symbol | Baseline add/sub/mul | Required |
| `aarch64-unknown-linux-gnu` | Complete inspection symbol | Baseline add/sub/mul | Required |
| `x86_64-unknown-linux-gnu` | Complete inspection symbol | Baseline add/sub/mul and BMI2 mul | Required |
| `x86_64-apple-darwin-inspection-only` | Exact Darwin wrapper around proved inner sequence | Baseline add/sub/mul and BMI2 mul | Inspection only |

The Darwin x86-64 entry is intentionally weaker: the checker verifies the exact wrapper, inner sequence, and post-return padding, but the existing System V theorem proves only the inner callable object. Linux x86-64 requires exact code through `ret`; the only permitted unreachable trailing instruction is the registered one-byte `int3` padding.

## Certified build matrix

`proofs/hol-light/fp64-certified-builds.json` is the source of truth for target and proof policy. It records:

- exact target triples and target components;
- baseline and BMI2 feature sets;
- Rust 1.95.0, Cargo 1.95.0, LLVM 22.1.2, and recorded revisions;
- release-profile fields including optimization, LTO, codegen units, overflow checks, panic behavior, stripping, and incremental compilation;
- expected bytes, object source, correctness source, body theorem, callable theorem, wrapper policy, and native-CI obligation for every operation.

The validator rejects duplicate, incomplete, or unknown entries. Proof linkage is accepted only when the matrix runner supplies the current contract. Unknown targets, toolchain or feature drift, inherited Rust flags, Cargo profile overrides, and direct proof-feature builds outside the runner fail closed. Ordinary non-proof builds do not run this gate.

The matrix narrows the certified build identity; exact byte comparison remains the final authority. The proof and artifact checker keep independent expected byte lists so compiler or encoding changes cannot be accepted merely by regenerating one artifact.

## Production linkage and trust boundary

The claim has four distinct layers:

1. HOL Light executes exact imported bytes in the pinned architecture model and proves their modular arithmetic.
2. The artifact checker confirms that the certified inspection symbol has those bytes under the registered build identity.
3. The inspection symbol invokes the same generic field operation used by normal Rust callers.
4. The final-executable checker finds exact instances or reports structurally similar candidates without calling candidates proved.

The theorem does not automatically transfer to every inlined caller. LLVM may assign different registers, move constants, merge input/output storage, or choose a different schedule. A structural candidate still needs a theorem for its actual register mapping and evidence that the caller prepares external values correctly. The scanner also does not establish reachability from a verifier entry point.

The JSON proof-run record is unauthenticated evidence. Trust comes from replaying the proof command or trusted CI provenance, not from accepting the record alone.

The current Jolt verifier does not instantiate `Prime64Offset59`; the legacy `akita` feature uses the external Akita Fp128 field. This PR does not add an unused call merely to manufacture reachability. A release ratchet should become required only when a downstream verifier selects this field path.

## Differential testing

The `fp64_scalar_differential` fuzz target reduces arbitrary 128-bit inputs into `Prime64Offset59`, then compares add, subtract, multiply, and square with independent `u128` modular arithmetic. CI runs it on:

- native Linux AArch64, covering the certified Linux compiler schedules;
- native x86-64 with BMI2, covering the optimized multiply schedule.

The target is wired into the repository `jolt-field` fuzz package and pinned by CI target-selection checks, so formal proof and randomized testing remain independent obligations.

## Other safety changes

Generated canonical constructors are now `unsafe fn` because callers must guarantee that the input is below the modulus. Debug assertions remain, and internal callers document the safety reason. This adds no release hot-path check.

Proof assembly wrappers reject unsupported object formats rather than treating every non-Darwin target as Linux. The matrix remains the build authority; an unsupported OS or ABI cannot inherit a theorem by resemblance.

## Developer and CI workflow

- `proofs/hol-light/check-fp64.sh bytes <arch> --matrix-entry <id>` performs the fast certified build and exact byte check.
- `proofs/hol-light/dev-fp64.sh <arch> <operation> --matrix-entry <id>` starts a persistent HOL Light session.
- `proofs/hol-light/check-fp64.sh all <arch> --matrix-entry <id> --clean` performs a fresh replay and writes an unauthenticated run record.
- `scripts/fp64_certified_matrix.py validate` checks the complete matrix contract.
- `scripts/check_field_proof_final_binary.py` scans a linked application and can save a JSON report.

CI validates the matrix and negative cases, replays native proofs on Linux AArch64, Linux x86-64, and Apple Silicon, runs architecture-specific differential fuzzing, and preserves proof logs and run records.

## Commit map

- `b5edc28d1` — prove scalar Fp64 arithmetic and add inspection workflow
- `f22fb8b6b` — add linked-executable inspection and document the final application boundary
- `a7d605694` — add separate Linux AArch64 compiler sequences
- `621e5a281` — add certified build matrix, fail-closed linkage, native Apple CI, run records, and negative tests
- `cd175c6b9`, `05a557d9a` — isolate proof builds from inherited Cargo/Rust flags
- `47c883ac4` — certify unreachable x86 trap padding
- `76623e1e1` — add independent scalar Fp64 differential fuzzing
- `f0cab8f1e` — pin the Fp64 fuzz targets in CI
- `a1b748ddd` — merge the squash-merged #1797 result and current `main` without changing the 68-file child surface

## Validation at `a1b748ddd`

The squash-merge restack was validated at the exact head as follows:

- the tree differs from current `main` in exactly the original 68 child files, with the original 6,184 insertions and 88 deletions;
- every file modified by the child PR is byte-for-byte identical to its pre-restack version;
- `cargo fmt --all --check` and `git diff --check`;
- 118 of 118 `jolt-field` Nextest tests with `--no-default-features --features solinas`;
- warnings-denied release Clippy for all `jolt-field` targets with the `solinas` feature;
- validation of all four registered Fp64 proof targets;
- 34 focused Python tests for the certified matrix, proof artifacts, and final-binary inspector.

At the previous child head, every formal-proof and differential-fuzz CI job was green. At description publication for the exact restacked head, 11 GitHub checks had passed, none had failed, and 35 were running, including fresh native HOL Light and differential-fuzz replays.

## Deferred coverage

This PR does not cover:

- generic machine theorems for arbitrary Fp64 moduli or offsets;
- the subword `Prime40Offset195`, `Prime48Offset59`, and `Prime56Offset27` aliases, which use different reduction schedules from the full-width `Prime64Offset59` kernel;
- degree-two extension, packed, subword, delayed, or unreduced kernels;
- squaring, inversion, decoding, or serialization as complete machine-code operations;
- every optimized downstream caller or verifier reachability;
- a complete callable theorem for the Darwin x86-64 wrapper;
- authentication of proof-run records outside trusted CI or local replay.

## Reviewer map

Suggested review order:

1. `book/src/how/formal-verification/field-kernels-fp64.md`
2. `proofs/hol-light/fp64-certified-builds.json` and `scripts/fp64_certified_matrix.py`
3. `proofs/hol-light/fp64_common.ml` and the architecture theorem statements
4. `crates/jolt-field/examples/fp64_production_witness.rs` and proof objects
5. `scripts/check_fp64_proof_artifacts.py` and proof runners
6. `crates/jolt-field/fuzz/fuzz_targets/fp64_scalar_differential.rs`
7. `.github/workflows/fp64-formal-verification.yml`
8. `scripts/check_field_proof_final_binary.py` and trust-boundary documentation


markosg04 (approved):
Reviewed and fixed the findings in 3feae4df282cba276d12ccf89b18434cd1a1fc55: architecture-specific assembly watches now reference existing files, and the final-binary inspector tests and path filters are wired into CI. Corrected the modulus typo in the PR description.

Validation passed: all 118 jolt-field Nextest tests, all 34 focused Python tests, four-target matrix validation, formatting, and warnings-denied workspace Clippy in both host and host,zk modes. A repeated focused Clippy run reports Fresh jolt-field, confirming the rebuild issue is resolved. The patch also received an independent static review.

No blocking findings remain from this review. Native HOL Light replay and differential fuzzing were not rerun locally; fresh remote CI remains separate from this approval.

PR #1806, opened 2026-08-26
ci: codify maintainer style invariants; enforce the mechanical subset

0xAndoroid:
Codifies maintainer style rules from recent review and polish passes, documents the judgment-based rules in `CLAUDE.md`, and runs the mechanical subset as a cheap CI job.

## Machine-checked rules

Repo-wide across tracked Rust files:

- `folded-cfg-attr`: adjacent `#[cfg_attr(P, A)]` / `#[cfg_attr(P, B)]` attributes with the same predicate must be folded.
- `native-visit`: `#[allocative(visit = ...)]` cannot decorate primitive-element containers; their native implementations report element types and unused capacity.

Added lines relative to the PR merge base:

- `nominal-imports`: import types, traits, enums, constants, and PascalCase macros and use their short names. Keep enum variants qualified by the enum type (`Kind::ADD`, never a bare imported `ADD`). Import singleton paths directly (`use x::Kind`, never `use x::Kind::{self}`). Lowercase namespace free functions and macros may remain qualified. Once a path is imported, do not spell it qualified elsewhere in the file.

The former `TODO`/`FIXME` issue-link rule was removed.

## Checker behavior

`scripts/check_style_invariants.py` handles multiline attributes and `use` trees, top-level versus local imports, module aliases used for disambiguation, char literals during scope tracking, macro/attribute exclusions, singleton-self diagnostics, and duplicate-report suppression. `scripts/tests/test_check_style_invariants.py` pins these cases.

The branch also normalizes the existing singleton `std::fmt::{self}` import and aliases `ark_ff::BigInt!` as `ArkBigInt!`, avoiding its collision with `num_bigint::BigInt`.

The nominal-import rule remains syntax-based. Semantic ambiguity cannot be inferred; import aliases are the deterministic escape.

## Judgment-based rules

`CLAUDE.md` now has one `Code Style Invariants` bullet list covering:

- single ownership for protocol formulas, geometry/sizing laws, schedules, and state transitions;
- typed correlated state instead of sentinels or parallel options;
- append-only serialized/ordinal enums, with feature/test variants last;
- release-mode boundary checks with typed error taxonomy;
- derive-first trait implementations and local manual exceptions;
- pure/shared free functions; otherwise methods or inlining;
- no speculative public surface or lifecycle state;
- honest docs, current semantic names, and necessary-only cfg gates;
- one-pass hot trace processing and one owner for trace-sized storage;
- removal of temporary or redundant test artifacts before handoff.

These remain review rules: generic enforcement would require semantic resolution, call graphs, contract declarations, or a performance cost model.

## CI and verification

- New `style` job in `.github/workflows/rust.yml`, with full checkout history for exact merge-base diffing.
- `uv run python -m unittest discover -s scripts/tests` - 19 tests pass.
- `uv run python scripts/check_style_invariants.py --base origin/main` - passes on this branch.
- `cargo clippy` passes for both touched Rust crates; the pre-commit workspace formatting and Clippy checks pass.

`crates/jolt-field/src/solinas/ext.rs` retains one explicit exemption for its pre-existing split `cfg_attr` pairs; remove it when those sites are folded.


CLAUDE.md
```diff
@@ -241,14 +241,38 @@ Concrete implementations: `OuterRemainingSumcheckParams` (spartan/outer.rs), `Ra
 - Alias an instruction-kind enum as `Kind` at emitter call sites and write `Kind::INSTRUCTION`; never qualify emitted instructions with `SourceInstructionKind`, `JoltInstructionKind`, or a module path.
 - Before PR handoff, audit every added test and helper. Remove development-only probes, ignored tests, temporary benchmarks, diagnostic counters or histograms, and one-off fuzz or parity scaffolding. Keep permanent tests only when they add a distinct failure signal beyond existing tests, golden fixtures, or CI. If a manual diagnostic is worth keeping, make it an intentional tool or benchmark with a documented command.
 
+### Style Invariants
+
+Distilled from maintainer style passes on #1732. `scripts/check_style_invariants.py` enforces the machine-checkable subset (CI job `style`); run it locally with `python3 scripts/check_style_invariants.py --base origin/main`. Diff-scoped rules apply to added lines only — pre-existing debt stays until touched.
```
0xAndoroid: remove this line

0xAndoroid: Removed in aa7e0825a.

markosg04 (approved):

PR #1808, opened 2026-08-26
feat: field-inline — native field arithmetic on the modular stack

markosg04:
## Summary

Adds native field arithmetic to the modular Jolt stack behind `field-inline`: a 16-slot field-register file, field instructions and integer/field bridges, composed Spartan relations, and field-register commitments through Dory and packed Akita openings. Clear and BlindFold verification use the same composed symbolic claim expressions.

Memory-sourced `FIELD_LOAD_WORD` and `FIELD_LOAD_WORD_HI` instructions load one limb per VM row through the ordinary RAM/register checks. `FIELD_ADVICE_LIMB` supplies a bounded limb and a field quotient. The composed eq-polynomial guest exercises memory loads, limb readout against a fixed integer, and nontrivial inversion.

This PR contains the ISA, proving/verifying machinery, and correctness coverage. Guest-program optimization work is maintained separately.

## Design and security

- Protocol-local claim families remain separate in `jolt-claims`; composition lives at the verifier boundary. Kernels return complete typed claims through the regular batch interface. Stage 2 commits each field product opening once and reconstructs its reduction aliases.
- The ISA owns integer operand projection, shared by tracing and bytecode folds. Memory-sourced field loads retain both their integer address source and scratch destination; field-register slots are excluded from ordinary register checks. Preprocessing rejects integer write destinations of x0, matching the tracer.
- Limb advice establishes `source = limb + 2^64 * quotient` in the proof field and bounds the limb to 64 bits. It does not establish a canonical integer decomposition. Consumers requiring canonical limbs must check the reconstructed integer against the modulus.
- Protocol and payload checks precede preprocessing walks. Full preprocessing re-derives field bytecode metadata and rejects disagreement. Committed field programs remain unsupported and are rejected before metadata can be discarded.
- `FIELD_STORE_TO_X` binds the field value, integer write, and lookup operand to the RangeCheck output. The adversarial clear/ZK tests bypass the tracer and exercise malformed values against the lookup endpoint and production BlindFold constraints.

## Upstream alignment and compatibility

Includes Jolt main `95c898d3d2bbcc178fd8b18603662fce69cfa37d`. Akita dependencies use public upstream `38dbf031013a2b8f7ff93252e190dac2b3353706`, including input/round-message validation, cache-lifetime, and SIS-pricing fixes. Schedule catalogs are regenerated for that planner. No local dependency paths are required.

Field-inline arithmetic is native to the selected proof field. This is a new proof-format revision; there is no decoder fallback for pre-#1808 proofs. The memory/advice instructions extend the stage-1 field opening layout, so proofs from earlier draft versions must be regenerated. Byte parity with the updated legacy prover establishes agreement within this revision.

## Validation

All 55 CI checks pass at `a48725375`, including clear/ZK/packed field-inline proofs, verifier fixture and tamper suites, modular/legacy byte parity, workspace Clippy, field portability and formal verification, and Akita catalog regeneration.

Local checks also pass: field-inline and workspace Clippy in standard/ZK/Akita lanes; 127 verifier tests; 28 clear, 18 ZK, and 26 Akita field-inline prover tests; 23 ordinary prover-fixture tests; 11 schedule tests; formatting and style guards. All 152 catalog rows were regenerated from the public Akita pin.

No recursion-cycle improvement is claimed by this infrastructure refresh.


markosg04:
The packed axis is now complete: field-inline works under the akita commitment mode over fp128, per the ruling that akita is fp128-only. Eight commits.

FR execution is configuration-selected — the akita feature chain repoints the tracer's proof field to fp128 and the value encoding to the two-limb variant; dory keeps BN254. FieldRdInc's canonical representative commits as two u64 limb columns riding the balanced-digit machinery, recomposed by one exact linear identity; the stage-6b reduced claim feeds a stage-8 reconstruction member (mirroring the untrusted-advice one) instead of the homomorphic splice. One disposition worth review: the limb object's presence is claim-gated (an all-zero one-hot object cannot be opened by the catalogued fold schedules), enforced both ways fail-closed — Schwartz-Zippel over the reduction chain covers the soundness direction. The field-inline x akita compile error is removed; the spec status section records all of this.

Fixtures: eqpoly-field re-pinned over fp128 (host-side derivation in the test) plus the FR-inactive muldiv accept, and five wire tampers reject, including a shifted digit column recommitted through the real commit path. Also fixed in passing: two feature pairs that had never compiled or run (kernels allocative+field-inline, kernels/profiling field-inline+akita) — those gaps now have a CI lane.

Full battery green at af086bf2b, including both byte-diff ratchets, the dory FR e2e unchanged, and the new packed FR e2e. Not verified: the new CI lanes on the actual runner, and packed FR behavior above the fixture scales.

markosg04:
The packed field-inline design changed: #1798 merged, so I reworked the FR commitment onto its dense-group heterogeneous batching. This supersedes my earlier comment about the reconstruction member and claim-gated presence — both are gone.

FieldRdInc's two u64 limb-word columns now commit as an independent dense precommitted group (PrecommittedRole order [UntrustedAdvice, TrustedAdvice, FieldIncLimbs, OneHotTrace]) and open in the single native batch. The stage-6b reduced claim binds by one fail-closed linear check, v == e0 + 2^64*e1, before the packing reduction; the group is always present on FR-on builds (all-zero content opens fine — dense schedules are shape-keyed), so proof shape is uniform again. No reconstruction sumcheck, no booleanity legs, no Schwartz-Zippel presence gate. The linear check pins only the weighted sum — limb columns need not be the canonical split; that's benign since no relation consumes a limb individually, and the seam doc says so.

Provisioning rides the #1798 registry: one FR profile per reachable trace arity, worst case 124 of the 128-row cap (pinned by a sweep test); resolution never invokes the planner; FR-off provisioning is cfg-isolated. The merge itself resolved toward main's stage-8 wholesale and briefly restored the field-inline x akita gate; it comes out again in the enable commit with the new fixtures.

An agent that didn't write the code re-verified the soundness surface (absorb-before-challenge ordering, both-ways presence, trusted-shape-first allocation, strict role order) and independently recomputed the eqpoly pin (23*20*263*9 = 1,088,820, reduction-free over fp128). E2e: both accepts (nonzero two-limb-live eqpoly; all-zero-present muldiv) and six tampers reject, including a spurious duplicate FR group. 48 lanes green including both byte-diff ratchets and the dory FR fixtures unchanged at zero source diff.

Not verified: the new CI lanes on an actual runner, FR-off proof bytes diffed against the merge base directly (cfg inspection + live ratchets stand in), and the akita backend's norm-bound enforcement for the u64 envelope, which the advice words already trust identically.

markosg04:
I ran a census-gated campaign to shrink the verifier-closure footprint (specs/field-inline-verifier-surface-reduction.md, added then closed on this branch). Outcome: one small refactor landed (a018601b3 — the four FR relation impls now derive their claim fan-out and id projections from the existing derive tables, −20 lines net), and the other four mechanisms were no-gos with the census as evidence, written into the spec's Outcome section.

The useful result is the negative one: the field-inline verifier surface measures ~4,800 production code lines (docs and inline tests excluded) across jolt-verifier and its deps, and three independent censuses found that number is the floor under the protocol-separation architecture — the symbolic layer, the generic BlindFold lowering, and the geometry tables already are the compressed form, and what remains is genuine per-stage math, wire-bearing curated absorb orders, and fail-closed validation. For review purposes: ~4.6k of it concentrates in jolt-verifier + jolt-claims; jolt-r1cs, jolt-openings, jolt-poly, jolt-sumcheck, and jolt-transcript are untouched by this PR.

PR #1809, opened 2026-08-26
perf: flatten BytecodePCMapper to packed one-word slots

0xAndoroid:
## What

`BytecodePCMapper` stores per-address PC mappings as a flat `Vec<PcSlot>` — an
8-byte `#[repr(C)]` struct of `first_pc: u32` and `virtual_sequence_length: u16`
— instead of `Vec<Vec<(u16, usize)>>`. `get_pc` becomes one flat-array load plus
arithmetic instead of two dependent heap hops and an inner-list scan.

## Why

`get_pc` runs once per trace row on the witness path (PC materialization in
witness generation, the packed bytecode walk), so its cost is host-shared —
every prover backend pays it. Measured on real sha2-chain rows:

| | before | after |
|---|---|---|
| `get_pc` alone | 220.5 ns/row | 10.5 ns/row |
| full per-row extraction around it | 221.6 ns/row | 3.4 ns/row |

The index array shrinks too: 24 bytes per address (an empty `Vec` header each)
plus one heap allocation per mapped address becomes 8 bytes and no inner
allocations.

The packing is lossless because every validated address holds a single
descending inline-vsr run at consecutive bytecode indices, so
`(first_pc, virtual_sequence_length)` determines every `(address, vsr) → pc`
pair.

## How

`try_new` walks maximal runs of equal addresses (`slice::chunk_by`) rather than
individual rows, because "each such run is exactly one complete inline sequence"
is the invariant being validated. Per run: check it counts down by one to its
anchor at 0, then claim the address's slot — an already-occupied slot *is* the
non-contiguity error.

A mapped address always has at least one row, so `virtual_sequence_length == 0`
is the unmapped marker. That keeps `PcSlot::default()` as the empty slot, leaves
no reserved PC value, and folds the emptiness test into `get_pc`'s bounds check
(a zero-length slot fails `vsr < length` for every `vsr`), so the lookup is a
single comparison.

## Semantics

Four new build-time rejects. None is reachable from the expander, whose
`stamp_sequence_metadata` stamps every materialized sequence as
`len - index - 1` onto one contiguous `source.address` block; the byte-diff
ratchets against the legacy prover pass unchanged.

- **`NonContiguousInlineSequence`** — an address appears in more than one run:
  an inline sequence interleaved with other addresses, or an address revisited
  after its sequence completed. Unrepresentable in packed form; the old
  structure accepted the interleaved case silently.
- **`UnterminatedInlineSequence`** — a run stops above
  `virtual_sequence_remaining == 0`. Previously such a slot was built anyway and
  left un-queryable for every vsr below where it stopped.
- **`InlineSequenceTooLong`** — a run longer than `u16::MAX` rows. `try_new` is
  public, so the slot's own limit is enforced there rather than relying on the
  expander.
- **`BytecodeTooLarge`** — the PC counter no longer fits `u32` (previously
  unbounded `usize`).

Reserving length 0 costs one representable sequence length, so
`MAX_INLINE_ROWS_PER_SOURCE` and `MAX_METADATA_SEQUENCE_ROWS` move from
`u16::MAX + 1` to `u16::MAX`. Both were derived from the old vsr encoding and
the packed slot is now the binding limit; the largest registered inline is
`KECCAK256_INLINE` at 3260 rows, per
`jolt-inlines/fixtures/fixtures/registered_inline_expand_parity_hashes.jsonl`.

Also fixed: a row at address 0 past the leading no-op sentinel used to be
skipped silently *without* advancing the PC counter, so every PC after it was
off by one against its own index in `bytecode` — and `get_pc`'s result is
exactly that index. It is now `InvalidBytecodeAddress(0)`.

## Tests

- `cargo nextest run -p jolt-program` — 43/43, including new
  `rejects_interleaved_inline_sequences`,
  `rejects_unterminated_inline_sequences`, and
  `rejects_zero_address_outside_sentinel` cases
- `cargo nextest run -p jolt-inlines-fixtures` — 3/3 (all 96 registered inline
  expansions re-checked against the golden fixture under the new cap)
- `cargo clippy --all --features host` and `--features host,zk` with
  `-D warnings` — clean
- `cargo nextest run -p jolt-prover --features prover-fixtures` (legacy
  byte-diff ratchets) — 21/21
- `cargo nextest run -p jolt-prover-legacy muldiv --features host` — 3/3

### DIFF
--- a/crates/jolt-program/src/preprocess/bytecode.rs
+++ b/crates/jolt-program/src/preprocess/bytecode.rs
@@ -85,10 +85,11 @@ impl BytecodePreprocessing {
 }
 
 /// An address expands to a full descending `virtual_sequence_remaining` run —
-/// `first_vsr` down to 0 — at consecutive PCs, so these two bounds determine
-/// every `(address, vsr) -> pc`. `try_new` rejects any bytecode that violates it.
+/// `virtual_sequence_length - 1` down to 0 — at consecutive PCs, so these two
+/// fields determine every `(address, vsr) -> pc`. `try_new` rejects any bytecode
+/// that violates it.
 #[repr(C)]
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[derive(Default, Debug, Clone, Copy, PartialEq, Eq)]
 #[cfg_attr(
     feature = "serialization",
     derive(
@@ -99,21 +100,16 @@ impl BytecodePreprocessing {
     )
 )]
 struct PcSlot {
-    /// PC of the address's first row; `u32::MAX` marks an unmapped slot.
+    /// PC of the address's first row.
     first_pc: u32,
-    /// `virtual_sequence_remaining` of that first row — one less than the number
-    /// of rows the address expands to.
-    first_vsr: u16,
+    /// Number of bytecode rows the address expands to; 0 marks an unmapped slot,
+    /// which is why `MAX_INLINE_ROWS_PER_SOURCE` stops one short of `u16` range.
+    virtual_sequence_length: u16,
 }
-
-const EMPTY_SLOT: PcSlot = PcSlot {
-    first_pc: u32::MAX,
-    first_vsr: 0,
-};
 
 impl PcSlot {
     const fn is_empty(&self) -> bool {
-        self.first_pc == EMPTY_SLOT.first_pc
+        self.virtual_sequence_length == 0
     }
 }
 
@@ -135,9 +131,9 @@ impl BytecodePCMapper {
     pub fn try_new(bytecode: &[JoltInstructionRow]) -> Result<Self, PreprocessingError> {
         // One allocation at the final size; the no-op sentinel lives in the
         // first slot (`index_count` is always >= 1).
-        let mut slots = vec![EMPTY_SLOT; Self::index_count(bytecode)?];
+        let mut slots = vec![PcSlot::default(); Self::index_count(bytecode)?];
         if let Some(first) = slots.first_mut() {
-            first.first_pc = 0;
+            first.virtual_sequence_length = 1;
         }
 
         // The leading no-op sentinel is the only row allowed at address 0.
@@ -150,7 +146,7 @@ impl BytecodePCMapper {
         // equal addresses is exactly one inline sequence.
         let mut last_pc = 0u32;
         for run in rows.chunk_by(|a, b| a.address == b.address) {
-            let Some(first_row) = run.first() else {
+            let Some((first_row, rest)) = run.split_first() else {
                 continue;
             };
             let address = first_row.address;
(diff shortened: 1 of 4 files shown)


0xAndoroid:
FYI, this has speed up the metal implementation by a couple of %.

crates/jolt-program/src/preprocess/bytecode.rs
```diff
@@ -84,6 +84,20 @@ impl BytecodePreprocessing {
...
 
+/// The per-address mapping, packed into one word: bits 0..16 `first_vsr`,
+/// 16..32 `last_vsr`, 32..64 `first_pc` (`u32::MAX` = unmapped address; the
+/// build keeps every real pc below it). A validated address holds a single
+/// descending inline run `first_vsr, first_vsr−1, …, last_vsr` at consecutive
+/// bytecode indices `first_pc, first_pc+1, …`, so `get_pc` is one flat-array
+/// load plus arithmetic on the per-row witness path.
+type PackedSlot = u64;
+
+const EMPTY_SLOT: PackedSlot = (u32::MAX as u64) << 32;
```
moodlezoup: The slot bit layout has one owner on the pack side (`pack_slot`) but the unpack side is open-coded three times (`try_new`, `get_pc`, `get_first_pc`), `EMPTY_SLOT` spells the layout a fourth time, and "unmapped" is tested two different ways (`*slot == EMPTY_SLOT` in the build vs `first_pc == u32::MAX` in the getters) — equivalent only via the every-real-pc-below-`u32::MAX` invariant a reader has to reconstruct. Consider the packed-word-accessor shape the repo already uses (`SourceInlineKey::packed()` in `jolt-riscv/src/row.rs`, the `pack_meta` accessors in `jolt-riscv/src/trace_row.rs`): a `PackedSlot` newtype with `first_vsr()`/`last_vsr()`/`first_pc() -> Option<u32>` would give the layout one owner and fold both sentinel spellings into the accessor. At minimum, `pack_slot` is already `const fn`, so this line can be `pack_slot(0, 0, u32::MAX)`.

---

crates/jolt-program/src/preprocess/bytecode.rs
```diff
@@ -95,47 +109,92 @@ impl BytecodePreprocessing {
...
+                    address: instruction.address,
+                    previous_sequence: last_vsr,
+                    expected_sequence,
+                    new_sequence: sequence,
+                });
+            }
+            // The packed run assumes the address's bytecode indices are
+            // consecutive; an inline sequence interleaved with another
+            // address would break that silently, so reject it.
+            if last_pc - first_pc != (first_vsr - sequence) as u32 {
```
moodlezoup: This check is blind to `address == 0` rows: they're `continue`d without advancing `last_pc`, so `[A vsr=1, addr-0 row, A vsr=0]` passes both the decrement check and this one, yet the run's rows are not at consecutive indices of the bytecode vec — which is what consumers index by pc. Contiguity is enforced in pc-count space, not vec-index space; the two coincide only because the prepended noop is the sole address-0 row real producers emit. Unreachable from the expansion pipeline today, but since this comment (and the `PackedSlot` doc) now advertise consecutiveness as build-enforced, either reject address-0 rows beyond bytecode position 0 in `try_new` or scope the claim to the no-mid-stream-address-0 assumption.

---

moodlezoup (commented):
Reviewed for soundness, correctness, performance, and maintainability with parallel deep-review passes. No soundness or correctness issues: the mapper is derived from public bytecode on both prover and verifier sides (verifier dependence is digest-bound into Fiat-Shamir), the packed lookup is byte-for-byte equivalent to the old `Vec<Vec<_>>` scan on every input the old code accepted, the `BytecodeTooLarge` guard makes real-slot/`EMPTY_SLOT` collision unreachable, and all new subtractions are underflow-safe under the build invariants. The two inline comments are maintainability-level. One heads-up not tied to a diff line: the serialized shape of `pc_map` changes, so persisted preprocessing artifacts and the program digest change with it — cached `.dat` files need regeneration, and there's still no format-version field to turn a stale file into a clean error.

---

moodlezoup (approved):

PR #1810, opened 2026-08-26
feat(field): expose allocation-free coefficient access

quangvdao:
## Summary

Add allocation-free canonical coefficient construction and access to ExtField. The existing slice and Vec helpers become defaults derived from these primitives.

This is required to preserve Akita main's dense EOR tensor path while completing the shared jolt-field ownership cutover in LayerZero-Labs/akita#443.

## Validation

- cargo fmt --all --check
- focused solinas_ext_differential coefficient-access test
- jolt-field Clippy with warnings denied


crates/jolt-field/src/extension.rs
```diff
@@ -27,15 +27,36 @@ pub trait ExtField<F: Field>: Field {
     /// base coordinate directly (no full extension multiply).
     fn mul_base(self, x: F) -> Self;
 
+    /// Constructs from a base-coefficient generator.
+    fn from_base_fn<G>(f: G) -> Self
```
Acentelles, line 34: The doc for `from_base_fn` specifies neither the index range, the call order, nor exactly-once semantics, while the `FnMut` bound invites stateful generators, e.g. `let mut it = src.iter(); E::from_base_fn(|_| *it.next().unwrap())`. Today correctness rests on unpromised facts: the degree-1 impl calls `f(0)` once and ignores the rest (so that stateful caller silently truncates when monomorphized at `E = F`), `FpExt2` relies on left-to-right evaluation of `Self::new(f(0), f(1))`, and `FpExt4`/`FpExt8` on `std::array::from_fn`'s ascending order. A future impl that fills coefficients in a different order would silently permute elements, with no panic. `specs/unify-field-hierarchy.md:665` already notes this follow-up "must explicitly specify coefficient order". Suggest a one-sentence contract here: `f` is called exactly once for each index in `0..Self::DEGREE`, in ascending order. The defaulted `from_base_slice` below is load-bearing on that convention.

crates/jolt-field/src/solinas/ext.rs
```diff
@@ -552,14 +552,17 @@ impl<F: PseudoMersenne> ExtField<F> for F {
...
+    where
+        G: FnMut(usize) -> F,
+    {
+        f(0)
     }
 
     #[inline]
-    fn to_base_vec(&self) -> Vec<F> {
-        vec![*self]
+    fn base_coefficient(&self, index: usize) -> F {
```
Acentelles, line 563: The newly documented panic contract for `base_coefficient` (panics if `index >= Self::DEGREE`) is enforced two different ways: `assert_eq!(index, 0)` here (panic message "assertion `left == right` failed") vs bare `self.coeffs[index]` in the `FpExt2`/`FpExt4`/`FpExt8` impls ("index out of bounds"). No test exercises the out-of-range path, in a file that carries three dedicated `should_panic` tests for the sibling `from_base_slice` length contract. A later change, e.g. dropping this assert as apparently redundant for degree 1, or switching an array impl to `get_unchecked`, would silently void the documented contract with no failing test; the message divergence also blocks a uniform `#[should_panic(expected = ...)]` test across degrees. Suggest unifying the check (e.g. `std::slice::from_ref(self)[index]`) and adding an out-of-range `should_panic` test.

crates/jolt-field/tests/solinas_ext_differential.rs
```diff
@@ -629,6 +629,28 @@ fn degree_one_reflexive_ext_matches() {
...
+    type F = two::Prime32Offset99;
+
+    fn check<E: ExtField<F>>() {
+        let value = E::from_base_fn(|index| F::from_u64((index + 3) as u64));
+        let expected = (0..E::DEGREE)
+            .map(|index| F::from_u64((index + 3) as u64))
+            .collect::<Vec<_>>();
+
+        assert_eq!(value.to_base_vec(), expected);
+        for (index, expected) in expected.into_iter().enumerate() {
```
Acentelles: This loop is tautological: no impl overrides the new default `to_base_vec`, which is defined as `(0..Self::DEGREE).map(|i| self.base_coefficient(i)).collect()`, so these assertions re-check element by element exactly what the `to_base_vec` comparison on the line above just proved. The loop cannot fail independently. The round trip itself is already covered by `check_ext!` (line ~175) across 5 primes x 4 ext types x 48 random iterations. Meanwhile neither genuinely new contract is exercised: the documented `base_coefficient` panic, and coefficient ordering (a mutant `FpExt4::from_base_fn = Self::new(array::from_fn(|i| f(3 - i)))` paired with `base_coefficient = self.coeffs[3 - index]` stays green here). Consider folding a `base_coefficient` assertion into `check_ext!` and adding a `should_panic` out-of-range test instead of this standalone test.

Acentelles (commented):
Reviewed the trait refactor for correctness. The four rewritten impls are semantically equivalent to what they replace, the length assertions survive in the defaulted `from_base_slice`, and there are no other `ExtField` implementors in the workspace, so no compile or behavioral break. Three findings below, all addressable with small additions.

Acentelles (approved):
Minor comments from the AI. Looks good to me

PR #1818, opened 2026-08-27
refactor: remove jolt-prover-legacy

markosg04:
## Summary

Jolt still carried the 100k-line legacy prover after production paths moved to the modular stack. This finishes the cutover and removes `jolt-prover-legacy`.

## Changes

- Extract host program build, trace, and analysis APIs into `jolt-host`, then migrate the SDK, CLI, examples, evaluation tools, and formal tooling.
- Add modular Dory and Akita preprocessing for full and committed programs, including advice, and move profiling and fixtures to `jolt-prover`.
- Preserve the deployed full, Dory-committed, and Akita-committed preprocessing transcript digests byte-for-byte, then remove the legacy crate and parity scaffolding.

## Testing

- [x] Standard, Akita, and ZK prover suites and verifier tamper fixtures pass; wasm, zklean, SDK compute-advice, and dependency checks pass.

## Security Considerations

This does not change proof equations or verifier protocol logic. Snapshot tests pin all three preprocessing transcript encodings to the legacy implementation, and Akita committed-program plus advice has explicit end-to-end coverage.

## Breaking Changes

Direct users of `jolt-prover-legacy` must use `jolt-prover` and `jolt-host`. The deployed proof and preprocessing transcript formats are unchanged.


crates/jolt-verifier/tests/support/verifier_fixtures.rs
```diff
@@ -270,6 +247,43 @@ pub fn standard_committed_muldiv_case() -> VerifierFixtureCase {
...
+}
+
+#[cfg(not(feature = "zk"))]
+pub fn fresh_standard_committed_advice_case() -> VerifierFixtureCase {
+    let _guard = verifier_fixture_lock();
+    fresh_case_from_accepted_fixture(generate_committed_advice_consumer)
+}
+
+#[cfg(not(feature = "zk"))]
+fn fresh_case_from_accepted_fixture(
```
0xAndoroid, line 278: Coverage regression: legacy `initial_pc_is_constrained_to_entry_point` (prover.rs:4125 at the merge-base) was the only live test that the verifier rejects a proof when its preprocessing carries a different `entry_address`. The modular tamper manifest lists `preprocessing.program.bytecode.entry_address` as `TamperCoverage::Deferred` ("needs a real bytecode fixture mutation"), so this PR deletes the property's only test.

`fresh_case_from_accepted_fixture` is the seam for it: `prepare_guest(muldiv)`, clone `program_preprocessing` with `bytecode.entry_address += 4`, build the verifier preprocessing via `from_shared`, and assert `verify` errors on the honest proof; then flip the manifest entry from `Deferred`. The other deleted negative e2es have only partial analogs: `malicious_trace` ≈ `verifier_rejects_public_io_mismatch`, `truncated_trace` ≈ `tampered_trace_length_reject` + the output tamper targets, and `max_advice_with_small_trace` (4 KiB advice against a minimal trace) has none.

0xAndoroid (commented):
Review scope: deletion audit (what was removed / repointed), migration of every former `jolt-prover-legacy` dependent, local build + tests, and an independent byte-parity probe against the legacy prover at the merge-base.

**Verified**
- No live references to `jolt-prover-legacy` remain outside `specs/` (historical design docs) and one dead alternative in the `.semgrep` regex. Workspace members/deps, CI jobs, nextest overrides, README / CLAUDE.md / CONTRIBUTING / book / scripts / agent skills are repointed; all former dependents (jolt-sdk, jolt-prover, jolt-verifier, jolt-lookup-tables, jolt-eval, z3-verifier, zklean-extractor, jolt-profiling, 26 examples, `jolt new` templates, wasm-e2e script) build against `jolt-host` / `jolt-prover`.
- Feature map: `host ⊃ prover ⊃ minimal` gone; `transcript-*` → `jolt-transcript`; `field-inline` → `jolt-host` + `jolt-witness`; `monitor`/`pprof`/`allocative` already lived in `jolt-profiling`. Dropped without replacement (documented): the legacy `profile --name sha2|sha3` single-shot workloads, the legacy `benches/` (criterion + iai), `--features monitor` + `scripts/postprocess_trace.py` flow.
- `jolt-verifier/src` and `tracer/src` changes vs the merge-base are doc-only; the `jolt-claims` `UnbatchedRelation` refactor is value-preserving (normalized `Expr` terms; checked RAM RWC and Shift by hand) and exists to feed `zklean-extractor`.
- Local: `cargo build --workspace` (only the `*-guest` bins fail to link on host — pre-existing `_main`), `clippy --all --features host` and `host,zk` with `-D warnings` clean, `nextest -p jolt-prover --features prover-fixtures` 13/13, `-p jolt-verifier --features prover-fixtures` green except the digest probes discussed inline, `-p jolt-host -p jolt-sdk -p jolt-lookup-tables -p jolt-transcript` 485/485.
- Byte parity: the legacy `RV64IMACProver` at merge-base `756bddce3` produces exactly HEAD's proof bytes for muldiv, committed muldiv (2 chunks) and advice-consumer (Blake2b-256 of the bincode wire encoding). Details and the values in the `standard.rs` comment.

**Not attachable inline (files untouched by the PR)**
- `book/src/dev/testing-gates.md:56-62` still tells readers to run `prover_matches_legacy_on_committed_muldiv_akita` "byte-parity gate"; that test went with `akita_byte_diff.rs`. Drop the command and the byte-parity wording.
- Root `Cargo.toml` `[workspace.dependencies] inferno` is now unreferenced (legacy's flamegraph dep).
- `book/src/usage/troubleshooting.md` deletes the whole fat-LTO miscompile section. The repro was legacy-specific, but the workaround (`lto = "thin"` default, `-C no-prepopulate-passes` for fat LTO) still applies to this workspace; keep a generic version.
- `crates/jolt-kernels/src/**` doc comments cite legacy symbols as if navigable (`RaPolynomial`, `SharedRaPolynomials`, `BooleanityAddressSumcheckProver`, `zkvm/claim_reductions/ram_ra.rs`, `compute_rs2_ra_claim`, ...). Follow-up sweep, or point them at a SHA.
- Akita wire parity was not re-verified here: `akita_byte_diff.rs` is deleted and #1817 changed the packed format on main, so the dropped akita digests were stale for a legitimate reason.

Review produced with Claude Code on Andrew's behalf.

.github/workflows/rust.yml
```diff
@@ -219,8 +215,8 @@ jobs:
         run: cargo install --path . --locked --force --profile ci --target-dir target
       - name: Install nextest
         uses: taiki-e/install-action@nextest
-      - name: Run jolt-prover-legacy tests
-        run: cargo nextest run --cargo-profile ci -p jolt-prover-legacy
+      - name: Run jolt-prover tests
```
0xAndoroid, line 204: With the byte_diff arms deleted and the zk/akita e2es behind their own features, `-p jolt-prover --features prover-fixtures` is 13 unit tests (13 passed, 2 skipped locally) — no proving. This job still installs the RISC-V toolchain and the `jolt` CLI for them, and the clear-mode e2e proving now lives only in `test-verifier-fixtures`. Either drop the toolchain/CLI setup here, or move the clear-mode prover e2es into this job so its name matches what it runs.

jolt-sdk/macros/src/lib.rs
```diff
@@ -70,7 +70,6 @@ impl MacroBuilder {
         let trace_to_file_fn = self.make_trace_to_file_func();
         let compile_fn = self.make_compile_func();
         let preprocess_shared_fn = self.make_preprocess_shared_func();
```
0xAndoroid, line 72: SDK surface removed without a note in Breaking Changes (nothing in-repo used it, but it was `pub`): the generated `preprocess_shared_committed_<fn>`; `preprocess_verifier_<fn>` now takes `Option<jolt::BlindfoldSetup>` (was `BlindfoldSetup<Curve>`); `JoltProverPreprocessing` lost its `<F, Curve, PCS>` generics; and the `jolt_sdk` re-exports `RV64IMACProver`, `guest`, `Curve`/`JoltCurve`, `Serializable`, `DoryGlobals`/`DoryContext`/`DoryLayout`, `MultilinearPolynomial`, `populate_memory_states`, `ProofCommitmentScheme` are gone (`AdviceTape` is gone for `guest-verifier`-only builds). Either list these in the PR's Breaking Changes or keep `preprocess_shared_committed_<fn>` as a wrapper over `preprocess_program(.., Some(chunks))`.

crates/jolt-prover/Cargo.toml
```diff
@@ -102,19 +90,18 @@ tracing.workspace = true
 
 # Optional: the `profiling` harness (bins cannot use dev-dependencies).
 allocative = { workspace = true, optional = true }
-bincode = { workspace = true, optional = true }
+bincode.workspace = true
```
0xAndoroid: `bincode` was optional under `profiling`; it is now unconditional, but the only non-test use is still `profile.rs` (`profiling`-gated) — the other use is the `#[cfg(test)]` round-trip in `preprocessing.rs`. Every consumer (SDK, wasm build) now pulls it. Keep `bincode = { workspace = true, optional = true }` + `dep:bincode` in `profiling`, and add `bincode.workspace = true` under `[dev-dependencies]` for the test.

crates/jolt-prover/src/dory/preprocessing.rs
```diff
@@ -0,0 +1,439 @@
...
+        1usize << CommitmentMatrixShape::balanced(bytecode_total_vars).column_vars();
+    let program_image_words = program
+        .meta
+        .program_image_len_words
+        .next_power_of_two()
+        .max(2);
+    let program_image_columns = 1usize
+        << CommitmentMatrixShape::balanced(program_image_words.ilog2() as usize).column_vars();
+    let max_log_t = program.max_padded_trace_length.next_power_of_two().ilog2() as usize;
+    let max_log_k_chunk = if max_log_t >= ONEHOT_CHUNK_THRESHOLD_LOG_T {
```
0xAndoroid: Third copy of the `log_T >= ONEHOT_CHUNK_THRESHOLD_LOG_T ? 8 : 4` rule (`config.rs::one_hot_config`, here, `setup_total_vars` below), and this one is hashed into `preprocessing_digest`. If the threshold rule moves in `config.rs`, the digest silently stops describing the prover's actual chunking. Expose one `pub(crate) fn committed_log_k_chunk(log_T: usize) -> u8` next to `one_hot_config` and call it from all three sites.

jolt-eval/README.md
```diff
@@ -33,7 +33,7 @@ The motivation is twofold:
 
 | Objective | Units | Description |
 |---|---|---|
-| `lloc` | lines | Total logical lines of code in `crates/jolt-prover-legacy/src/` |
+| `lloc` | lines | Total logical lines of code in `crates/jolt-prover/src/` |
```
0xAndoroid: `lloc` (and `cognitive_complexity_avg` / `halstead_bugs`) now walk `PROOF_SYSTEM_CRATE_DIRS` — 19 crates — not `crates/jolt-prover/src/`. Fix the table row; the `description()` string in `lloc.rs` already says "the modular proof system".

jolt-sdk/src/host_utils.rs
```diff
@@ -110,11 +232,139 @@ pub fn serialize_and_print_size(
...
+        program_size: Some(memory_layout.program_size),
+    };
+    let trace_inputs = TraceInputs::new(
+        inputs.to_vec(),
+        untrusted_advice.to_vec(),
+        trusted_advice.to_vec(),
+        memory_config,
+    )
+    .with_advice_tape(advice_tape);
+    #[cfg(not(feature = "field-inline"))]
```
0xAndoroid, line 312: Follow-up, not a blocker: `prove_program` interleaves three `#[cfg(feature = "field-inline")]` / `not(...)` pairs (trace, `ProverConfig`, witness) with the shared code. One cfg-gated helper returning `(trace, config, witness)` would leave a single straight-line proving path here.

crates/jolt-transcript/src/legacy.rs
omibo, line 1: Should we rename legacy.rs or it is intended to be removed?

crates/jolt-prover/tests/dory_byte_diff.rs
mtbadakhshan, line 1: Should we add dory_e2e.rs for jolt-prover? similar to what we currently have for akita_e2e.rs?

moodlezoup: This should be addressed by https://github.com/a16z/jolt/pull/1904 but lmk if you have something more in mind

moodlezoup (approved):

PR #1824, opened 2026-09-01
ci: same-runner A/B comparison for crate benchmarks

moodlezoup:
## Problem

The bench comparison this workflow posts is dominated by runner noise, not code changes. From the last 15 PRs with a `<!-- bench-crates -->` comment, 10 touched **none** of the four benched crates, yet 9 of those got flagged rows — 380 spurious rows total, median ~46 flagged benches per PR out of a 62-ID suite. Spurious ratios: median 1.14x, p90 1.34x, max 1.87x, with 61/67 bench names flagged in *both* directions across PRs (e.g. `pairing` 1.55x slower on #1705, 1.20x faster on #1748). The signature is a whole-suite uniform shift (#1790: all 51 rows in a 1.25–1.34x band): baseline and PR runs land on different VMs from a heterogeneous CPU fleet (`fp64 packed_mul` measures 1.5ns on one run and 2.8ns on another). The baseline artifact is also stale by construction — one Aug 27 artifact served every PR for 4+ days, and it tracks latest main, not the PR's merge target.

Criterion's 3s default warmup applies to every bench (no custom configs anywhere), so warmup was ruled out as the cause.

## Fix: bench base and head back-to-back on the same runner

The PR job now checks out the merge commit's first parent (the exact main tip the PR merges against), benches it as `base_run`, checks out the merge commit, benches `pr_run`, and compares locally with critcmp. Same-VM back-to-back runs measure ~1.5–3% drift vs the 14%/34% cross-VM offsets, so the comparison becomes meaningful.

Details:

- **PR-only trigger** — the push-to-main baseline job, artifact upload/download, and the `dawidd6/action-download-artifact` dependency are gone. Existing `criterion-baseline` artifacts age out via their 30-day retention.
- **Paths narrowed** to the four benched crates + `jolt-utils` (their only in-repo dependency, per `cargo metadata`) + `Cargo.toml`/`Cargo.lock`/`rust-toolchain.toml` + the workflow itself. Most spurious comments were on PRs that could not have changed these numbers.
- **Threshold 5% → 10%** — the 11 `poly_ops` benches are rayon-parallel on 4 shared vCPUs and still see within-job tenancy drift; 10% keeps them quiet while cross-VM offsets no longer apply. The first validation run on this PR confirmed the design: **zero** of the 62 benchmarks differed by more than 10% between the two same-VM legs, where the old cross-VM comparison flagged ~46 rows on a median unrelated PR.
- **critcmp replaced by `scripts/ci/bench_comment.py`** — the comment is now a Markdown table instead of a critcmp code fence: the meaningless `? ?/sec` throughput column is gone (no bench declares a Criterion throughput), the faster run is **bolded** per benchmark, and entries whose delta exceeds the noise threshold are colored green (PR faster) / red (PR slower) via GitHub's math rendering (the only inline text color GFM supports; verified against the markdown API, including the `\%`-in-math escaping gotcha). Significant rows are listed up top sorted by magnitude; the full table is collapsed in a `<details>` block. This also removes critcmp's exit-1-on-quiet behavior, which failed the first validation run.
- **`RUSTFLAGS: -C target-feature=+bmi2,+adx,+avx2`** — with no target-feature config, the fp128 `mulx`/`adcx` asm specialization and the packed SIMD paths were compile-time dead code in CI, so the "packed"/asm-oriented benches timed scalar fallbacks. Every CPU observed in the hosted x64 fleet (Xeon 8370C/8573C, EPYC 7763/9V74) supports all three features. Verified: `cargo check --benches` for all four crates and a full-codegen `cargo build` of `jolt-field --features solinas,asm` both pass for `x86_64-unknown-linux-gnu` with these flags.
- **`--noplot`** — nothing consumes the HTML reports; skipping plotting saves time in each leg.
- **`timeout-minutes: 90`** backstop. The job roughly doubles (~14 → ~28–30 min warm; the first run is cold because RUSTFLAGS changes the rust-cache key).
- The sticky comment keeps its marker and now names both SHAs and the threshold.
- A non-merge `HEAD` (e.g. conflicted PR) fails loudly at the resolve step instead of benching the wrong ref.

This PR modifies the workflow file, which is itself in the new path filter — so the run on this PR exercises the new A/B flow end-to-end.

## Not in this PR

Instruction-count gating (CodSpeed simulation mode or wiring up jolt-eval's existing `callgrind:*` objectives) would get the noise floor to ~0.00001% and is the natural next step; this change fixes the comparison methodology without new dependencies or services.

### DIFF
--- a/.github/workflows/bench-crates.yml
+++ b/.github/workflows/bench-crates.yml
@@ -1,17 +1,27 @@
 name: Benchmark modular crates
 
 on:
-  push:
-    branches: [main]
-    paths:
-      - "crates/**"
   pull_request:
     branches: [main]
+    # Only paths that can change the benched code: the four benched crates,
+    # their sole in-repo dependency (jolt-utils), and the build/toolchain inputs.
     paths:
-      - "crates/**"
+      - "crates/jolt-crypto/**"
+      - "crates/jolt-field/**"
+      - "crates/jolt-poly/**"
+      - "crates/jolt-transcript/**"
+      - "crates/jolt-utils/**"
+      - "Cargo.toml"
+      - "Cargo.lock"
+      - "rust-toolchain.toml"
+      - ".github/workflows/bench-crates.yml"
 
 env:
   CARGO_TERM_COLOR: always
+  # Every CPU observed in the hosted x64 fleet (Xeon 8370C/8573C, EPYC 7763/9V74)
+  # supports BMI2+ADX+AVX2. Without these flags the fp128 mulx/adx asm and the
+  # packed SIMD kernels are compile-time dead code and the benches time fallbacks.
+  RUSTFLAGS: -C target-feature=+bmi2,+adx,+avx2
 
 concurrency:
   group: bench-crates-${{ github.head_ref || github.ref }}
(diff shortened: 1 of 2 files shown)


0xAndoroid:
Pushed 139658cd7 on top of 5bef8804a (no force-push):

- `bench_comment.py`: only the Δ cell is colored now — the faster **base** value turned red on a regression; the TeX time-cell branch and TeX unit suffixes are gone, and tiny negatives render as `+0.0%` instead of `-0.0%`.
- Comment header shows the PR head SHA (`HEAD^2`) instead of the synthetic merge-commit SHA; the PR leg still benches the merge commit.
- The two identical `cargo bench` blocks are one looped step (`::group::` per leg keeps the log split), with a comment on the base-first order effect: on the validation run the second leg came out ~0.2% faster (median; 42 of 57 non-zero rows) — fine at 10%, but don't tighten the threshold without an A/B/A order.
- `permissions: contents: read, pull-requests: write`; `listComments` is paginated so the sticky comment is found on PRs with >30 comments.

Checked: actionlint + shellcheck clean, py_compile, synthetic-fixture smoke run of the renderer. CI run: https://github.com/a16z/jolt/actions/runs/33557164819

0xAndoroid (approved):

PR #1826, opened 2026-09-02
fix(z3): model RV64 word right shifts

### LINKED ISSUE #1803: Nightly Z3 formal verification failed

zakazaka95:
## Summary

Closes #1803. The nightly verifier aborted because `symbolic_exec` had no models for the RV64 word right-shift expansion steps, so this adds the missing sound models instead of treating those instructions as unsupported.

## Changes

- Model register and immediate arithmetic/logical word right shifts with 32-bit extraction and RV64 sign extension.
- Model the word-sized right-shift bitmask, including reduced-bit verifier mode.

## Testing

- Exact eight failing Z3 checks: 8/8 pass in default 64-bit mode and 8/8 with `Z3_VERIFIER_BV_BITS=8`.
- Full `z3-verifier` suite: 104/104 pass; 20 existing solver-heavy tests skipped.
- Scoped strict Clippy and rustfmt pass. The two workspace-wide host Clippy commands were also attempted, but an existing Windows-only `zeroos-vfs-core` POSIX-constant error stops them before this crate.

## Security Considerations

This restores formal verification coverage for five RV64 word-shift virtual instructions. The models preserve RISC-V word semantics: operate on the low 32 bits, mask the shift amount, and sign-extend the 32-bit result to XLEN.

## Breaking Changes

None

### DIFF
--- a/z3-verifier/src/virtual_sequences.rs
+++ b/z3-verifier/src/virtual_sequences.rs
@@ -608,8 +608,7 @@ fn symbolic_exec(instr: &Instruction, cpu: &mut SymbolicCpu) {
             // Bitmask immediate: compute trailing_zeros, then scale the shift amount
             let shift_amt = operands.imm.trailing_zeros();
 
-            // Word instructions (SRLIW, SRAIW) encode as (base_shift + 32)
-            // Decompose shifts >= 32 to handle this pattern
+            // Preserve RV64 shifts of 32..63 across reduced verifier widths.
             let scaled_shift = if shift_amt >= 32 {
                 let base = (shift_amt - 32) & (cpu.word_bits - 1);
                 (cpu.word_bits + base) as u64
--- a/z3-verifier/src/virtual_sequences.rs
+++ b/z3-verifier/src/virtual_sequences.rs
@@ -78,11 +78,16 @@ use tracer::{
         virtual_shift_data_h::VirtualShiftDataH,
         virtual_shift_data_w::VirtualShiftDataW,
         virtual_shift_right_bitmask::VirtualShiftRightBitmask,
+        virtual_shift_right_bitmask_w::VirtualShiftRightBitmaskW,
         virtual_sign_extend_word::VirtualSignExtendWord,
         virtual_sra::VirtualSRA,
         virtual_srai::VirtualSRAI,
+        virtual_sraiw::VirtualSRAIW,
+        virtual_sraw::VirtualSRAW,
         virtual_srl::VirtualSRL,
         virtual_srli::VirtualSRLI,
+        virtual_srliw::VirtualSRLIW,
+        virtual_srlw::VirtualSRLW,
         virtual_window_mask_b::VirtualWindowMaskB,
         virtual_window_mask_h::VirtualWindowMaskH,
         virtual_window_mask_w::VirtualWindowMaskW,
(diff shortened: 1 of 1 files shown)


0xAndoroid (commented):
Reproduced the fix locally on `b43f6ead6` (system Z3 5.1.0):

| Check | Result |
|---|---|
| `cargo fmt --check` (z3-verifier) | clean |
| `cargo clippy --all-targets -- -D warnings` (z3-verifier) | clean |
| Nightly step 1: full suite, 64-bit | 104 passed, 20 skipped |
| Nightly step 2: DIV/REM ignored set, `Z3_VERIFIER_BV_BITS=8` | 8 passed |
| Extra: full non-ignored suite at `Z3_VERIFIER_BV_BITS=8` | 104 passed |

The nightly builds vendored Z3 5.0.0 rather than 5.1.0, but the eight new obligations each discharge in under 100 ms, so the solver version cannot change the verdict.

**Model review.** Each of the five new `symbolic_exec` arms matches the tracer `exec` and the `jolt-lookup-tables` reference functions on every input the expansions can produce:

- `VirtualShiftRightBitmaskW`: `(1 << word_bits) - (1 << (rs1 & (word_bits-1)))` is the tracer's `(1<<32) - (1<<(rs1 & 0x1f))` at width 64.
- `VirtualSRLW` / `VirtualSRAW`: trailing zeros of the *low word* of rs2, where the tracer uses the full 64-bit register. These agree because `VirtualShiftRightBitmaskW` always emits a mask below 2^32 with a nonzero low word.
- `VirtualSRLIW` / `VirtualSRAIW`: at width 64 the `& (word_bits-1)` is the identity since the expansion's shamt is already `imm & 0x1f`; at reduced width it folds the shamt mod `word_bits`, the same convention the existing spec closures and the `VirtualSRLI` model use.

**One nit.** The comment above the `VirtualSRLI` model is stale now that SRLIW/SRAIW have their own arms:

https://github.com/a16z/jolt/blob/b43f6ead6fdb8911f99111dfb25e2d6a1631b0e5/z3-verifier/src/virtual_sequences.rs#L611-L612

The `shift_amt >= 32` branch is still needed for RV64 `SRLI` with shamt 32..63 under reduced width, so only the comment should change. Worth fixing here since it now sits next to the dedicated `VirtualSRLIW` arm.

**Post-merge.** Trigger the nightly via `workflow_dispatch` rather than waiting for the scheduled run: the failure-issue step dedupes by title, so nothing else will confirm the fix.

z3-verifier/src/virtual_sequences.rs
```diff
@@ -600,6 +624,11 @@ fn symbolic_exec(instr: &Instruction, cpu: &mut SymbolicCpu) {
             cpu.x[operands.rd as usize] =
                 cpu.sign_extend(&cpu.unsigned_data(&rs1).bvlshr(scaled_shift));
         }
+        Instruction::VirtualSRLIW(VirtualSRLIW { operands, .. }) => {
```
0xAndoroid, line 626: With this arm in place, the comment on the `VirtualSRLI` model above is stale:

https://github.com/a16z/jolt/blob/b43f6ead6fdb8911f99111dfb25e2d6a1631b0e5/z3-verifier/src/virtual_sequences.rs#L611-L612

SRLIW/SRAIW no longer route through `VirtualSRLI`. The `shift_amt >= 32` branch there is still needed for RV64 `SRLI` with shamt 32..63 under reduced width, so only the comment should change (e.g. "RV64 shifts >= 32 map onto the reduced-width word boundary"). Anchored here because GitHub only allows inline comments on diff lines.

zakazaka95: Updated the stale comment in `8719397`; it now describes RV64 shifts 32..63 under reduced verifier widths. No behavior changed.

0xAndoroid (approved):
LGTM, thanks a lot.

PR #1830, opened 2026-09-03
refactor(kernels): retire the scalar-table visit helpers

### LINKED ISSUE #1805: jolt-kernels: retire the scalar-table visit helpers once the akita-field cutover lands
Follow-up from #1732 (Allocative derive refactor in jolt-kernels).

## Problem

Kernel structs derive `Allocative` under `#[allocative(bound = "F: JoltField")]`, and every field naming the bare generic `F` — `Vec<F>`, `Vec<Vec<F>>`, `[Vec<F>; N]`, **86 sites** — carries `#[allocative(visit = jolt_poly::visit_scalars)]`.

All of it exists for one reason: `Allocative` cannot be implied by `JoltField`.

The clean form is a supertrait:

```rust
pub trait JoltField: Field + CanonicalEncoding + WithAccumulator + MaybeAllocative {}
```

Then `F: JoltField ⟹ F: Allocative`, the derive's default bound is satisfied at every call site, and the attributes plus `jolt_poly::visit_scalars`/`visit_scalar_rows` delete themselves.

It compiles for BN254 and jolt-field's own Solinas types. It fails on the akita bootstrap edge:

```
$ cargo check -p jolt-akita --features jolt-field/allocative
error[E0277]: the trait bound `Fp128<340282366920938463463374607427473266697>: Allocative` is not satisfied
    = note: required for `akita_field::Fp128<…>` to implement `jolt_field::algebra::MaybeAllocative`
    = note: required for `akita_field::Fp128<…>` to implement `JoltField`
```

`AkitaField` is `akita_config::proof_optimized::fp128::Field` — foreign, so no `Allocative` impl for it can exist in this workspace (orphan rule). The supertrait does not error in jolt-field; it silently strips `AkitaField` of its blanket `JoltField` impl and jolt-akita stops compiling.

The direct form is a non-starter for the same reason one level up:

```
$ # impl<F: JoltField> allocative::Allocative for F { … }
error[E0210]: type parameter `F` must be used as the type parameter for some local type
    = note: only traits defined in the current crate can be implemented for a type parameter
```

A `where` bound on a local trait does not satisfy coherence — only the `allocative` crate could write that impl.

Field elements are heap-free, so the *semantics* were never in question; one foreign scalar type is the whole blocker.

## Trigger

`crates/jolt-field/src/akita.rs` is explicitly temporary:

> Temporary bootstrap adapter for the pre-cutover `akita-field` type. […] it is a bootstrap edge, not the target architecture, and is removed in the final migration PR together with the `akita` feature.

Its replacement — jolt-field's own `solinas::Fp128` — already derives `Allocative` (`crates/jolt-field/src/solinas/fp128.rs:69`). Once the cutover removes the foreign scalar this unblocks with no further upstream work.

## Work

- Add `MaybeAllocative` to `JoltField`'s supertraits in `crates/jolt-field/src/algebra.rs` (cfg-gated blanket, same shape as `jolt_kernels::backend::MaybeAllocative`).
- Delete `crates/jolt-poly/src/allocative_support.rs`, its re-exports, and the `jolt-poly/allocative` feature.
- Strip the 86 `F`-naming `#[allocative(visit = …)]` attributes and the 42 `allocative(bound = "F: JoltField")` / 9 `allocative(bound = "F")` attributes from jolt-kernels.

What stays, and why:

- 20 `visit` attributes whose *element* type is foreign or unimplemented (`Vec<JoltOpeningId>`, `Vec<PrefixEval<F>>`, `Vec<StageVal>`, the row newtypes). Deriving `Allocative` on those would cascade into jolt-claims, jolt-lookup-tables, and jolt-witness for types that own no heap.
- The three foreign-keyed container helpers in `crates/jolt-kernels/src/backend.rs` (`visit_keyed_polys`, `visit_poly_map`, `visit_scalar_map`) — `LookupTableKind` and the `JoltOpeningId`-keyed maps have the same cascade problem.
- `bound` on `PrecommittedReductionCarry<F, R>`, `CycleReductionKernel<F, R>`, `AddressReductionKernel<F, R>`, `NaiveSumcheckProver<F, R>`, `BundleStore<B>`, and `LazyFoldedRa<F, S>`, where the derive would otherwise demand `Allocative` of a generic that is not a field.

## Alternative

One `impl Allocative for Fp128` behind an optional feature upstream in `LayerZero-Labs/akita` unblocks this today, but the dependency is pinned to rev `1d48114e6f0aefd1384a57be167037c57de99d22`, so it costs a coordination round-trip for something the cutover removes anyway.

moodlezoup:
## Summary

Closes #1805.

The akita-field cutover (#1796) deleted `crates/jolt-field/src/akita.rs` and with it the last foreign scalar reaching `JoltField`. That was the only thing blocking the supertrait, so this PR takes it:

```rust
pub trait JoltField:
    Field + CanonicalEncoding + WithAccumulator + Serialize + DeserializeOwned + MaybeAllocative
```

`MaybeAllocative` is the same cfg-gated blanket shape as `jolt_kernels::backend::MaybeAllocative`. With `F: JoltField ⟹ F: Allocative`, every field-generic container renders through allocative's native impls and the hand-written byte-sizing visitors delete themselves.

| | before | after |
|---|---|---|
| `allocative(visit = …)` attributes | 110 | 21 |
| `allocative(bound = …)` attributes | 60 | 9 |
| visit helpers | 4 | 2 |

Net **+165 / −498** across 43 files.

## What went, and what stayed

**Deleted:** `crates/jolt-poly/src/allocative_support.rs` (`visit_scalars`, `visit_scalar_rows`) and `jolt_kernels::backend::visit_scalar_pairs` — allocative 0.3.6 has native `Vec<T>`, `[T; N]`, and tuple impls, so every scalar shape including `[(Vec<F>, Vec<F>); 4]` covers itself. jolt-poly's three hand-rolled container visitors (`Polynomial<T>`, `TensorEqTable`, `GruenSplitEqPolynomial`) became plain derives.

**Kept — 21 `visit` attributes** whose *element* type owns no heap but has no `Allocative` impl: `Vec<JoltOpeningId>`, `Vec<PrefixEval<F>>`, `Vec<StageVal>`, the row newtypes, and the two `LookupTableKind`-keyed tables. Deriving `Allocative` on those cascades into jolt-claims, jolt-lookup-tables, and jolt-witness for types that own no heap. Their visitor moves next to `visit_keyed_polys` in jolt-kernels' backend as **`visit_heap_free_elements`** — the old name stopped being accurate once the elements stopped being scalars.

**Kept — 9 `bound` attributes** on generics that are not fields: `PrecommittedReductionCarry`/`CycleReductionKernel`/`AddressReductionKernel`'s relation marker `R`, `BundleStore<B>`'s row type, `LazyFoldedRa`'s chunk source `S`, jolt-field's extension towers, and the legacy prover's `bound = ""`. Each was confirmed load-bearing by trial deletion — removing them all reintroduces exactly these errors and no others.

One bound was **narrowed** rather than dropped: `PrecommittedReductionCarry`'s `"F, R"` → `"F: JoltField, R"`, which now discharges `F: Allocative` through the supertrait while leaving `R` unbounded.

## Deviations from the issue text

Three things the issue anticipated differently:

1. **`jolt-poly/allocative` cannot be deleted.** It is no longer there for the helper module, but jolt-poly's own container derives still need `dep:allocative`. It survives.
2. **Feature wiring was required.** Only `jolt-prover/allocative` forwarded `jolt-field/allocative`; `jolt-poly` and `jolt-kernels` did not. Without the forward the supertrait is vacuous while the derives are live, so `cargo check -p jolt-kernels --features allocative` fails. Both manifests now forward.
3. **jolt-field needed an `allocative` dev-dependency.** `tests/spine.rs`'s oracle field `M61` must satisfy the widened bundle, and integration tests do not see optional regular dependencies.

`jolt-kernels` deliberately keeps its **own** `MaybeAllocative` rather than re-exporting jolt-field's: the two are gated on different features, and `cargo check -p jolt-kernels --features jolt-field/allocative` (kernels' own feature off) must keep compiling — a re-export would demand `Allocative` of kernels whose derives are cfg'd out.

## Heap accounting

Bytes are unchanged. Both paths report `capacity * size_of::<F>()` and both count unused capacity — allocative's `visit_iter` takes an O(1) `visit_simple` fast path for `!needs_drop` element types. Verified against the real `FlameGraphBuilder`:

```
Polynomial<Fr> (len 512, cap 1024)
  …;evals;Vec<Fr>;ptr;jolt_field::bn254::Fr   16384
  …;evals;Vec<Fr>;ptr;unused_capacity         16384   = 32768 heap
  warnings: none
```

Reported *totals* rise by the container headers (24 B per converted field, plus outer-Vec slack for `Vec<Vec<F>>`) that the helpers' oversized inline child previously swallowed via a negative `rem_size` — which also retires the "Incorrect size declaration" warnings the helpers emitted at every site. Frame labels now name the element type instead of a flat `elements`. `parse_folded`'s `roots` keys are unaffected, and there are no golden `.folded` fixtures; `telemetry:*:heap:*` values shift on the order of 10⁻⁶.

## Validation

- `cargo clippy --all --all-targets` — `host` ✅, `host,zk` ✅
- All four CI allocative lanes: `--all …--features allocative,host` ✅, `allocative,host,zk` ✅, `-p jolt-prover …profiling,allocative` ✅, `-p jolt-prover …akita,profiling,allocative` ✅
- `cargo fmt --all --check`, `git diff --check`, `scripts/check_style_invariants.py --base origin/main` ✅
- `cargo nextest run -p jolt-field -p jolt-poly -p jolt-kernels --features allocative` — 347 passed
- muldiv e2e: `host` 3/3 ✅, `host,zk` 3/3 ✅
- Modular clear acceptance: **21/21**, including all four byte-diff ratchets against the legacy prover
- Feature-matrix checks incl. `jolt-field --no-default-features --features allocative,solinas` and the mismatched `-p jolt-kernels --features jolt-field/allocative`

**Note on `jolt-prover --features prover-fixtures,zk`:** this suite fails locally for me with `ThreadPoolBuildError { IOError(EAGAIN) }` at `zk_e2e.rs:300`. It fails *identically on unmodified `origin/main`*, so it is pre-existing and unrelated to this PR — my sandbox can't run 14 concurrent 128 MB-stack ZK proofs. Each test passes in isolation. Worth noting separately: `.config/nextest.toml` serializes `package(jolt-verifier) and test(zk)` into `zk-core-fixtures`, but `jolt-prover`'s `zk_e2e` uses the same wide-stack pattern and is not in any serial group.


quangvdao:
I traced the `JoltField`/`Allocative` history more closely. The new supertrait is clearly intentional: issue #1805 planned this exact change once #1796 removed the foreign Akita scalar. Before this PR, kernel objects had a separate `MaybeAllocative` requirement while scalar tables bypassed `F: Allocative` through manual size visitors. I therefore don’t consider the widened `JoltField` bundle a blocker.

I found two remaining cleanup items:

1. The [`SumcheckKernel` documentation](https://github.com/a16z/jolt/blob/7c375837120ac467440163529da8843ad9995b6e/crates/jolt-kernels/src/kernel.rs#L66-L72) still instructs implementers to use manual size arithmetic so `F` remains unbounded. That describes the design this PR removes. Please update it to recommend native derives for field containers and reserve `visit_heap_free_elements` for heap-free types without `Allocative`.

2. The additional [`allocative` dev-dependency and cargo-machete exemption](https://github.com/a16z/jolt/blob/7c375837120ac467440163529da8843ad9995b6e/crates/jolt-field/Cargo.toml#L11-L42) appear redundant. When `jolt-field/allocative` is enabled, the existing optional normal dependency is activated and is available to the integration test. Please remove the duplicate dev-dependency and exemption if `cargo check -p jolt-field --tests --features allocative` still passes.

It would also be helpful to update `specs/jolt-field-rebuild.md`, whose documented `JoltField` bundle and third-party-field recipe still describe the pre-PR contract.

Overall, the intended cleanup looks sound; these are non-blocking documentation and manifest follow-ups.

0xAndoroid (approved):
i'm fixing a couple of nits, but otherwise lgtm

0xAndoroid (approved):

PR #1845, opened 2026-09-04
perf(inlines): fuse keccak rho/theta-apply XOR+ROTL steps (3,087 -> 2,511 rows/perm)

0xAndoroid:
## What

Fuses the XOR+ROTL lane-steps in the keccak-f inline — the rho/pi rotations and the theta-apply XORs — using 23 new variants of the existing generic XOR-ROT lookup construction (same table shape, no new polynomial form). Builds directly on the scheduling/absorb work from [#1749](https://github.com/a16z/jolt/pull/1749) and keeps its theta-D layout intact.

**3,087 → 2,511 rows per permutation (−18.66%).** Fused absorb path: 3,121 → 2,545.

The added table variants overflowed the 6-bit packed table-ID field; this PR widens it to 7 bits by taking one PC bit (56 → 55). Packed row stays 32 B clear / 40 B Akita; RAF/sign keep their top bits. The existing max-table roundtrip test is strengthened to max-table-ID × max-packed-PC in both clear and Akita modes. Tag assignment accounts for the reserved range 0x009e–0x00a0 filled by [#1768](https://github.com/a16z/jolt/pull/1768).

Not attempted here: one-row theta parity and one-row chi — both need more than two register inputs per lookup, which the current construction doesn't support. Chi already uses the ANDN form.

## Verification

- Legacy suite: 613 clear + 626 ZK tests green.
- Modular prover acceptance: 21 clear + 14 ZK green, including the inline-bearing guest.
- Inline/table/tracer suites, ABI checks, packed-row roundtrips (clear + Akita), 5-guest × 5-config trace parity: green.
- Golden keccak vectors bit-exact; workspace clippy/fmt/style both modes; Linux x86 cross-check (runtime execution left to CI).
- Rows/perm is the claimed improvement; prover wall-time is not measured here.

## Measured downstream impact

On a stateless Ethereum mainnet block-validation guest (jeth), a fixed 10-block sweep (25905781–25905790):

- Gas-weighted self-verifying cost: **21.685 → 20.094 cycles/gas (−7.34%)**.
- Keccak permutation counts unchanged on all 10 blocks; per-block row savings equal exactly 576 × permutation count (residual zero).


0xAndoroid:
Closing: the row savings don't justify adding this many lookup tables / instructions. Keccak acceleration will go through a batched precompile design instead.

PR #1846, opened 2026-09-05
docs(agents): clarify authorization and verification scope

0xAndoroid:
## Summary

Model-neutral guidance changes after auditing the [official GPT-6 Astra guide](https://developers.openai.com/api/docs/guides/latest-model?model=gpt-6-astra).

- `CLAUDE.md` (also read through the tracked `AGENTS.md` symlink): explicit user/skill precedence, continued work within authorized scope, evidence-first searches, independent read batching, sequential Cargo, concise results, and bounded verification.
- `implement-spec`, `new-invariant`, `new-objective`, `update-docs`, and the exported Jolt skill: reuse supplied answers and authorization; retain questions for missing requirements and approval when edits were not requested.
- `ci-code-review`: remove the instruction to run tests that contradicted its CI-only testing rule. `new-spec`: require concrete acceptance evidence without demanding a new test for every change.

## Preserved

- Required host/ZK checks, protocol invariants, command examples, skill metadata, and the AGENTS symlink.
- Claude model names, agent counts, tool names, routine dispatch, and interactive spec interviews. Astra-specific API changes and blanket delegation are outside this shared-guidance patch.
- No new untracked guidance was found. All eight modified files were tracked; changes are isolated in one branch.

## Verification

- `git diff --check` and staged diff inspection.
- Compared every fenced code block and skill frontmatter with the base: unchanged.
- Verified the tracked AGENTS symlink and the original main checkout's index/status are unchanged.
- Commit hooks passed: typos, style invariants, formatting, and `cargo clippy --all --all-targets -q --message-format=short -- -D warnings` (default features). No tests added or run.

No model A/B evaluation performed; this PR changes instructions, not the agent provider.

### DIFF
--- a/.claude/skills/ci-code-review/SKILL.md
+++ b/.claude/skills/ci-code-review/SKILL.md
@@ -54,7 +54,7 @@ Follow these steps:
 
    For each issue from the agents:
    - For complex logic/semantic issues, reason through whether the bug is real and exploitable.
-   - For issues you can verify mechanically (e.g., a failing test), prefer direct verification (run the test).
+   - For mechanically checkable issues, inspect existing tests and CI results; do not run builds or tests unless the user explicitly requests them.
 
    Score each issue 0-100 AFTER validation:
    - 0: False positive, doesn't stand up to scrutiny, or pre-existing issue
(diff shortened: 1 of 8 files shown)


ozereray:
The authorization/verification distinction here is especially interesting for agent systems.

I think a useful invariant is to keep four things explicitly separate at the execution boundary:

1. what the user authorized,
2. what the agent proposed,
3. what policy allowed,
4. what actually executed.

That separation makes it possible to audit cases where the agent stays inside its general scope but a specific tool call, target, or resource would still be unauthorized.

We're building Aegisora around that model as well: ALLOW / BLOCK / ESCALATE is produced immediately before execution, with the decision bound to the agent/tool/target and an auditable evidence record.

I'd be interested in whether Jolt's guidance is intentionally aiming for this kind of execution-time decision contract, or primarily for documentation/agent-behavior discipline.

PR #1851, opened 2026-09-09
fix(riscv): enforce absolute LD/SD alignment

felix314159:
## Summary

this PR enforces absolute eight-byte alignment for RV64 LD and SD, independently of the RAM remapping base.

## Changes

- Adds a DoublewordAlignment lookup to the modular and legacy provers.
- Binds the lookup to rs1 + imm.
- Requires the alignment result through the existing assertion constraint.
- Adds an end-to-end regression using a six-instruction RV64 trace with a misaligned RAM remapping base.

## Testing

- Modular clear suite: 22/22 passed.
- Modular ZK suite: 14/14 passed.
- Legacy muldiv with host: 3/3 passed.
- Legacy muldiv with host,zk: 3/3 passed.
- Workspace clippy passed with host and host,zk.
- cargo fmt -q passed.

## Security Considerations

For layouts generated by MemoryLayout::new, RAM RAF and the R1CS address constraint already enforce alignment because the remapping base is eight-byte aligned.

Caller-constructed or deserialized verifier preprocessing can contain a remapping base where `lowest_address % 8 == 4`. In that case, RAM RAF admits addresses congruent to four modulo eight, allowing architecturally misaligned LD and SD executions.

The regression demonstrates that the previous verifier accepts such a proof, while this change rejects the same witness.

## Breaking Changes

The lookup-table ABI changes by appending `DoublewordAlignment`. This means proofs and preprocessing artifacts created against the previous lookup-table catalog are not compatible with this version.

### DIFF
--- a/crates/jolt-lookup-tables/src/instructions/riscv/ld.rs
+++ b/crates/jolt-lookup-tables/src/instructions/riscv/ld.rs
@@ -3,14 +3,38 @@ use crate::traits::LookupQuery;
 use jolt_riscv::instructions::Ld;
 use jolt_riscv::JoltCycle;
 
-impl_lookup_table!(Ld, None);
+impl_lookup_table!(Ld, Some(DoublewordAlignment));
 
 impl<const XLEN: usize, C: JoltCycle> LookupQuery<XLEN> for Ld<C> {
     fn to_instruction_inputs(&self) -> (u64, i128) {
-        (0, 0)
+        super::doubleword_memory_instruction_inputs::<XLEN, _>(&self.0)
+    }
+
+    fn to_lookup_operands(&self) -> (u64, u128) {
+        super::doubleword_memory_lookup_operands::<XLEN, _>(&self.0)
+    }
+
+    fn to_lookup_index(&self) -> u128 {
+        LookupQuery::<XLEN>::to_lookup_operands(self).1
     }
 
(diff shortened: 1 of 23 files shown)


moodlezoup:
Do you have an example demonstrating the bug here? I'm pretty sure we're already covered by the RAM raf-evaluation sumcheck plus the R1CS constraint that `RamAddress = Rs1Value + Imm`

felix314159:
> Do you have an example demonstrating the bug here? I'm pretty sure we're already covered by the RAM raf-evaluation sumcheck plus the R1CS constraint that `RamAddress = Rs1Value + Imm`

Thanks for the fast feedback, and you are right for layouts produced by `MemoryLayout::new` specifically. I have narrowed the PR and added the proof of concept. Basically what remains valid is that caller-constructed or deserialized verifier preprocessing can use a misaligned RAM remapping base. When `lowest_address % 8 == 4`, RAF reconstructs addresses as 8k + lowest_address, so RAF plus RamAddress = Rs1Value + Imm still admits misaligned LD/SD accesses. The PoC produces and verifies such a proof before this change and the updated PR rejects the same witness.

moodlezoup:
Maliciously constructed preprocessing is outside of Jolt's security model. Closing this unless there's a PoC that doesn't involve tampering with the preprocessing.

PR #1860, opened 2026-09-10
fix(akita): allow configuring CPU commitment scratch budget

anylots:
Large one-hot commitments can exceed the fixed 8 MiB CPU scratch budget even when the host has ample memory. A Nitro
 attestation trace required 9,449,508 bytes per worker and failed before commitment allocation.

 Allow overriding the budget through `JOLT_AKITA_COMMIT_SCRATCH_BYTES_PER_WORKER`, read once and shared by setup and
 proving. Defaults remain unchanged; invalid values return a setup error.

 Validated with Clippy, 24 unit tests, and 3 existing roundtrip tests using a 16 MiB override.

### DIFF
--- a/crates/jolt-akita/src/adapters.rs
+++ b/crates/jolt-akita/src/adapters.rs
@@ -1,4 +1,5 @@
 use std::{
+    env::VarError,
     fmt,
     io::Cursor,
     path::{Path, PathBuf},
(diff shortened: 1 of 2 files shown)


moodlezoup (approved):
thanks!

PR #1863, opened 2026-09-14
feat(inlines): keccak256 INIT and unaligned-block variants, word-wise SDK padding

0xAndoroid:
Implements `specs/keccak256-inline-init-unaligned.md` (included in this PR).

## Why

Every hash has exactly one first block, and the SDK paid for it twice: 25 stores to zero the state, then 25 loads of those zeros inside the absorb inline. Misaligned input was copied block by block into an aligned stack buffer, and the final block was padded with sub-word stores. jeth carries a ~300-line word-wise shim to beat this; the tricks belong in the sequence builder, where containing-word loads are plain `LD`s on flat RAM.

## Change 1: INIT variant (funct3 0x02)

`Keccak256InitAbsorbPermutation`: lanes 0..17 are the block at `rs2`, lanes 17..25 are zero, `rs1` is only written. The eight zero lanes are never materialized: round 0's θ skips them in the column parities, and each column's first zero lane takes over its D register (the other three copy it), so INIT costs 17 `LD` and *fewer* round rows than the permutation, not 17 `LD` + 8 zero moves.

## Change 2: unaligned-block variants (funct3 0x03 absorb, 0x04 INIT)

`Keccak256AbsorbPermutationUnaligned` / `Keccak256InitAbsorbPermutationUnaligned`: `rs2` at any alignment. The sequence reads the 18 aligned doublewords at `rs2 & !7` and funnel-shifts them: with `sh = 8 * (rs2 & 7)`, lane `i = w[i] >> sh | w[i+1] << (64 - sh)`. Both shift operands are computed once in a 9-row prologue — the right shift is a `VirtualSRL` by the bitmask `-(2^sh)`, the left shift a `MUL` by `2^(64 - sh) mod 2^64` — so each lane costs `LD + VirtualSRL + MUL + OR` (+ `XOR` for absorb). The multiplier is zero for an aligned `rs2`, so the sequence is also correct there (reading one doubleword past the block); the SDK never issues that case. For a misaligned `rs2` the 18 doublewords are exactly the ones containing the block. Temporaries reuse the C/D registers (dead until θ): still 37 virtual registers, no new instruction kinds.

## Change 3: SDK

- `Keccak256::digest`: first block through INIT (aligned or unaligned by `ptr & 7`) into a `MaybeUninit` state — the state is never zeroed in memory; further full blocks through absorb or its unaligned twin, decided once (stride 136); the tail through a word-wise padded block. `< 136` bytes is one INIT of the padded block.
- `absorb_full_blocks` (also `update`): the per-block `copy_nonoverlapping` path is gone; misaligned input feeds the unaligned inline directly.
- `finalize`: pads word-wise (clear the partial word's dead bytes and the words after it, then `block[len/8] |= 0x01 << 8*(len%8)`, `block[16] |= 1 << 63`). No `u8` store remains on any digest path; the only sub-word traffic is the platform `memcpy` of the tail.
- New `pub unsafe` entry points next to `keccak256_absorb_permute`, same three cfg arms: `keccak256_init_absorb_permute`, `keccak256_absorb_permute_unaligned`, `keccak256_init_absorb_permute_unaligned`. The host arms model the ops in `exec.rs` (`execute_absorb_permute`, `execute_init_absorb_permute`).
- Book: inline table rows for funct3 0x01–0x04 (0x01 was missing).

## Rows

`test_keccak256_inline_rows` pins the exact counts (24 rounds + traffic + 37 register resets):

| funct3 | op | rows | vs absorb (0x01) |
|---|---|---:|---:|
| 0x00 | permutation | 3087 | unchanged (byte-identical sequence) |
| 0x01 | absorb | 3121 | unchanged (byte-identical sequence) |
| 0x02 | INIT | 3066 | −55 |
| 0x03 | absorb, unaligned | 3181 | +60 |
| 0x04 | INIT, unaligned | 3126 | +5 |

Byte-identity of 0x00/0x01 checked by diffing the full `Debug` dump of both sequences before/after.

`sha3-chain` guest (100 × `Keccak256::digest` of a 32-byte input, i.e. the padded-INIT path), traced via `analyze_sha3_chain`: **332,140 → 314,541 rows (−17,599, −176 per hash, −5.3%)**. Per hash: −55 in the inline, −25 `SD` (no state zeroing), the rest from the dropped byte-store padding and the smaller `memset`/`memcpy` loops (`LD` −27, `SD` −27, `BLTU` −28, `ADDI` −34, `XOR` −33, `VirtualWindowMaskB`/`VirtualShiftDataB` −2 each).

## Tests

- `spec.rs`/`tests.rs`: `InlineSpec` impls for the three ops through `assert_edge_cases_match_reference` / `assert_random_cases_match_reference` (100 random each). Unaligned inputs are `(state, block, offset)`; the harness input region is 144 bytes, the block is written `offset` bytes in with junk around it and `rs2` bumped by `offset`. Edge cases cover offsets 0, 1, 7 (0 pins the aligned-`rs2` behaviour). The INIT specs load junk into `rs1` to prove it is not read.
- `sdk.rs`: `test_keccak256_every_offset_and_length` — every pointer offset mod 8 × every length 0..=600 plus 1024, 8192, through `digest` and six `update` chunkings, against the `sha3` crate.
- `differential_fuzz.rs`: `fuzz_init_and_unaligned_inlines_vs_tiny_keccak` — the three new sequences in the emulator vs `tiny_keccak::keccakf`, 2,000 cases (scales with `KECCAK_FUZZ_ITERS`).
- Row ratchet as above. Existing tests (trace-file round trip, XKCP vectors, `test_execute_keccak256`, the digest/update fuzzes) unchanged and green.

## Checks

```
cargo fmt -q
cargo clippy --all --features host -q --all-targets -- -D warnings
cargo clippy --all --features host,zk -q --all-targets -- -D warnings
cargo nextest run -p jolt-inlines-keccak256 --features host --cargo-quiet   # 22 passed
cargo nextest run -p jolt-inlines-sdk --features host --cargo-quiet
```

## Spec deviations

- **Zero lanes folded, not moved.** The spec's INIT loads the block and emits 8 zero-moves; this PR tracks the zero lanes through round 0's θ instead (−8 parity `XOR`s, −8 moves, 5 of the remaining copies replaced by register adoption). INIT is 3066 rows, not the spec's `rounds + 50`.
- **Shift lowering.** In Jolt `SRL`/`SLL` expand to two rows each (`VirtualShiftRightBitmask` + `VirtualSRL`, `VirtualPow2` + `MUL`), so the spec's per-lane `SRL; SLLI 1; SLL; OR` would be 6 rows. The bitmask and the left-shift multiplier are hoisted into the prologue (`VirtualAlignAddr`, `ANDI`, `VirtualMULI`, `XORI`, `VirtualPow2`, `ADD`, `VirtualPow2`, `SUB`, `LD`), and `2^(64-sh) mod 2^64` replaces the spec's `<< 1; << nsh` pair. `VirtualShiftRightBitmask` has no `Kind::` constant reachable from an inline crate (its variant name collides), so the bitmask is `SUB x0, 2^sh` (one extra prologue row).
- **No trace fixture to regenerate.** `*.joltinline` files are gitignored and nothing in the repo calls `store_inlines()`; `test_keccak256_trace_file_matches_generated` round-trips the writer on the fly. Nothing to check in.
- **Trace-length measurement.** The modular `profile` command writes only the padded trace length to `timings.csv`, and its default `reference` backend proves 2^22 in hours, so the delta above comes from `analyze_sha3_chain` (`ProgramSummary::trace_len`) on the same guest with 100 iterations, run on `origin/main` and on this branch. Same guest, same input; only the iteration count differs from the profile's.
- **Book table:** four rows added (the existing absorb op 0x01 had no row), not three.
- **`test_keccak256_aligned_vs_unaligned`** is renamed to `test_keccak256_every_offset_and_length`; it no longer compares aligned against unaligned but every case against the reference.


0xAndoroid:
Superseded, not pursuing (INIT/unaligned variants ≈0.2% of rows; jeth keeps its own shim).

PR #1865, opened 2026-09-15
feat(sumcheck): support standalone use

omibo:
## Motivation

`jolt-sumcheck` could not be used as a lightweight external dependency. Even with `default-features = false`, its dependency graph selected BN254, Arkworks, polynomial commitments, stock transcript backends, and Rayon through unconditional dependencies and inherited defaults.

This change makes clear sumcheck proving and verification usable with a caller-provided field and transcript, without selecting a curve backend, commitment scheme, or thread pool. Jolt's existing committed and ZK modes remain available through explicit features.

## Changes

- Give `jolt-sumcheck` an empty default feature set and add explicit features for parallelism, stock transcript backends, committed rounds, and R1CS lowering.
- Disable dependency defaults throughout the standalone sumcheck closure and make crypto and R1CS integrations optional.
- Move `EvaluationClaim` to `jolt-poly`, with compatible re-exports from `jolt-sumcheck` and `jolt-openings`, so clear sumcheck no longer depends on the openings crate.
- Split `jolt-transcript` into a dependency-light trait surface and feature-gated Blake2b, Keccak, Poseidon, Spongefish, and BN254 capabilities. Direct transcript consumers retain the existing stock-backend defaults, while Jolt workspace dependency edges disable defaults and select their required backends explicitly.
- Separate `jolt-crypto`'s BN254 and parallel features, while preserving both in that crate's own defaults. BN254 code now has sequential implementations when Rayon is disabled.
- Make R1CS claim lowering optional so sumcheck can use only the core R1CS builder.
- Update Jolt crates to select the capabilities they use explicitly, preserving existing clear, committed, and ZK configurations.
- Keep `jolt-akita` free of BN254, Arkworks, Poseidon, and Dory dependencies, including dev dependencies. The mixed Akita/Dory comparison benchmark now belongs to `jolt-dory`.
- Add a custom-transcript example, external-consumer fixture, dependency checks, and usage documentation for pinned Git consumers.

## What to review

1. **Feature boundaries:** confirm that the default sumcheck graph contains only the clear, sequential surface and that each optional feature activates the intended capability.
2. **Compatibility:** check the `EvaluationClaim` move and transcript refactor for unchanged public paths, serialization, and Fiat-Shamir absorption order.
3. **Committed and ZK behavior:** confirm that optionalizing crypto and R1CS does not weaken or alter Jolt's committed proof path.
4. **Parallel fallbacks:** review the sequential BN254 batch-addition and GLV paths used when `jolt-crypto/parallel` is disabled.
5. **Akita isolation:** verify that Akita's manifests and relocated benchmark keep every Akita feature combination curve-free.
6. **External contract:** review the example and isolated fixture as the supported pattern for downstream Git dependencies.

## Validation

- `scripts/check-external-sumcheck.sh`
- `scripts/check-akita-dependencies.sh`
- `jolt-sumcheck`: minimal and committed/R1CS suites
- `jolt-transcript`, `jolt-crypto`, `jolt-openings`, `jolt-r1cs`, `jolt-blindfold`, `jolt-dory`, and `jolt-akita` package suites
- Modular prover clear fixture suite, including legacy byte-diff checks
- Modular prover ZK acceptance cases, with concurrent stack-overflowing cases rerun successfully in isolation
- Legacy `muldiv` end-to-end tests with `host` and `host,zk`
- Strict package-level Clippy checks and formatting

The full workspace Clippy command still encounters an unrelated pre-existing invalid attribute emitted by the recursion guest macro. Package-level checks for the affected dependency changes pass.

## Compatibility notes

- Jolt's proof encoding and transcript schedule are unchanged.
- `jolt-crypto` still defaults to BN254 plus parallel execution for existing direct consumers.
- `jolt-transcript` still defaults to the three stock backends for existing direct consumers; `default-features = false` exposes only the generic transcript surface.
- Poseidon explicitly selects BN254 because the current implementation is field-specific.
- External consumers should pin all Jolt crates to the same full Git revision. Cargo workspace-root patches are not inherited by downstream projects; minimal clear sumcheck does not require Jolt's Arkworks patches.


crates/jolt-sumcheck/src/lib.rs
```diff
@@ -96,9 +97,9 @@ pub mod round_proof;
 pub mod scalar;
 pub mod verifier;
 
-#[cfg(test)]
+#[cfg(all(test, feature = "committed"))]
 mod round_scheduler_tests;
-#[cfg(test)]
+#[cfg(all(test, feature = "committed"))]
 mod tests;
```
moodlezoup, line 103: Gating `mod tests` on `committed` silently drops this crate's unit-test suite (~60 tests in tests.rs plus round_scheduler_tests) from CI: the per-crate shard in rust.yml runs `cargo nextest run -p jolt-sumcheck` with the now-empty default features and `--no-tests=pass`, no CI invocation enables `committed` for this crate, and there is no workspace-wide nextest run to pick it up via feature unification. The new `[[test]] name = "committed"` integration target never runs in CI either. Most of tests.rs doesn't touch jolt-crypto — consider splitting the crypto-free tests out of the gate, or adding a `-p jolt-sumcheck --features committed` lane to the shard.

---

scripts/check-external-sumcheck.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
```
moodlezoup: Neither this script nor `check-akita-dependencies.sh` is wired into CI (or referenced anywhere outside the jolt-sumcheck README), so the PR's core guarantees — the minimal standalone graph staying free of ark-*/rayon/spongefish/getrandom, the external fixture's committed Cargo.lock staying valid, Akita staying curve-free — are enforced by nothing and will regress silently, since workspace feature unification means regular CI never compiles the `committed`/`r1cs`/backend seams in their off state. Both scripts should get a workflow step.

---

crates/jolt-crypto/src/lib.rs
```diff
@@ -21,8 +21,32 @@
...
+macro_rules! cfg_iter_mut {
+    ($values:expr) => {{
+        #[cfg(feature = "parallel")]
+        let it = $values.par_iter_mut();
+        #[cfg(not(feature = "parallel"))]
+        let it = $values.iter_mut();
+        it
+    }};
+}
+
```
moodlezoup: This reimplements `ark_std::cfg_iter!`/`cfg_iter_mut!`, which have identical bodies and resolve `feature = "parallel"` at the expansion site (i.e. against jolt-crypto's own feature) — and ark-std is already a dependency under the same `bn254` gate that guards these macros. Importing them from ark_std at the use sites deletes both definitions. There's also a third copy of this macro family already exported from `jolt-field/src/solinas/parallel.rs`; per the one-owner convention, worth consuming one of the existing ones rather than adding another.

---

crates/jolt-sumcheck/src/prover.rs
```diff
@@ -216,28 +220,13 @@ where
...
-    }
 
-    #[expect(
-        clippy::unwrap_used,
-        reason = "2 is invertible in any field of characteristic != 2, and Jolt fields are large-prime"
-    )]
-    let two_inv = F::from_u64(2).inverse().unwrap();
+    let two_inv = F::from_u64(2)
+        .inverse()
+        .ok_or(SumcheckError::TwoNotInvertible)?;
```
moodlezoup: `Ring::two_inv()` already owns this formula (jolt-field/src/algebra.rs), documents characteristic-two as unsupported at the trait level, and lets fields override it together with `half()` for a cheap shift — open-coding `from_u64(2).inverse()` bypasses that override, and the new public `TwoNotInvertible` variant is unreachable for any field honoring the trait contract. `let two_inv = F::two_inv();` and dropping the variant is less code.

Related, below: the `coefficient_count` `checked_add` is loop-invariant but recomputed every round, and `DegreeOverflow` only fires at `max_degree == usize::MAX` (anything slightly smaller aborts in the `vec[]` allocation one line later). If kept, hoist it above the round loop — or into `validate_batch_dimensions` with the other dimension checks.

---

crates/jolt-sumcheck/examples/clear_custom_transcript.rs
```diff
@@ -0,0 +1,102 @@
...
+        &mut prover_transcript,
+    )?;
+    let recorded = recorder.finish(&proved.member_claims, &mut prover_transcript)?;
+    let SumcheckProof::Clear(ClearProof::Compressed(proof)) = recorded.proof else {
+        unreachable!("the clear recorder always returns a compressed clear proof")
+    };
+
+    let mut verifier_transcript = FixedTranscript::new(b"external-sumcheck-example");
+    verifier_transcript.append_labeled(SUMCHECK_CLAIM_TRANSCRIPT_LABEL, &claim.claimed_sum);
+    let reduced = SumcheckVerifier::verify_compressed(
```
moodlezoup, line 113: Two Fiat-Shamir deviations from the crate's own contract, invisible here only because `FixedTranscript` ignores every absorb: the verifier never mirrors the `b"opening_claim"` absorbs that `recorder.finish` performs for each member claim (`ClearSumcheckRecorder`'s doc says its writes are byte-identical to what the clear verifier reads back), and the batching coefficient is hardcoded to 1 at line 64 rather than drawn from the transcript as the reference path does (`crates/jolt-kernels/src/reference/naive.rs`). Since this file is the template for the external audience (and doubles as the fixture's main.rs), a consumer copying it with a real transcript gets prover/verifier divergence on the next squeeze, and a fixed coefficient in a multi-member batch lets per-member errors cancel. Worth mirroring the absorbs and deriving the coefficient — or commenting both deviations explicitly.

---

crates/jolt-sumcheck/src/error.rs
```diff
@@ -40,6 +40,47 @@ pub enum SumcheckError<F: Field> {
         max_num_vars: usize,
     },
 
+    /// A batch member declared more rounds than the batch contains.
+    #[error("batch member {member} has {rounds} rounds, exceeding batch size {max_num_vars}")]
```
moodlezoup: `max_num_vars` is the batch's round count, not its member count — "batch size" will mislead external consumers debugging `try_new` (the sibling `BatchMemberWindowOutOfRange` renders it as rounds).

```suggestion
    #[error("batch member {member} has {rounds} rounds, exceeding the batch's {max_num_vars} rounds")]
```

---

moodlezoup (commented):
Reviewed the standalone-sumcheck refactor with focus on soundness, Fiat-Shamir compatibility, and the feature matrix. The core moves check out: `prelude.validate()` is a strict superset of the removed inline checks and runs before any transcript interaction; the `EvaluationClaim` move is byte-identical on the transcript (fs_inventory consistent); the committed/ZK paths change only by cfg gating; the sequential GLV/batch-addition fallbacks are order-independent and result-identical; and every prover-path consumer of jolt-crypto still ends up with `parallel` enabled.

---

scripts/check-akita-dependencies.sh
moodlezoup, line 1: let's delete this

moodlezoup:
Suggestion: could we express the external-consumer round trip as a regular Rust `#[test]` inside the existing `crates/jolt-sumcheck/tests/external-consumer` fixture, then run that fixture explicitly in CI?

Keeping its separate `[workspace]` and minimal dependencies preserves the downstream isolation we want. A normal integration test inside `jolt-sumcheck` would not give the same assurance, because the package's dev-dependencies can enable extra features and the parent workspace's patches still apply.

The fixture could reuse the example's logic and be run with:

```sh
cargo nextest run \
  --manifest-path crates/jolt-sumcheck/tests/external-consumer/Cargo.toml \
  --locked
```

That gives us a persistent test of the supported external-consumer behavior with normal Rust test reporting, without needing a dedicated shell script for it. The parent workspace's test run will not discover the isolated fixture, so it needs an explicit CI invocation. The dependency-tree assertions and the separate transcript feature check can be considered independently.

crates/jolt-sumcheck/src/lib.rs
omibo: Fixed in a68788501. Added an explicit CI lane for `jolt-sumcheck --features committed`; it runs the gated unit and scheduler suites.

crates/jolt-crypto/src/lib.rs
omibo: Fixed in a68788501. Removed the local macro copies and imported `ark_std::cfg_iter!` / `cfg_iter_mut!` at the BN254 use sites. Serial and parallel suites pass.

crates/jolt-sumcheck/src/prover.rs
omibo: Fixed in a68788501. The prover now uses `F::two_inv()`. Coefficient-count overflow is validated once in `BatchPrelude` validation and covered by a boundary test.

crates/jolt-sumcheck/examples/clear_custom_transcript.rs
omibo: Fixed in a68788501. The example derives the batching coefficient after input absorption, mirrors output-claim absorption in the verifier, and asserts final transcript-state equality.

crates/jolt-sumcheck/src/error.rs
omibo: Fixed in a68788501. The error now says the member exceeds the batch round count.

scripts/check-akita-dependencies.sh
omibo: Deleted in a68788501.

omibo:
Addressed the isolated-consumer suggestion in a68788501: the fixture is now a regular `nextest` integration test with an explicit CI invocation. The dependency-tree assertions remain a separate check.

moodlezoup (approved):

PR #1872, opened 2026-09-18
fix(crypto): reject G1/G2 encodings with trailing bytes

kriss39:
## Summary

`Bn254GT`'s serde `Deserialize` gates its payload on the exact compressed size so that a group element has exactly one wire encoding (`gt.rs`: "makes the encoding canonical (no trailing bytes)"). The G1/G2 wrapper macro in `ec/bn254/mod.rs` does not: it reads a `Vec<u8>` and runs `deserialize_compressed(&buf[..])`, and arkworks stops after one point, so any valid point followed by junk bytes still decodes.

In ZK mode every `Vec<Bn254G1>` in a proof goes through this path — the `BlindFoldProof` commitment vectors, the committed sumcheck round commitments of each stage, and Dory's `y_com` — so one proof has many accepted byte encodings. Verification itself is unaffected (the transcript re-serializes canonically), but anything that treats proof bytes or their hash as an identity (dedup, replay tracking) can be fooled, and it contradicts the canonicality the GT decoder and the `DoryProof` decoder (trailing-byte check) already enforce.

This applies the same exact-size gate to G1 and G2.

## Reproducer

Before this change, both of the new tests decode successfully:

```rust
let mut bytes = Vec::new();
Bn254::g1_generator().0.serialize_compressed(&mut bytes).unwrap();
bytes.push(0);
let json = serde_json::to_string(&bytes).unwrap();
serde_json::from_str::<Bn254G1>(&json) // Ok(Bn254G1((1, 2)))
```

The same holds through bincode with the junk inside the length prefix, which is the layout a serialized `JoltProof` uses.

## Testing

- `cargo nextest run -p jolt-crypto` (136 passed, incl. the three new tests)
- `cargo nextest run -p jolt-dory -p jolt-blindfold -p jolt-openings` (141 passed)
- `cargo nextest run -p jolt-verifier --features prover-fixtures,zk completeness::zk` — the live muldiv and committed-muldiv ZK proofs are written to the fixture files and read back through the new decoder (`zk_muldiv_verifier_proof_is_accepted`, `zk_committed_muldiv_verifier_proof_is_accepted` pass; `zk_muldiv_blindfold_shape_audit_matches_modular_protocol` fails 227 vs 221 exactly as on unmodified `main`)
- `cargo clippy -p jolt-crypto --all-targets -- -D warnings` clean


moodlezoup (approved):
ty!

PR #1873, opened 2026-09-18
refactor(sumcheck): minimize algebra boundary

omibo:
## Summary

Make the clear sumcheck implementation usable with fields that implement Jolt's algebraic `Field` interface without requiring the full `JoltField` capability bundle.

Previously, public polynomial, proving, verification, and proof-container APIs inherited `JoltField`. That also required canonical encoding, optimized accumulation, Serde, and optional allocation profiling, even when the underlying operation only performed field arithmetic. This prevented direct use with Jolt's Solinas extension fields and made integration with externally defined fields unnecessarily difficult.

This change places each capability bound at the layer that consumes it:

| Surface | Required capability |
| --- | --- |
| Polynomial storage, evaluation, compression, and interpolation | `Field` |
| Round generation, scheduling, batching, and proof containers | `Field` |
| Stock clear transcript recording and verification | `Field + AppendToTranscript` |
| Concrete digest and sponge challenge generation | `CanonicalEncoding` where currently required |
| Serialization | Conditional `Serialize` / `Deserialize<'de>` |
| Committed proving, optimized kernels, and R1CS adapters | Existing stronger bounds |

Jolt's existing callers continue to satisfy these relaxed bounds. The proving algorithm, transcript labels, compression convention, and sumcheck equations are unchanged.

## Changes

- Remove the redundant public `SumcheckScalar` marker and use `Field` consistently across the prover and verifier.
- Relax `UnivariatePoly`, `CompressedPoly`, and the required Lagrange/interpolation helpers from `JoltField` to `Field`.
- Relax the clear prover, batch scheduler, recorder interface, proof containers, and verifier to their actual algebraic requirements.
- Require `AppendToTranscript` only on implementations that absorb scalar values.
- Remove the canonical-encoding requirement from `Transcript::Challenge`.
- Relax `challenge_scalar_powers` to `Ring`, which is all its power-generation logic needs.
- Keep `JoltField` on committed proving, vector commitments, and R1CS paths that still require its broader capabilities.
- Replace `DeserializeOwned` and higher-ranked deserialization bounds with conditional `Deserialize<'de>` bounds where sufficient.
- Add `CanonicalBytes` implementations for:
  - `FpExt2<F, C>` as `c0 || c1`
  - `FpExt4<F>` as `c0 || c1 || c2 || c3`
- Document the new capability boundary in the sumcheck crate README and crate-level documentation.

The extension encodings concatenate fixed-width little-endian base-field coefficients in the extension's existing basis order. The existing `AppendToTranscript` blanket implementation reverses the complete payload before absorption, so the transcript representation remains explicitly distinct from the canonical little-endian payload.

## External-field and extension coverage

The integration tests exercise the real local field types used to represent the requested field shapes:

- `Prime128Offset275`
- `Ext2<Prime64Offset59>`
- `FpExt4<Prime32Offset99>`

Each field runs through the stock clear prover, recorder, proof container, and verifier. The fixture uses products of affine multilinear polynomials, computes the initial claim by summing the Boolean hypercube, and independently checks the final evaluation by direct substitution.

The tests also cover:

- Nonzero higher extension coordinates in polynomial values and challenges.
- Prover/verifier transcript replay.
- Conditional proof serialization.
- Centered-domain uni-skip proving and verification.
- A tampered quartic coefficient rejected by the application's final evaluation check.
- Golden canonical and transcript byte ordering.

The isolated external-consumer package defines its own field wrapper and deliberately omits:

- `CanonicalBytes`
- `CanonicalEncoding`
- `WithAccumulator`
- Serde
- Allocation profiling

That field still runs the stock clear prover and compressed verifier, showing that unrelated Jolt capabilities no longer leak into the public algebra boundary.

## Testing

- [x] `cargo nextest run -p jolt-poly -p jolt-transcript -p jolt-sumcheck --cargo-quiet`
  - 279 tests passed
- [x] `cargo nextest run -p jolt-field --features solinas --cargo-quiet`
  - 146 tests passed
- [x] Committed, R1CS, and parallel sumcheck suite
  - 89 tests passed
- [x] Isolated external-consumer tests and dependency-isolation script
- [x] Legacy muldiv acceptance tests in clear and ZK modes
- [x] Modular prover fixture suites in clear and ZK modes
- [x] Strict clippy for `jolt-field`, `jolt-poly`, `jolt-transcript`, and `jolt-sumcheck`
- [x] `cargo fmt`
- [x] `git diff --check`

The repository-wide clippy commands remain blocked by the existing `#[jolt::provable]` `invalid attribute` failure in `examples/recursion/guest/src/provable_macro.rs`.

## Security Considerations

This change does not modify sumcheck identities, claims, transcript labels, challenge order, polynomial compression, or Jolt's optimized commitment kernels.

Sumcheck verification still produces a final evaluation claim that the application must check against its polynomial oracle or commitment scheme. The quartic tampering test exercises this obligation explicitly.

The new extension byte implementations define absorption compatibility for the stock clear recorder. They do not define a production challenge sampler for those extension fields. External integrations remain responsible for providing a reviewed transcript with an appropriate challenge distribution.

## Breaking Changes

`jolt_sumcheck::SumcheckScalar` and the public `jolt_sumcheck::scalar` module are removed. Downstream code should use `jolt_field::Field` directly.

`Transcript::Challenge` no longer implies `CanonicalEncoding`. Generic downstream code that relied on that associated-type bound must state an encoding bound locally where it uses canonical encoding.

Existing Jolt field and transcript combinations retain their serialization and transcript behavior. No proof format or existing field encoding is changed.


crates/jolt-sumcheck/src/scalar.rs
```diff
@@ -1,42 +1,6 @@
-use std::{
-    fmt::{Debug, Display},
-    hash::Hash,
-};
+use jolt_field::Field;
 
-use jolt_field::{CanonicalEncoding, Field, Ring};
+/// Algebraic scalar capabilities shared by clear sumcheck proving and verification.
+pub trait SumcheckScalar: Field {}
```
moodlezoup: Let's remove `SumcheckScalar` entirely in this PR. With `SumcheckScalar: Field` and a blanket implementation for every `Field`, it adds no capabilities or restrictions. The prover already uses `Field` directly, so retaining a second name on the verifier side suggests a distinction that no longer exists.

Replace the eight remaining bounds in `domain.rs`, `round_proof.rs`, and `verifier.rs` with `Field`, delete this module and its re-export, and update the spec references. Keep `AppendToTranscript` local to the implementations that absorb scalars, as this PR already does. Note the removed public import in the breaking changes.

I verified this removal locally: all 90 sumcheck tests with `committed,r1cs,parallel`, both external-consumer tests, and strict workspace clippy with `host` and `host,zk` passed.

omibo: Good call — I verified that the marker was now exactly equivalent to `Field` and removed it in d3352a56f.

The eight remaining bounds now use `Field` directly, and the `scalar` module and `SumcheckScalar` re-export are gone. I also updated the PR description's breaking-changes section to call out the removed public import.

Validated with the 90-test `committed,r1cs,parallel` sumcheck suite, both isolated external-consumer tests, strict affected-crate clippy, the style-invariant checker, formatting, and diff checks.

moodlezoup (approved):

